PR #307, opened 2014-04-24
add current_time option to out_stdout

sonots:
Current `out_stdout` outputs messages as

```
2010-05-04 12:02:01 +0900 test: {"test":"test"}
```

where the time expresses the time parameter of messages, i.e., the time when the message is created or, the time written in application logs. 

`current_time` option enables to output the current time of *logging* which looks like

```
2014-04-24 18:49:23 +0900 2010-05-04 12:02:01 +0900 test: {"test":"test"}
```

so that we can tell the latency between receiving (logging) and sending (creating) of messages. 

PS. I am still wondering the log format and the option name.

### DIFF
--- a/lib/fluent/formatter.rb
+++ b/lib/fluent/formatter.rb
@@ -84,6 +84,22 @@ module Fluent
       end
     end
 
+    class OutStdoutFormatter < Formatter
+      config_param :output_type, :string, :default => 'json'
+
+      def configure(conf)
+        super
+
+        @formatter = Plugin.new_formatter(@output_type)
+        @formatter.configure(conf)
+      end
+
+      def format(tag, time, record)
+        header = "#{Time.now.localtime} #{tag}: "
+        "#{header}#{@formatter.format(tag, time, record)}"
+      end
+    end
+
     module StructuredFormatMixin
       def self.included(klass)
         klass.instance_eval {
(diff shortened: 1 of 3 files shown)


sonots:
@kiyoto @repeatedly @tagomoris ping

tagomoris:
It seems to be not so general that OutStdoutFormatter exists ... but I also found OutFileFormatter. Ooooooops.
I think better that output plugin specific formatter exists in out_xxx.rb.

kiyoto:
@sonots The code itself looks good to me.

@tagomoris agree with the lack of generality of out_stdout/out_file formats. As a matter of fact, I think both formats are pretty unpopular/confusing for the newcomers (see, for example, "Problem #2" in [this blog post](http://www.stormacq.com/storing-apache-access-and-error-logs-to-amazon-s3-with-fluentd/). That said, I am against breaking out _just_ out_file and out_stdout formatters. If we are going to break out individual parsers/formatters, we should apply the same principle as output/input plugins and put them all under `lib/fluent/plugin` with the prefixes `parser_` and `formatter_`.

sonots:
I completely agree that out_file and out_stdout formatter are not general, but this was my compromise to support old version compatibility.

repeatedly:
`out_file` format is also used in `out_s3` so this format cannot be located in out_file.rb.

sonots:
So, need to move to out_stdout.rb or not?

repeatedly:
stdout filter has this feature so close this PR.

Issue #98, opened 2013-01-18
Why log level was changed from debug to warn when text does not match the pattern?


nabeken:
I'm implementing a input plugin which extends in_tail plugin.
- the target log file is written by various processes via syslog
- the plugin does pattern matching against some of processes in the file

In 317284e35432cb7bff95e9daba45319db0f69f2e log level was changed from debug to warn when text does not mattch the pattern. so I received too many warns like this:

```
2013-01-18 15:01:49 +0900: pattern not match:  .......
2013-01-18 15:01:49 +0900: pattern not match:  .......
```

To supress these warns I added '-qq' to TD_AGENT_ARGS in /etc/init.d/td-agent but I noticed that other useful warns also supressed.

Why the log level changed? I'd like to supress 'pattern not match' only not to supress all warns. any ideas?

PR #362, opened 2014-06-21
Add suppress_parse_error_log option to ignore parsing error in in_tail

sabottenda:
Add `suppress_parse_error_log` option to in_tail plugin. This is a same feature to fluent-plugin-parser.


tagomoris:
I think we can close this, because not so many vote for this feature here and #98.
I'll reject this change.

PR #401, opened 2014-08-11
add file writable check in #configure

sonots:
`--config-test` is very useful, but it does not check file permission. This proposition is to make a rule to check file permissions in #configure method of each plugin because I thought it is impossible to check file permissions for each configuration parameter of each plugin from Fluentd side (Fluentd can not know what each plugin wants to do)

This pull request is just the first example of the proposition.

@repeatedly What do you think? Yes, each plugin author should support file permission check in #configure method of each plugin when we decide as this is an official way. 

PS. @repeatedly Could you tell me if you have a right place to put the file permission check utility function?

### DIFF
--- a/lib/fluent/plugin/file_util.rb
+++ b/lib/fluent/plugin/file_util.rb
@@ -1,0 +1,30 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Fluent
+  module FileUtil
+    # Check file is writable if file exists
+    # Check directory is writable if file does not exist
+    #
+    # @param [String] path File path
+    # @return [Boolean] file is writable or not
+    def writable?(path)
+      path = File.exist?(path) ? path : File.dirname(path)
+      File.writable?(path)
+    end
+    module_function :writable?
+  end
+end
--- a/lib/fluent/plugin/in_debug_agent.rb
+++ b/lib/fluent/plugin/in_debug_agent.rb
@@ -20,6 +20,7 @@ module Fluent
 
     def initialize
       require 'drb/drb'
+      require 'fluent/plugin/file_util'
       super
     end
 
@@ -31,6 +32,11 @@ module Fluent
 
     def configure(conf)
       super
+      if @unix_path
+        unless ::Fluent::FileUtil.writable?(@unix_path)
+          raise ConfigError, "in_debug_agent: `#{@unix_path}` is not writable"
+        end
+      end
     end
 
     def start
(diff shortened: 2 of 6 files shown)


sonots:
@repeatedly ping

lib/fluent/plugin/file_util.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module Fluent
+  module FileUtil
+    # Check file is writable if file exists
+    # Check directory is writable if file does not exist
+    #
+    # @param [String] path File path
+    # @return [Boolean] file is writable or not
+    def writable?(path)
+      if File.exists?(path)
```
repeatedly: Should use `exist?` instead of `exists?`

lib/fluent/plugin/in_debug_agent.rb
```diff
@@ -31,6 +32,11 @@ def initialize
 
     def configure(conf)
       super
+      if @unix_path
+        unless ::Fluent::FileUtil.writable?(@unix_path)
+          raise ConfigError, "in_debug_agent: `#{@unix_path}` is not writable."
```
repeatedly: `.` is not needed when error message is one sentence.
This is consistency for other places.

lib/fluent/plugin/out_file.rb
```diff
@@ -59,6 +60,16 @@ def configure(conf)
...
       end
 
+      test_path =
+        if @append
+          Time.now.strftime("#{@path_prefix}#{@time_slice_format}#{@path_suffix}#{suffix}")
+        else
+          Time.now.strftime("#{@path_prefix}#{@time_slice_format}_0#{@path_suffix}#{suffix}")
+        end
+      unless ::Fluent::FileUtil.writable?(test_path)
+        raise ConfigError, "out_file: `#{test_path}` is not writable."
```
repeatedly: ditto

lib/fluent/plugin/out_file.rb
```diff
@@ -59,6 +60,16 @@ def configure(conf)
         conf['buffer_path'] ||= "#{@path}.*"
       end
 
+      test_path =
+        if @append
```
repeatedly: Why this condition is needed?
This is for permission test. Adding `_0` causes different result?

sonots: When the file exists already, we should check the **file** permission.

lib/fluent/plugin/in_debug_agent.rb
sonots: Okay, I did not know the rule. I will fix.

lib/fluent/plugin/file_util.rb
sonots: I do not mind. I will fix.

lib/fluent/plugin/out_file.rb
sonots: rewrote to use `generate_path`

sonots:
@repeatedly fixed!

lib/fluent/plugin/file_util.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module Fluent
+  module FileUtil
+    # Check file is writable if file exists
+    # Check directory is writable if file does not exist
+    #
+    # @param [String] path File path
+    # @return [Boolean] file is writable or not
+    def writable?(path)
+      if File.exist?(path)
```
repeatedly: How about following code?

```rb
path = File.dirname(path) unless File.exist?(path)
File.writable?(path)
```

sonots: Good. Fixed

PR #423, opened 2014-08-25
Fix bundler_injection for environment that has a embedded ruby

rail44:
Hello.

In environment with embedded ruby, such as td-agent, the `--gemfile` option may be unavailable.
It will be caused by `bundler_injection.rb` trying to exec system's `bundle` command.
This PR is going to make it find `bundle` from current ruby's gems.

If this looks good to merge, there are two questions.

1. I couldn't find tests about `bundle_inject.rb`. Should I add it?
2. When it cannot find `bundle` command,  Should I make it will put some error message?

Thank you.

### DIFF
--- a/lib/fluent/command/bundler_injection.rb
+++ b/lib/fluent/command/bundler_injection.rb
@@ -14,18 +14,20 @@
 #    limitations under the License.
 #
 
-system("bundle install")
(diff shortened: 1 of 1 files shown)


repeatedly:
I confirmed this patch works on pure fluentd and inside td-agent.
Thanks!

rail44:
Thanks too :star:

PR #503, opened 2014-12-09
Accept '@type' and '@id' in v1 configuration. fixes #500

### LINKED ISSUE #500: '@' prefix for type and id
In v1 configuration, we add `@` prefix for system keyword or symbol to avoid parameter conflict, `@label`, `@include` and etc. I think `type` and `id` are also `@` prefix parameters.

Forcing `@` is high-cost so accepting `type` / `id` and writing document with `@` are better for smooth migration.

What do you think?

repeatedly:
Plugin type and id are system reserved parameters like `@label`.
So using '@' prefix is better for separating normal plugin parameters.

I will add this test to Agent tests.

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -76,7 +76,7 @@ module Fluent
           opts[:pretty_json] = true
         end
 
-        if tags = qs['tag'] and tag = tags[0]
+        if tag = get_search_parameter(qs, 'tag'.freeze)
           # ?tag= to search an output plugin by match pattern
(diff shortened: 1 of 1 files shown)


sonots:
One more cherry-pick to v0.10 branch. https://github.com/fluent/fluentd/commit/c97dedfb3b0c675df24a0081d005b877983e2569

sonots:
Needed one more fix https://github.com/fluent/fluentd/commit/513bd332683d7bbcbd6bdd1d811c135ddd358eec

sonots:
Revert fluent.conf too https://github.com/fluent/fluentd/commit/aba7deb44206bbcff181769d227b1ae2e60d87e4

PR #518, opened 2014-12-17
Buffered plugins for tests

tagomoris:
Add buffered output plugins to test buffer plugins
 * buffered_stdout: show records in log console w/ buffering
 * buffered_null: show errors only brought by buffer plugins

These plugins are useful to develop 3rd party buffer plugins, so should be included in fluent/plugin/ directory.

And fix in_dummy not to stop to emit records beside any errors of buffer/output plugins to reproduce emit errors continuously.

### DIFF
--- a/example/out_buffered_null.conf
+++ b/example/out_buffered_null.conf
@@ -1,0 +1,32 @@
+#
+# bundle exec bin/fluentd -c example/out_buffered_null.conf
+#   (+ --emit-error-log-interval 10)
+<source>
+  @type dummy
+  tag dummy
+  rate 500000000
+  dummy [
+    {"message": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"},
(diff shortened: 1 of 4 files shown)


cosmo0920:
Is this pull request still active?
If so, I'll try to adopt `Formatter` feature.

cosmo0920:
Using Formatter feature, out_buffered_stdout plugin becomes more simple like this:

```ruby
# 
# License header snipped.
# 
module Fluent
  class BufferedStdoutOutput < BufferedOutput
    Plugin.register_output('buffered_stdout', self)

    config_param :output_type, :default => 'json'

    def configure(conf)
      super
      @formatter = Plugin.new_formatter(@output_type)
      @formatter.configure(conf)
    end

    def write(chunk)
      chunk.msgpack_each do |time, tag, record|
        log.write "#{time} #{tag}: #{@formatter.format(time.localtime.to_s, tag, record)}\n"
      end
      log.flush
    end
  end
end
```

tagomoris:
@cosmo0920 Sorry for late reply, and your points looks correct.
Could you add commits for it and re-create another pull-request? I think it's better to do so.

cosmo0920:
OK. I'll send another PR.

tagomoris:
#847 is now going on.

PR #524, opened 2014-12-19
remove the unused local variable

kiyoto:
noticed this while writing a docs.

Another option is to have

```rb
config_param :flush_at_shutdown, :bool, :default => false
```

but that can break backward compatibility (for example, `flush_at_shutdown false` returns different vals.


sonots:
cherry-picked to v0.10

PR #526, opened 2014-12-21
Stop in_exec immediately at shutdown

sabottenda:
in_exec plugin with `run_interval` option does not stop immediately when sending shutdown signal.
It waits at most `run_interval` seconds because the process is sleeping and not interrupted.
`Thread#run` interrupts sleep, so this fixes the problem.

### DIFF
--- a/lib/fluent/plugin/in_exec.rb
+++ b/lib/fluent/plugin/in_exec.rb
@@ -99,6 +99,7 @@ module Fluent
     def shutdown
       if @run_interval
         @finished = true
+        # call Thread#run which interupts sleep in order to stop run_periodic thread immediately.
         @thread.run
         @thread.join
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_exec.rb
```diff
@@ -99,6 +99,7 @@ def start
     def shutdown
       if @run_interval
         @finished = true
+        @thread.run
```
sonots: Want a comment in **codes** why we need this `run`

sabottenda: Comments addressed.

repeatedly:
@sonots @tagomoris okay? If so, I will merge this PR and release v0.12.18.

PR #531, opened 2014-12-30
fix to flush a buffer by USR1 signal on retrying

sonots:
Continued from https://github.com/fluent/fluentd/pull/468

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -384,6 +384,9 @@ module Fluent
     end
 
     def force_flush
+      @num_errors_lock.synchronize do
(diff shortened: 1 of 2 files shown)


lib/fluent/output.rb
```diff
@@ -383,6 +383,7 @@ def try_flush
     end
 
     def force_flush
+      @next_retry_time = Engine.now - 1
```
repeatedly: This line should be wrapped with ` @num_errors_lock.synchronize` ?

sonots: Thanks! wrapped with it.

sonots:
cherry-picked to v0.10

Issue #532, opened 2014-12-30
TimeSlicedOutput does not flush on USR1


sonots:
Separated from https://github.com/fluent/fluentd/pull/468

It looks TimeSlicedOutput does not push chunks to queues until specified time comes. No chunk is pushed, then no chunk is flushed. 

https://github.com/fluent/fluentd/blob/a5d489b2123faf69b156b58602990ae040dc8161/lib/fluent/output.rb#L526-L531

```ruby
          nowslice = @time_slicer.call(Engine.now.to_i - @time_slice_wait)
          @buffer.keys.each {|key|
            if key < nowslice
              @buffer.push(key)
            end
          }
```

Voices from users

* https://twitter.com/fd0/status/165995569834242049
* http://blog.kyanny.me/entry/2012/02/09/002430

PR #533, opened 2014-12-30
Fix TimeSlicedOutput does not flush with SIGUSR1

sonots:
Fix https://github.com/fluent/fluentd/issues/532


repeatedly:
Maybe, adding `true` argument to `BufferedOutput#force_flush` is needed for consistency.

sonots:
cherry-picked to v0.10

PR #535, opened 2015-01-10
Fix Nginx and Apache2 format to parse access log without http-version

sabottenda:
Some hardwares such as LoadBalancer access to http servers without http-version in Request-Line. It's like following:

```
$ telnet localhost 80
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
GET /index.html
```

The access generates a log for Nginx and Apach2 servers.
```
# for Nginx
127.0.0.1 - - [10/Jan/2015:17:52:50 +0900] "GET /index.html" 200 612 "-" "-"
# for Aapache2
127.0.0.1 - - [10/Jan/2015:19:17:08 +0900] "GET /index.html" 200 11510 "-" "-"
```

Embedded parser format for Nginx and Aapach2 cannot parse these logs. This request fixes the regexps to parse correctly.

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -407,7 +407,7 @@ module Fluent
     end
 
     class ApacheParser < Parser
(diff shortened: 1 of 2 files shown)


repeatedly:
Good catch!

sonots:
cherry-picked to v0.10

PR #536, opened 2015-01-10
Add nginx_error Parser

sabottenda:
Embedded parser supports both access and error logs for  Apache in `apache` and `apache_error`, but only access log for Nginx.  This patch adds support to parse error log for Nginx.


repeatedly:
I don't merge this PR for now.
I saw many users try to collect Apache error logs but nginx error doesn't.
If some users want this parser, I will include it as built-in.

Keep to open this PR for new users.

icy:
@sabottenda I don't think this work will be able all nginx error logs. For example when you run `php-fpm` behind `nginx (+fastcgi)`. Messages streamed from `php-fpm` are quite random and I couldn't catch them by any regular expression. Let's see

````
2015/12/08 20:03:54 [error] 365#0: *9 FastCGI sent in stderr: "PHP message: PHP Warning:  require(test): failed to open stream: No such file or directory in /home/port/80/xxzzyy/kyanh.net/index.php on line 15
PHP message: PHP Fatal error:  require(): Failed opening required 'test' (include_path='.:/usr/share/pear') in /home/port/80/xxzzyy/example.net/index.php on line 15" while reading response header from upstream, client: 127.0.0.1, server: kyanh.net, request: "HEAD / HTTP/1.1", upstream: "fastcgi://unix:///run/php-fpm/php-fpm.sock:", host: "example.net"
````

My colleage suggests to use multiple-line format, but I'm not sure if that can catch all kind of errors from nginx. For now I use `format none` to get all original lines :)

repeatedly:
nginx errors somtimes contain multi line logs so we should use multiline format by default.
fluentd-docs also mentions this problem: http://docs.fluentd.org/articles/common-log-formats

So close this PR.

PR #537, opened 2015-01-10
fix SIGHUP does not reload config

sonots:
Reported by https://groups.google.com/forum/#!topic/fluentd/jFQTeNONmCA
Fix a bug by https://github.com/fluent/fluentd/pull/496

This patch makes system config be also reloadable (it was not before).


repeatedly:
Does it cause double memory problem again?

sonots:
HUP is caught by supervisor process and reloading config happens on the supervisor process, not on the master process. So, CoW should not occur.

sonots:
cherry-picked to v0.10

Issue #540, opened 2015-01-19
Old fluentd processes leaked after reloading configuration with SIGHUP.


jperville:
After upgrading fluentd to version 0.12.3, I noticed that the old fluentd processes are not reclaimed after reloading the configuration with SIGHUP. Is this expected behavior? With version 0.12.2, the configuration was not reloaded at all on SIGHUP (fixed in #537), so I was looking forward to the new release.

About my setup: I am running fluentd inside a docker image that watches the docker API. When a docker container is started or stopped, the `/etc/fluentd/fluentd.conf` file is regenerated from a template and the fluentd process is asked to reload its configuration by receiving SIGHUP.

As a workaround for this issue, I will regenerate my image to reload the configuration by completely restarting the fluentd process (SIGINT) but that is far from optimal.

Here is a lot that shows the problem (I filtered out unrelated entries).

```
# ps x
  PID TTY      STAT   TIME COMMAND
    1 ?        Ss     0:00 /usr/bin/python3 -u /sbin/my_init
   15 ?        S      0:00 /usr/bin/runsvdir -P /etc/service
   19 ?        Ss     0:00 runsv fluentd
   20 ?        Ss     0:00 runsv docker-gen
   22 ?        Sl     0:00 /usr/local/bin/docker-gen -watch -notify sv hup fluentd /root/fluentd.tmpl /etc/fluent
   39 ?        Sl     0:00 /usr/bin/ruby2.1 /usr/local/bin/fluentd -c /etc/fluentd/fluentd.conf
  104 ?        Sl     0:00 /usr/bin/ruby2.1 /usr/local/bin/fluentd -c /etc/fluentd/fluentd.conf
  156 ?        Sl     0:00 /usr/bin/ruby2.1 /usr/local/bin/fluentd -c /etc/fluentd/fluentd.conf
  158 ?        Sl     0:00 /usr/bin/ruby2.1 /usr/local/bin/fluentd -c /etc/fluentd/fluentd.conf
  207 ?        R+     0:00 ps x
# sv hup fluentd
# ps aux | grep fluentd
root        19  0.0  0.0    168     4 ?        Ss   02:44   0:00 runsv fluentd
root        22  0.1  0.0 189884  4664 ?        Sl   02:44   0:00 /usr/local/bin/docker-gen -watch -notify sv hup fluentd /root/fluentd.tmpl /etc/fluentd/fluentd.conf
[...]

PR #541, opened 2015-01-19
Fix ThreadError occuring on Signal.trap SIGHUP

sonots:
Fix https://github.com/fluent/fluentd/issues/540

This is the first easiest implementation to fix the issue.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -321,12 +321,16 @@ module Fluent
       trap :HUP do
         $log.debug "fluentd supervisor process get SIGHUP"
         $log.info "restarting"
-        read_config
-        apply_system_config
-        if pid = @main_pid
-          Process.kill(:TERM, pid)
-          # don't resuce Erro::ESRSH here (invalid status)
-        end
+        # Creating new thread due to mutex can't lock
+        # in main thread during trap context
+        Thread.new {
+          read_config
+          apply_system_config
+          if pid = @main_pid
+            Process.kill(:TERM, pid)
+            # don't resuce Erro::ESRSH here (invalid status)
+          end
+        }.run
       end
 
       trap :USR1 do


repeatedly:
If works on @jperville environment, I will merge it.

These code are replaced with ServerEngine's in v0.14.

jperville:
I have applied the patch by hand (wrapping the reloading code in a Thread) and it seems to be working.

Here is the fluentd output after applying the fix:

```
2015-01-19 07:59:37 +0000 [info]: restarting
2015-01-19 07:59:37 +0000 [info]: reading config file path="/etc/fluentd/fluentd.conf"
2015-01-19 07:59:38 +0000 [info]: shutting down fluentd
2015-01-19 07:59:38 +0000 [info]: process finished code=0
2015-01-19 07:59:38 +0000 [error]: fluentd main process died unexpectedly. restarting.
2015-01-19 07:59:38 +0000 [info]: starting fluentd-0.12.3
2015-01-19 07:59:38 +0000 [info]: gem 'fluent-mixin-config-placeholders' version '0.3.0'
2015-01-19 07:59:38 +0000 [info]: gem 'fluent-plugin-elasticsearch' version '0.7.0'
2015-01-19 07:59:38 +0000 [info]: gem 'fluent-plugin-exclude-filter' version '0.0.1'
2015-01-19 07:59:38 +0000 [info]: gem 'fluent-plugin-record-modifier' version '0.2.0'
2015-01-19 07:59:38 +0000 [info]: gem 'fluentd' version '0.12.3'
2015-01-19 07:59:38 +0000 [info]: gem 'fluentd' version '0.10.58'
2015-01-19 07:59:38 +0000 [info]: using configuration file: <ROOT>
  <match docker.filtered.**>
    type elasticsearch
    host localhost
    port 9999
    index_name fluentd
    type_name fluentd
    logstash_format true
  </match>
</ROOT>
2015-01-19 07:59:38 +0000 [info]: adding match pattern="docker.filtered.**" type="elasticsearch"
```

The `fluentd main process died unexpectedly. restarting` message is logged as expected.

I have also tried spamming `sv hup restart` to see how the program behaves when the HUP signal is received while the reloading code is still working, and it seems to behave sanely.

repeatedly:
@jperville Thanks for the confirmation!

@sonots Is this no problem on regular environment? If so, please merge this patch!

sonots:
It seems sending multiple HUP invokes bad situation with this code. I need SignalThread maybe. Please wait.

repeatedly:
> I need SignalThread

In v0.14, we have a plan to replace supervisor code with ServerEngine.
At that time, we use SignalThread for these signal handling.
So there is no additional patch for this problem.

sonots:
Okay

jperville:
Thank you very much @sonots and @repeatedly . Any chance for a new gem release with the fix?

repeatedly:
I want to include https://github.com/fluent/fluentd/pull/542 for next version.
I will merge https://github.com/fluent/fluentd/pull/542 soon.

sonots:
cherry-picked to v0.10 branch

repeatedly:
@jperville I released v0.12.4. Please try!

PR #542, opened 2015-01-22
Support 64bit inode environment in in_tail

repeatedly:
For example, XFS uses 64bit integer for inode.
in_tail should work on these environments.

See this thread: https://groups.google.com/d/topic/fluentd/YYrshTNGQ8c/discussion

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -597,7 +597,7 @@ module Fluent
         @file.write path
         @file.write "\t"
         seek = @file.pos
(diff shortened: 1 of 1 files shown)


repeatedly:
We need some testers for

- test on XFS like 64bit inode environment
- test existing pos_file updated to 64bit inode automatically

I tested latter and it seems to work.

repeatedly:
Okay, some users tested and this patch doesn't break existing environment.
Ready to merge.

sonots:
cherry-picked to v0.10 branch

PR #543, opened 2015-01-28
MultilineCSVParser: parse CSVs with newlines

tiandrey:
I've written new parser based on MultilineParser and CSVParser for simple purpose - parse CSVs with newlines (used to parse PostrgreSQL csvlog).

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -621,6 +621,54 @@ module Fluent
       end
     end
 
+    class MultilineCSVParser < ValuesParser
+      config_param :format_firstline, :string, :default => nil
+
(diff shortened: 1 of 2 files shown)


repeatedly:
I don't merge this PR for now.
The reason is here: https://github.com/fluent/fluentd/pull/536#issuecomment-69874394

But in_tail uses multiline mode when `multiline` prefix is specified is good.
Could you send another PR for checking multiline prefix?

tiandrey:
Done. https://github.com/fluent/fluentd/pull/545

repeatedly:
Released multline prefix support version before.

PR #547, opened 2015-01-31
Add tag information to emit log

repeatedly:
This is useful for identifying which input / application causes error events.


sonots:
:+1:

tagomoris:
:+1:

PR #551, opened 2015-02-12
Fixed the error handling upon exception at heartbeat with TCP

tatsu-yam:
If the fluentd process of the destination node is not running, `ForwardOutput#send_heartbeat_tcp` throws `Errno::ECONNREFUSED`. If the exception at line 364 was not rescued, it would be excluded from `@nodes.each{}` and fall to send heartbeats to subsequent nodes. For this reason, if the fluentd process of a destination node stops, the fluentd process of the localhost detaches from subsequent nodes.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -361,7 +361,7 @@ module Fluent
           else
             @usock.send "\0", 0, Socket.pack_sockaddr_in(n.port, n.resolved_host)
           end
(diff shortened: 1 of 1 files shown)


tagomoris:
LGTM :)
Failure of TravisCI seems timing problem, not bug.

repeatedly:
travis-ci seems slow and it seems no problem.
Just merged.

repeatedly:
@sonots Could you backport to v0.10?

sonots:
cherry-picked to v0.10

PR #552, opened 2015-02-13
Use container to reduce testing time

repeatedly:


sonots:
Did this make effect? Cherry-picked to v0.10

PR #556, opened 2015-03-02
BufferedOutputTestDriver must use correct buffer chunk type

repeatedly:
Only MemoryBufferChunk is not good for plugin test.


repeatedly:
@sonots Could you also backport this? I want to release new v0.12 and v0.10.

sonots:
cherry-picked to v0.10

repeatedly:
Thanks!

PR #558, opened 2015-03-05
record_transformer: support placeholders in record keys

sonots:
@kiyoto @repeatedly 

This patch is to add a feature to support placeholders in record keys for filter_record_transformer, which is requested by @fujiwara cf. https://twitter.com/fujiwara/status/572990626930638848

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -108,27 +108,29 @@ module Fluent
 
       new_record = @renew_record ? {} : record.dup
       @keep_keys.each {|k| new_record[k] = record[k]} if @keep_keys and @renew_record
-      @map.each_pair {|k, v| new_record[k] = interpolate(v)}
+      new_record.merge!(expand_placeholders(@map))
       @remove_keys.each {|k| new_record.delete(k) } if @remove_keys
 
       new_record
     end
 
-    def interpolate(value)
+    def expand_placeholders(value)
       if value.is_a?(String)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -108,27 +108,29 @@ def reform(time, record, opts)
...
       @keep_keys.each {|k| new_record[k] = record[k]} if @keep_keys and @renew_record
-      @map.each_pair {|k, v| new_record[k] = interpolate(v)}
+      new_record.merge!(expand_placeholders(@map))
       @remove_keys.each {|k| new_record.delete(k) } if @remove_keys
 
       new_record
     end
 
-    def interpolate(value)
+    def expand_placeholders(value)
```
sonots: just changing the method name

lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -108,27 +108,29 @@ def reform(time, record, opts)
...
         value.each_pair do |k, v|
-          new_value[@placeholder_expander.expand(k)] = interpolate(v)
+          new_value[@placeholder_expander.expand(k)] = expand_placeholders(v)
         end
-        value = new_value
       elsif value.is_a?(Array)
+        new_value = []
         value.each_with_index do |v, i|
-          value[i] = interpolate(v)
+          new_value[i] = expand_placeholders(v)
```
sonots: here, we should not have modified the value of `@map` directly, fixed.

PR #561, opened 2015-03-09
560 - exclude files from tail

ajohnstone:
560 - exclude files from tail

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -26,6 +26,7 @@ module Fluent
 
     config_param :path, :string
     config_param :tag, :string
+    config_param :exclude, :array, :default => []
     config_param :rotate_wait, :time, :default => 5
     config_param :pos_file, :string, :default => nil
     config_param :read_from_head, :bool, :default => false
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks!
I will check it later.

lib/fluent/plugin/in_tail.rb
```diff
@@ -26,6 +26,7 @@ def initialize
 
     config_param :path, :string
     config_param :tag, :string
+    config_param :exclude, :array, :default => []
```
repeatedly: `exclude_path` is good because exclude is too confusion name.

ajohnstone: Thanks, I've made this change...

Lance N on the group mailing list mentioned that you can use format null using a tag to filter this, however this still means your processing the files and tailing them, at least this is my assumption from what the code I've seen is doing.

In my current example I still think the patch is appropriate to avoid any overhead.
The docker container in my case handles all logging. This means all logs would be duplicated by tailing this particular containers logs.

repeatedly:
I will merge this PR today if there is no any concern from other users.

tagomoris:
LGTM

ajohnstone:
Need to add some docs for this, but will be good to get merged.

Thanks

repeatedly:
Thanks!

sonots:
cherry-picked to v0.10

Issue #309, opened 2014-04-26
[WIP] Idea notes for next generation Buffer/Input/Output plugins


tagomoris:
Related with #287

tagomoris:
`buffer_dir_path` instead of `buffer_path`, to manage specified directory exclusively against other plugin instance. All files in specified directory are to be flushed by specifying plugin instance.

* That is useful for fluent-plugin-forest to flush all buffer files at startup
* `buffer_path` has priority than `buffer_dir_path`.

tagomoris:
New configuration parameter set for buffer plugins:

* buffer_space_limit (buffer plugin)
  * to specify total size of buffer chunks (`== buffer_chunk_limit * buffer_queue_limit`)
  * to replace specifying buffer_queue_limit with this parameter to show disk/memory usage explicitly
* `<buffer>` section to specify parameters of buffer plugins (output plugin)
  * by using `config_section`

tagomoris:
Fluent::PluginSupport::* mixin modules instead of Actor
* Fluent::PluginSupport::Thread
* Fluent::PluginSupport::ChildProcess (for exec_filter and others)
* Fluent::PluginSupport::Timer
* Fluent::PluginSupport::Socket (provides listener, tcpsocket(?), sslsocket(?))

tagomoris:
Common API to generate paths with timestamp placeholders like `%Y`, `%m` and `%d`.
Related:
 * buffer chunk id
 * time slice key generation

tagomoris:
Benchmark plugins:
 * [in/out] forward(w/, w/o SSL)
 * [in/out] scribe
 * [out] td, bigquery, s3, mongo, file, exec
 * [in] exec, http, syslog, unix
 * [filter] exec, grep, record_transformer

tagomoris:
Simple datastore for plugin instances & plugin types, to store on-memory data at shutdown (and to resume it at startup).
* Serialization format? (msgpack or json?)
* Datastore? (plain text file, sqlite or pluggable?)

My idea:
 * format: JSON for human readability
 * datastore: Pluggable datastore, but plain text file only at first

tagomoris:
v11 actors: async_actor, background_actor, io_actor, socket_actor, timer_actor

Notes:
* async and background are actually same? (run once or many times)
* socket actor should be separated into tcp, udp and unix socket utils.
* file io actor is needed? it seems to be required only by out_file

tagomoris:
[...]

Issue #287, opened 2014-04-02
Implement Actor


repeatedly:
v11 has Actor.
It improves plugin APIs and makes abstraction layer for Fluentd event loop.
We should port v11 actor to v0.10.x for v1 release.

One concern is Actor is good name or not ;p

repeatedly:
This will be done by PR 562.

PR #562, opened 2015-03-16
[WIP] v0.14 plugin APIs

tagomoris:
Plugins and PluginSupport APIs for v0.14 plugin system.

Related: #309  #287


example/in_status.conf
```diff
@@ -0,0 +1,10 @@
+<source>
```
repeatedly: in_status will be revmoed. so no need this configuration.

repeatedly:
`plugin_support` should be `plugin/support`?
I'm not sure Ruby's manner.

@nurse What do you think?

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,352 @@
...
+
+        if @buffer_config
+          @chunk_bytes_limit = @buffer_config.chunk_bytes_limit
+          @total_bytes_limit = @buffer_config.total_bytes_limit
+
+          @chunk_records_limit = @buffer_config.chunk_records_limit
+
+          @queue_length_limit = @buffer_config.queue_length_limit
+          if @queue_length_limit && @total_bytes_limit > @chunk_bytes_limit * @queue_length_limit
+            @total_bytes_limit = @chunk_bytes_limit * @queue_length_limit
```
repeatedly: `log.warn` is needed.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -0,0 +1,300 @@
...
+    class Buffer
+      class FileChunk < Chunk
+        ### buffer path user specified : /path/to/directory/user_specified_prefix.*.log
+        ### buffer chunk path          : /path/to/directory/user_specified_prefix.b513b61c9791029c2513b61c9791029c2.log
+        ### buffer chunk metadata path : /path/to/directory/user_specified_prefix.b513b61c9791029c2513b61c9791029c2.log.meta
+
+        # NOTE: Old style buffer path of time sliced output plugins had a part of key: prefix.20150414.b513b61...suffix
+        #       But this part is not used for any purpose. (Now metadata is used instead.)
+
+        # state: b/q - 'b'(on stage), 'q'(enqueued)
```
repeatedly: Why use 'b'? 's' for stage is better?

lib/fluent/plugin/buffered_output.rb
```diff
@@ -0,0 +1,522 @@
...
+      end
+
+      def check_as_secondary(primary)
+        if primary.class != self.class
+          log.warn "type of secondary output should be same as primary output", primary: primary.class.to_s, secondary: self.class.to_s
+        end
+      end
+
+      def start
+        @num_errors = 0
```
repeatedly: `num_errors` should be in OutputThread.
Sharing `num_errors` between flush threads causes inconsistent state.

tagomoris: all `@num_errors` (and `@retry`) operations are protected by `@retry_mutex`

lib/fluent/plugin/buffer/file_chunk.rb
tagomoris: For interoperability between old version file buffer chunks and `v0.14`'s one

lib/fluent/plugin/buffered_output.rb
repeatedly: So sharing `@num_errors` is intended?
It means 2 chunks hits errors, hitting retry limit is 2x faster, right?

tagomoris: Yes > intended
In my idea, main control factor for how long buffered plugins continue to retry, is time (minutes/hours), not retry times (but it will also be considered for compatibility).

repeatedly: I see. I checked code again. There is no `@num_errors` in condition.
Seems no problem. Thanks for explanation.

lib/fluent/test/driver/input.rb
```diff
@@ -0,0 +1,114 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/test/driver/base'
+require 'fluent/test/driver/event_router'
+
+module Fluent
+  # TODO: Chekc What is this for?
```
repeatedly: During running multiple test, buffer_path conflicts previous test.
Maybe, Base is better than Input.

nurse:
> plugin_support should be plugin/support?
> I'm not sure Ruby's manner.
Ruby doesn't have related manner, but on Rails "plugin_helper" or "helpers" seem related words.

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,352 @@
...
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def metadata_list
+        synchronize do
+          @metadata_list.dup
+        end
+      end
+
+      def metadata(timekey: nil, tag: nil, key_value_pairs={})
```
cosmo0920: typo? `key_value_pairs` should be keyword argument.

```diff
-def metadata(timekey: nil, tag: nil, key_value_pairs={})
+def metadata(timekey: nil, tag: nil, key_value_pairs: {})
```

lib/fluent/buffer.rb
```diff
@@ -1,336 +0,0 @@
-#
```
cosmo0920: I think this change should be follow in `lib/fluent/load.rb`:

```diff
diff --git a/lib/fluent/load.rb b/lib/fluent/load.rb
index c1f408c..f03866f 100644
--- a/lib/fluent/load.rb
+++ b/lib/fluent/load.rb
@@ -28,7 +28,7 @@ require 'fluent/plugin'
 require 'fluent/parser'
 require 'fluent/formatter'
 require 'fluent/event'
-require 'fluent/buffer'
+require 'fluent/plugin/buffer'
 require 'fluent/input'
 require 'fluent/output'
 require 'fluent/filter'
```

lib/fluent/buffer.rb
```diff
@@ -1,336 +0,0 @@
...
-        begin
-          u.each(&block)
-        rescue EOFError
-        end
-      }
-    end
-  end
-
-
-  class BasicBuffer < Buffer
```
cosmo0920: Where has `BasicBuffer` gone?
This class will be replace near future?

tagomoris: At `v0.14`, we will just use `Fluent::Plugin::Buffer` as a base class for all buffer plugins.

cosmo0920: I see. Thanks for the explanation!

tagomoris:
Now I'm rebasing on current master.

tagomoris:
TODO:
* add `raise NotImplementedError` in methods to be overridden in sub classes (currently commented out)

tagomoris:
Rebased.

lib/fluent/plugin/in_http.rb
```diff
@@ -355,46 +318,30 @@ def on_message_complete
...
+      def send_response(code, header, body=nil)
+        unless @keep_alive
+          @io_handler.closing = true # keepalive disabled for HTTP/1.0 (or earlier)
+        end
 
-      def send_response(code, header, body)
-        header['Content-length'] ||= body.bytesize
-        header['Content-type'] ||= 'text/plain'
+        if body
+          header['Content-length'] ||= body.bytesize
```
cosmo0920: In RFC 7230, content length header is described as [Content-**L**ength](http://tools.ietf.org/html/rfc7230#section-3.3.2).
Should we follow the RFC or keep as-is for backward compatibility?

lib/fluent/plugin/in_http.rb
```diff
@@ -355,46 +318,30 @@ def on_message_complete
...
+          header['Content-type'] ||= 'text/plain'
```
cosmo0920: In RFC 7231, content type header is described as [Content-**T**ype](http://tools.ietf.org/html/rfc7231#section-3.1.1.5).

tagomoris:
I found some inconsistent changes after rebase, and now, I think that it's inappropriate to push this change as a huge pull-request.
So I'll create some different patches and pull-requests for diffs which should be contained to this.

Anyway, this change is good reference to re-create diffs. Please leave this as is.

PR #568, opened 2015-03-27
fix try_flush_interval < 1 and queued_chunk_flush_interval < 1 to work

sonots:
We were supposed to support `try_flush_interval < 1` and `queue_chunk_flush_interval < 1`, but the interval to call `try_flush` was still minimum 1 sec because `Engine.now` is `Time.now.to_i`. 

Illustration: The problem of the old logic:

```ruby
        until @finish
          time = Engine.now

          if @next_time <= time # (4) have to wait to be 1427488240.1 <= 1427488241
            @mutex.unlock
            begin
              @next_time = @output.try_flush # (1) 1427488240.1
            ensure
              @mutex.lock
            end
            next_wait = @next_time - Engine.now # (2) 0.1
          else
            next_wait = @next_time - time
          end

          # (3) will wait 0.1 sec
          cond_wait(next_wait) if next_wait > 0
        end
```

`@next_time` can be like `1427488240.1` when `queued_chunk_flush_interval 0.1`. 
But, `time` exceeds the `@next_time` only when it became `1427488241` (1 sec passed) because `Engine.now` is integer.

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -98,7 +98,7 @@ module Fluent
     def initialize(output)
       @output = output
       @finish = false
-      @next_time = Engine.now + 1.0
+      @next_time = Time.now.to_f + 1.0
     end
 
     def configure(conf)
@@ -132,7 +132,7 @@ module Fluent
       @mutex.lock
       begin
         until @finish
-          time = Engine.now
+          time = Time.now.to_f
 
           if @next_time <= time
             @mutex.unlock
@@ -141,7 +141,7 @@ module Fluent
             ensure
               @mutex.lock
             end
-            next_wait = @next_time - Engine.now
+            next_wait = @next_time - Time.now.to_f
           else
             next_wait = @next_time - time
           end
@@ -231,7 +231,7 @@ module Fluent
     end
 
     def start
-      @next_flush_time = Engine.now + @flush_interval
+      @next_flush_time = Time.now.to_f + @flush_interval
       @buffer.start
       @secondary.start if @secondary
       @writers.each {|writer| writer.start }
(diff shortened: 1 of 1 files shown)


sonots:
cc: @shyouhei

lib/fluent/output.rb
```diff
@@ -98,7 +98,7 @@ class OutputThread
     def initialize(output)
       @output = output
       @finish = false
-      @next_time = Engine.now + 1.0
+      @next_time = Time.now.to_f + 1.0
```
shyouhei: What about just directly use `Time.now + 1.0`.  Time objects can be compared or subtracted as-is.

```ruby
irb(main):001:0> next_time = Time.now + 1.0
=> 2015-03-28 11:24:03 +0900
irb(main):002:0> next_wait = next_time - Time.now
=> 0.719691
```

sonots: float + float is faster than time + float, and time < float results in error

```
irb(main):001:0> Time.now < Time.now.to_f
ArgumentError: comparison of Time with 1427512006.7472174 failed
        from (irb):1:in `<'
```

but I think we can change it to time object here. let me try.

sonots: I made change like this to use time object > https://gist.github.com/sonots/3cd09b3022460ec1aac8

But, test did not pass. I felt it makes code complicated. 
Also, note that float + float is faster than time + float.

My opinion is to use float rather than time object. Any opinions? > ALL

sonots:
@repeatedly ping

lib/fluent/output.rb
shyouhei: OK.  I now feel that changing to Time objects here is kinda overkill to this original issue.  It is a good thing to have, but we can live without them.

shyouhei:
:+1: LGTM

sonots:
I am going to merge this if there is no objection because this is critical for us.

sonots:
Merged!

sonots:
cherry-picked to v0.10

repeatedly:
I hope code itself is no problem.
For safety, adding new method like `Engine.now` is better because refactoring `Time.now.to_f` is hard for sub-second support.

shyouhei:
Thank you.

In case you plan to refactor, I recommend you to just use `Time.now` directly as I wrote earlier.  That should suffice every needs, if not that is a ruby's design flaw.

Issue #279, opened 2014-03-22
Need to work Fluentd with ignoring input plugins


sabottenda:
I want to work fluentd without input plugins if the configuration has `<source>` options. It means the Fluentd does not receive new messages and only work with current remaining buffers. The use case is:
- Flushing all buffers without receiving new messages for retiring servers.
- Debugging output plugins

Of course this feature is accomplished by modifying configurations manually, but I want it to be supported officially. For example, it is difficult to modify a part of configurations for specific servers if they are provided via http.

There are two candidates for the implementation I think.
- Fluentd starts normally, then send signal such as a `SIGUSR2` to a working process to disable input plugins.
- Fluentd starts with a option such as a `--without-source` to disable input plugins.

I think second one is better because first one requires new signal.

Actually it is good for debugging purpose that disabled input plugin is selectable, but I don't care it for now.

Are there any comments or objections for this feature?

PR #569, opened 2015-04-02
Support `stop_file` option in in_forward

sonots:
My colleague was writing this patch. 

This `stop_file` feature enables to stop receiving new incoming data, which is typically used before shutdown.

This is related with https://github.com/fluent/fluentd/issues/279, but `--without-source` is insufficient because we want to stop receiving input **dynamically** without restarting fluentd. 

EDIT: The problem of `--without-source` is

１. We have to take procedures as followings to ensure all received buffers are flushed (FileBuffer)

> -> Shutdown fluentd. We may still be left with some buffer data
> -> Start fluentd again with --without-source option
> -> Shutdown again ensuring that the entire buffer is flushed

２. The above procedure does not work for MemoryBuffer. `stop_file` feature is required for MemoryBuffer.

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -34,6 +34,9 @@ module Fluent
     config_param :chunk_size_warn_limit, :size, :default => nil
     config_param :chunk_size_limit, :size, :default => nil
 
+    config_param :stop_file, :string, :default => nil
+    config_param :stop_check_interval, :time, :default => 5
+
     def configure(conf)
       super
     end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_forward.rb
```diff
@@ -65,9 +72,9 @@ def shutdown
       # As a workaround, check if watchers are attached before detaching them.
       @loop.watchers.each {|w| w.detach if w.attached? }
       @loop.stop
-      @usock.close
+      @usock.close unless @usock.closed?
       @thread.join
-      @lsock.close
+      @lsock.close unless @lsock.instance_variable_get(:@listen_socket).closed?
```
sonots: Want Coolio::TCPServer#closed?

sonots:
In this patch, socket is closed when `stop_file` is found. 

I also have another patch which does not close socket, but stops responding heartbeat when `stop_file` is found although it works for only `heartbeat_type udp`. 
https://github.com/fluent/fluentd/blob/22e85adf7e638f5d108d38b0cdb1e550da5b9567/lib/fluent/plugin/in_forward.rb#L337. 

Any opinions are welcome!

sonots:
Implemented an enchancement version https://github.com/fluent/fluentd/pull/570, so I will close this one.

PR #570, opened 2015-04-02
add --stop-source option

sonots:
This is an enhancement version of https://github.com/fluent/fluentd/pull/569

At #569 , I added an option `stop_file` to **in_forward** plugin, but this version adds `--stop-source` option to fluentd command, and generalized its effectiveness to all input plugins. 

-----

This is a kind of an enhancement version of --without-source option.

This --stop-source feature enables to stop all input plugins
by touching the specified file. This feature is useful when users want
to stop receiving new data until users are sure that all buffers were flushed.

Removing the file restarts all input plugins, and the file existing check runs
every --stop-source-interval seconds (default: 5).

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -83,6 +83,14 @@ op.on('--without-source', "invoke a fluentd without input plugins", TrueClass) {
   opts[:without_source] = b
 }
 
+op.on('--stop-source PATH', "stop input plugins if specified file exists") {|s|
+  opts[:stop_source] = s
+}
+
+op.on('--stop-source-interval SECONDS', "check the stop file every specified seconds") {|s|
+  opts[:stop_source_interval] = s.to_i
+}
+
 op.on('--use-v1-config', "Use v1 configuration format (default)", TrueClass) {|b|
   opts[:use_v1_config] = b
 }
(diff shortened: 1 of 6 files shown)


tagomoris:
I think that it's better to add RPC control feature on Fluentd itself (configured in system section) for many requirements.
And v0.14 seems good step to add something like that.
How do you think about this idea? @sonots @repeatedly

sonots:
I agree that RPC is good for v0.14, but want this feature for v0.12, even for v0.10. Hmm

sonots:
When will the v0.14 be published? Are there any scheduled plans?

sonots:
Modified to use `--without-source PATH` option instead of `--stop-source PATH`. So, now this PR is really an enhancement of the `--without-source` option. 

With this patch, `--without-source` option behaves as follows:

1. `--without-source` invokes fluentd without input plugins (Same with before)
2. `--without-source PATH` invokes fluentd without input plugins if the PATH file exists. It starts input plugins if the file disappears, and stops input plugins if the file appears (New!)

repeatedly:
If this feature is replaced with v0.14's RPC mechanizm, merge is hard.
Because if we add, we should maintain this feature until v2 for backward compatibility.

Is this feature worthy to support in v0.12?
If not, we should focus on developing v0.14 for release early.

repeatedly:
or RPC mechazim can backport to v0.12?

tagomoris:
I think that RPC mechanism is independent from v0.14, so we can implement it on v0.12.
(If someone have enough time to do it.)

repeatedly:
Yeah. Supporting RPC in supervisor/engine and putting related parameters in `<system>` are realistic way.

repeatedly:
I will support this feature after merged this PR: https://github.com/fluent/fluentd/pull/585

repeatedly:
I close this issue.
If you want to implement this feature, please use RPC instead.

PR #571, opened 2015-04-06
Fix bug to overwrite all subsections definition by subclass

tagomoris:
* overwriting config_section options like required/multi/alias breaks behavior of `super`
  * so subclass must not overwrite these options
* subsection definitions of subsections (Hash) is simply merged
  * it is bug, because same section definition like <a> is overwritten by subclass
  * under this bug, subclass MUST repeat superclass's definition, and add original definition
  * but this is impossible in the real world
  * subsection definitions should be merged correctly

This bug especially make effect to the situations such like `<buffer>` configuration sections.
* Once Buffer base class define `<buffer>` section as non-required and non-multi
* Next AwesomeBuffer class (subclass of base class) define `<buffer>` section as required and multi
* `super` of AwesomeBuffer must raises error always

With automatically merged `config_section`, such that case doesn't occur.

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -60,16 +60,29 @@ module Fluent
 
       def merge(other) # self is base class, other is subclass
         options = {
-          param_name: other.param_name,
-          required: (other.required.nil? ? self.required : other.required),
-          multi: (other.multi.nil? ? self.multi : other.multi)
+          param_name: other.param_name, # param_name is used not to ovewrite plugin's instance varible, so this should be able to be overwritten
+          required: (@required.nil? ? other.required : self.required), # subclass cannot overwrite base class's definition
+          multi: (@multi.nil? ? other.multi : self.multi),
+          alias: (@alias.nil? ? other.alias : self.alias),
         }
         merged = self.class.new(other.name, options)
 
         merged.argument = other.argument || self.argument
-        merged.params = self.params.merge(other.params)
+        merged.params = other.params.merge(self.params)
         merged.defaults = self.defaults.merge(other.defaults)
-        merged.sections = self.sections.merge(other.sections)
+        merged.sections = {}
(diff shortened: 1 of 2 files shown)


test/config/test_configurable.rb
```diff
@@ -493,6 +499,59 @@ def e(name, arg = '', attrs = {}, elements = [])
...
+          # <detail> address </detail> is required by Example1 config_param spec
+          assert_raise(Fluent::ConfigError.new("'address' parameter is required, in section detail")) { ConfigurableSpec::Example2.new.configure(conf2) }
+
+          conf3 = e('ROOT', '', {
+              'name' => 'tagomoris',
+              'bool' => true,
+            },
+            [e('detail', '', { 'address' => "Chiyoda Tokyo Japan" }, [])],
+          )
+          # <detail> phone_no </detail is required by Example2 config_param spec
```
repeatedly: `</detail` lack of `>`

tagomoris: Ah, fixing...

tagomoris:
typo fixed.

lib/fluent/config/configure_proxy.rb
```diff
@@ -148,11 +161,16 @@ def config_section(name, *args, &block)
           raise ArgumentError, "#{self.name}: unknown config_section arguments: #{args.inspect}"
         end
 
-        sub_proxy = ConfigureProxy.new(name, *args)
-        sub_proxy.instance_exec(&block)
+        # if @sections[name]
```
repeatedly: Why these lines are commented out?
Can't be deleted?

tagomoris: Ah, it's remains of try-and-errors. I'll delete it.

tagomoris:
fixed and force pushed to remove remains of try&errors.

sonots:
cherry-picked to v0.10

PR #580, opened 2015-04-20
out_forward should resend on dropped connection

### LINKED ISSUE #576: out_forward with require_ack_response drops messages when destination server restarted
I'm building an at-least-once delivery log shipping system with fluentd, and I have the following setup:

    local td-agent --> local haproxy   ----->   remote td-agent
                                         `-->   remote td-agent

The local td-agent uses `out_forward` with `require_ack_response`.

I'm using the local HAProxy because our infrastructure uses [SmartStack](http://nerds.airbnb.com/smartstack-service-discovery-cloud/) and specifically the [Synapse](https://github.com/airbnb/synapse) component to configure HAProxy with the list of dynamically changing aggregator hosts. This means I don't need to reconfigure and restart td-agent when downstream hosts change, and this haproxy setup works well for a number of services.

When testing various failure scenarios of this logging system, I found that if I gracefully stopped the aggregator td-agent (with `/etc/init.d/td-agent stop`) and then started again a few seconds later, some messages would be dropped (ie. not delivered to the outputs on the aggregator box). Exploring the code, I found this in [`out_forward.rb`](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward.rb#L300-L312):

```rb
# When connection is closed by remote host, socket is ready to read and #recv returns an empty string that means EOF.
# In this case, the node is available and successfully close connection without sending responses.
# ForwardInput is not expected to do so, but some alternatives may do so.
# Therefore do not send the chunk again.
unless raw_data.empty?
  # Serialization type of the response is same as sent data.
  res = MessagePack.unpack(raw_data)

  if res['ack'] != option['chunk']
    # Some errors may have occured when ack and chunk id is different, so send the chunk again.
    raise ForwardOutputResponseError, "ack in response and chunk id in sent data are different"
  end
end
```

I believe this behavior to be wrong: it is my expectation that if an ack is not received for _any reason_, then delivery has failed and fluentd should retry delivery of that chunk later.

To test this out, I monkey-patched the `ForwardOutput` class to do this:

```rb
# When connection is closed by remote host, socket is ready to read and #recv returns an empty string that means EOF.
# If this happens we assume the data wasn't delivered and retry it.
if raw_data.empty?
  @log.warn "node #{node.host}:#{node.port} closed the connection. regard it as unavailable."
  node.disable!
  raise ForwardOutputConnectionClosedError, "node #{node.host}:#{node.port} closed connection"
else
  # Serialization type of the response is same as sent data.
  res = MessagePack.unpack(raw_data)

  if res['ack'] != option['chunk']
    # Some errors may have occured when ack and chunk id is different, so send the chunk again.
    raise ForwardOutputResponseError, "ack in response and chunk id in sent data are different"
  end
end
```

Testing again with this code, when stopping and starting the aggregator, the upstream td-agent correctly detects the connection close, reconnects, resends the chunk, and there are no dropped messages.

Should this be a config option? Happy to create a PR if you like the solution.

georgebashi:
Fixes #576. I haven't had time to add tests, but here's a PR for the change itself.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -298,10 +298,12 @@ module Fluent
               raw_data = sock.recv(1024)
 
               # When connection is closed by remote host, socket is ready to read and #recv returns an empty string that means EOF.
-              # In this case, the node is available and successfully close connection without sending responses.
-              # ForwardInput is not expected to do so, but some alternatives may do so.
-              # Therefore do not send the chunk again.
-              unless raw_data.empty?
+              # If this happens we assume the data wasn't delivered and retry it.
+              if raw_data.empty?
+                @log.warn "node #{node.host}:#{node.port} closed the connection. regard it as unavailable."
+                node.disable!
+                raise ForwardOutputConnectionClosedError, "node #{node.host}:#{node.port} closed connection"
+              else
                 # Serialization type of the response is same as sent data.
                 res = MessagePack.unpack(raw_data)
 


repeatedly:
@tagomoris I assigned you to this PR. Feel free to merge.

tagomoris:
I merged and added tests for cases of this situation.
Thank you @georgebashi !

repeatedly:
Release v0.12.8 including this fix.
Please try!

Issue #582, opened 2015-04-21
config_param regression in subclasses


shyouhei:
Current master breaks [riywo/fluent-plugin-hash-forward](https://github.com/riywo/fluent-plugin-hash-forward) behaviour.

### How to reproduce

```bash
bash-3.2$ git clone git@github.com:shyouhei/fluent-config-param.git
bash-3.2$ cd fluent-config-param
bash-3.2$ bundle install -j 32 --path=vendor/bundle
bash-3.2$ bundle exec fluentd -c fluentd.conf 
2015-04-21 18:39:15 +0900 [info]: reading config file path="fluentd.conf"
2015-04-21 18:39:15 +0900 [info]: starting fluentd-0.12.7
2015-04-21 18:39:15 +0900 [info]: gem 'fluentd' version '0.12.7'
2015-04-21 18:39:15 +0900 [info]: gem 'fluent-plugin-hash-forward' version '0.3.5'
2015-04-21 18:39:15 +0900 [info]: using configuration file: <ROOT>
  <match **>
    type copy
    <store>
      type hash_forward
      heartbeat_type none
    </store>
  </match>
</ROOT>
2015-04-21 18:39:15 +0900 [info]: adding match pattern="**" type="copy"
2015-04-21 18:39:15 +0900 [error]: config error file="fluentd.conf" error="forward output heartbeat type should be 'tcp' or 'udp'"
2015-04-21 18:39:15 +0900 [info]: process finished code=256
2015-04-21 18:39:15 +0900 [warn]: process died within 1 second. exit.
bash-3.2$ 
```

Here, the error was from [lib/fluent/plugin/out_forward.rb:47](https://github.com/fluent/fluentd/blob/bdf1f4f104c00a791aa94dc20087fe2011e1fd83/lib/fluent/plugin/out_forward.rb#L47) but the `type` is not forward but `hash_forward`.  It _does_ support `none` heartbeat type [here](https://github.com/riywo/fluent-plugin-hash-forward/blob/ddbbca55848428af405a41e984d549af7c1a8536/lib/fluent/plugin/out_hash_forward.rb#L21).

So the `config_param` in the subclass is now magically ignored.

tagomoris:
fixed at #583

shyouhei:
:beers: awesome

PR #583, opened 2015-04-21
Refix subsection merging rule

tagomoris:
#582 shows that #571 breaks compatibility of existing 3rd party plugins.
This pull-request does:
* revert #582 at first
* add tests for past behavior
* add `final` option to specify sections not to be overwritten

3rd party plugins should be able to overwrite super-class's behavior. But some should prohibit overwriting by sub classes. Ex: `<buffer>`.
Plugin authors should specify `final: true` for these sections.

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -60,29 +60,16 @@ module Fluent
 
       def merge(other) # self is base class, other is subclass
         options = {
-          param_name: other.param_name, # param_name is used not to ovewrite plugin's instance varible, so this should be able to be overwritten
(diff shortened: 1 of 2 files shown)


tagomoris:
@sonots please review this changes

shyouhei:
Deployed this to our shadow servers.  Are working fluently so far.

repeatedly:
Test passed and it seems OK for me.
I assigned to @sonots for handling this problem.

sonots:
Looked codes carefully, and LGTM. Since it is working fine as shyouhei-san says, let me merge this.

sonots:
cherry-picked to v0.10

Issue #575, opened 2015-04-11
RPC for instance management


repeatedly:
We have a plan to add RPC feature to v0.14 but this is an independent from v0.14 features so it can be implemented at v0.12.
The merit is using RPC is better than signal on several environment, e.g. JRuby and Windows.

## Features

JSON + HTTP API is better because user can call APIs via command line tools like curl.

- Change fluentd state at runtime like `--without-source`
  - `POST /plugins/stop_sources`
- Send actions instead of singal. This is needed for JRuby / Windows support
  - `POST /buffers/flush` or `/signal/hup`? Former is readable.
- Integrate monitor_agent (needed?)
- misc

## Configuration

If set `rpc_port` in `<system>`, fluentd launches RPC server.

What do you think?

tagomoris:
All switches should be toggled, like below:
* `POST /plugins/sources/on` & `POST /plugins/sources/off` or so

Readable path is better, but path consistency is also important: `POST /plugins/buffers/flush`

sonots:
Just an idea, but what about to provide API interfaces for plugins liks

    POST /plugins/out_grep/:plugin_id
    {
       "tag":"foo"
    }

which enables post any json to plugin instance.

This would call new plugin API like

    def rpc(params)
    end

so that plugin authors freely define its behaviors.

repeatedly:
Implemented

PR #585, opened 2015-04-23
Implement RPC server for better instance management

repeatedly:
RPC server is launched when an user sets rpc_endpoint in system directive.
Each handler returns JSON response by default.

In the future, we use pipe between supervisor and workers.

Related to https://github.com/fluent/fluentd/issues/575 issue.


repeatedly:
This is prototype so some features, e.g. stop sources, are not implemented yet.

tagomoris:
I think that RPC handlers (mounted on webrick) should be separated from `supervisor.rb`.

repeatedly:
@tagomoris What do you mean? Mount handler code should be in rpc/handler.rb or similar?

tagomoris:
@repeatedly yes. `fluent/supervisor.rb` is too big for me now.

repeatedly:
@tagomoris I agree > too big.
How to separate code? Open class? module mixin?
For separating code, need to access internal state of Supervisor.

tagomoris:
How about to do `@rpc_server.supervisor = self` just before `@rpc_server.start`?

repeatedly:
@tagomoris RPC server is not only for supervisor so your approach also seems a bit tricky.
For separating code, we have several approaches in rpc/handlers/supervisor.rb:

```rb
module Fluent::RPC::Supervisor
  def setup_handlers(rpc_server, supervisor)
    # signal handlers are private
    rpc_server.mount_proc { |req, res|
      supervisor.__send__(:xxx)
    }
    # ...
  end
end
```

or

```
module Fluent
  class Supervisor
     # same as current patch
  end
end
```

or

```
module Fluent::RPC::Supervisor
  # include in actual supervisor code.
  module SetupHandlersMixin
     # ... setup code
  end
end
```

Separating code is good but it seems to bring another complexity for me.
Does someone have a good idea?

repeatedly:
Does someone have more feedback?
I want a feedback of RPC endpoints. `/plugins/buffers/flush` should be `/buffers/flush` or another?

repeatedly:
I use `rpc_endpoit` for launching RPC server.
This should be separated into rpc_bind / rpc_port with default values like `rpc_bind` is `localhost` and `rpc_port` is `23224`?

repeatedly:
I have a plan to introduce ServerEngine and it replaces some supervisor code with serverengine's part so
I delayed separating code.

repeatedly:
Another approach is Slack like HTTP RPC Style API: https://api.slack.com/methods

- `/api/plugins.flushBuffers`
- `/api/processes.killWorkers`
- `/api/processes.interruptWorkers`
- `/api/config.reload`

It seems better than REST like API for me.
What do you think?

sonots:
HTTP RPC Style looks nice :)

repeatedly:
No additional comment?
If nothing, I will merge this PR today.

tagomoris:
LGTM

repeatedly:
@sonots You can implement `/api/plugins.stopSources` :)

JindongChen:
Is there any plan to implement /api/plugins.stopSources? I'm looking forward to stop source when output to Elasticsearch is blocking.

PR #586, opened 2015-04-27
add filter_stdout

sonots:
This implementation is almost same with the implementation at https://github.com/fluent/fluentd/pull/307#issuecomment-68285384, but filter plugin not output plugin.

### DIFF
--- a/example/filter_stdout.conf
+++ b/example/filter_stdout.conf
@@ -1,0 +1,22 @@
+<source>
+  type dummy
+  tag dummy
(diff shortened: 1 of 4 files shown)


sonots:
I'm going to merge if there is not objection.

repeatedly:
It seems useful. No need copy and stdout trick :+1:

sonots:
Merged!

repeatedly:
Great!

PR #587, opened 2015-04-28
Added keep_time_key option to parsers

sonots:
This patch allows to use `keep_time_key` option in `in_tail` plugin. 

Now, writing config like:

```apache
<source>
  type tail
  path /var/log/nginx/access.log
  pos_file tmp/access.log.pos
  tag accesslog
  format /^(?<remote>[^ ]*) (?<host>[^ ]*) (?<remoteuser>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)" "(?<forwardedfor>[^"]*)")?$/
  time_format %d/%b/%Y:%H:%M:%S %z
  keep_time_key true
</source>
```

keeps the `time` field in records.

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -27,6 +27,8 @@ module Fluent
     # 'configure()' may raise errors for unexpected configurations
     attr_accessor :estimate_current_event
 
+    config_param :keep_time_key, :bool, :default => false
+
     def initialize
       super
       @estimate_current_event = true
@@ -201,6 +203,13 @@ module Fluent
             case name
             when "time"
               time = @mutex.synchronize { @time_parser.parse(value) }
+              if @keep_time_key
+                record[name] = if @type_converters.nil?
+                                 value
+                               else
+                                 convert_type(name, value)
+                               end
+              end
             else
               record[name] = if @type_converters.nil?
                                value
@@ -239,7 +248,8 @@ module Fluent
       def parse(text)
         record = Yajl.load(text)
 
-        if value = record.delete(@time_key)
+        value = @keep_time_key ? record[@time_key] : record.delete(@time_key)
+        if value
           if @time_format
             time = @mutex.synchronize { @time_parser.parse(value) }
           else
@@ -299,7 +309,7 @@ module Fluent
         record = Hash[keys.zip(values)]
 
         if @time_key
-          value = record.delete(@time_key)
+          value = @keep_time_key ? record[@time_key] : record.delete(@time_key)
           time = if value.nil?
                    if @estimate_current_event
                      Engine.now
@@ -465,6 +475,7 @@ module Fluent
           "referer" => referer,
           "agent" => agent,
         }
+        record["time"] = m['time'] if @keep_time_key
 
         if block_given?
           yield time, record
@@ -520,6 +531,7 @@ module Fluent
               record['pri'] = value.to_i
             when "time"
               time = @mutex.synchronize { @time_parser.parse(value.gsub(/ +/, ' ')) }
+              record[name] = value if @keep_time_key
             else
               record[name] = value
             end
--- a/test/test_parser.rb
+++ b/test/test_parser.rb
@@ -131,6 +131,31 @@ module ParserTest
         assert_nil time, "parser returns nil if configured so"
       }
     end
+
+    def test_parse_with_keep_time_key
+      parser = TextParser::RegexpParser.new(
+        Regexp.new(%q!(?<time>.*)!),
+        'time_format'=>"%d/%b/%Y:%H:%M:%S %z",
+        'keep_time_key'=>'true',
+      )
+      text = '28/Feb/2013:12:00:00 +0900'
+      parser.parse(text) do |time, record|
+        assert_equal text, record['time']
+      end
+    end
+
+    def test_parse_with_keep_time_key_with_typecast
+      parser = TextParser::RegexpParser.new(
+        Regexp.new(%q!(?<time>.*)!),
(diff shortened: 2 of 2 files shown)


sonots:
I am wondering that whether I should add the `time` field with `to_i`ed value or the original string value.
Currently, the original string is the default, and we can specify `include_time_key integer` to get `to_i` value. 

Any opinions?

kenjiskywalker:
Great pull request! Thanks sonots-san!

> Currently, the original string is the default, and we can specify include_time_key integer to get to_i value.

:+1: 

If  added option I going to use string(default).

sonots:
Decided

1. Change the option name `include_time_key` to `keep_time_key` avoid possible option name conflictions.
2. Drop the `integer` option. Now, `keep_time_key` is a bool option.

Changed the PR title.

sonots:
Added `time_key` option for RegexpParser to change the record key name.

repeatedly:
I still don't know time_key option in RegexpParser is useful or not
If someone wants to set time to any filed, record_transformer does.

In my fluentd experience, I didn't receive time_key like feature request.
So I can't imagine the useful situation.

sonots:
If we add keep_time_key, users may feel as they want to change the name of the time field. In such situation time_key option is useful.

I think keep_time_key is useful because I often am asked how to add time field in records. This option enables to do it without using any filter plugins.

I am rather thinking as the behavior of keep_time_key true can be a default, but to keep lower version compatibility, I am setting it as default false.

sonots:
Some lists of users' voides (I will add if I find more):

* http://qiita.com/rch850/items/3b7ce04e38c85a1ce5d0
* https://twitter.com/fujiwara/status/592602918756294656
* https://github.com/fluent/fluentd/pull/587#issuecomment-97014238

repeatedly:
I agree keep_time_key is useful but I don't agree time_key option is useful.

sonots:
Why? LTSVParser or other parsers has it. Only RegexpParser does not have it.

repeatedly:
Because RegexpParser can specify `(?<time>)` for log time but
other formats are not. LTSV, JSON and others are already parsed.

sonots:
I have to store the JSON key name as `logtime` to insert into my company's hadoop. This is my company's tradition. 

Using `keep_time_key and `(?<time>)`, the record key becomes `time` and it does not fit with the use case.

repeatedly:
It seems to be resolved by record_transformer filter.
Hmm... does someone want to use this feature?
If some users want, I will re-consider it.

sonots:
As you know, keep_time_key itself can be resolved with record-transformer.

The aim of this PR is to achieve the goal without using record-transformer or othe filters by providing a simple interface for users, especially beginners.

sonots:
Without time_key option for RegexpParser, some users including me still have to use record-transformer, I want to avoid such inconvenience.

repeatedly:
> As you know, keep_time_key itself can be resolved with record-transformer.

Yes but first parser behaviour is time key is included.
I'm not sure why time key is removed.
I think first behaviour is better so I agree keep_time_key is useful.

repeatedly:
BTW, including two features in one PR is wrong way.

sonots:
time_key for RegexpParser without keep_time_key is useless as you said. So, I moved it to this as I wrote.

sonots:
The aim of this PR is not to use record-transformer to add time field is records, so they should be considered together.

sonots:
By the way, can `keep_time_key true` be default by any chance? We may be able to remove this option itself in the future.

sonots:
https://groups.google.com/forum/#!topic/fluentd/U85Qjll2Y-o

This guy wants to treat `time` field as a normal field. To achieve this, `time_key foo` for RegexpParser is more suitable than `keep_time_key` option. Adding `time_key` option for RegexpParser should go.

repeatedly:
> To achieve this, time_key foo for RegexpParser is more suitable than keep_time_key option.

In his case, keep_time_key is enough because he says "How to keep the 'time' field of fluentd message" in thread title. He doesn't seem to want to change time key.

repeatedly:
> By the way, can keep_time_key true be default by any chance?

v1 or v2. Maybe, v1 is better.

sonots:
He is saying as

> I don't want it to be consider as time, I want to treat it as normal field

So, it looks he wants to skip `TIme.parse` itself. `keep_time_key` does not to allow to skip `Time.parse`. 

BTW, I remembered that I had conf files configuring as `(?<vtime>)` to skip `Time.parse`, and then rename to `time` for performance issue.

repeatedly:
> So, it looks he wants to skip TIme.parse itself.

We can think this because we know fluentd's internal deeply. But in user side, user doesn't know such action in general.
Parsing time is needed for fluentd event so it is no problem. His problem seems parsed time is removed from the record. So the best is keeping time with original string like `{"msg":"log message", "time":"May 4, 2015..."}`.

Skip parsing time causes time gap between log generated time and log collected time. Using log collected time is unexpected behaviour for almost users.

sonots:
summary: time_key has two effects:

1. to chang key field name on keep_time_key true
2. skip Time.parse itself

I think this option is worth to add.

sonots:
Reopened https://github.com/fluent/fluentd/pull/588, and moved the `time_key` option patch from here.

repeatedly:
Okay. Let's merge.

PR #588, opened 2015-04-29
Add time_key option for RegexpParser

sonots:
RegexpParser does not have `time_key` option currently, but it is useful if it has it. 

EDIT: `time_key` option is useful in two aspects:

１. We add `keep_time_key` option at https://github.com/fluent/fluentd/pull/587. However, without `time_key` option for RegexpParser, we still have to write a conf as follows when we need to change the time field name from `time` to `logtime` (or anything). 

```apache
<source>
  type tail
  format /^\[(?<time>[^\]]*)\]$/
  tag accesslog
</source>

<filter accesslog>
  type record_transformer
  enable_ruby true
  <record>
    logtime ${time.to_i}
  </record>
</filter>

<filter accesslog>
  type typecast
  types logtime:integer
</filter>

<match accesslog>
  type stdout
  #=> {"logtime":123456789}
</match>
```

By adding `time_key` option, this conf becomes as follows, which is pretty simple:

```apache
<source>
  type tail
  format /^\[(?<logtime>[^\]]*)\]$/
  tag accesslog
  time_key logtime
  types logtime:time
</source>

<match accesslog>
  type stdout
  #=> {"logtime":123456789}
</match>

```

２. `time_key` option allows to skip `Time.parse` (`Time.parse` in ruby is slow)

```apache
<source>
  type tail
  format /^\[(?<time>[^\]]*)\]$/
  time_key foo
  tag accesslog
</source>
```

The main purpose of this PR is 1, though.

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -160,6 +160,7 @@ module Fluent
     class RegexpParser < Parser
       include TypeConverter
 
+      config_param :time_key, :string, :default => 'time'
       config_param :time_format, :string, :default => nil
 
       def initialize(regexp, conf={})
@@ -199,7 +200,7 @@ module Fluent
         m.names.each {|name|
           if value = m[name]
             case name
-            when "time"
+            when @time_key
               time = @mutex.synchronize { @time_parser.parse(value) }
             else
               record[name] = if @type_converters.nil?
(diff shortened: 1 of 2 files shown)


repeatedly:
I can't image the useful situation.
Could you show me an actual example?
Your test is just not realistic.

sonots:
Ah, I will include this into #587. This becomes useful using together with #587

sonots:
Moved into #587

sonots:
Reopened, and wrote about use-cases.

sonots:
I will collect users' opinions for necessity about `time_key`.

kenjiskywalker:
I wonder why `time_key` was not necessary.

repeatedly:
I received some requests from users to adapt existing sysmtes without trick.
gogo~.

repeatedly:
Can we merge this without changes?

sonots:
I think so

PR #589, opened 2015-05-01
add filter_typecast plugin

sonots:
I created https://github.com/sonots/fluent-plugin-filter_typecast a few days ago, and it seems very useful. 

I want to propose to have filter_typecast plugin as a built-in plugin because this plugin is just using `Fluent::TextParser::TypeConverter` which Fluentd has for `in_tail` plugin. 

I will make obsolete the above my plugin if this PR is merged.

### DIFF
--- a/example/filter_typecast.conf
+++ b/example/filter_typecast.conf
@@ -1,0 +1,14 @@
+<source>
+  type dummy
+  tag dummy
+  dummy {"field1":"1","field2":"2","field3":"2013-02-12 22:04:14 UTC","field4":"true","field5":"a,b,c","other":"other"}
+</source>
+
+<filter **>
+  type typecast
+  types field1:integer,field2:string,field3:time,field4:bool,field5:array
+</filter>
+
+<match **>
+  type stdout
+</match>
--- a/lib/fluent/plugin/filter_typecast.rb
+++ b/lib/fluent/plugin/filter_typecast.rb
@@ -1,0 +1,14 @@
+module Fluent
+  class TypecastFilter < Filter
+    Fluent::Plugin.register_filter('typecast', self)
+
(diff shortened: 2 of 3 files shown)


repeatedly:
Hmm... does someone vote this?
I know some users used typecast with in_tail before.
But after supported types parameter in in_tail, the number of users are decreased.

Personally, I can't judge this plugin should be included as built-in or not...

sonots:
types is supported only by RegexpParser, and typecast should also be useful for record_transformer because all of its outputs become String.

toyama0919:
+1

sonots:
Oh, there was a vote!

example/filter_typecast.conf
```diff
@@ -0,0 +1,14 @@
+<source>
+  type dummy
```
tagomoris: use `@type`

example/filter_typecast.conf
```diff
@@ -0,0 +1,14 @@
...
+  tag dummy
+  dummy {"field1":"1","field2":"2","field3":"2013-02-12 22:04:14 UTC","field4":"true","field5":"a,b,c","other":"other"}
+</source>
+
+<filter **>
+  type typecast
```
tagomoris: ditto

example/filter_typecast.conf
```diff
@@ -0,0 +1,14 @@
...
+  types field1:integer,field2:string,field3:time,field4:bool,field5:array
+</filter>
+
+<match **>
+  type stdout
```
tagomoris: ditto

tagomoris:
Major usage of that plugin seems for string values from tail and record_transformer, but now these plugins can emit values of other types.
Third party plugins can depends on other third party plugins, so I'm going to reject this proposal.

Issue #590, opened 2015-05-06
Preceeding space gets added to key


dmfenton:
This bug was messing me up with trying to index some CSVs into Elasticsearch. Every time I would start indexing, what looked like duplicate mappings were being created.

The reason was that a set of keys like: "key1, key2, key3" was turned into "key1", " key2", " key3". This was really hard to ferret out. At a minimum there should be docs about this in big bold letters, or better yet, detect and ignore the spaces between keys.

Here's a GIST that shows the consequences of this bug:
https://gist.github.com/dmfenton/23197b0729b6735f47ab

repeatedly:
I heard this before but we keep this behaviour because few users go with this behaviour.
Yeah, this is error prone...

`,` separated approach is the legacy of old configuration format.
We have a plan to support array based parameter, `keys ["key1", "key2"]`.

dmfenton:
So maybe some big red letters in your documentation would be useful until there is a new configuration format available?

repeatedly:
I just added NOTE message to csv / tsv section.

http://docs.fluentd.org/articles/in_tail

I will add json array support later.

repeatedly:
PR are merged. v0.12.9 ship with json array parameter.

PR #592, opened 2015-05-13
Parser's keys parameter accept json array configuration.

repeatedly:
'keys ["key1", "key2"]' works. ',' separated value is also supported.

Related to https://github.com/fluent/fluentd/issues/590


repeatedly:
@sonots Could you check this?

sonots:
LGTM :ok_woman:

PR #593, opened 2015-05-14
Add read_lines_limit parameter to control chunk size in in_tail

repeatedly:
Related to https://groups.google.com/forum/#!topic/fluentd/-6Q8pPMU-Kc
back-pressure can't resolve this problem so adding option is better.

This limit is rough so the number of emit records succeeds `read_lines_limit`.
For example, set 100 but the number of emit records becomes 105. It depends on line length of log.

Edit: Change parameter name to read_lines_limit


repeatedly:
@sonots @tagomoris Do you have any concern to merge this PR? If no problem, I will merge this.

PR #597, opened 2015-05-19
Use none parser when json parser failed

Jackyjjc:
In the project my team are working on, we do not want to lose any logs in production simply because the logs are not in json format. We think it is good to preserve all the logs even when the parser fails.

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -239,6 +239,7 @@ module Fluent
       def configure(conf)
         super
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Maybe, multi-format-parser resolve this problem.
https://github.com/repeatedly/fluent-plugin-multi-format-parser

Jackyjjc:
@repeatedly I didn't know about that, thank you! That plugin seem to fit our needs.

PR #600, opened 2015-05-22
Add config_param type `enum` and its tests, Fix #599

### LINKED ISSUE #599: config parameter type `enum`
Enum -like configuration parameters are used in many places.
Now many plugin implement it with `config_param` with blocks, but it's troublesome.

This issue propose new `config_param` type `enum` like this:

```ruby
config_param :heartbeat_type, :enum, list: [:tcp, :udp], default: :tcp
```

This is easy to understand for both of developer and users, and reduces number of use of `config_param` with blocks, so it is helpful for #584.

tagomoris:
Type `enum` is to:
* handle symbols only for values
* require `list: [symbols]` parameter, which specifies the list of valid options
* raises `Fluent::ConfigError` if specified value in configuration file

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -67,6 +67,14 @@ module Fluent
     val
   })
 
+  Configurable.register_type(:enum, Proc.new { |val, opts|
+    s = val.to_sym
+    unless opts[:list].include?(s)
+      raise ConfigError, "valid options are #{opts[:list].join(',')} but got #{val}"
+    end
+    s
+  })
+
   Configurable.register_type(:integer, Proc.new { |val, opts|
     val.to_i
   })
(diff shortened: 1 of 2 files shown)


lib/fluent/config/types.rb
```diff
@@ -67,6 +67,14 @@ def self.bool_value(str)
     val
   })
 
+  Configurable.register_type(:enum, Proc.new { |val, opts|
+    s = val.to_sym
+    unless opts[:list].include?(s)
```
repeatedly: Maybe, checking `:list` is needed. Current code raises nil related exception without `:list` option.

tagomoris: But there're no point to check `opts` before here...

PR #603, opened 2015-05-28
use params when body is empty

wd:
As discussed at https://groups.google.com/forum/#!topic/fluentd/GCfGSP-rBpA, I simply create this PR. I don't master ruby and fluentd well, maybe someone can give a better solution :)

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -317,7 +317,7 @@ module Fluent
         params = WEBrick::HTTPUtils.parse_query(uri.query)
 
         if @format != 'default'
-          params[EVENT_RECORD_PARAMETER] = @body
+          params[EVENT_RECORD_PARAMETER] = @body.empty? ? params : @body
         elsif @content_type =~ /^application\/x-www-form-urlencoded/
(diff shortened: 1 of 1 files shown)


repeatedly:
Does it work on your environment?
`params` refers to `params`, so reference loop happens.

wd:
Haven't noticed before... but it's worked..

repeatedly:
Hmm... In your implementation, `content` value is `params` so it is a Hash object: https://github.com/fluent/fluentd/blob/1a95bde27cd7bae918ac03f3f45fc6542da8a72c/lib/fluent/plugin/in_http.rb#L196
But `Parser#parse` requires given argument is String object, so your patch seems to always raise an exeception: https://github.com/fluent/fluentd/blob/1a95bde27cd7bae918ac03f3f45fc6542da8a72c/lib/fluent/plugin/in_http.rb#L197
What the format do you use in your `in_http` configuration?

wd:
Oh, I use a custom parser... this commit seems buggy, I forget to test use default parser. Close, and will do more test.

repeatedly:
> I use a custom parser

Ah, I see. Okay, I wait new PR.

PR #604, opened 2015-05-29
Add secret option

repeatedly:
We sometimes got this feature request.
If plugins add `:secret => true` in `config_param`, the values are masked with `xxxxxx` like below.

```
2015-05-29 19:50:10 +0900 [info]: using configuration file: <ROOT>
  <source>
    type http
    port 8080
    format csv
  </source>
  <match http.**>
    type copy
    deep_copy xxxxxx # for test. deep_copy is not secret
    <store>
      type stdout
      output_type hash
    </store>
  </match>
</ROOT>
```

This changes two existing behaviours.
 
- in_monitor_agent returns masked configuration.
- Log order is changed during fluentd startup. Before: config dump -> register. After: register -> config dump

From my experience, these changes are no problem.

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -117,6 +117,14 @@ module Fluent
         out
       end
 
+      def to_masked_element
+        element = Element.new(@name, @arg, {}, @elements, @unused)
+        each_pair { |k, v|
+          element[k] = secret_param?(k) ? 'xxxxxx' : v
+        }
+        element
+      end
+
       def secret_param?(key)
         return false if @corresponding_proxies.empty?
 
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -29,9 +29,10 @@ module Fluent
         }
         @unused = unused || attrs.keys
         @v1_config = false
+        @corresponding_proxies = [] # some plugins use flat parameters, e.g. in_http doesn't provide <format> section for parser.
       end
 
-      attr_accessor :name, :arg, :elements, :unused, :v1_config
+      attr_accessor :name, :arg, :elements, :unused, :v1_config, :corresponding_proxies
 
       def add_element(name, arg='')
         e = Element.new(name, arg, {}, [])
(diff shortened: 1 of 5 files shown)


repeatedly:
@tagomoris @sonots @frsyuki How about this?

lib/fluent/config/element.rb
```diff
@@ -112,6 +117,28 @@ def to_s(nest = 0)
...
+        element
+      end
+
+      def secret_param?(key)
+        return false if @corresponding_proxies.empty?
+
+        param_key = key.to_sym
+        @corresponding_proxies.each { |proxy|
+          block, opts = proxy.params[param_key]
+          if opts
```
tagomoris: Should it be `opts && opts[:secret]`?
If opts has other option's key, this method returns opts[:secret] if it's not specified.

lib/fluent/engine.rb
```diff
@@ -90,12 +90,12 @@ def configure(conf)
         $log.info "gem '#{spec.name}' version '#{spec.version}'"
       end
 
+      @root_agent.configure(conf)
```
tagomoris: What for this diff is?

lib/fluent/config/element.rb
repeatedly: You mean following code is better?

```
if opts && opts.has_key?(:secret)
  return opts[:secret]
end
```

tagomoris: Ah, yes. Exactly.

tagomoris:
Almost ok.
Good feature to be merged :+1:

lib/fluent/engine.rb
repeatedly: We must call `configure` before configuration dump because
we need to link between element and configure_proxy.

tagomoris: got it.

repeatedly:
Are there is problem or counter opinion?
If so, I will merge this PR tomorrow.

lib/fluent/config/element.rb
```diff
@@ -29,9 +29,10 @@ def initialize(name, arg, attrs, elements, unused = nil)
         }
         @unused = unused || attrs.keys
         @v1_config = false
+        @corresponding_proxies = [] # some plugins use flat parameters, e.g. in_http doesn't provide <format> section for parser.
```
sonots: What does "corresponding" mean here? Corresponding with what??

sonots:
Commented.

lib/fluent/config/element.rb
repeatedly: configure proxies.

repeatedly:
Can I merge this?

sonots:
どうぞどうぞ

PR #610, opened 2015-06-09
Update upper msgpack bound

jamielennox:
msgpack 0.6.0 was released with a fix that allows for correct encoding
of binary messages. This is required by a plugin I'm working on.

There is no incompatibilities between the current 0.5 release and the
0.6 release.

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 1.9.3'
 
-  gem.add_runtime_dependency("msgpack", [">= 0.5.11", "< 0.6.0"])
+  gem.add_runtime_dependency("msgpack", [">= 0.5.11", "< 0.7.0"])
   gem.add_runtime_dependency("json", [">= 1.4.3"])
   gem.add_runtime_dependency("yajl-ruby", ["~> 1.0"])
   gem.add_runtime_dependency("cool.io", [">= 1.2.2", "< 2.0.0"])


repeatedly:
@tagomoris @frsyuki Can we merge this?

tagomoris:
It's not true that "There is no incompatibilities between ...", but anyway, Fluentd must do `force_encoding("ASCII-8BIT")` for all unpacking of msgpack binaries.
So, we can merge this pull-request if we don't miss to do it.

jamielennox:
@tagomoris for my own work why would we need to force_encoding here? My understanding of the change in msgpack was that it handles encoding to ascii-8bit for us. Anything not encoded as 8bit will not be affected by the update.

tagomoris:
@jamielennox exactly, in ideal. But if we missed to do `force_encoding` and someone handle data as string with utf-8 encoding information, it will break (because it will be ascii-8bit string w/ msgpack v0.6.0).

tagomoris:
I think that this change is to be done with major version up of Fluentd.
With this change, Fluentd `out_forward` sends binaries typed as `bin8` (and others). If any tools which supports Fluentd forward protocol doesn't support bin8/bin16/... types, it will break.

repeatedly:
v0.14 updates msgpack-ruby dependency to v0.7.0 for supporting extension type.

repeatedly:
fluentd v0.14.0.pre1 uses msgpack 0.7.0.
So I close this PR. Thanks for the report :)

PR #611, opened 2015-06-10
Add chunk format check in in_forward

repeatedly:
This is useful for which forwarder agent sends broken chunk to aggregators.


lib/fluent/plugin/in_forward.rb
```diff
@@ -125,7 +125,12 @@ def on_message(msg, chunk_size, source)
         return
       end
 
-      # TODO format error
+      # TODO: raise an exception if broken chunk is generated by recoverable situation
+      unless msg.is_a?(Array)
+        log.warn "incoming chunk is broken:", source: source, msg: msg
```
tagomoris: `msg` should be `inspect` -ed for various values, ex: strings with escape code or others.

repeatedly: `inspect` is done in log methods.

https://github.com/fluent/fluentd/blob/7bd529cc8007f04592c971f6b2fd9d216c60cf5d/lib/fluent/log.rb#L279

tagomoris: ah, okay.

repeatedly:
Are there another concern?
If not, I will merge the patch.

tagomoris:
LGTM

PR #615, opened 2015-06-12
Fix SingleValueFormatter to prevent record from modified

maxwellzdm:
when i use `out_copy` plugin to copy events to multiple outputs, I found out that even when I set the `deep_copy` to be true, original data would still  be modified.
The reason for such behavior is that when use `<<` to append new_line code to text, original record would also be modified.
So in order to keep original record unmodified, duplicate is need.


repeatedly:
Good catch!

PR #617, opened 2015-06-13
Support formatter plugins in out_stdout

takebayashi:
Make `output_type` option of `out_stdout` accept not only `json` or `hash` but also any other formatter plugin.


repeatedly:
Does someone see any concern?
Using Formatter is good for me.

sonots:
:+1:

tagomoris:
:+1:

PR #625, opened 2015-07-01
out_forward: Support dns round robin

sonots:
I want to support DNS round robin.

Ruby's `Socket.pack_sockaddr_in` or `Resolv.getaddress` or `IPSocket.getaddress`, and so on does not support DNS round robin. To do it, we have to `Resolv.getaddresses(hostname).shuffle!.first`. 

This code still works on non dns round robin environment, so I did not add any new options. 

EDIT: Writing tests is difficult because we need to set up a DNS round robin environment, but existing tests passed, and I verified this works manually.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -460,9 +460,8 @@ module Fluent
       end
 
       def resolve_dns!
-        @sockaddr = Socket.pack_sockaddr_in(@port, @host)
-        port, resolved_host = Socket.unpack_sockaddr_in(@sockaddr)
-        return resolved_host
+        # shuffle to support dns round robin
+        Resolv.getaddresses(@host).shuffle!.first
       end
       private :resolve_dns!
 
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -460,8 +460,8 @@ module Fluent
       end
 
       def resolve_dns!
-        # shuffle to support dns round robin
-        Resolv.getaddresses(@host).shuffle!.first
+        # sample to support dns round robin
+        Socket.getaddrinfo(@host, @port).sample[3]
       end
       private :resolve_dns!
 
--- a/example/in_forward.conf
+++ b/example/in_forward.conf
@@ -1,0 +1,7 @@
+<source>
+  type forward
+</source>
+
+<match test>
+  type stdout
+</match>
--- a/example/out_forward.conf
+++ b/example/out_forward.conf
@@ -1,5 +1,6 @@
 <source>
-  type forward
+  type dummy
+  tag test
 </source>
 
 <match test>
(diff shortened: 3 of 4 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -460,9 +460,8 @@ def resolved_host
       end
 
       def resolve_dns!
-        @sockaddr = Socket.pack_sockaddr_in(@port, @host)
-        port, resolved_host = Socket.unpack_sockaddr_in(@sockaddr)
-        return resolved_host
+        # shuffle to support dns round robin
+        Resolv.getaddresses(@host).shuffle!.first
```
kou: How about using `Array#sample` like `Resolv.getaddresses(@host).sample`? It doesn't generate a temporary `Array`.

nurse:
You can use [Array#sample](http://docs.ruby-lang.org/ja/1.9.3/method/Array/i/sample.html) for Ruby 1.9.3+. (note that current fluentd master is Ruby 1.9.3+)

Also Ruby has a getaddrinfo(3) wrapper, [Socket#getaddrinfo](http://docs.ruby-lang.org/ja/1.9.3/method/Socket/s/getaddrinfo.html).

Therefore you can write as `Socket.getaddrinfo(@host, @port).sample[3]`

lib/fluent/plugin/out_forward.rb
sonots: Cool, I will fix so.

sonots:
Thanks, using Socket#getaddrinfo is less expensive, I will use it.

sonots:
fixed

repeatedly:
Thanks for reviewing!
Ruby has several host to IP convert approaches so this is very helpful.

nurse:
Anyway as far as I understand, getaddrinfo(3) returns results by suitable order referring RFC6724. Therefore this feature shouldn't be default, and simply use IPSocket.getaddress seems correct.

Even though you want to do round robin for some reason ([I know DeNA uses it](http://www.atmarkit.co.jp/ad/dena/dena1004/)), `Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM).sample[3]` or `Socket.getaddrinfo(@host, @port, Socket::AF_INET, Socket::SOCK_STREAM).sample[3]` to reject IPv6 addresses.

repeatedly:
In this case, `Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM).sample[3] ` seems better for me.
I'm not sure we should specify address family. If so, we should use AF_INET6 for IPv6 address.

sonots:
Okay, committed > https://github.com/fluent/fluentd/commit/cbfa7438cd052c54b18b1fc3d0d3c28427b086bb

repeatedly:
Thanks!
If someone hit the IPv4/IPv6 problem, we will add AF_INET/AF_INET6 to getaddrinfo.

PR #626, opened 2015-07-01
out_forward: heartbeat_type none

sonots:
Added an option to disable the heartbeat. 

Using another round robin system such as [DNS round robin](https://github.com/fluent/fluentd/pull/625), HAProxy, or something, we typically want to do health check with another system. 

BTW: This is a kind of porting from [fluent-plugin-keep-forward](https://github.com/sonots/fluent-plugin-keep-forward)

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -43,8 +43,10 @@ module Fluent
         :tcp
       when 'udp'
         :udp
+      when 'none'
+        :none
       else
-        raise ConfigError, "forward output heartbeat type should be 'tcp' or 'udp'"
+        raise ConfigError, "forward output heartbeat type should be 'tcp', 'udp', or 'none'"
       end
     end
     config_param :heartbeat_interval, :time, :default => 1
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -43,8 +43,10 @@ def initialize
         :tcp
       when 'udp'
         :udp
+      when 'none'
+        :none
       else
-        raise ConfigError, "forward output heartbeat type should be 'tcp' or 'udp'"
+        raise ConfigError, "forward output heartbeat type should be 'tcp' or 'udp', or 'none'"
```
repeatedly: ```'tcp', 'udp' or 'none'``` ?

lib/fluent/plugin/out_forward.rb
```diff
@@ -123,32 +130,36 @@ def start
...
+        @loop.watchers.each {|w| w.detach }
+        @loop.stop
+      end
+      @thread.join if @thread
       @usock.close if @usock
     end
 
     def run
-      @loop.run
+      @loop.run if @loop
```
repeatedly: When `heartbeat_type none`, `run` is never called, right?

sonots: Right, this is just for safe and symmetry with `#shutdown` (everything is nil checked in shutdown now).

lib/fluent/plugin/out_forward.rb
sonots: fixed as `'tcp', 'udp', or 'none'`

lib/fluent/plugin/out_forward.rb
repeatedly: I see.

repeatedly:
It seems good for me.
Does someone have any problem?

tagomoris:
:+1:

PR #628, opened 2015-07-06
Fix secret option to support config_section

repeatedly:
Currently, config_section's secret is ignored.

@sonots Could you test this with your large configuration file? This increases memory usage.

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -118,11 +118,12 @@ module Fluent
       end
 
       def to_masked_element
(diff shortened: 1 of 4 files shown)


sonots:
I transfered from my previous department on April. @shyouhei succeeded the large configuration system, so he would be able to help us to test this.

repeatedly:
@sonots @shyouhei Ah, okay. If there is no problem, I wll merge this and release new version.

repeatedly:
Today, I will merge this patch.

PR #629, opened 2015-07-13
Log unused section configuration as warn level

repeatedly:

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -30,7 +30,7 @@ module Fluent
         @unused = unused || attrs.keys
         @v1_config = false
         @corresponding_proxies = [] # some plugins use flat parameters, e.g. in_http doesn't provide <format> section for parser.
-        @unused_in = false # if this element is not used in plugins, correspoing plugin name is set. otherwise nil.
+        @unused_in = false # if this element is not used in plugins, correspoing plugin name and parent element name is set, e.g. [source, plugin class].
       end
 
       attr_accessor :name, :arg, :elements, :unused, :v1_config, :corresponding_proxies, :unused_in
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -30,9 +30,10 @@ module Fluent
         @unused = unused || attrs.keys
         @v1_config = false
         @corresponding_proxies = [] # some plugins use flat parameters, e.g. in_http doesn't provide <format> section for parser.
+        @unused_in = false # if this element is not used in plugins, correspoing plugin name is set. otherwise nil.
       end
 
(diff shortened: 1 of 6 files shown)


lib/fluent/config/element.rb
```diff
@@ -30,9 +30,10 @@ def initialize(name, arg, attrs, elements, unused = nil)
         @unused = unused || attrs.keys
         @v1_config = false
         @corresponding_proxies = [] # some plugins use flat parameters, e.g. in_http doesn't provide <format> section for parser.
+        @unused_in = false # if this element is not used in plugins, correspoing plugin name is set. otherwise nil.
```
tagomoris: default value `false` is against to comment.

repeatedly: other wise refers to "if this element is not used". I will add more line.

lib/fluent/engine.rb
```diff
@@ -76,6 +76,10 @@ def parse_config(io, fname, basepath = Dir.pwd, v1_config = false)
     def run_configure(conf)
       configure(conf)
       conf.check_not_fetched { |key, e|
+        if plugin_name = e.unused_in
+          $log.warn "section <#{e.name}> is not used in #{plugin_name}."
```
tagomoris: IMO, this warning should contain parent section name, ex: `is not used in <section> of #{plugin_name}`, `.. in <server> of #{plugin_name}`.

repeatedly:
Apply review. Please check again.

@sonots Could you check this too?

repeatedly:
Here is logged example:

```
2015-07-13 17:54:45 +0900 [info]: using configuration file: <ROOT>
  <source>
    type secure_forward
    self_hostname server.fqdn.example.com
    shared_key secret_string
    secure no
    private_key_passphrase xxxxxx
    <user>
      username test
      password test_pass
      <foo>
        a b
      </foo>
    </user>
    <test>
      foo bar
    </test>
  </source>
  <match debug.**>
    type stdout
    <test>
      c d
    </test>
  </match>
  <fil>
    type foo
  </fil>
  <matc foo.**>
    type bar
  </matc>
</ROOT>
2015-07-13 17:54:45 +0900 [warn]: section <foo> is not used in <user> of secure_forward plugin
2015-07-13 17:54:45 +0900 [warn]: section <test> is not used in <source> of secure_forward plugin
2015-07-13 17:54:45 +0900 [warn]: section <test> is not used in <match debug.**>
2015-07-13 17:54:45 +0900 [warn]: section <fil> is not used in <ROOT>
2015-07-13 17:54:45 +0900 [warn]: section <matc> is not used in <ROOT>
```

repeatedly:
If no problem, I will merge it today.

repeatedly:
If you ok, please merge this.

PR #631, opened 2015-07-15
Fix UDP heartbeat regression caused by #625

repeatedly:
@sonots Please check this patch ASAP.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -472,7 +472,9 @@ module Fluent
 
       def resolve_dns!
         # sample to support dns round robin
-        Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM).sample[3]
+        addrinfo = Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM).sample
+        @sockaddr = Socket.pack_sockaddr_in(addrinfo[1], addrinfo[3])
+        addrinfo[3]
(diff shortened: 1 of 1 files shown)


sonots:
Oh, so `@sockaddr` was required. Thanks.

sonots:
It is okay to merge this, but I need to examine more. I have to check DNS round robin with UDP heartbeat works well, then.

sonots:
It looked working fine on my development environment, but theoretically, situations like following would happen?

1. Resolve host, get ipaddress of A. Send message to A
2. On timer, Send UDP heartbeat to A
3. Resolve host, get ipaddress of B, Send message to B
4. Receive UDP heatbeat response from A, But node.packaddr points to B now, so heartbeat check fails

heartbeat `tcp` must be used with DNS round robin.

sonots:
Sent another PR

repeatedly:
Non DNS round-robin environment, does 3 and 4 happen?
getaddrinfo seems to return only one addrinfo.

Anyway, adding dns_round_robin opton is safer.

sonots:
> Non DNS round-robin environment, does 3 and 4 happen?

I believe this would happen even on Non DNS round-robin environment at the moment DNS cache is expired. 
I think that the original code which memoizes `@sockaddr` like this is making things difficult, but improving it requires refactoring of out_forward / in_forward.

repeatedly:
@sonots I see. Okay.

PR #632, opened 2015-07-15
dns_round_robin option

sonots:
See https://github.com/fluent/fluentd/pull/631. I added `dns_round_robin` option to prevent using dns round robin with udp heartbeat.

### DIFF
--- a/lib/fluent/config/section.rb
+++ b/lib/fluent/config/section.rb
@@ -18,6 +18,7 @@ require 'json'
 
 module Fluent
   require 'fluent/config/error'
+  require 'fluent/config/v1_parser'
 
   module Config
     class Section < BasicObject
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -62,6 +62,7 @@ module Fluent
     config_param :ack_response_timeout, :time, :default => 190  # 0 means do not wait for ack responses
     # Linux default tcp_syn_retries is 5 (in many environment)
     # 3 + 6 + 12 + 24 + 48 + 96 -> 189 (sec)
+    config_param :dns_round_robin, :bool, :default => false # heartbeat_type 'udp' is not available for this
 
     attr_reader :nodes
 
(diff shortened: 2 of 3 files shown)


lib/fluent/config/section.rb
```diff
@@ -18,6 +18,7 @@
 
 module Fluent
   require 'fluent/config/error'
+  require 'fluent/config/v1_parser'
```
sonots: I got error `uninitialized constant Fluent::Config::V1Parser`  on running `ruby test/plugin/test_out_forward.rb`, so fixed.

lib/fluent/plugin/out_forward.rb
```diff
@@ -62,6 +62,7 @@ def initialize
     config_param :ack_response_timeout, :time, :default => 190  # 0 means do not wait for ack responses
     # Linux default tcp_syn_retries is 5 (in many environment)
     # 3 + 6 + 12 + 24 + 48 + 96 -> 189 (sec)
+    config_param :dns_round_robin,      :bool, :default => false # heartbeat_type 'udp' is not available for this
```
repeatedly: No need additional space between `:dns_round_robin` and `:bool`.

lib/fluent/plugin/out_forward.rb
```diff
@@ -471,8 +479,13 @@ def resolved_host
       end
 
       def resolve_dns!
-        # sample to support dns round robin
-        Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM).sample[3]
+        if @conf.dns_round_robin
```
repeatedly:
```rb
addrinfo_list = Socket.getaddrinfo(@host, @port, nil, Socket::SOCK_STREAM)
addrinfo = @conf.dns_round_robin ? addrinfo_list.sample : addrinfo.first
```

or similar code is better for reducing same function call code.

sonots: fixed

lib/fluent/plugin/out_forward.rb
sonots: fixed

repeatedly:
@sonots Could you send a PR to add `dns_round_robin` to fluentd-docs?

sonots:
Sure, i'll do it later

Issue #520, opened 2014-12-17
Remove in_status plugin


repeatedly:
`in_status` is un-documented plugin.
In addition, we have well-behaved `in_monitor_agent` so `in_status` is no longer needed.

I think we can remove `in_status` from core plugins.

PR #635, opened 2015-07-22
Warn deprecation message for in_status plugin

repeatedly:
Related to https://github.com/fluent/fluentd/issues/520


tagomoris:
:+1:

Issue #504, opened 2014-12-10
support the compressing stream in forward and secure forward


haroboro:
It's to save the bandwidth across the data centers, also accelerate the transfer speed and save the time.

repeatedly:
We will consider supporting compression in v0.14's new APIs: https://github.com/fluent/fluentd/issues/309#issuecomment-83390980

cswilliams:
+1, this would be a great feature to help reduce our bandwidth charges on AWS

deitch:
Yes, I was just looking for this. You would need to support it in `out_forward` as well, so the receiver can decompress it.

I have a use case where the compressed logs are ~10% of the uncompressed logs. For LAN, that doesn't matter, but for shipping inter-DC or to the cloud, that is a huge deal.

+1

repeatedly:
@deitch Do you use Gzip or Snappy like compression?

deitch:
Don't actually care. We have gotten similar results with gzip and plain zip. But as long as in_forward / in_secure_forward and out_forward both support it, I can put a fluentd on each host to collect and forward, and have another cluster of them in the cloud, and everyone is happy.

tsailiming:
+1

repeatedly:
Does someone try this patch?

https://github.com/fluent/fluentd/pull/637

Manikantan22:
I am looking for a option to compress a message during out_forward. Could someone help me?

tagomoris:
refer #1179

PR #637, opened 2015-07-27
Support compression in forward transfer

repeatedly:
Experimental implementation.

Adding `compress true` in out_forward configuration.

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -20,6 +20,8 @@ module Fluent
 
     def initialize
       super
+      require 'zlib'
+      require 'stringio'
       require 'fluent/plugin/socket_util'
     end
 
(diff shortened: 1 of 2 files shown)


dhiltonp:
Can we get feedback on this PR? Is it going to be merged? Is there a reason for it to *not* be merged? Is there something that needs to be fixed in this code before it is accepted?

Thanks!

repeatedly:
We need feedback on several enviroment for checking performance and compression ratio with actual logs.
I post it on #504 but there is no feedback so I'm pending the merge.

repeatedly:
BTW, now this branch conflicts with master.
I will update the patch later.

repeatedly:
rebased for testing.

tagomoris:
At least, this change doesn't include tests.

repeatedly:
Yes because we need test with production traffic.
If this change is useful, I will add tests.

Manikantan22:
Do we still have "compress true" for out_forward plugin?

I am using td-agent2 and tried to use compress true but getting below error in logs.

[warn]: parameter 'compress' in <match pattern> is not used.

repeatedly:
compress doesn't support yet in the upstream.

tagomoris:
This feature will be supported with compressed buffer. Pending.

repeatedly:
https://github.com/fluent/fluentd/pull/1179 is more better way. I close this PR.

PR #640, opened 2015-07-29
WIP - DO NOT MERGE - Supports RFC 5425 and RFC 6587 msg-len fields

huntc:
In order to support long messages, and messages containing the full range of UTF-8 messages a message length may be conveyed at the beginning of a message. As defined by the following RFCs:

  https://tools.ietf.org/html/rfc5425
  https://tools.ietf.org/html/rfc6587

The enhancement changes the behaviour of fluentd's TcpHandler to accept a function that determines whether the tcp message has been completely read. In the absence of a message length field in the message, the original behaviour of using a delimiter ("\n") is used. Otherwise the message length is honoured.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ test/config/tmp/*
 make_dist.sh
 Gemfile.local
 .ruby-version
+.idea
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -164,7 +164,20 @@ module Fluent
       log.error_backtrace
     end
 
-    private
+    def test(buffer, offset)
+      # If there's a <pri> then we will assume that a newline terminates the syslog message.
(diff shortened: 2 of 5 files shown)


repeatedly:
In your PR, it doesn't consider RFC5424's [foo bar="baz"] structured message?

repeatedly:
ping. How about the progress of this PR?

huntc:
Sorry @repeatedly - I decided on another route whereby we communicate directly with ES. It became important to have a back-pressure mechanism so that ES signalled the message provider when it could no longer process messages; otherwise there is a danger of losing messages. We also send messages once per second, or every 5MB depending on which occurs first.

I'll close out this issue, but hopefully this feedback is useful.

repeatedly:
@huntc Interesting. What the tool do you use for back-pressure?

huntc:
We home grew our own solution using akka-streams which is an implementation of reactive-streams.

PR #657, opened 2015-08-17
Add null_value_pattern and null_empty_string option for LabeledTSVParser, CSVParser, TSVParser.

toyama0919:
Hello.

Cannot express NULL in LTSV.

ltsv library for Ruby blank to NULL.
https://github.com/condor/ltsv/blob/master/lib/ltsv.rb#L97

I use LTSV with apache.
Apache log is If value not exists with hyphen('-').

I think hyphens to null, any of the following methods.

1. LabeledTSVParser add parameter with @null_value
※this PR

settings example.
```
<source>
  type tail
  path /var/log/httpd/access_log
  format ltsv
  time_key time
  null_value -
  time_format %d/%b/%Y:%H:%M:%S %z
  tag td.apache.access
  pos_file /var/log/td-agent/apache_access.pos
</source>
```

2. specifications similar to the Ruby library(blank to null)

3. Make your own filter plugin.

Which one is good?

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -295,6 +295,8 @@ module Fluent
       end
       config_param :time_key, :string, :default => nil
       config_param :time_format, :string, :default => nil
+      config_param :null_value_pattern, :string, :default => nil
+      config_param :null_empty_string, :bool, :default => false
 
       def configure(conf)
         super
@@ -308,11 +310,16 @@ module Fluent
         end
 
         @time_parser = TimeParser.new(@time_format)
+
+        if @null_value_pattern
+          @null_value_pattern = /#{@null_value_pattern}/
+        end
+
         @mutex = Mutex.new
       end
 
       def values_map(values)
-        record = Hash[keys.zip(values)]
+        record = Hash[keys.zip(values.map { |value| convert_value_to_nil(value) })]
 
         if @time_key
           value = @keep_time_key ? record[@time_key] : record.delete(@time_key)
@@ -345,6 +352,16 @@ module Fluent
           end
         }
       end
+
+      def convert_value_to_nil(value)
+        if @null_empty_string
+          value = (value == '') ? nil : value
+        end
+        if @null_value_pattern
+          value = (value =~ @null_value_pattern) ? nil : value
+        end
+        value
+      end
     end
 
     class TSVParser < ValuesParser
@@ -367,8 +384,6 @@ module Fluent
     class LabeledTSVParser < ValuesParser
       config_param :delimiter,       :string, :default => "\t"
       config_param :label_delimiter, :string, :default =>  ":"
-      config_param :null_value_pattern, :string, :default => nil
-      config_param :null_empty_string, :bool, :default => false
       config_param :time_key, :string, :default =>  "time"
 
       def configure(conf)
@@ -383,12 +398,6 @@ module Fluent
         text.split(delimiter).each do |pair|
           key, value = pair.split(label_delimiter, 2)
           @keys.push(key)
-          if null_empty_string
-            value = (value == '') ? nil : value
-          end
-          if null_value_pattern
-            value = (value =~ /#{null_value_pattern}/) ? nil : value
-          end
           values.push(value)
         end
 
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -354,10 +354,10 @@ module Fluent
       end
 
       def convert_value_to_nil(value)
-        if @null_empty_string
+        if value and @null_empty_string
           value = (value == '') ? nil : value
         end
-        if @null_value_pattern
+        if value and @null_value_pattern
           value = (@null_value_pattern.match(value)) ? nil : value
         end
         value
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -367,7 +367,8 @@ module Fluent
     class LabeledTSVParser < ValuesParser
       config_param :delimiter,       :string, :default => "\t"
       config_param :label_delimiter, :string, :default =>  ":"
-      config_param :null_value, :string, :default => nil
+      config_param :null_value_pattern, :string, :default => nil
+      config_param :null_empty_string, :bool, :default => false
       config_param :time_key, :string, :default =>  "time"
 
       def configure(conf)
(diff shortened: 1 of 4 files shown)


frsyuki:
I think 1. is good and it is good enough for now.


I may want to use using 1. and 2. together occasionally. So, alternative idea is to add `null_value_pattern: pattern` option instead of `null_value` to accept a regexp, or add `null_empty_string: true` option in addition to `null_value`.

frsyuki:
I think this PR is simple enough and good to merge. Any concerns, @repeatedly?

toyama0919:
@frsyuki
Thank you for your reply!

>I may want to use using 1. and 2. together occasionally. So, alternative idea is to add null_value_pattern: pattern option instead of null_value to accept a regexp, or add null_empty_string: true option in addition to null_value.

I see.

Please wait a moment...

toyama0919:
Thank you for advice!

I add null_value_pattern and null_empty_string option for LabeledTSVParser.

settings example.
```
<source>
  type tail
  path /var/log/httpd/access_log
  format ltsv
  time_key time
  null_value_pattern ^(-|null|NULL)$
  null_empty_string true
  time_format %d/%b/%Y:%H:%M:%S %z
  tag td.apache.access
  pos_file /var/log/td-agent/apache_access.pos
</source>
```

## null_value_pattern
* default is nil.
* regular expression syntax.

## null_empty_string
* default is false.
* empty string to null.
* [this](https://github.com/condor/ltsv/blob/master/lib/ltsv.rb#L97) is same behavior.

lib/fluent/parser.rb
```diff
@@ -381,6 +383,12 @@ def parse(text)
         text.split(delimiter).each do |pair|
           key, value = pair.split(label_delimiter, 2)
           @keys.push(key)
+          if null_empty_string
+            value = (value == '') ? nil : value
+          end
+          if null_value_pattern
+            value = (value =~ /#{null_value_pattern}/) ? nil : value
```
repeatedly: Creating regexp object at each record is high cost.
It should be done in configure method.

sonots:
Why don't we add these options for super class (ValuesParser)?

lib/fluent/parser.rb
toyama-h: thank you!
I'll fix it.

toyama-h:
@sonots
thank you!
Oh, ValuesParser definition, affects CSVParser and TSVParser.
Null in regular expression definition, I think Good idea!
But empty to NULL casting is LTSV specification.

@repeatedly
I think no problem, Is it OK?

toyama-h:
Certainly, in the format like below with MySQL restore

```
"1";"2";"3";"4";NULL
```

sonots:
> empty to NULL casting is LTSV specification.

Is it official LTSV specification? I thought it is just a custom.

toyama0919:
>Is it official LTSV specification? I thought it is just a custom.

That's right!
No such process in other languages.

### java
https://github.com/making/ltsv4j/blob/master/src/main/java/am/ik/ltsv4j/LTSVParser.java

### python
https://github.com/hekyou/python-ltsv/blob/master/ltsv/_reader.py

toyama0919:
>Why don't we add these options for super class (ValuesParser)?
I'll fix fix in this way.
Please wait a moment.

toyama-h:
fixed.
add null_value_pattern and null_empty_string option, support format is tsv,csv,ltsv.

lib/fluent/parser.rb
```diff
@@ -308,11 +310,16 @@ def configure(conf)
         end
 
         @time_parser = TimeParser.new(@time_format)
+
+        if @null_value_pattern
+          @null_value_pattern = /#{@null_value_pattern}/
```
sonots: `Regexp.new` is better because users do not need to care of escaping `/` with it.

lib/fluent/parser.rb
```diff
@@ -345,6 +352,16 @@ def convert_field_type!(record)
           end
         }
       end
+
+      def convert_value_to_nil(value)
+        if @null_empty_string
+          value = (value == '') ? nil : value
+        end
+        if @null_value_pattern
```
sonots: `if value and @null_value_pattern` is more clear to tell returning nil when `nil =~ @null_value_pattern`

lib/fluent/parser.rb
```diff
@@ -345,6 +352,16 @@ def convert_field_type!(record)
...
+          value = (value =~ @null_value_pattern) ? nil : value
```
sonots: String scrub is needed? cf. http://blog.livedoor.jp/sonots/archives/34702351.html

sonots:
reviewed

lib/fluent/parser.rb
sonots: I usually does String#scrub in my plugin as https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/filter_grep.rb#L73-L83. 
I am not sure for the case of fluentd itself. What do you think? > @repeatedly

toyama-h:
thank you for review.
I 'll fix it.

lib/fluent/parser.rb
toyama0919: sorry.

I think this private method add ValuesParser, Is it OK?

```
      def match(regexp, string)
        begin
          return regexp.match(string)
        rescue ArgumentError => e
          raise e unless e.message.index("invalid byte sequence in".freeze).zero?
          log.info "invalid byte sequence is replaced in `#{string}`"
          string = string.scrub('?')
          retry
        end
        return true
      end

```

toyama0919:
Apply this review.
scrub Referenced filter-grep-plugin.

thanks.

lib/fluent/parser.rb
sonots: Creating Fluent::StringUtil should be better.

toyama0919: Ah OK!
thanks.

toyama0919:
added Fluent::StringUtil, filter-grep-plugin and this feature common procesing.
thanks.

sonots:
Looks good to me. Any concerns? > @repeatedly

repeatedly:
Nothing.

sonots:
Merged!

toyama-h:
@repeatedly @sonots @frsyuki thank you for many advise! :blush:

repeatedly:
Released v0.12.16!

toyama-h:
:blush:

tagomoris:
I found that `log` is not defined in `Fluent::StringUtil.match_regexp`. So this rescue block will raise NameError every time.

toyama0919:
@tagomoris Sorry and Thanks. I will check it.

Issue #584, opened 2015-04-22
Request feature to get options of fluent-plugin


toru-takahashi:
### Descriptions
Each fluent-plugin has a lot of option.
Almost users need to choice these options for own use-case.
But, almost user don't know these options. 
Therefore, it would be useful for users to show all options of fluent-plugin.

### Expected Results
Ex.
 
```
> fluentd --show fluent-plugin-s3
    config_param :path, :string, :default => ""
    config_param :use_ssl, :bool, :default => true
    config_param :use_server_side_encryption, :string, :default => nil
    config_param :aws_key_id, :string, :default => nil
    config_param :aws_sec_key, :string, :default => nil
    config_param :s3_bucket, :string
    config_param :s3_region, :string, :default => nil
    config_param :s3_endpoint, :string, :default => nil
    config_param :s3_object_key_format, :string, :default => "%{path}%{time_slice}_%{index}.%{file_extension}"
    config_param :store_as, :string, :default => "gzip"
    config_param :auto_create_bucket, :bool, :default => true
    config_param :check_apikey_on_start, :bool, :default => true
    config_param :proxy_uri, :string, :default => nil
    config_param :reduced_redundancy, :bool, :default => false
    config_param :format, :string, :default => 'out_file'
```

kazegusuri:
I like this feature. I and my colleagues often spend time checking how to use plugins by reading upstream source code.
If possible (with self documented api and for supported plugins), it's better to show some examples together.

tagomoris:
I have some idea to realize this feature:
* load plugin
* build configure_proxy tree
* show it on console

Problem is:
* `config_param` should accept `description` option to describe what it is
* default can be overwritten in `configure()` method in non-declarative way
* type may not be decided in declarative way with `config_param` w/ blocks

Some of these problems may be able to be solved by v0.14 plugin APIs.
Related: #309

repeatedly:
Adding `desc` to `config_param` could be implemented in v0.12 but it needs extra cost...

okkez:
I'm trying to implement this feature.

Work like following, for now.

```text
$ bundle exec fluentd --show-plugin-config output:s3
2015-08-26 13:59:43 +0900 [info]: Show config for output:s3
2015-08-26 13:59:43 +0900 [info]: 
path: string: <"">
use_server_side_encryption: string: <nil>
aws_key_id: string: <nil>
aws_sec_key: string: <nil>
s3_bucket: string: <nil>
[...]

PR #663, opened 2015-08-26
Add show plugin config

okkez:
See #584


repeatedly:
@toru-takahashi @kazegusuri How about this patch?
Do you want more information?

toru-takahashi:
It's enough for me.

Issue #633, opened 2015-07-16
Dump configuration in memory by signal or rpc


tagomoris:
Long running fluentd process may have configuration, which is different from configuration file on local filesystem.
Something to dump configuration tree in memory may help to shoot troubles.
* Adding RPC method
* Signal?

cosmo0920:
I try to implement this feature in this branch: https://github.com/fluent/fluentd/compare/master...cosmo0920:implement-dump-in-memory-configure

Questions: 

1. Which signal is suitable for? For now, I used `USR2` signal for this purpose.
2. How do we test? Run fluentd and send signal?

cosmo0920:
I'm not sure about signal related unit test....

tagomoris:
@cosmo0920 

1. IMO, signals are very precious to use, and I'm not sure about this feature is enough important to use one of rest of available signals for fluentd. It seems better to implement only as a RPC handler.
2. RPC handlers are tested in current Fluentd code, right? Signal handlers may not be tested, but there's no problem if we don't use signals.

Adding signal handlers later is not so hard. So I think it's good to add RPC handlers only at first.

cosmo0920:
1. OK, thanks.
2. I see. I'll remove signal handler implementation.

> Adding signal handlers later is not so hard. So I think it's good to add RPC handlers only at first.

I'll send PR which contains only rpc handler ASAP.

tagomoris:
Cool!

repeatedly:
PR is merged

PR #666, opened 2015-08-28
Implement RPC to dump in memory config

cosmo0920:
Related to #633.

Add `<endpoint>/api/config.dump` to RPC.

e.g)

example conf:

```conf
<source>
  type http
  bind 0.0.0.0
  port 9880
  body_size_limit 32MB
  keepalive_timeout 10
  # backlog 0
  add_http_headers false
  format default
</source>

<match test>
  type stdout
</match>
<system>
  rpc_endpoint 127.0.0.1:9000
</system>
```

### Dump API

```bash
$ curl http://127.0.0.1:9000//api/config.dump
{"ok":true}
```

then, got:

```log
2015-08-28 15:45:28 +0900 [info]: dump in-memory config
2015-08-28 15:45:28 +0900 [info]: <ROOT>
  <source>
    type http
    bind 0.0.0.0
    port 9880
    body_size_limit 32MB
    keepalive_timeout 10
    add_http_headers false
    format default
  </source>
  <match test>
    type stdout
  </match>
  <system>
    rpc_endpoint 127.0.0.1:9000
  </system>
</ROOT>
```

### via HTTP

```bash
$ curl http://127.0.0.1:9000/api/config.getDump
{"conf":"<ROOT>\n  <source>\n    type http\n    bind 0.0.0.0\n    port 9880\n    body_size_limit 32MB\n    keepalive_timeout 10\n    add_http_headers false\n    format default\n  </source>\n  <match test>\n    type stdout\n  </match>\n  <system>\n    rpc_endpoint 127.0.0.1:9000\n  </system>\n</ROOT>\n","ok":true}
```

### DIFF
--- a/lib/fluent/rpc.rb
+++ b/lib/fluent/rpc.rb
@@ -41,7 +41,7 @@ module Fluent
       def mount_proc(path, &block)
         @server.mount_proc(path) { |req, res|
           begin
-            code, header, body = block.call(req, res)
+            code, header, response = block.call(req, res)
           rescue => e
             @log.warn "failed to handle RPC request", :path => path, :error => e.to_s
             @log.warn_backtrace e.backtrace
@@ -56,11 +56,11 @@ module Fluent
 
           code = 200 if code.nil?
           header = {'Content-Type' => 'application/json'} if header.nil?
-          body = if body.nil?
+          body = if response.nil?
                    '{"ok":true}'
                  else
-                   body['ok'] = code == 200
-                   body.to_json
+                   response.body['ok'] = code == 200
+                   response.body.to_json
                  end
 
           res.status = code
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -263,6 +263,12 @@ module Fluent
         supervisor_sighup_handler
         nil
       }
+      @rpc_server.mount_proc('/api/config.dump') { |req, res|
+        $log.debug "fluentd RPC got /api/config.dump request"
+        $log.info "dump in-memory config"
+        supervisor_dump_config_handler
+        nil
+      }
     end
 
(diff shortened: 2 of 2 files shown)


sonots:
Do we really need two ways to dump?

cosmo0920:
I'm not sure for now.

I think that dump in-memory config in log is really needed.
But, I think that via HTTP version needs to pay attention to implement. (detail as below)

At least, both ways makes easy to dump in-memory configurations.
Two ways to dump has different advantages and disadvantages.

## config.dump

### Advantage

* Display in-memory config with pretty print

### Disadvantage

* Need to login or open terminal where fluentd running

## config.getDump

### Advantage

* Be able to get dump without login or open terminal where fluentd running

### Disadvantage

* Too easy to see in-memory config when someone know Fluentd RPC endpoints (maybe violate security???)

tagomoris:
In my opinion, we can monitor Fluentd configuration from remote point continuously with `config.getDump`. It's helpful to find Fluentd node/process which are running with old/outdated configuration.

If all plugins uses `:secret` feature to mask sensitive values, there seems no security risk. But right now, we can't expect it.
I think that `config.getDump` should be disabled in default, and it's better to be enabled by command line option or `<system>` configuration. Fluentd v0.14 may be able to enable it in default.

cosmo0920:
> I think that config.getDump should be disabled in default, and it's better to be enabled by command line option or <system> configuration. Fluentd v0.14 may be able to enable it in default.

:+1:
I think so, too.
I want to toggle enable/disable it by `<system>` configuration.

repeatedly:
@sonots @tagomoris If you are Okay, please merge this PR.

sonots:
:+1:

repeatedly:
Can we merge this?

PR #673, opened 2015-09-15
Fix add_remote_addr and add_http_headers options not working on bulk request

tarokkk:
This small patch solve bulk request parsing. Previously the code tried to insert a string into a list
which caused "no implicit conversion of String into Integer".

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -132,7 +132,7 @@ module Fluent
         end
 
         unless record.is_a?(Array)
-          if @add_http_headers and 
+          if @add_http_headers
             params.each_pair { |k,v|
               if k.start_with?("HTTP_")
                 record[k] = v
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -131,18 +131,18 @@ module Fluent
           end
         end
 
-        if @add_http_headers
-          params.each_pair { |k,v|
-            if k.start_with?("HTTP_")
-              record[k] = v
-            end
-          }
+        if not record.is_a?(Array)
+          if @add_http_headers and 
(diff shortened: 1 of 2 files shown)


Tyrael:
@repeatedly any chance you could review this PR?

lib/fluent/plugin/in_http.rb
```diff
@@ -131,18 +131,18 @@ def on_request(path_info, params)
...
-          params.each_pair { |k,v|
-            if k.start_with?("HTTP_")
-              record[k] = v
-            end
-          }
-        end
-
-        if @add_remote_addr
-          record['REMOTE_ADDR'] = params['REMOTE_ADDR']
+        if not record.is_a?(Array)
```
repeatedly: In fluentd, we use `unless` instead of `if not`.

repeatedly:
Sorry for the delay.

Please add test to `test/plugin/test_in_http.rb`.

tarokkk:
@repeatedly I fixed the code based on your comments and added tests for the headers.

lib/fluent/plugin/in_http.rb
```diff
@@ -131,18 +131,18 @@ def on_request(path_info, params)
...
-            if k.start_with?("HTTP_")
-              record[k] = v
-            end
-          }
-        end
-
-        if @add_remote_addr
-          record['REMOTE_ADDR'] = params['REMOTE_ADDR']
+        unless record.is_a?(Array)
+          if @add_http_headers and
```
repeatedly: `and` is needed?

repeatedly:
Commented. Others are OK :+1:

repeatedly:
Thanks for the work! I will release new version soon.

repeatedly:
Released v0.12.16!

PR #674, opened 2015-09-16
Add Windows support 

nurse:
To use this see http://qiita.com/nurse/items/8bbb194982c685f1e795


repeatedly:
@nurse Tests are failed. Could you check it?

sonots:
Why don't you add CI for windows? http://www.appveyor.com/

cosmo0920:
:+1: Add CI for Windows for me.

test/plugin/test_buf_file.rb
```diff
@@ -28,6 +26,7 @@ def filebufferchunk(key, unique, opts={})
     end
 
     def test_init
+      omit "Windows doesn't spport symlink" if RUBY_PLATFORM.include?('mswin')
```
cosmo0920: It should be use `omit "Windows doesn't spport symlink" if $platformwin` here.

cosmo0920:
Cross compiled msgpack 0.5.12 seems to be broken for 32bit Windows platform.
Should we use 0.6.12 (current stable) or 0.7.0dev1?

lib/fluent/plugin/buf_file.rb
```diff
@@ -65,8 +66,20 @@ def open(&block)
     end
 
     def mv(path)
-      File.rename(@path, path)
-      @path = path
+      if defined?(Windows)
+        pos = @file.pos
```
cosmo0920: It seems curious indentations....

test/plugin/test_in_stream.rb
```diff
@@ -122,4 +122,4 @@ def test_configure
   def connect
     UNIXSocket.new("#{TMP_DIR}/unix")
   end
-end
+end unless /mswin|mingw/ =~ RUBY_PLATFORM
```
cosmo0920: It seems to be able to use `$platformwin` here.

test/plugin/test_out_file.rb
```diff
@@ -232,6 +232,7 @@ def test_write_with_append
   end
 
   def test_write_with_symlink
+    omit "Windows doesn't spport symlink" if $platformwin
```
cosmo0920: typo

```diff
-spport
+support
```

test/plugin/test_buf_file.rb
```diff
@@ -28,6 +26,7 @@ def filebufferchunk(key, unique, opts={})
     end
 
     def test_init
+      omit "Windows doesn't spport symlink" if $platformwin
```
cosmo0920: also typo

```diff
-spport
+support
```

lib/fluent/supervisor.rb
```diff
@@ -16,6 +16,14 @@
 
 require 'fluent/load'
 require 'etc'
+if $platformwin
```
cosmo0920: `$platformwin` is not initialized here. So it always evaluated to `false`.

nurse: Defined $platformwin in fluent/load.rb.

lib/fluent/plugin/buf_file.rb
nurse: fixed!

nurse:
> Cross compiled msgpack 0.5.12 seems to be broken for Ruby 2.0/2.1 32bit Windows platform.
Should we use 0.6.12 (current stable) or 0.7.0dev1?
Simply 0.5.12 doesn't have 32bit binary gem.

cosmo0920:
>> Cross compiled msgpack 0.5.12 seems to be broken for Ruby 2.0/2.1 32bit Windows platform.
>> Should we use 0.6.12 (current stable) or 0.7.0dev1?

> Simply 0.5.12 doesn't have 32bit binary gem.

Oh..., that's too bad.

BTW, string-scrub gem also doesn't provide fat gem.
For now, Should we use DevKit for Windows CI?

nurse:
Remaining failures are because of Ruby's bug which is fixed on latest Ruby 2.2.3/2.1.7.
https://github.com/appveyor/ci/issues/368

Gemfile
```diff
@@ -2,6 +2,9 @@ source 'https://rubygems.org/'
 
 gemspec
 
+gem 'test-unit', github: 'nurse/test-unit', branch: 'code-snippet-fetcher-set_encoding'
+gem 'cool.io', github: 'nurse/cool.io', branch: 'windows-select-fix'
```
repeatedly: Release cool.io v1.4.1 including this fix. Could you remove this line?

lib/fluent/plugin/in_tail.rb
```diff
@@ -161,7 +162,7 @@ def stop_watchers(paths, immediate = false, unwatched = false)
         if tw
           tw.unwatched = unwatched
           if immediate
-            close_watcher(tw, false)
+            close_watcher(tw, true)
```
repeatedly: Why change this value?

lib/fluent/command/fluentd.rb
```diff
@@ -164,4 +188,48 @@
   exit 0
 end
 
+  FLUENTD_WINSVC_NAME="fluentdwinsvc"
```
repeatedly: These lines should be in inside `if winsvcinstmode = opts[:regwinsvc]` ?

lib/fluent/load.rb
```diff
@@ -15,6 +15,15 @@
   # ignore setup error on Win or similar platform which doesn't support signal
 end
 require 'cool.io'
+
+$platformwin = /mswin|mingw/ === RUBY_PLATFORM
+if $platformwin
+  require_relative 'winfile'
+else
+  module Fluent
+    Win32File = File
```
repeatedly: How about using `FileWrapper` or something name?
Using `Win32File` in everywhere seems to misread `this code doesn't work on Linux?`

sodabrew:
AppVeyor has just deployed Ruby 2.1.7 / 2.2.3 today, so you can re-run the CI now.

nurse:
@sodabrew thanks for info.
now CI becomes green!

repeatedly:
@tagomoris @sonots @frsyuki Check this PR. I have a plan to merge this PR and master branch will be for v0.14 development, e.g. nanosecond, SocketManager, ServerEngine.

lib/fluent/load.rb
```diff
@@ -15,6 +15,8 @@
   # ignore setup error on Win or similar platform which doesn't support signal
 end
 require 'cool.io'
+
+$platformwin = /mswin|mingw/ === RUBY_PLATFORM
```
repeatedly: From sada's comment on serverengine, how about using this approach like `Fluent.on_windows?` ? > https://github.com/fluent/serverengine/pull/26#discussion_r40830182

We have fluent/env.rb so this file seems good place.

nurse: changed

lib/fluent/plugin/buf_file.rb
```diff
@@ -65,8 +66,20 @@ def open(&block)
     end
 
     def mv(path)
-      File.rename(@path, path)
-      @path = path
+      if defined?(Windows)
```
repeatedly: `Fluent.windows?` is better?

lib/fluent/plugin/file_wrapper.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    def ino
+      by_handle_file_information = '\0'*(4+8+8+8+4+4+4+4+4+4)   #72bytes
+
+      unless GetFileInformationByHandle.call(@file_handle, by_handle_file_information)
+        return 0
+      end
+
+      fileindex = by_handle_file_information.unpack("I11Q1")[11]
+      return fileindex
```
repeatedly: `fileindex` and `return fileindex` seem to no needed. Only `by_handle_file_information.unpack("I11Q1")[11]` is enough.

lib/fluent/plugin/in_tail.rb
```diff
@@ -494,9 +500,9 @@ def on_notify
             begin
               while true
                 if @buffer.empty?
-                  @io.read_nonblock(2048, @buffer)
+                  @io.readpartial(2048, @buffer)
```
repeatedly: Does `readpartial` behave like `read_nonblock` on linux?

repeatedly:
Commented. The correctness of windows related code relies on patch!

lib/fluent/plugin/in_tail.rb
nurse: Almost yes, and in this case the difference is not an issue (or better than original).

#### read_nonblock(nbytes) behaves:
* If Ruby's internal IO buffer has chars, read (0, nbytes] bytes from it, and return it.
* Otherwise call read(2) **with** setting the io as O_NONBLOCK.
  * If it can read string, return it.
  * Otherwise, raise an exception Errno::EAGAIN or Errno::EWOULDBLOCK

#### read_partial behaves
* If Ruby's internal IO buffer has chars, read (0, nbytes] bytes from it, and return it.
* Otherwise call read(2) **without** setting the io as O_NONBLOCK.
  * If it can read string, return it even if the length is less than nbytes.
  * Otherwise, wait until it gets at least one byte char or EOF.

#### read(nbytes)
* If Ruby's internal IO buffer has at least 2048 chars, read and return it.
* Otherwise read rest of bytes with read(2) with blocking.

### In this case
The code is in infinite loop with reading `@io`. Therefore using read_nonblock without something blocking call like select may consume extra CPU. IO#readpartial realize such efficient logic without struggling IO.select. (Moreover read_nonblock may raise EWOULDBLOCK if there's enough bytes. I don't know why previous code works well. (Won't happen such case in current use case?))

repeatedly: > (Won't happen such case in current use case?)

In in_tail case, `read_nonblock` always raises EOFError. Is this unexpected with `read_nonblock` or ok with local file?

repeatedly:
@tagomoris @frsyuki @sonots Are there any concerns? I want to merge this PR before merging nano-second support.

lib/fluent/plugin/in_tail.rb
frsyuki: At least on Mac OS X and Linux., read_nonblock raises EOFError if IO is for a local file and the file reaches end-of-file. It doesn't raise Errno::EAGAIN or Errno::EWOULDBLOCK. Although it seems better to rescue Errno::EAGAIN and Errno::EWOULDBLOCK as well as EOFError.

If we aware of it, current `IOHandler#on_notify` is not a infinite loop because it returns immediately when file has no extra bytes.

`IOHandler#on_notify` is called by TimerWatcher (trigger using timer) as well as StatWatcher (trigger using Coolio::StatWatcher). When on_notify is called by TimerWatcher, file may have 0 bytes.

So, doesn't this change cause issue when 2 or more TimerWatcher is registered and one of them is blocked at IO#readpartial?

lib/fluent/command/fluentd.rb
```diff
@@ -115,6 +126,19 @@
...
+}
+
+op.on('--reg-winsvc MODE', "install/uninstall as Windows Service. (i: install, u: uninstall) (Windows only)") {|s|
+  opts[:regwinsvc] = s
+}
+
+op.on('--reg-winsvc-fluentdopt OPTION', "specify fluentd option paramters for Windows Service. (Windows only)") {|s|
+  opts[:fluentdopt] = s
+}
+
```
frsyuki: How about showing above windows-only options only when `Fluentd.windows?` is true?

frsyuki:
Great job :+1:
I added 2 comments.

lib/fluent/plugin/in_tail.rb
nurse: Ah, sure, in this case input is always a file and it reaches EOF; thank you for explanation.

> When on_notify is called by TimerWatcher, file may have 0 bytes.

When readpartial for 0 byte file or already reached EOF, it raises EOFError.
So it won't block.

repeatedly:
If there is no concerns, I will merge this PR tomorrow!

serialoverflow:
Hi,

the docs lead me to believe that fluentd does not support Windows, see here:
http://docs.fluentd.org/articles/windows

If that is wrong, you might want to correct the documentation. If it's true, then i don't understand this pull request.
Would you mind clarifying exactly how i can use fluentd on Windows?

tagomoris:
That document is about Fluentd v0.12. Fluentd v0.14 (and master branch) supports Windows environment.
Documents about Fluentd v0.14 is under construction... there's not enough resource to do it :(

repeatedly:
I will add windows installation article on fluentd.org, not docs.fluentd.org.

wolfgangihloff:
@repeatedly I do not see any details on documentation yet? All I find as of now is https://docs.fluentd.org/v0.12/articles/windows

PR #683, opened 2015-10-16
Improve describing plugin parameters

okkez:
We can write long description easily.


repeatedly:
Good :+1:

okkez:
Thank you!!

PR #687, opened 2015-10-26
Don't use popular name to prevent field overwrite when enable_ruby is true

repeatedly:
@sonots please check this patch.

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -230,18 +230,17 @@ module Fluent
         @placeholders = struct
       end
 
-      def expand(str, force_stringify=false)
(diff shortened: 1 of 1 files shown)


sonots:
Ah, so local variables are still available even we do `instance_eval { binding }`. 
Is there no way to off local variable scope?

sonots:
It looks better to add some random hex like `6c8047b87c44f301eeae6e522a42d059` than adding `_`.

repeatedly:
I asked @nurse and he said there is no good way for now.
We will consider to add pre-eval based formatter so this patch is enough for temporary solution.

repeatedly:
@sonots if you okay, please merge this.

sonots:
Hmm, no good, but okay to merge...

PR #693, opened 2015-10-29
Add Engine#msgpack_factory for v0.14 compatibility

repeatedly:
In v0.14, core and plugin use `Engine#msgpack_factory` for creating MessagePack Packer / Unpacker.
For supporting v0.12 and v0.14 in plugin code, providing same API is better.

Related to https://github.com/fluent/fluentd/pull/653

### DIFF
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -19,6 +19,16 @@ module Fluent
   require 'fluent/root_agent'
 
   class EngineClass
+    class DummyMessagePackFactory
+      def packer(io = nil)
+        MessagePack::Packer.new(io)
+      end
+
+      def unpacker(io = nil)
+        MessagePack::Unpacker.new(io)
+      end
+    end
+
     def initialize
       @root_agent = nil
       @event_router = nil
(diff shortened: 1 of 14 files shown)


repeatedly:
@frsyuki @tagomoris @sonots Could you check this PR? If no problem, I will merge it and release fluentd v0.12.17.

lib/fluent/engine.rb
```diff
@@ -19,6 +19,16 @@ module Fluent
   require 'fluent/root_agent'
 
   class EngineClass
+    class DummyMessagePackFactory
+      def packer(io = nil)
+        MessagePack::Packer.new(io)
+      end
+
+      def unpacker(io = nil)
+        MessagePack::Unpacker.new(io)
```
frsyuki: I think packer and unpacker methods should be:

```
def packer(*args)
  MessagePack::Packer.new(*args)
end
```

because Packer and Unpacker accepts some options such as `allow_unknown_ext: true` in addition to io.

frsyuki:
I commented one thing. Other than that, looks good to me. Good job.

lib/fluent/engine.rb
repeatedly: Ah, I see. I will update the patch. Thanks

repeatedly:
Applied. Does someone have any concern?

repeatedly:
if there are no any concerns, I will merge this today.

Issue #699, opened 2015-11-09
Will in_tail input loss messages when router emitting fail


zheolong:
In in_tail.rb, the final emitting codes are as follows.

``` ruby
  227     def receive_lines(lines, tail_watcher)
  228       es = @receive_handler.call(lines, tail_watcher)
  229       unless es.empty?
  230         tag = if @tag_prefix || @tag_suffix
  231                 @tag_prefix + tail_watcher.tag + @tag_suffix
  232               else
  233                 @tag
  234               end
  235         begin
  236           router.emit_stream(tag, es)
  237         rescue
  238           # ignore errors. Engine shows logs and backtraces.
  239         end
  240       end
  241     end
```

I wonder if router emitting fail will cause messages to be lost.

repeatedly:
`emit` fail is very rare case, e.g. plugin has bug or no error handling in output plugin.
If you want to catch such error records in `emit`, you can use `<label @ERROR>`.

```
<label @ERROR>
  <match foo.**>
    @type file
    path /path/to/backup_records
  </match>
</label
```

repeatedly:
Latest patch: https://github.com/fluent/fluentd/pull/810

Issue #567, opened 2015-03-27
[PROPOSAL] back pressure-like flow control for in/out_forward protocol


tagomoris:
Using opts (possibly v0.12)

repeatedly:
It means in_forward returns own busy status to out_forward and out_forward coordinates buffer handling?

tagomoris:
I don't have clear answer what we can do about this feature.
If buffer plugins has any APIs to explain its status, we can use it.

lam-juice:
it means telling input plugins to not ACK / update pos file when the buffer is reaching its limits?

lance-edmodo:
A possible way: create a mutex class and allow plugins to refer to it.

Usually, there are many plugins bug only a few places that need to implement backpressure. Under this approach, in the configuration for an input or output plugin you would explicitly refer to the mutex. Otherwise, plugins work as they do now.

donato:
Is this a thing? Due to the us-east-1 outage from AWS today, we are experiencing some headaches with td-agent.

Td-agent reads as fast as it can, even though we can't write output to S3. The buffer fills until we run out of memory and td-agent gets reapered. If td-agent would use backpressure to not read past a certain level until the buffer is being flushed, then this wouldn't occur.

fragatina:
+1

seungyongshim:
+1

rverma-jm:
Much required feature

PR #700, opened 2015-11-09
Fix in_tail input messages loss when receive_lines fail

zheolong:
Fix https://github.com/fluent/fluentd/issues/699

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -236,8 +236,11 @@ module Fluent
           router.emit_stream(tag, es)
         rescue
           # ignore errors. Engine shows logs and backtraces.
+          return false
         end
       end
+
+      return true
     end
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks for the patch!
I will check it later.

repeatedly:
This problem should be fixed by https://github.com/fluent/fluentd/issues/567
But this patch seems acceptable. I will test it.

BTW, test code is needed.

zheolong:
@repeatedly Ok, I will check the code and add test for it.

repeatedly:
Hmm writing test seems difficult.
I will try it.

repeatedly:
This will be included in next v0.12.20.

repeatedly:
I will try to write this test.

repeatedly:
@zheolong Update your patch on #810. Could you check it?

repeatedly:
#810 includes this patch.

PR #701, opened 2015-11-10
Add descriptions to out_file plugin parameters

cosmo0920:
I tried to add descriptions in out_file plugin.

Related to #584.

### DIFF
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -23,11 +23,11 @@ module Fluent
       'gzip' => :gz,
     }
 
-    desc "Specify filepath to output."
+    desc "The Path of the file."
     config_param :path, :string
-    desc "Choose output format. You can choose it from registered formatter plugins."
+    desc "The format of the file content. The default is out_file."
     config_param :format, :string, :default => 'out_file'
-    desc "Disable path increment."
+    desc "The flushed chunk is appended to existence file or not."
     config_param :append, :bool, :default => false
     desc "Compress flushed file."
     config_param :compress, :default => nil do |val|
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_file.rb
```diff
@@ -23,9 +23,13 @@ class FileOutput < TimeSlicedOutput
       'gzip' => :gz,
     }
 
+    desc "Specify filepath to output."
     config_param :path, :string
+    desc "Choose output format. You can choose it from registered formatter plugins."
     config_param :format, :string, :default => 'out_file'
+    desc "Disable path increment."
     config_param :append, :bool, :default => false
+    desc "Compress chunks with specified compression algorithm."
```
sonots: Should be

    Compress output file

lib/fluent/plugin/out_file.rb
```diff
@@ -33,6 +37,7 @@ class FileOutput < TimeSlicedOutput
       end
       c
     end
+    desc "A parameter is used for pointing to latest written file."
     config_param :symlink_path, :string, :default => nil
```
sonots: Should be like

    Create symlink to temporary buffered file when buffer_type is file.

sonots:
Could you use same words with doc? http://docs.fluentd.org/articles/out_file

lib/fluent/plugin/out_file.rb
cosmo0920: Document says "Compresses flushed files using gzip.".

Should I use this one or your suggestion?

sonots: No. For this, docs should be fixed (not always gzip)

cosmo0920: Hmm..., OK I'll fix this line with your suggestion.

sonots: Ah, let me write more descriptive. 

Currently, `out_file` supports only gzip, so `Compresses flushed files using gzip` is not wrong, but I think 

    Compress flushed file with specified compression algorithm.

is fine. In addition, documents should list what algorithms can be specified.

PR #702, opened 2015-11-13
Add ParserTestDriver

cosmo0920:
Currently, fluentd does not have `ParserTestDriver`.
This PR adds this test driver.

### Remaining tasks

- [x] Frozen ParserTestDriver API
  - [x] ~~Decide whether to include utility methods such as `str2time` or not.~~
- [x] Replace a few parser test cases with `ParserTestDriver`
- [ ] Documentations?

### DIFF
--- a/lib/fluent/test.rb
+++ b/lib/fluent/test.rb
@@ -20,5 +20,6 @@ require 'fluent/test/base'
 require 'fluent/test/input_test'
 require 'fluent/test/output_test'
 require 'fluent/test/filter_test'
+require 'fluent/test/parser_test'
 
 $log ||= Fluent::Log.new(Fluent::Test::DummyLogDevice.new)
--- a/lib/fluent/test/parser_test.rb
+++ b/lib/fluent/test/parser_test.rb
@@ -34,7 +34,7 @@ module Fluent
 
       attr_reader :instance, :config
 
-      def configure(conf, use_v1 = false)
+      def configure(conf)
         if conf.is_a?(Fluent::Config::Element)
           @config = conf
         else
--- a/lib/fluent/test/parser_test.rb
+++ b/lib/fluent/test/parser_test.rb
@@ -17,7 +17,7 @@
 module Fluent
   module Test
     class ParserTestDriver
-      def initialize(klass_or_str, format=nil, conf={},  &block)
+      def initialize(klass_or_str, format=nil, conf={}, &block)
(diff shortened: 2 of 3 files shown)


repeatedly:
Could you replace one actual parser test using this driver?

cosmo0920:
I try to replace `TimeParser` and `SyslogParse` test cases with `ParserTestDriver`.

cosmo0920:
I've found that parser test driver in this commit https://github.com/cosmo0920/fluentd/commit/d5d0846e8af444cbddbb3e3a03e6df19b7875ef0 does not handle to instantiate `RegexParser`.
How do we abstract this parser with this test driver?

okkez:
I want to write parser test like following:

```ruby
def test_multiline
  conf = %[
  <grok>
    pattern %{DATA:message1}\n
  </grok>
  <grok>
    pattern %{DATA:message2}\n
  </grok>
  ]
  text = <<TEXT
host1 message1
  message2
TEXT
  d = Fluent::Parser::TestDriver.new(Fluent::TextParser::MultilineGrokParser).configure(conf)
  d.parse(text) do |time, record|
     assert_equal(expected, record)
  end
end
```

This is same as other test drivers.

repeatedly:
TimeParser itself is not a Parser plugin so I think supporting TimeParser API is not needed.

cosmo0920:
Rebased. I removed a commit which replaces TimeParse test cases with ParserTestDriver.

repeatedly:
@okkez How about this? If you ok, I will merge this patch.

okkez:
How to use nested config_section and cnofig_param?
How about this?

But this is no tested.

```ruby
def configure(conf)
  case conf
  when Fluent::Config::Element
    @config = conf
  when String
    io = StringIO.new(conf)
    @config = Cofig::Parser.parse(io, 'fluent.conf')
  when Hash
    @config = Config::Element.new('ROOT', '', conf, [])
  else
    raise "Unknown type... #{conf}"
  end
  @instance.configure(@config)
  self
end
```

cosmo0920:
As a result of @okkez and I discussion, I added commits to drop `TimeParser` support in this test driver, fix usage of `Class.new`, and string literal support in `#configure`.

Could you review this PR again?

repeatedly:
If there are no concerns from other developers, I will merge it.

okkez:
Looks good to me :+1:

PR #704, opened 2015-11-18
Support built-in text parsers in in_exec. fix #698

### LINKED ISSUE #698: in_exec does not support regex?
Excuse me if I have misunderstood the correct way to achieve this (the documentation is kind of vague on this point) - but as far as I am given to understand the only way to get the output of a command (short of writing it to a temp file and tailing the temp file) would be using the `in_exec` plugin .e.g 

```
<source>
  type exec
  command heroku "logs -p postgres -t --app my_app"
</source>
```

Unfortunately from the looks of it the `in_exec` plugin does not support regex formatting (https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_exec.rb#L34) as does the [tail plugin where it is documented](http://docs.fluentd.org/articles/parser-plugin-overview).

I'm not sure why the in_exec plugin has it's own custom formatting limitations rather than using already existing support? Or am I on the wrong track here?

repeatedly:
It seems useful.

### DIFF
--- a/lib/fluent/plugin/exec_util.rb
+++ b/lib/fluent/plugin/exec_util.rb
@@ -16,6 +16,8 @@
 
 module Fluent
   module ExecUtil
+    require 'fluent/parser'
+
     SUPPORTED_FORMAT = {
       'tsv' => :tsv,
       'json' => :json,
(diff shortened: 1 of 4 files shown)


repeatedly:
@pik Could you try this patch?

sonots:
Can you support built-in formatter for out_exec and out_exec_filter similarly?

repeatedly:
I'm not sure. If someone want it, I will consider it.

PR #707, opened 2015-11-20
out_forward raises ZeroDivisionError when no <server> is available

frsyuki:
When no <server> is set to out_forward plugin,
`Fluent::ForwardOutput#rebuild_weight_array` raises ZeroDivisionError
at `coe = (regular_nodes.size * 6) / weight_array.size`. This change
gives more useful error message to uses instead of unfriendly
ZeroDivisionError message.


repeatedly:
Could you add a test for this change?

sonots:
Added test

repeatedly:
@sonots good :+1:

PR #708, opened 2015-11-20
Add process_name parameter in system to change fluentd's process name

repeatedly:
Like https://github.com/cubicdaiya/fluent-plugin-setproctitle plugin.
Set `process_name`,

```aconf
<system>
  process_name foo
</system>
```

ps result is below.

```
% ps aux | grep foo
repeatedly      45673   0.4  0.2  2523252  38620 s001  S+    7:04AM   0:00.44 worker:foo
repeatedly      45647   0.0  0.1  2481260  23700 s001  S+    7:04AM   0:00.40 supervisor:foo
```

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -121,6 +121,7 @@ module Fluent
       @chgroup = opt[:chgroup]
       @chuser = opt[:chuser]
       @rpc_server = nil
+      @process_name = nil
 
       @log_level = opt[:log_level]
       @suppress_interval = opt[:suppress_interval]
@@ -325,6 +326,7 @@ module Fluent
     def supervise(&block)
       start_time = Time.now
 
+      Process.setproctitle("supervisor:#{@process_name}") if @process_name
       $log.info "starting fluentd-#{Fluent::VERSION}"
 
       if !Fluent.windows?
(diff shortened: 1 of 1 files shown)


tagomoris:
:+1:

sonots:
I can imagine that someone requests that he wants to set entire proctitle. How about providing supervisor process name, and worker process options?

Also, should note how it behaves on Windows (maybe, on docs)

cubicdaiya:
:+1:

repeatedly:
@sonots 

> I can imagine that someone requests that he wants to set entire proctitle.

Hmm... any situation? I sometimes received this feature request to better fluentd process management because current process name is bit messy.
Current patch identifies supervisor and worker using a prefix. Could you show me the good example?

sonots:
It just depends on users' preference. He may want to name as `fluentd worker` rather than `worker:fluentd`.

repeatedly:
If it is `may`, I think we don't need to support two parameters for now.

repeatedly:
My first implementation was `fluentd-supervisor` and `fluentd-worker`.
I'm not sure which is better.

repeatedly:
No against for `worker` and `supervisor` prefix, I will merge it.

PR #709, opened 2015-11-21
Add chunk_key_type option with buffered_time support for TimeSlicedOutput

sonots:
Originally, chunk keys of TimeSlicedOutput are generated based on event times.
This option allows us to genearte chunk keys based on buffering time,
i.e., TimeSlicedOutput#emit.

USE CASE:
1) fluent-plugin-s3

out_s3 allows us to use chunk key for s3 object path. But, with event-time
based chunk keys, out_s3 would put objects to old dated directories if
dalayed events arrive. This makes us difficult to decide when we should collect
s3 objects from a dated directory.

With `chunk_key_type buffered_time`, chunk keys are generated based on
buffering time (that is, TimeSlicedOutput#emit). This assures out_s3
does not put objects to old dated directories, thus, we can easily
decide when we should collect s3 objects.

cf. https://twitter.com/sonots/status/667568964575236096

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -481,16 +481,7 @@ module Fluent
 
     config_param :time_slice_format, :string, :default => '%Y%m%d'
     config_param :time_slice_wait, :time, :default => 10*60
-    config_param :chunk_key_type, :default => :event_time do |val|
-      case val.downcase
-      when 'event_time'
-        :event_time
-      when 'buffered_time'
-        :buffered_time
-      else
-        raise ConfigError, "chunk_key_type should be 'event_time', or 'buffered_time'"
-      end
-    end
+    config_param :chunk_key_type, :string, :default => 'event_time'
     config_param :timezone, :string, :default => nil
     config_set_default :buffer_type, 'file'  # overwrite default buffer_type
     config_set_default :buffer_chunk_limit, 256*1024*1024  # overwrite default buffer_chunk_limit
@@ -550,6 +541,13 @@ module Fluent
           }
         end
       end
+
+      @chunk_key_type = @chunk_key_type.downcase.to_sym
+      supported_types = [:event_time]
+      supported_types << :buffered_time if self.class.support_chunk_key_type_buffered_time?
+      unless supported_types.include?(@chunk_key_type)
+        raise ConfigError, "chunk_key_type should be one of #{supported_types}"
+      end
     end
 
     def emit(tag, es, chain)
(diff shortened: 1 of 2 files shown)


sonots:
BTW, I started to use power-assert.

sonots:
@repeatedly @tagomoris any concerns?

tagomoris:
Why don't you create s3 output plugin based on BufferedOutput? It's just BufferedOutput that arrival time decides buffer chunks to be written.

sonots:
Why do we have to create and maintain exactly same plugin except chunk key behaviors? Maintaining two similar plugins is nonsense. Providing an option is clearly better.

sonots:
And, TimeSlicedOutput is more suitable for out_s3 because we want to put objects with time slice window, with naming s3 object path based on the time sliced chunk key.

tagomoris:
"clearly better" is not clear for me.
This change breaks consistency that: "chunks from TimeSlicedOutput plugins contains records with "time" in the range of chunk key". It's impossible to guarantee that this change doesn't break existing TimeSliceOutput plugins' output behavior.
Will you notify all plugin developers to check this option work well or break compatibilities (and to add *DO NOT use `chunk_key_type buffered_time` in README?)

sonots:
In my understanding, this change does not break any existing TimeSlicedBuffer plugins. If you can show me an example, I can consider about it and take care.

tagomoris:
A fictional plugin for example to calculate values for specified time range:
* plugin store all events in a chunk for specified time range (a minute)
* at flush time, calculate summarized value from all records in a chunk

This plugin will produce wrong value for a time range by past records in a chunk if `chunk_key_time buffered_time` specified.

Plugin authors should not write such code, but can write it. IMO, Fluentd should not break compatibilities even with such cases.

tagomoris:
IMHO, acceptable solution is:
* `chunk_key_time buffered_time` options is enabled only when plugin explicitly allow it
* "allow it" can be implemented by implementing a method like `emit_with_buffered_time` or something else

sonots:
It looks just a behavioral change for me, not wrong value. Users choose the buffered_time behavior with his will if he wants to calculate summary based on arriving time, the summary is not wrong value then.

sonots:
Wait, okay. Understood what you meant.

sonots:
Added `support_chunk_key_type_buffered_time?`. To "allow" `chunk_key_time buffered_time`, a plugin must override it and return true.

tagomoris:
LGTM, but rebase needed.

lib/fluent/output.rb
```diff
@@ -481,6 +481,7 @@ def initialize
 
     config_param :time_slice_format, :string, :default => '%Y%m%d'
     config_param :time_slice_wait, :time, :default => 10*60
+    config_param :chunk_key_type, :string, :default => 'event_time'
```
repeatedly: should use `:enum`

tagomoris:
Chunk keys are configurable (and easy to modify for plugins). So this pull-request looks already outdated.
Please reopen this if you will add it only in `v0.12` branch.

PR #706, opened 2015-11-20
Show plugin config ancestors

okkez:

PR #715, opened 2015-11-27
Add description to log level

okkez:
We can see this via `--show-plugin-config` after merge #706

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -343,6 +343,7 @@ module Fluent
   module PluginLoggerMixin
     def self.included(klass)
       klass.instance_eval {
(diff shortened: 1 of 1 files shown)


sonots:
:+1:

lib/fluent/log.rb
```diff
@@ -343,6 +343,7 @@ def enable_color(b = true)
   module PluginLoggerMixin
     def self.included(klass)
       klass.instance_eval {
+        desc 'Allows the user to set different levels of logging for each plugin.'
```
sonots: `Allows to set` is fine.

sonots: It seems this is using the same wording with docs. Let me merge although I don't know who is "the" user.

PR #717, opened 2015-11-27
Add descriptions to in_forward

okkez:
Following document does not describe `backlog` and `blocking_timeout`.

See http://docs.fluentd.org/articles/in_forward


sonots:
BTW, we are assuming usual users do not configure `backlog`, and `blocking_timeout`. That's why we are not writing them on docs.

okkez:
OK, I don't write about them `backlog` and `blocking_timeout`.
Thanks!

PR #718, opened 2015-11-27
Add descriptions to in_http

okkez:
See http://docs.fluentd.org/articles/in_http

I cannot describe some options:

  * backlog
  * blocking_timeout
  * cors_allow_origins


sonots:
It looks we should write about `cors_allow_origins` in doc. cf. https://github.com/fluent/fluentd/pull/473
cc: @repeatedly ?

repeatedly:
@sonots Yes, we should write. If this PR add `desc`, fluentd-docs follows it.

okkez:
I've added description for `cors_allow_origins`.

PR #719, opened 2015-11-27
Add descriptions to in_tail

okkez:
See http://docs.fluentd.org/articles/in_tail

I cannot find description for `read_lines_limit`.


sonots:
It looks we should write about `read_lines_limit` on doc, too. cf. https://github.com/fluent/fluentd/pull/593 @repeatedly ?

repeatedly:
I added to fluend-docs.

okkez:
I've added description for `read_lines_limit`.

PR #721, opened 2015-11-27
Add descriptions to Fluent::SocketUtil::BaseInput

okkez:
See followings:

  * http://docs.fluentd.org/articles/in_tcp
  * http://docs.fluentd.org/articles/in_udp

I cannot find description for `blocking_timeout`


sonots:
We are not assuming that usual users configure `blocking_timeout`. That is why we did not write on docs.

PR #722, opened 2015-11-27
Add descriptions buffer

okkez:


okkez:
I've added descriptions for buffer plugins.

PR #724, opened 2015-11-28
Fix DelayedForwarder#run and add unit test

komamitsu:
I found https://github.com/fluent/fluentd/commit/66e618e6d93ccc2928e378b8b5eb702723797bf7 included a typo and it didn't work. Also, I added an unit test for the event lost issue.


repeatedly:
Good catch!

Issue #726, opened 2015-12-01
in_tail: refresh_interval doesn't work if there is a "permission denied" issue


icy:
Quoted from http://docs.fluentd.org/articles/in_tail

> * and strftime format can be included to add/remove watch file dynamically. At interval of refresh_interval, Fluentd refreshes the list of watch file.

However, if any file in list has a "permission denied" issue, `fluentd` keeps generating error every second. This will create a hug amount of useless log entries

Example

````
2015-12-01 09:33:43 +0000 [error]: Permission denied @ rb_sysopen - /var/mail/mail
  2015-12-01 09:33:43 +0000 [error]: suppressed same stacktrace
2015-12-01 09:33:43 +0000 [error]: Permission denied @ rb_sysopen - /var/mail/root
  2015-12-01 09:33:43 +0000 [error]: suppressed same stacktrace
2015-12-01 09:33:43 +0000 fluent.error: {"message":"Permission denied @ rb_sysopen - /var/mail/mail","container_id":"bc4cacb9c7f4","ident":"docker"}
2015-12-01 09:33:43 +0000 fluent.error: {"message":"Permission denied @ rb_sysopen - /var/mail/root","container_id":"bc4cacb9c7f4","ident":"docker"}
2015-12-01 09:33:44 +0000 [error]: Permission denied @ rb_sysopen - /var/mail/mail
  2015-12-01 09:33:44 +0000 [error]: suppressed same stacktrace
2015-12-01 09:33:44 +0000 [error]: Permission denied @ rb_sysopen - /var/mail/root
  2015-12-01 09:33:44 +0000 [error]: suppressed same stacktrace
2015-12-01 09:33:44 +0000 fluent.error: {"message":"Permission denied @ rb_sysopen - /var/mail/mail","container_id":"bc4cacb9c7f4","ident":"docker"}
2015-12-01 09:33:44 +0000 fluent.error: {"message":"Permission denied @ rb_sysopen - /var/mail/root","container_id":"bc4cacb9c7f4","ident":"docker"}
````

repeatedly:
You should add right permission or invoke fluentd with correct user.
This is not fluentd problem. Please check your setting.

icy:
There are two problems here

1. File permission issue
2. Fluentd doesn't respect users' setting "refresh_interval", and update the list of files every 1 second

[...]

PR #734, opened 2015-12-04
in_tail: expand_paths exclude unreadable files

icy:
This is to avoid generating too many error messages from file's watcher
(which runs every second.) Each unreadable file will cause a warning.
If it becomes readable, it would be included in the next turn.

See also `refresh_interval` and #726.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -131,7 +131,7 @@ module Fluent
         if File.readable?(path)
           true
         else
-          $log.warn "#{path} unreadable. It is excluded and would be examined next time."
+          log.warn "#{path} unreadable. It is excluded and would be examined next time."
           false
         end
       }
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -120,21 +120,20 @@ module Fluent
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
-          paths += Dir.glob(path)
+          paths += Dir.glob(path).select { |p|
+            if File.readable?(p)
+              true
+            else
+              log.warn "#{p} unreadable. It is excluded and would be examined next time."
+              false
+            end
+          }
         else
           # When file is not created yet, Dir.glob returns an empty array. So just add when path is static.
           paths << path
         end
       }
-
-      (paths - excluded).select { |path|
-        if File.readable?(path)
-          true
-        else
-          log.warn "#{path} unreadable. It is excluded and would be examined next time."
-          false
-        end
-      }
+      paths - excluded
     end
 
     # in_tail with '*' path doesn't check rotation file equality at refresh phase.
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -126,7 +126,15 @@ def expand_paths
           paths << path
         end
       }
-      paths - excluded
+
+      (paths - excluded).select { |path|
+        if File.readable?(path)
+          true
+        else
+          $log.warn "#{path} unreadable. It is excluded and would be examined next time."
```
repeatedly: Should use `log` instead of `$log`.

icy:
> Should use `log` instead of `$log`.

I have updated the PR, @repeatedly . Could you please explain why `log` should be used here?

repeatedly:
Using `$log`, plugin level `log_level` doesn't effect.

repeatedly:
All tests are filed. Could you check it?

icy:
> All tests are failed. Could you check it?

I will take a look.

icy:
> Tests on travis-ci are failed. Could you check it?

Tests are failed on my local machine (I should have executed the tests first!)

The problem is that test driver creates new temporary file (`"#{TMP_DIR}/tail.txt"`) *after* the invocation of `expand_paths` method. (In the code below, `expand_paths` is invoked before `d.run`.) The file will not be watched as it's excluded.

Do you have any idea to get rid of this problem?

````
  data('1' => [1, 2], '10' => [10, 1])
  def test_emit_with_read_lines_limit(data)
    limit, num_emits = data
    d = create_driver(CONFIG_READ_FROM_HEAD + SINGLE_LINE_CONFIG + "read_lines_limit #{limit}")
    msg = 'test' * 500 # in_tail reads 2048 bytes at once.

    d.run do
      sleep 1
      File.open("#{TMP_DIR}/tail.txt", "ab") {|f|
        f.puts msg
        f.puts msg
      }
       sleep 1
    end

    emits = d.emits
    assert_equal(true, emits.length > 0)
    assert_equal({"message" => msg}, emits[0][2])
    assert_equal({"message" => msg}, emits[1][2])
    assert_equal(num_emits, d.emit_streams.size)
  end
````

**Comment by repeatedly** [comment]
How about following code?

```diff
diff --git a/lib/fluent/plugin/in_tail.rb b/lib/fluent/plugin/in_tail.rb
index f2f2448..73843f4 100644
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -126,7 +126,14 @@ module Fluent
           paths << path
         end
       }
-      paths - excluded
+
+      (paths - excluded).select { |path|
+        unless File.exist?(path)
+          true # 'file not found' means users use fixed path configuration, not '*' included dynamic path. Such files should be watched.
+        else
+          File.readable?(path) ? true : false
+        end
+      }
     end
 
     # in_tail with '*' path doesn't check rotation file equality at refresh phase.
```

icy:
Thank you @repeatedly . I understand the idea here. I have updated the PR. All tests are passed now.

repeatedly:
@sonots @tagomoris Do you see any concerns this patch?

sonots:
Fine

repeatedly:
Thanks

icy:
Thanks a lot, @repeatedly and @sonots!

PR #735, opened 2015-12-04
Add skip_network_error_at_init option to forward output. fix #731

### LINKED ISSUE #731: Fluentd does not honor dns cache expiration when hostname is initially unresolvable
Fluentd has an option expire_dns_cache. When the hostname of the Fluentd server is unresolvable, I would expect Fluentd to retry later. Instead, the whole process stops with an exception. This is undesirable.

repeatedly:
This option is available when `heartbeat_type` is `tcp` for now.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -63,6 +63,8 @@ module Fluent
     config_param :phi_threshold, :integer, :default => 16
     desc 'Use the "Phi accrual failure detector" to detect server failure.'
     config_param :phi_failure_detector, :bool, :default => true
+    desc 'Skip network related error, e.g. DNS error, during plugin setup'
+    config_param :skip_network_error_at_init, :bool, :default => false
 
     # if any options added that requires extended forward api, fix @extend_internal_protocol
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Do we need to support udp heartbeat?

harmw:
I would say this feature should be protocol-agnostic, thus yes UDP should be supported, though you could just merge it now with a disclaimer of some sorts and only support TCP :smile:

To my understanding, the heartbeat is used to ping/check a backend or forwarder, whereas this PR is about mere DNS resolving.

mgcrea:
:+1: this is really needed!

tagomoris:
I think UDP should be supported too.
Any progress?

repeatedly:
We can support UDP but it makes code complicated, e.g. lazy initialization for socket and Heartbeater.
It needs re-writing socket handling of out_forward.
So I want to delay UDP support after other important changes are merged.

tagomoris:
I will come back to this issue after implementing forward plugins with socket/server plugin helpers and v0.14 APIs.

tagomoris:
I created #1399 based on v0.14 code base.

PR #743, opened 2015-12-10
standardize logging level for source configs

skippy:
I've found the in_forward standard of logging, at startup, where it is
expecting incoming msgs to be very helpful.


repeatedly:
Good catch!

PR #745, opened 2015-12-12
Add 'buffer_queue_full_action'  parameter to control buffer actions

repeatedly:
https://groups.google.com/forum/#!topic/fluentd/JWlpijPrsqI

Suggestion from Micheal on the above thread.

`block` action is sometimes dangerous because it causes log lost so default action is still `exception`.
I commented implementation note in the code.

## TODO

Should we add `drop_last` action to drop events for non-important destination?

### DIFF
--- a/lib/fluent/buffer.rb
+++ b/lib/fluent/buffer.rb
@@ -143,6 +143,8 @@ module Fluent
     config_param :buffer_chunk_limit, :size, :default => 8*1024*1024
     desc 'The length limit of the chunk queue.'
     config_param :buffer_queue_limit, :integer, :default => 256
+    desc 'The action when the size of buffer queue exceeds the buffer_queue_limit.'
+    config_param :buffer_queue_full_action, :enum, :list => [:exception, :block], :default => :exception
 
     alias chunk_limit buffer_chunk_limit
     alias chunk_limit= buffer_chunk_limit=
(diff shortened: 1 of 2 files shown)


tatsu-yam:
It is a just idea.
How about using plugin configured at `<secondary>`?

sonots:
LGTM

repeatedly:
@tatsu-yam I'm considering `secondary` support but buffer doesn't know output. So sending records to secondary from buffer is hard without modification for now.

@tagomoris @frsyuki How about this? This is rough implemenetation but it covers most cases with exsiting plugins.

repeatedly:
Current patch works but no need ConditionVariable. So I will update the patch.

repeatedly:
Update patch with `rescue/retry/sleep` way.
I'm considering `while/break` way but it i increases the overhead in normal path.

repeatedly:
@tagomoris @sonots patch updated. please check again.

sonots:
The comment says as "rough implementation". Under this comment, it looks fine for me.

tagomoris:
IMO, `#configure` MUST warn if it's configured as `:block`.
And buffers should log in info or debug level before sleeping when it exceeds limits.
Others are looks good to me.

repeatedly:
@tagomoris added

PR #752, opened 2015-12-18
fixed problem in watch multiple file by "*" on Windows

masato25:
HI , I found a problem in Windows.

When I using tail plugin on window for watch files in a folder.it's not work.
ex. C:\test\*.csv 
that cause of the "\" making Dir parse the file path error. and when I  change the path to "C:/test/*.csv".
It's works well now, so I add a if condition for checking if it's a File-path of Windows will automatic fix the file path.
```
  #if this is file path of windows
  if(path =~ /^\s*[C-Z]:\\/)
    #fix folder watch issue of windows
    path.gsub!("\\","/")
  end
  paths += Dir.glob(path)
```

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -106,6 +106,11 @@ module Fluent
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
+          #if this is file path of windows
+          if(path =~ /^\s*[C-Z]:\\/)
+            #fix folder watch issue of windows
+            path.gsub!("\\","/")
+          end
           paths += Dir.glob(path)
         else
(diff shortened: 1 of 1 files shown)


repeatedly:
So this problem doesn't happen with fixed path like 'C:\test.csv', right?

repeatedly:
Ah, wait. Please send a patch to `master`.
`v0.14` is a sandbox branch for new buffer API.

masato25:
HI repeatedly, thanks for quick response.

the current only solution is change the "c:\test\*.csv" to "c:/test/*.csv" in windows.

this is a hotfix for windows , I will patch it into `windows` branch.

repeatedly:
> this is a hotfix for windows , I will patch it into windows branch.

`windows` branch is already merged into `master`. So please use `master`.

PR #753, opened 2015-12-18
fix watch folder failed issue of winodws

masato25:
if we set "c:\test\*.csv" as path of tail plugin, fluentd can't get watch files list.
user need change "/" to "\" on windows. this patch is for fixed this problem.


masato25:
sorry , again. please reject this one. I will submit a new patch into `master`.

PR #754, opened 2015-12-18
fix fetch files list issue on windows

masato25:
Fix fetch files list issue on Windows.

sorry for the wrong pull request many times. :(

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -120,6 +120,12 @@ module Fluent
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
+          
+          #fix fetch files list issue on windows
+          if(path =~ /^\s*[C-Z]:\\/)
+            path.gsub!("\\","/")
+          end
+
           paths += Dir.glob(path).select { |p|
             if File.readable?(p)
               true


lib/fluent/plugin/in_tail.rb
```diff
@@ -120,6 +120,12 @@ def expand_paths
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
+          
+          #fix fetch files list issue on windows
+          if(path =~ /^\s*[C-Z]:\\/)
```
repeatedly: Could you remove `()` like above `if` condition?

repeatedly:
@nurse Could you check this approach? Is this good or does Ruby have proper way?

nurse:
@masato25 What issue do you want to solve? Could you show an example case?

masato25:
HI @nurse , nice to talk with you.
I had tried the tail plugin to watch bunch of files in a folder on Windows OS.
ex. C:\test\*.csv 
But it's seems doesn't works. so I look into the source code of tail plugin.
```
            if File.readable?(p)
              true
            else
              log.warn "#{p} unreadable. It is excluded and would be examined next time."
              false
            end
          }
```
and I checked [Ruby Doc](http://ruby-doc.org/core-2.2.0/Dir.html), the document mention 
```
\
Escapes the next metacharacter.

Note that this means you cannot use backslash on windows as part of a glob, i.e. Dir["c:\foo*"] will not work, use Dir["c:/foo*"] instead.
``` 

I followed the doc's solutions to change "c:\test\*.csv" into "c:/test/*.csv" and that is works.

I just think If fluentd can native support for automatic fixed this kind of problem will be more friendly for Winodws users, so I did this pull request.

Thanks 
Masato

nurse:
It sounds reasonable.
You can use `Fluent.windows?` instead of `(path =~ /^\s*[C-Z]:\\/)` as the begining of the file, and tr!('\\', '/') instead of gsub. The code should be in def configure.

nurse:
As wrote in https://github.com/fluent/fluentd/pull/919#issuecomment-214678118, it breaks if there's a file which has glob meta character.

PR #757, opened 2015-12-22
Add on_exit_process(i) hook point

riywo:
Hi,

I wonder how to call `shutdown` method before the child process exits when I use `detach_process` because I'd like to shutdown my plugin properly.

After looking the code, I hope this hook point just before `exit!`. Then I can call any logic before exiting like below:
```
def on_exit_process(i)
  shutdown
end
```

Is there any better way to do this?

Best,

### DIFF
--- a/lib/fluent/process.rb
+++ b/lib/fluent/process.rb
@@ -289,6 +289,9 @@ module Fluent
     def on_detach_process(i)
     end
 
+    def on_exit_process(i)
+    end
+
     private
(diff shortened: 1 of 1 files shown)


repeatedly:
It seems reasonable.

repeatedly:
If there is no other suggestion, I wll merge this patch.

riywo:
Thanks!

PR #759, opened 2015-12-24
Add descriptions to BufferedOutput

okkez:
I cannot find description for `:try_flush_interval`.

See http://docs.fluentd.org/articles/output-plugin-overview


sonots:
`try_flush_interval` is a kind of hidden parameters. That is why you could not find the description.

okkez:
Thanks!

PR #761, opened 2015-12-29
Use `Timeout.timeout` instead of `Object#timeout`

okkez:
Because Object#timeout has been deprecated since Ruby 2.3.0.


sonots:
Thanks!

PR #762, opened 2015-12-29
Enable to control wait period of d.run in input and output test driver

sonots:
I want to control wait period of d.run in input and output test driver. 

Please see what I am doing every time > https://github.com/sonots/fluent-plugin-grep/blob/master/test/helper.rb. I want to drop this monkey patch.


sonots:
I expect new plugin API gives us better test helpers in the future :-)

PR #770, opened 2016-01-12
Add filter and label example to fluent.conf. fix #512

### LINKED ISSUE #512: update the default config file to include filters and labels
The current example config generated by `--setup` does not include a filter or label example.

- I feel pretty strongly about having a couple of filter examples.
- I am less concerned about labels since it's more "advanced" and not needed until more complex event routing is required. but it probably doesn't hurt to have an example toward the bottom of `fluent.conf`.

repeatedly:
@kiyoto could you check?

### DIFF
--- a/fluent.conf
+++ b/fluent.conf
@@ -31,6 +31,24 @@
 #  tag apache.access
 #</source>
 
+## Mutating event filter
+## Add hostname and tag fields to apache.access tag events
+#<filter apache.access>
+#  @type record_transformer
+#  <record>
+#    hostname ${hostname}
+#    tag ${tag}
+#  </record>
+#</filter>
+
+## Selecting event filter
+## Remove unnecessary events from apache prefixed tag events
+#<filter apache.**>
+#  @type grep
+#  include1 method GET # pass only GET in 'method' field
+#  exclude1 message debug # remove debug event
+#</filter>
+
 # Listen HTTP for monitoring
 # http://localhost:24220/api/plugins
 # http://localhost:24220/api/plugins?type=TYPE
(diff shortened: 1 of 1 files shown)


repeatedly:
Or @sonots @tagomoris how  about this?

fluent.conf
```diff
@@ -31,6 +31,24 @@
...
+#    hostname ${hostname}
+#    tag ${tag}
+#  </record>
+#</filter>
+
+## Selecting event filter
+## Remove unnecessary events from apache prefixed tag events
+#<filter apache.**>
+#  @grep
+#  include1 method GET # pass only GET in 'method' field
```
sonots: regexp1

sonots:
Need an exmple to emit messages with label (source)

repeatedly:
@sonots Please commit changes to this branch directly :)

fluent.conf
```diff
@@ -31,6 +31,24 @@
 #  tag apache.access
 #</source>
 
+## Mutating event filter
+## Add hostname and tag fields to apache.access tag events
+#<filter apache.access>
+#  @record_transformer
```
tagomoris: `@type record_transformer`?

fluent.conf
```diff
@@ -31,6 +31,24 @@
...
+#  <record>
+#    hostname ${hostname}
+#    tag ${tag}
+#  </record>
+#</filter>
+
+## Selecting event filter
+## Remove unnecessary events from apache prefixed tag events
+#<filter apache.**>
+#  @grep
```
tagomoris: ditto

repeatedly:
Fix typo of missing `@type`

repeatedly:
Merged

PR #776, opened 2016-01-15
Add skip_invalid_event parameter to in_forward. fixes #774

### LINKED ISSUE #774: Add check and skip event option in in_forward
```
App - 1 -> fluentd-a - 2 -> fluentd-b
```

In the above topology, `2` transfer should be safe but `1` sometimes has a problem, e.g. client send an invalid event or something.
For debugging and stability, adding check and skip invalid event option is better.
This is helpful for identifying where is the cause of invalid event, 1 or 2? 

If set `skip_invalid_event true` in in_forward, in_forward checks received event like below.

```
if tag.is_a?(String)
if time.is_a?(Integer)
if record.is_a?(Hash)
```

If detect invalid event, in_forward logs information and skip its event.
This feature also prevents broken records are stored into buffer with PackedForward.

Enable this option on aggreagator may be hard because aggregator receives lots of events from lots of servers.
On the other hard, enable it on forwarder side is reasonable.

repeatedly:
See #774

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -38,6 +38,8 @@ module Fluent
     config_param :chunk_size_warn_limit, :size, :default => nil
     desc 'Received chunk is dropped if it is larger than this value.'
     config_param :chunk_size_limit, :size, :default => nil
+    desc 'Skip an event if incoming event is invalid.'
+    config_param :skip_invalid_event, :bool, :default => false
 
     def configure(conf)
       super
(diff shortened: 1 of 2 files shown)


repeatedly:
@sonots @tagomoris 
I will add tests but how about this?

frsyuki:
I commented on #774.

repeatedly:
Add tests

repeatedly:
Can I merge this?
After merge this PR, I will release v0.12.20.

repeatedly:
@frsyuki @tagomoris @sonots if no problem, I will merge this today.

PR #781, opened 2016-01-21
Add some objects cannot be parsed

ysksn:
Regexp, Proc, Symbol, and String (empty string) objects added.
thanks :)


repeatedly:
Thanks!

PR #788, opened 2016-01-31
in_tail: avoid unnecessary open()s

dwoo4dwoo:
As discussed in https://groups.google.com/forum/#!topic/fluentd/BBRde4Lyl9E

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -586,8 +586,7 @@ module Fluent
 
         def on_notify
           begin
-            io = FileWrapper.open(@path)
-            stat = io.stat
+            stat = FileWrapper.stat(@path)
             inode = stat.ino
             fsize = stat.size
           rescue Errno::ENOENT
@@ -599,13 +598,14 @@ module Fluent
           begin
             if @inode != inode || fsize < @fsize
               # rotated or truncated
+              begin
+                io = FileWrapper.open(@path)
+              rescue Errno::ENOENT
+              end
               @on_rotate.call(io)
-              io = nil
             end
             @inode = inode
             @fsize = fsize
-          ensure
-            io.close if io
           end
 
         rescue


dwoo4dwoo:
The Travis CI build failures seem unrelated to my change ... (UdpInputTest EADDRINUSE)

repeatedly:
Possible file rotation cases:

1. File is rotated before `stat = FileWrapper.stat(@path)`

No problem.

2. File is rotated ater `stat = FileWrapper.stat(@path)`

2a. `if @inode != inode || fsize < @fsize` is false

No problem. Next trigger should detect file rotation.

2b. `if @inode != inode || fsize < @fsize` is true

It seems no problem. `on_rotate` also checks stat info with previous file.
So this patch should not effect existing behaviour.

Does anyone have any concern for this patch?

repeatedly:
There are no problem, I will merge this patch today.

lib/fluent/plugin/in_tail.rb
```diff
@@ -593,13 +592,11 @@ def on_notify
           begin
             if @inode != inode || fsize < @fsize
               # rotated or truncated
+              io = FileWrapper.open(@path)
```
sonots: When Errno::ENOENT occurs on `FileWrapper.stat(@path)`, it seems Errno::ENOENT occurs again at here. Do not you need to handle it?

repeatedly: So adding `rescue nil` is better?

dwoo4dwoo: @sonots : good catch - thanks!
@repeatedly : I'll update the pull request with your suggestion

repeatedly:
@sonots Could you check again?

lib/fluent/plugin/in_tail.rb
```diff
@@ -593,13 +592,11 @@ def on_notify
           begin
             if @inode != inode || fsize < @fsize
               # rotated or truncated
+              io = FileWrapper.open(@path) rescue nil
```
sonots: Could you tell me when this io object will be closed? Or, not necessary to close? Why?

sonots: I am not sure why there was

     io = nil

before

sonots: You should ignore only Errno::ENOENT as repeatedly wrote.

dwoo4dwoo: Hi @sonots ,

> Could you tell me when this io object will be closed?

The file descriptor is consumed here:

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L384

The first time through ... an io_handler is created here:

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L416

The file descriptor remains open until the next file rotation, removal, or fluentd shutdown.  The logic to deal with some of these situations begins here:

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L420

Other cases are handled here:

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L188

For example, shutdown calls stop_watchers().

> You should ignore only Errno::ENOENT as repeatedly wrote

Sorry, I thought @repeatedly was giving me a choice, and I picked the first option.  I'll update the pull request using the second option.  I'll also rebase against master.

David

sonots:
LGTM

repeatedly:
Thanks for the patch!

PR #791, opened 2016-02-04
Use error stream in TimeSlicedOutput#emit

repeatedly:
Currently, TimeSlicedOutput#emit raises an exeception when one event is invalid.
If one event in event stream has a problem, it should be treated as an error event instead of entire error.
Input plugins hard to handle such error because input plugins don't know how to recover invalid event.
So send it to error stream is better.

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -566,11 +566,12 @@ module Fluent
             key = @time_slicer.call(time)
             @before_key = key
           end
-          formatted_data[key] ||= ''
-          formatted_data[key] << format(tag, time, record)
         rescue => e
           @router.emit_error_event(tag, Engine.now, {'time' => time, 'record' => record}, e)
         end
+
+        formatted_data[key] ||= ''
+        formatted_data[key] << format(tag, time, record)
       }
       formatted_data.each { |key, data|
         if @buffer.emit(key, data, chain)
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -557,16 +557,20 @@ module Fluent
       @emit_count += 1
       formatted_data = {}
       es.each {|time,record|
-        tc = time / @time_slice_cache_interval
-        if @before_tc == tc
-          key = @before_key
-        else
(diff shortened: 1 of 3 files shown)


lib/fluent/test/base.rb
```diff
@@ -54,6 +54,7 @@ def initialize(klass, &block)
         end
         @instance.router = Engine.root_agent.event_router
         @instance.log = TestLogger.new
+        Engine.root_agent.instance_variable_set(:@log, @instance.log)
```
repeatedly: This addition fixes holding different log object issue between `@instance` and `root_agent`.
If fix this problem, we should change Engine setup sequence in the tests.

lib/fluent/output.rb
```diff
@@ -557,16 +557,20 @@ def emit(tag, es, chain)
...
+          tc = time / @time_slice_cache_interval
+          if @before_tc == tc
+            key = @before_key
+          else
+            @before_tc = tc
+            key = @time_slicer.call(time)
+            @before_key = key
+          end
+          formatted_data[key] ||= ''
+          formatted_data[key] << format(tag, time, record)
```
frsyuki: `rescue` catches an exception happens at `format`. So, a TimeSlicedOutput plugin doesn't have to catch exception at `format`.
But `BufferedOutput#format_stream` doesn't catch exception. Thus a BufferedOutput plugin needs to catch exception at `format` to support error stream.
I think that is confusing. How about doing either of them?:

* a) TimeSliceOutput doesn't catch exceptions that happen at `format`
* b) BufferedOutput#format_stream catches exceptions that happen at `format` and calls `@router.emit_error_event`

I think a) is better at least with before v0.14 because BufferedOutput#format_stream can't decide message structure (like `{'record'=>record}` or `record`).

repeatedly: I see. a) is better because effect is minimum. I will change it.
v0.14 with new API, we should re-consider error stream use.

frsyuki:
Looks good :+1:

PR #794, opened 2016-02-08
Separate system config object from supervisor

tagomoris:
to pass system configuration to plugins

* v0.14 feature: new Plugin API expects these configuration parameters to system config
  * plugin storage path
  * default file/directory permissions
  * many others

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -15,6 +15,7 @@
 #
 
 require 'fluent/load'
+require 'fluent/system_config'
 require 'etc'
 if Fluent.windows?
   require 'windows/library'
@@ -146,7 +147,8 @@ module Fluent
       @log.init
       show_plugin_config if @show_plugin_config
       read_config
-      apply_system_config
+      @system_config = SystemConfig.create(@conf) # @conf is set in read_config
+      @system_config.apply(self)
 
       dry_run if @dry_run
       start_daemonize if @daemonize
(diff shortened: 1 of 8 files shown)


lib/fluent/supervisor.rb
```diff
@@ -585,8 +576,9 @@ def change_privilege
     end
 
     def init_engine
-      init_opts = {:suppress_interval => @suppress_interval, :suppress_config_dump => @suppress_config_dump, :without_source => @without_source}
-      Fluent::Engine.init(init_opts)
+      # init_opts = {:suppress_interval => @suppress_interval, :suppress_config_dump => @suppress_config_dump, :without_source => @without_source}
```
repeatedly: This comment is needed?

lib/fluent/system_config.rb
```diff
@@ -0,0 +1,92 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Fluent
+  require 'fluent/configurable'
+
+  module SystemConfigMixin
```
repeatedly: Who use this module?

tagomoris: Plugins (or PluginSupport modules to be introduced) which need to refer system configurations.

repeatedly: I see

repeatedly:
Looks good

tagomoris:
Merged.

PR #795, opened 2016-02-09
enforce new hash syntax by `rubocop --only HashSyntax --auto-correct`

tagomoris:
I did syntax check for all .rb files by `find . -name '*.rb' | xargs -n 1 ruby -c`


lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -60,8 +60,8 @@ def configure(conf)
       end
 
       placeholder_expander_params = {
-        :log           => log,
-        :auto_typecast => @auto_typecast,
+        log: log,
+        auto_typecast: @auto_typecast,
```
okkez: Breaks align. Is this ok?

test/plugin/test_filter_record_transformer.rb
```diff
@@ -411,35 +411,35 @@ def emit(config, msgs = [''])
...
         es.each_with_index do |(t, r), i|
           actual_results << {
-            :single      => r["single"],
-            :multiple    => r["multiple"],
-            :with_prefix => r["with_prefix"],
-            :with_suffix => r["with_suffix"],
+            single: r["single"],
+            multiple: r["multiple"],
+            with_prefix: r["with_prefix"],
+            with_suffix: r["with_suffix"],
```
okkez: Breaks align in this file. Is this ok?

lib/fluent/plugin/filter_record_transformer.rb
tagomoris: it has just 2 fields and these are far different from each other. There seems no problem.

test/plugin/test_filter_record_transformer.rb
tagomoris: It seems test code alignments, not to raise any errors.

tagomoris:
@repeatedly I'm about to merge this. Are you ok?

repeatedly:
Dows this syntax work on ruby 1.9?
We sometimes backport master changes to v0.12 and v0.12 should work on ruby 1.9.
I know ruby 1.9 is dead but some users use it on production.

tagomoris:
It's the new feature for Ruby 1.9.
http://effectif.com/ruby/update-your-project-for-ruby-19-hash-syntax

repeatedly:
I see. Okay.

tagomoris:
Merged.

PR #799, opened 2016-02-09
Fix dependencies between files

tagomoris:
Fix to do `require` for required files straightforward, per files
* Currently, all required files are loaded by `lib/fluent/load.rb` even for tests
  * This file creates the implicit order of files to be loaded
  * There's no explicit dependency definition, and there are many circular dependencies
  * This situation is totally broken
* Without this change, we can't run tests by `bundle exec rake test TEST=test/file.rb`
  * Running whole tests consumes long time, and it's not acceptable
    
This change is to make straightforward dependency from all files to others.
We can load any file in fluentd source code, make clean dependencies, and also can run each files itself to test it.

After this change, some 3rd party plugins might fail to run test code. But it's because these plugins' implicit dependencies to the other files, occasionally not loaded. It's danger situation, and should be fixed eventually.

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -13,10 +13,13 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 #
-module Fluent
-  require 'fluent/configurable'
-  require 'fluent/engine'
 
+require 'fluent/configurable'
+require 'fluent/engine'
+require 'fluent/plugin'
+require 'fluent/output'
+
+module Fluent
   #
   # Agent is a resource unit who manages emittable plugins
   #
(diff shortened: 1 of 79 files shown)


tagomoris:
FYI: I checked this change by:
* standard test: `bundle exec rake test`
* tests per each files: `find test -name 'test_*.rb' | xargs -I{} -n1 bundle exec rake test TEST={}`
* execution trial: `bundle exec bin/fluentd -c example/in_forward.conf`

sonots:
Great work

tagomoris:
@repeatedly I'll merge this if no objection exist.

tagomoris:
Merged.

PR #800, opened 2016-02-10
Extract OutputChain

cosmo0920:
I've extracted `OutputChain` from output.rb.

And I also keep to be able to run tests `test_output` alone.


lib/fluent/load.rb
```diff
@@ -32,6 +32,7 @@
 require 'fluent/event'
 require 'fluent/buffer'
 require 'fluent/input'
+require 'fluent/output_chain'
```
tagomoris: This `require` seems not needed because 'fluent/output' requires 'fluent/output_chain'.
If any code were broken, it depends on that file.

cosmo0920: I've confirmed that this require is needless.
Thank you!

repeatedly:
Can we merge this?

tagomoris:
LGTM

PR #807, opened 2016-02-17
in_forward: Add source_hostname_key parameter. fix #804

### LINKED ISSUE #804: Add option to enable source host key for in_forward plugin
Like other plugins, it would a really helpful to be able to use "source_host_key" for in_forward as is possible with in_tcp.

repeatedly:


lib/fluent/plugin/in_forward.rb
```diff
@@ -47,6 +47,8 @@ def initialize
     config_param :chunk_size_limit, :size, default: nil
     desc 'Skip an event if incoming event is invalid.'
     config_param :skip_invalid_event, :bool, default: false
+    desc "The field name of the client's hostname."
```
tagomoris: IMO it's better to have a not for performance degradation.

repeatedly: It seems document issue like this: https://github.com/fluent/fluentd-docs/pull/215/files

tagomoris:
This feature will be supported by porting into v0.14 API and using `inject` plugin helper.

repeatedly:
@tagomoris How to insert client's hostname by inject plugin helper?

repeatedly:
Rebased.

tagomoris:
Oh, I misunderstood the problem.

tagomoris:
IMO, parameter name should be `source_hostname_key` instead of `source_host_key` to show this feature injects hostname, not host address or others explicitly.

repeatedly:
It is derived from `in_tcp` so you also think in_tcp's parameter should be renamed?

tagomoris:
Adding newer name to `in_tcp` sounds good idea.
`in_forward` plugin is used much widely than `in_tcp`, so I think we should NOT follow the bad way of `in_tcp`.

repeatedly:
Renamed.

test/plugin/test_in_forward.rb
```diff
@@ -583,5 +583,53 @@ def send_data(data, try_to_receive_response=false, response_timeout=1)
...
+    }
+  end
+
+  test 'forward protocol with source_hostname_key' do
+    execute_test { |events|
+      entries = []
+      events.each {|tag,time,record|
+        entries << [time, record]
+      }
+      send_data ['tag1', entries].to_msgpack
```
tagomoris: I cannot understand why this code works well... `time` should be an ext typed value, but there are no info about it.

test/plugin/test_in_forward.rb
```diff
@@ -583,5 +583,53 @@ def send_data(data, try_to_receive_response=false, response_timeout=1)
...
+      ["tag1", time, {"a"=>1}],
+      ["tag1", time, {"a"=>2}]
+    ]
+    d.expected_emits_length = events.length
+
+    d.run do
+      block.call(events)
+    end
+
+    d.emits.each { |tag, _time, record|
```
tagomoris: Does `d.emits` really have 2 events? This test code doesn't wait/confirm to receive data via TCP socket.
I doubt that `d.emits` is empty here.

test/plugin/test_in_forward.rb
tagomoris: Correct way to do it is using `Fluent::MessagePackFactory`.

test/plugin/test_in_forward.rb
repeatedly: Test calls `d.expected_emits_length = events.length` before so `d.run` is finished after received 2 events.

test/plugin/test_in_forward.rb
repeatedly: EventTime is serialized to int via `to_msgpack` and this is why this test works.

repeatedly: Yeah, using Fluent::MessagePackFactory is better. "packed forward protocol with source_hostname_key" case does it.

tagomoris: It looks incorrect behavior to simulate `out_forward`... but not so serious for this test.

repeatedly:
Okay. After test passed, I will merge this patch.

PR #808, opened 2016-02-17
Fix to do `require` for required files straightforward, per files

okkez:
* Currently, all required files are loaded by `lib/fluent/load.rb` even for tests
  * This file creates the implicit order of files to be loaded
  * There's no explicit dependency definition, and there are many circular dependencies
  * This situation is totally broken
* Without this change, we can't run tests by `bundle exec rake test TEST=test/file.rb`
  * Running whole tests consumes long time, and it's not acceptable

This change is to make straightforward dependency from all files to others.
We can load any file in fluentd source code, make clean dependencies, and
also can run each files itself to test it.

For v0.12 branch,

* I don't merge Windows specific patches
* I don't merge v0.14 specific patches

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -13,10 +13,13 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 #
-module Fluent
-  require 'fluent/configurable'
-  require 'fluent/engine'
 
+require 'fluent/configurable'
+require 'fluent/engine'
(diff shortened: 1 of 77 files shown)


okkez:
Test on TravisCI is failed but this failure does not seem to be related to this PR.

repeatedly:
@tagomoris Can we merge this PR?

repeatedly:
Thanks!

PR #810, opened 2016-02-19
Fix in_tail input messages loss when receive_lines fail. fix #699

### LINKED ISSUE #699: Will in_tail input loss messages when router emitting fail

repeatedly:
Based on #700. Update patch and rebase to master.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -258,9 +258,11 @@ module Fluent
               end
         begin
           router.emit_stream(tag, es)
-        rescue
-          # ignore errors. Engine shows logs and backtraces.
+        rescue BufferQueueLimitError
           return false
+        rescue
+          # ignore non BufferQueueLimitError errors because in_tail can't recover. Engine shows logs and backtraces.
+          return true
         end
       end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -248,6 +248,7 @@ module Fluent
       log.error_backtrace
     end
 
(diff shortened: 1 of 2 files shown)


repeatedly:
@sonots @tagomoris @zheolong How about this?

zheolong:
@repeatedly I think it's great.

lib/fluent/plugin/in_tail.rb
```diff
@@ -258,10 +258,15 @@ def receive_lines(lines, tail_watcher)
...
+        rescue BufferQueueLimitError
+          return false
         rescue
-          # ignore errors. Engine shows logs and backtraces.
+          # ignore non BufferQueueLimitError errors because in_tail can't recover. Engine shows logs and backtraces.
+          return true
         end
       end
+
+      return true
```
sonots: true/false is ambiguous what it means. please write a yardoc comment for the return value.

sonots:
commented

repeatedly:
If there is no problem, I will merge this patch today and release new fluentd.

PR #811, opened 2016-02-24
Added enable_watch_timer option

dwoo4dwoo:
The enable_watch_timer option was discussed in:

    https://groups.google.com/forum/#!topic/fluentd/BBRde4Lyl9E

Setting this parameter to `false` will significantly reduce CPU and I/O consumption when tailing a large number of files on systems with inotify support.

Corresponding PR for documentation changes: https://github.com/fluent/fluentd-docs/pull/215

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -54,6 +54,8 @@ module Fluent
     config_param :read_lines_limit, :integer, default: 1000
     desc 'The interval of flushing the buffer for multiline format'
     config_param :multiline_flush_interval, :time, default: nil
+    desc 'Enable the additional watch timer.'
+    config_param :enable_watch_timer, :bool, default: true
 
     attr_reader :paths
 
@@ -162,7 +164,7 @@ module Fluent
 
     def setup_watcher(path, pe)
       line_buffer_timer_flusher = (@multiline_mode && @multiline_flush_interval) ? TailWatcher::LineBufferTimerFlusher.new(log, @multiline_flush_interval, &method(:flush_buffer)) : nil
-      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher,  &method(:receive_lines))
+      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher,  &method(:receive_lines))
       tw.attach(@loop)
       tw
     end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -324,16 +326,20 @@ def parse_multilines(lines, tail_watcher)
...
         @rotate_wait = rotate_wait
         @pe = pe || MemoryPositionEntry.new
         @read_from_head = read_from_head
+        @disable_watch_timer = disable_watch_timer
         @read_lines_limit = read_lines_limit
         @receive_lines = receive_lines
         @update_watcher = update_watcher
 
-        @timer_trigger = TimerWatcher.new(1, true, log, &method(:on_notify))
+        if ! @disable_watch_timer
```
repeatedly: Could you use `@timer_trigger = TimerWatcher.new(1, true, log, &method(:on_notify)) unless @disable_watch_timer`?

lib/fluent/plugin/in_tail.rb
```diff
@@ -356,13 +362,17 @@ def wrap_receive_lines(lines)
       end
 
       def attach(loop)
-        @timer_trigger.attach(loop)
+        if ! @disable_watch_timer
```
repeatedly: ditto

lib/fluent/plugin/in_tail.rb
```diff
@@ -356,13 +362,17 @@ def wrap_receive_lines(lines)
...
+          @timer_trigger.attach(loop)
+        end
         @stat_trigger.attach(loop)
         on_notify
       end
 
       def detach
-        @timer_trigger.detach if @timer_trigger.attached?
+        if ! @disable_watch_timer
```
repeatedly: `unless @disable_watch_timer` instead of `if ! @disable_watch_timer`

dwoo4dwoo: Sure @repeatedly ... I'll push a new version in a few minutes with the requested changes.

sonots:
Reading the thread on google group, it looks `disable_watch_timer` can be true by default on most of environments. I suggest to change the option name to `enable_watch_timer` and go like:

1. Add `enable_watch_timer` but default is `true` for backward compatibility
2. Change the default value to `false` someday
3. Remove the option and codes related with time watcher if we find that nobody requires the watch timer.

@repeatedly what do you think?

repeatedly:
@sonots Hm. `enable_watch_timer` also seems good. Either is fine :)

sonots:
@dwoo4dwoo could you change the option to `enable_watch_timer`?

dwoo4dwoo:
> could you change the option to enable_watch_timer?

Sure @sonots ... it's done!  David

sonots:
Thanks!! LGTM!

repeatedly:
Thanks for the work!

PR #821, opened 2016-03-02
Enum list must be of symbols

tagomoris:
And make types.rb testable.


repeatedly:
Testable is important :+1:

PR #822, opened 2016-03-02
Fix the regression of SystemConfig apply

repeatedly:
Introduced by https://github.com/fluent/fluentd/pull/794


repeatedly:
@tagomoris check this patch.

tagomoris:
LGTM

PR #823, opened 2016-03-03
Add missing require

okkez:
fluent/output uses fluent/status.


okkez:
CI fails, but this failure is not related to my change.

repeatedly:
Thanks

PR #827, opened 2016-03-04
fix to dup default values to make it safe if plugins modify provided values

tagomoris:
This problem makes bad effect especially for instantiated/configured twice or more in a process.
The most visible case is testing, but it might affect configuration reloaded with `HUP` signals.


tagomoris:
Can I merge this?

repeatedly:
LGTM

PR #832, opened 2016-03-08
separate PluginId module from config.rb

tagomoris:
* simply speaking, it's wrong file to write it
* added a feature to check whether id is intentionally set by user or not
  * it's important to check whether we can create many thing depends on id or not
  * for example, id-based auto generated buffer path, and plugin storage on local filesystem

### DIFF
--- a/lib/fluent/plugin_id.rb
+++ b/lib/fluent/plugin_id.rb
@@ -1,0 +1,57 @@
+#
+# Fluent
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin_id.rb
```diff
@@ -0,0 +1,52 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Fluent
+  module PluginId
+    @@configured_ids = {}
```
repeatedly: It seems stored values are not used.
Using Set or Array is better for memory usage and GC pressure.

tagomoris: Ah, that's right. I'll use it.

tagomoris:
I modified commits:
* to use Set instead of Hash
* to stringify `@id` not to duplicate values between strings and numbers (or other types)
* not to add nil if `@id` is missing

tagomoris:
Merged.

sonots:
LGTM

PR #833, opened 2016-03-08
Fix to be able to get any number of available ports

okkez:
This is based on 1f514600f9


okkez:
See also https://github.com/fluent/fluentd/wiki/Developer-notes-for-fixes-on-v0.14#tests

tagomoris:
LGTM.

PR #834, opened 2016-03-09
Add methods for `p` output

okkez:


okkez:
See also https://github.com/fluent/fluentd/wiki/Developer-notes-for-fixes-on-v0.14#configuration

lib/fluent/config/section.rb
```diff
@@ -62,6 +66,19 @@ def [](key)
         @params[key.to_sym]
       end
 
+      def respond_to?(symbol, include_all=false)
+        case symbol
+        when :inspect, :nil?, :to_h, :+, :instance_of?, :kind_of?, :[], :respond_to?, :respond_to_missing?, :method_missing,
+          true
+        when :!, :!= , :==, :equal?, :instance_eval, :instane_exec
```
uu59: `instane_exec` should be `instance_exec` ?

okkez: Thanks! I've fixed.

tagomoris:
Can you add some tests for this change?

okkez:
OK! I will add tests.

okkez:
I've added tests.

tagomoris:
@okkez tests are failing with syntax error on Ruby 2.0 & 2.1.

test/config/test_section.rb
```diff
@@ -106,6 +106,42 @@ class TestSection < ::Test::Unit::TestCase
...
+        sub_test_case '#to_s' do
+          test '#to_s == #inspect' do
+            h1 = {name: "s1", num: 10, class: "A"}
+            s1 = Fluent::Config::Section.new(h1)
+
+            assert_equal(s1.to_s, s1.inspect)
+          end
+        end
+
+        data("inspect": [:inspect, true],
```
cosmo0920: {"string": value} JSON syntax has been introduced in Ruby 2.3.
But Fluentd v0.14 targets Ruby 2.1 or later.
We should use {"string" => value} JSON syntax.

okkez: Good catch. I will fix this.

okkez:
I've fixed tests.

tagomoris:
LGTM.

tagomoris:
Merged. Thank you!

PR #836, opened 2016-03-09
In http: Use RFC suggested character cases of HTTP response header

cosmo0920:
* In RFC 7230, content length header is described as [Content-**L**ength](http://tools.ietf.org/html/rfc7230#section-3.3.2).
* In RFC 7231, content type header is described as [Content-**T**ype](http://tools.ietf.org/html/rfc7231#section-3.1.1.5).

Note that `Content-type` is already corrected into `Content-Type` in internally by Net::HTTP: https://github.com/ruby/ruby/blob/5728783a0458357105c1943a4b0a9e04c1879a04/lib/net/http/header.rb#L171.

So, this change is only code appearance.

### DIFF
--- a/test/plugin/test_in_http.rb
+++ b/test/plugin/test_in_http.rb
@@ -323,7 +323,7 @@ class HttpInputTest < Test::Unit::TestCase
       d.run do
         # Send two requests the second one has no Content-Type in Keep-Alive
         Net::HTTP.start("127.0.0.1", PORT) do |http|
-          req = Net::HTTP::Post.new("/foodb/bartbl", {"connection" => "keepalive", "content-type" => "application/json"})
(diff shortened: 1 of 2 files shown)


test/plugin/test_in_http.rb
```diff
@@ -323,7 +323,7 @@ def on_message_complete
       d.run do
         # Send two requests the second one has no Content-Type in Keep-Alive
         Net::HTTP.start("127.0.0.1", PORT) do |http|
-          req = Net::HTTP::Post.new("/foodb/bartbl", {"connection" => "keepalive", "content-type" => "application/json"})
+          req = Net::HTTP::Post.new("/foodb/bartbl", {"connection" => "keepalive", "content-Type" => "application/json"})
```
repeatedly: `Content-Type`?

cosmo0920: Ah, yes....

repeatedly:
Thanks

PR #838, opened 2016-03-09
Register plugin into registry

tagomoris:
To make straightforward dependencies from kinds of plugins to plugin.rb, and reduce duplicated implementation between plugin.rb and registry.rb.

### DIFF
--- a/lib/fluent/plugin.rb
+++ b/lib/fluent/plugin.rb
@@ -136,10 +136,10 @@ module Fluent
       case
       when obj.is_a?(Class)
         obj.new
-      when obj.respond_to?(:call)
+      when obj.respond_to?(:call) && obj.arity == 0
         obj.call
       else
-        raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable."
+        raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable (without arguments)."
       end
     end
   end
--- a/lib/fluent/plugin.rb
+++ b/lib/fluent/plugin.rb
@@ -14,166 +14,133 @@
 #    limitations under the License.
 #
 
+require 'fluent/registry'
 require 'fluent/config/error'
 
 module Fluent
-  class PluginClass
-    # This class is refactored using Fluent::Registry at v0.14
+  class Plugin
+    SEARCH_PATHS = []
 
-    def initialize
-      @input = {}
-      @output = {}
-      @filter = {}
-      @buffer = {}
-    end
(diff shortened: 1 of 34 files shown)


lib/fluent/plugin.rb
```diff
@@ -14,166 +14,133 @@
...
-        path = "fluent/plugin/filter_#{type}"
-      when 'buffer'
-        path = "fluent/plugin/buf_#{type}"
+    def self.new_impl(kind, registry, type)
+      # "'type' not found" is handled by registry
+      obj = registry.lookup(type)
+      case
+      when obj.is_a?(Class)
+        obj.new
+      when obj.respond_to?(:call)
```
repeatedly: The number of arity check is needed?

lib/fluent/registry.rb
```diff
@@ -43,11 +46,18 @@ def lookup(type)
...
+      end
+      nil
+    end
+
     def search(type)
       path = "#{@search_prefix}#{type}"
 
       # prefer LOAD_PATH than gems
-      files = $LOAD_PATH.map { |lp|
+      files = ($LOAD_PATH + @paths).map { |lp|
```
repeatedly: Use `(@paths + $LOAD_PATH)` instead.
Fluentd's plugin pathes should override default installed plugins.

lib/fluent/plugin.rb
tagomoris: Correct. Better to have.

lib/fluent/registry.rb
tagomoris: Okay.

tagomoris:
@repeatedly Can I merge after fixes of your points?

repeatedly:
Yes.

@sonots Could you also check this?

lib/fluent/registry.rb
tagomoris: Ah, this code sorts found list and take last member of it to get latest version of files (it has another incorrectness, but it's out of scope for now).
https://github.com/fluent/fluentd/pull/838/files#diff-da556f72b0de83b0e659240bf9d879bbR66

So we should check files separately for `@paths` and `$LOAD_PATH`.

repeatedly: okay

sonots:
LGTM

tagomoris:
OK. I'll merge this after green (Travis is already ok. Waiting appveyor).

tagomoris:
An Appveyor task in a job is failing weird error and there's no way to rebuild it... I'll merge this :P

PR #843, opened 2016-03-10
add new Plugin::Base class and PluginHelper modules

tagomoris:
PluginHelper provides utilities to write plugins:
* high level methods to hide underlying features and options (e.g., threads, cool.io, ...)
* pre-configured suite nice to use in plugin code (e.g., threads marked as abort_on_exception in default)
* well organized test driver integration (in future): test code runs after threads start to run correctly (no more sleeps in test code!)

And change Fluent::Plugin from class to module (because it can't be instantiated) and add Fluent::Plugin::Base class as a base class of all plugins, to define methods for plugin life cycles which are used from plugin helpers.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -245,6 +245,10 @@ module Fluent
       @out.flush
     end
 
+    def reset
+      @out.reset if @out.respond_to?(:reset)
+    end
+
     private
 
     def dump_stacktrace(backtrace, level)
--- a/lib/fluent/plugin.rb
+++ b/lib/fluent/plugin.rb
@@ -18,7 +18,7 @@ require 'fluent/registry'
 require 'fluent/config/error'
 
 module Fluent
-  class Plugin
+  module Plugin
     SEARCH_PATHS = []
 
     # plugins for fluentd:         fluent/plugin/type_NAME.rb
@@ -34,24 +34,6 @@ module Fluent
     # TODO: plugin storage
 
     REGISTRIES = [INPUT_REGISTRY, OUTPUT_REGISTRY, FILTER_REGISTRY, BUFFER_REGISTRY, PARSER_REGISTRY, FORMATTER_REGISTRY]
-
-    def self.lookup_type_from_class(klass_or_its_name)
-      klass = if klass_or_its_name.is_a? Class
-                klass_or_its_name
-              elsif klass_or_its_name.is_a? String
-                eval(klass_or_its_name) # const_get can't handle qualified klass name (ex: A::B)
-              else
-                raise ArgumentError, "invalid argument type #{klass_or_its_name.class}: #{klass_or_its_name}"
-              end
-      REGISTRIES.reduce(nil){|a, r| a || r.reverse_lookup(klass) }
-    end
-
-    def self.add_plugin_dir(dir)
-      REGISTRIES.each do |r|
-        r.paths.push(dir)
-      end
-      nil
-    end
 
     def self.register_input(type, klass)
       register_impl('input', INPUT_REGISTRY, type, klass)
@@ -87,6 +69,24 @@ module Fluent
       else
         register_impl('formatter', FORMATTER_REGISTRY, type, klass_or_proc)
       end
+    end
+
+    def self.lookup_type_from_class(klass_or_its_name)
+      klass = if klass_or_its_name.is_a? Class
+                klass_or_its_name
+              elsif klass_or_its_name.is_a? String
+                eval(klass_or_its_name) # const_get can't handle qualified klass name (ex: A::B)
+              else
+                raise ArgumentError, "invalid argument type #{klass_or_its_name.class}: #{klass_or_its_name}"
+              end
+      REGISTRIES.reduce(nil){|a, r| a || r.reverse_lookup(klass) }
+    end
+
+    def self.add_plugin_dir(dir)
+      REGISTRIES.each do |r|
+        r.paths.push(dir)
+      end
+      nil
     end
 
     def self.new_input(type)
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -28,13 +28,73 @@ module Fluent
       include PluginLoggerMixin
       include PluginHelper::Mixin
 
-      def initialize; end
-      def configure(conf); end
-      def start; end
-      def stop; end
-      def shutdown; end
-      def close; end
-      def terminate; end
+      State = Struct.new(:configure, :start, :stop, :shutdown, :close, :terminate)
+
+      def initialize
+        super
+        @state = State.new(false, false, false, false, false, false)
+      end
+
+      def has_router?
+        false
+      end
+
+      def configure(conf)
+        super
+        @state.configure = true
+        self
+      end
+
+      def start
+        @log.reset
+        @state.start = true
+        self
+      end
+
+      def stop
+        @state.stop = true
+        self
+      end
+
+      def shutdown
+        @state.shutdown = true
+        self
+      end
+
+      def close
+        @state.close = true
+        self
+      end
+
+      def terminate
+        @state.terminate = true
+        @log.reset
+        self
+      end
+
+      def configured?
+        @state.configure
+      end
+
+      def started?
+        @state.start
+      end
+
+      def stopped?
+        @state.stop
+      end
+
+      def shutdown?
+        @state.shutdown
+      end
+
+      def closed?
+        @state.close
+      end
+
+      def terminated?
+        @state.terminate
+      end
     end
   end
 end
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -1,0 +1,40 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/plugin'
+require 'fluent/configurable'
+require 'fluent/plugin_id'
+require 'fluent/log'
+require 'fluent/plugin_helper'
+
+module Fluent
+  module Plugin
+    class Base
+      include Configurable
+      include PluginId
+      include PluginLoggerMixin
+      include PluginHelper::Mixin
+
+      def initialize; end
+      def configure(conf); end
+      def start; end
+      def stop; end
+      def shutdown; end
+      def close; end
+      def terminate; end
+    end
+  end
+end
--- a/lib/fluent/plugin_helper.rb
+++ b/lib/fluent/plugin_helper.rb
@@ -1,0 +1,36 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/plugin_helper/event_emitter'
+require 'fluent/plugin_helper/thread'
+require 'fluent/plugin_helper/event_loop'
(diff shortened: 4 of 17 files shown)


tagomoris:
First bunch is for:
* EventEmitter: handles `@router`
  * to separate output plugin from input plugin
  * to show which output plugins may emit records explicitly
* Thread: create/manage threads
* EventLoop: create/manage event loops of cool.io (event loop per plugin instance)
* Timer: no more threads/sleeps for timers in plugin code
* ChildProcess: handle process execution and terminating

lib/fluent/plugin.rb
```diff
@@ -35,24 +35,6 @@ class Plugin
 
     REGISTRIES = [INPUT_REGISTRY, OUTPUT_REGISTRY, FILTER_REGISTRY, BUFFER_REGISTRY, PARSER_REGISTRY, FORMATTER_REGISTRY]
 
-    def self.lookup_type_from_class(klass_or_its_name)
```
tagomoris: Methods for internal use should be under public use.

tagomoris:
@repeatedly please take a glance before merge.

lib/fluent/log.rb
```diff
@@ -245,6 +245,10 @@ def flush
       @out.flush
     end
 
+    def reset
+      @out.reset if @out.respond_to?(:reset)
```
repeatedly: IO object doesn't have `reset` method.
What the object do you assume?

lib/fluent/plugin/base.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      def started?
+        @state.start
+      end
+
+      def stopped?
+        @state.stop
+      end
+
+      def shutdowned?
```
repeatedly: The past form of `shutdown` is `shutdown`

lib/fluent/plugin/base.rb
```diff
@@ -0,0 +1,96 @@
...
+        self
+      end
+
+      def terminate
+        @state.terminate = true
+        @log.reset
+        self
+      end
+
+      State = Struct.new(:configure, :start, :stop, :shutdown, :close, :terminate)
```
repeatedly: Define before `initialize` is better

lib/fluent/plugin_helper/child_process.rb
```diff
@@ -0,0 +1,208 @@
...
+            if !parallel && s.running
+              log.warn "previous child process is still running. skipped.", title: title, command: command, arguments: arguments, interval: interval, parallel: parallel
+            else
+              child_process_execute_once(title, command, arguments, subprocess_name, read, write, internal_encoding, external_encoding, &callback)
+            end
+          end
+        end
+      end
+
+      ProcState = Struct.new(:running)
```
repeatedly: Why use `Struct`?
Do you have a plan to add other attributes?

lib/fluent/plugin_helper/thread.rb
```diff
@@ -0,0 +1,116 @@
...
+          m.unlock
+          thread_exit = false
+          ::Thread.current[:_fluentd_plugin_helper_thread_title] = title
+          ::Thread.current[:_fluentd_plugin_helper_thread_started] = true
+          ::Thread.current[:_fluentd_plugin_helper_thread_running] = true
+          begin
+            yield *t_args
+            thread_exit = true
+          rescue => e
+            log.warn "thread exited by unexpected error", plugin: self.class, title: title, error_class: e.class, error: e
```
repeatedly: should set `true` to `thread_exit`?

lib/fluent/plugin_helper/event_emitter.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        if label_name = conf['@label']
+          label = Engine.root_agent.find_label(label_name)
+          @router = label.event_router
+        elsif @router.nil?
+          @router = Engine.root_agent.event_router
+        end
+      end
+
+      def emits?
```
repeatedly: I'm not sure when use this method but `emittable` is better?

lib/fluent/log.rb
repeatedly: Ah, Is this for test purpose?

lib/fluent/plugin_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+
+require 'fluent/plugin_helper/event_emitter'
+require 'fluent/plugin_helper/thread'
+require 'fluent/plugin_helper/event_loop'
+require 'fluent/plugin_helper/timer'
+require 'fluent/plugin_helper/child_process'
+
+module Fluent
+  module PluginHelper
+    module Mixin
```
repeatedly: If only one method, how about merging into `Base`?

lib/fluent/plugin_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+      def self.included(mod)
+        mod.extend(Fluent::PluginHelper)
+      end
+    end
+
+    def helpers(*snake_case_symbols)
```
repeatedly: If we support 3rd party plugin helper, `require "fluent/plugin/#{helper_module}"` and `include` it is better.

lib/fluent/plugin_helper/event_loop.rb
```diff
@@ -0,0 +1,114 @@
...
+      # Currently this plugin helper is only for other helpers, not plugins.
+      # there's no way to create customized watchers to attach event loops.
+      include Fluent::PluginHelper::Thread
+
+      # stop     : [-]
+      # shutdown : detach all event watchers on event loop
+      # close    : stop event loop
+      # terminate: initialize internal state
+
+      EVENT_LOOP_RUN_DEFAULT_TIMEOUT = 0.2
```
repeatedly: in_forward uses `0.5` by default.
Keep same value is better? > https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_forward.rb#L42

lib/fluent/log.rb
tagomoris: Yes. This is to initialize lines in logger between tests.

lib/fluent/plugin/base.rb
tagomoris: oh... does `shutdown?` look good to you?

lib/fluent/plugin/base.rb
tagomoris: I want not to write code for internal use if possible. But `State` is used in `initialize`... Okay, i'll fix it.

lib/fluent/plugin_helper.rb
tagomoris: Do you mean to add `self.helpers` method to `Fluent::Plugin::Base`?
I think here is better, because it keeps `Base` class a bit simpler than adding all to `Base`.
We should require all helpers from anywhere either way, and require the file which has many `require` for all helpers from `Fluent::Plugin::Base`.

lib/fluent/plugin_helper.rb
tagomoris: I have no plan to make it possible to add plugin helpers for 3rd party modules. Plugin helpers will provide many core features (to listen sockets shared between workers, to create event loops, ...) and will be connected strongly with test drivers. It is very hard to maintain compatibilities for 3rd party helpers.
Is it different from your opinion?

lib/fluent/plugin_helper/child_process.rb
tagomoris: Oh, it's because of my first plan... I'll fix it.

lib/fluent/plugin_helper/event_loop.rb
tagomoris: Okay.

lib/fluent/plugin_helper/thread.rb
tagomoris: Correct.

lib/fluent/plugin_helper/event_emitter.rb
tagomoris: This method is called when Fluentd core checks the plugin is to emit events or not.
Plugin does or doesn't emit, so it's not possibility, I think.

lib/fluent/plugin_helper.rb
repeatedly: Currently we have XXXMixin to help plugin implementaion.
Maybe, existing PluginHelper:::XXX and XXXMixin is very confusing.
Migrating into PluginHelper is better for developers.

lib/fluent/plugin_helper/event_emitter.rb
repeatedly: I see. But `emits` seems not clear name.
`emittable?` or `can_emit?` is better for me.

lib/fluent/plugin_helper.rb
tagomoris: Plugin helper doesn't provide methods to override plugin methods (like `format` or others), and also doesn't overwrite any behavior of plugins by just included.
On the other hand, mixins can do everything. In my opinion, these should be separated from each other.

(Along this opinion, EventEmitter looks not good as a plugin helper, but good as a mixin. Hmm.)

lib/fluent/plugin_helper/event_emitter.rb
tagomoris: How about `has_router?`

repeatedly: Ok for me.

tagomoris: OK. I'll add that method to `Fluent::Plugin::Base` too. (I missed it)

lib/fluent/plugin_helper/thread.rb
tagomoris: Fixed.

tagomoris:
I pushed fixed for comments.

lib/fluent/plugin_helper.rb
repeatedly: > Plugin helper doesn't provide methods to override plugin methods (like format or others), and also doesn't overwrite any behavior of plugins by just included.

Does this depend on PluginHelper implementation or do you have any validation mechanizm?
Plugins use both with  `include XXX` code.
Seperating responsibility is good but it introduces complicated mechanizm.
And `Mixin` is too ruby-ish name so users may assume PluginHelper is good for providing development support like TypeConverter.

lib/fluent/plugin_helper/event_loop.rb
```diff
@@ -0,0 +1,114 @@
...
+        end
+
+        super
+      end
+
+      def terminate
+        @_event_loop = nil
+        @_event_loop_running = false
+        @_event_loop_mutex = nil
+        @_event_loop_run_timeout = EVENT_LOOP_RUN_DEFAULT_TIMEOUT
```
repeatedly: Resetting this value is needed?
If needed, `nil` is better?

lib/fluent/plugin_helper/child_process.rb
```diff
@@ -0,0 +1,206 @@
...
+      # terminate: [-]
+
+      attr_reader :_child_process_processes # for tests
+
+      def child_process_execute(
+          title, command,
+          arguments: nil, subprocess_name: nil, interval: nil, immediate: false, parallel: false,
+          read: true, write: true, internal_encoding: 'utf-8', external_encoding: 'utf-8', &block
+      )
+        raise ArgumentError, "BUG: title must be a symbol" unless title.is_a? Symbol
```
repeatedly: There are lots of `raise ArgumentError, "BUG: message" if cond`.
How about providing check helper in Plugin like Guava's preconditions?

```rb
Plugin.check_param(cond, 'message');
# or 
PluginHelper.check_param(cond, 'message');
```

lib/fluent/plugin_helper/event_loop.rb
tagomoris: Correct. `nil` is better.
In basic design of new plugin API, plugins should be re-initialized after `terminate`. It's good for many things including tests.

lib/fluent/plugin_helper/child_process.rb
tagomoris: Calling checker methods and implementation methods twice looks much worse for me... I think these easy mistakes can be found by very simple testing.
I want to keep writing plugins as simple as possible.

lib/fluent/plugin_helper.rb
tagomoris: Migrating some mixins in fluentd core to Plugin Helpers are not bad idea if it can be recreated with limited APIs and no implicit overriding of methods. I agree about your opinion that `Mixin` is too ruby-ish and to separate responsibilities to each helpers (with changing APIs from mixin to helpers).
Validating mechanism doesn't exist, because Plugin Helpers are only in Fluentd core.
I have no plan to prohibit to create any mixins (for 3rd party authors), but moving toward the direction without any mixins looks good especially in Fluentd core.

tagomoris:
I pushed more fixes again.

lib/fluent/plugin_helper.rb
tagomoris: I think it's better to migrate mixins to plugin helpers, to change to use explicit call to wrap plugin behaviors:
```ruby
  helpers :recordfilter
  def start
    super
    recordfilter_wrap(:filter_stream, :set_time_key)
  end
```

tagomoris:
Now it's green on CIs.

tagomoris:
Merged!

PR #844, opened 2016-03-11
Fix overwrite subsection

okkez:
* overwriting config_section options like required/multi/alias breaks
  behavior of `super`
 * so subclass must not overwrite these options
* subsection definitions of subsections (Hash) is simply merged
 * it is bug, because same section definition like <a> is overwritten by
   subclass
 * under this bug, subclass MUST repeat superclass's definition, and add
   original definition
 * but this is impossible in the real world
 * subsection definitions should be merged correctly

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -69,15 +69,19 @@ module Fluent
       def merge(other) # self is base class, other is subclass
         return merge_for_finalized(other) if self.final?
 
-        options = {
-          param_name: other.param_name,
-          required: (other.required.nil? ? self.required : other.required),
-          multi: (other.multi.nil? ? self.multi : other.multi)
-        }
+        options = {}
+        # param_name is used not to ovewrite plugin's instance
+        # varible, so this should be able to be overwritten
+        options[:param_name] = other.param_name
+        # subclass cannot overwrite base class's definition
+        options[:required] = @required.nil? ? other.required : self.required
+        options[:multi] = @multi.nil? ? other.multi : self.multi
+        options[:alias] = @alias.nil? ? other.alias : self.alias
+
         merged = self.class.new(other.name, options)
 
         merged.argument = other.argument || self.argument
-        merged.params = self.params.merge(other.params)
+        merged.params = other.params.merge(self.params)
         merged.defaults = self.defaults.merge(other.defaults)
         merged.sections = {}
         (self.sections.keys + other.sections.keys).uniq.each do |section_key|
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -115,12 +115,12 @@ module Fluent
         #  * overwrite param_name to escape duplicated name of instance variable
         #  * append params/defaults/sections which are missing in superclass
 
-        options = {
-          param_name: other.param_name,
-          required: (self.required.nil? ? other.required : self.required),
-          multi: (self.multi.nil? ? other.multi : self.multi),
-          final: true,
-        }
+        options = {}
+        options[:param_name] = other.param_name
+        options[:required] = @required.nil? ? other.required : self.required
+        options[:multi] = @multi.nil? ? other.multi : self.multi
+        options[:alias] = @alias.nil? ? other.alias : self.alias
+        options[:final]  = true
         merged = self.class.new(other.name, options)
 
         merged.argument = self.argument || other.argument
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -87,6 +87,7 @@ module Fluent
         options[:required] = @required.nil? ? other.required : self.required
         options[:multi] = @multi.nil? ? other.multi : self.multi
         options[:alias] = @alias.nil? ? other.alias : self.alias
+        options[:final] = @final || other.final
 
         merged = self.class.new(other.name, options)
 
@@ -121,6 +122,7 @@ module Fluent
         options[:multi] = @multi.nil? ? other.multi : self.multi
         options[:alias] = @alias.nil? ? other.alias : self.alias
         options[:final]  = true
+
         merged = self.class.new(other.name, options)
 
         merged.argument = self.argument || other.argument
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -120,6 +120,16 @@ module Fluent
           raise ConfigError, "BUG: subclass cannot overwrite finalized base class's config_section"
         end
 
+        if overwrite?(other, :required)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: required"
+        end
+        if overwrite?(other, :multi)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: multi"
+        end
+        if overwrite?(other, :alias)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: alias"
+        end
+
         options = {}
         options[:param_name] = other.param_name
         options[:required] = @required.nil? ? other.required : self.required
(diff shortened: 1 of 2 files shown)


okkez:
@tagomoris Review this PR, please.

lib/fluent/config/configure_proxy.rb
```diff
@@ -63,21 +63,36 @@ def multi?
       end
 
       def final?
-        @final
+        @final == true
```
tagomoris: `!!@final` looks normal.

lib/fluent/config/configure_proxy.rb
```diff
@@ -63,21 +63,36 @@ def multi?
...
       def merge(other) # self is base class, other is subclass
         return merge_for_finalized(other) if self.final?
 
-        options = {
-          param_name: other.param_name,
-          required: (other.required.nil? ? self.required : other.required),
-          multi: (other.multi.nil? ? self.multi : other.multi)
-        }
+        if overwrite?(other, :required)
+          raise ConfigError, "subclass cannot overwrite base class's definition"
```
tagomoris: It's better to have field name `required` to fail to overwrite.

lib/fluent/config/configure_proxy.rb
```diff
@@ -63,21 +63,36 @@ def multi?
...
+        end
+        if overwrite?(other, :multi)
+          raise ConfigError, "subclass cannot overwrite base class's definition"
```
tagomoris: ditto.

lib/fluent/config/configure_proxy.rb
```diff
@@ -63,21 +63,36 @@ def multi?
...
+        end
+        if overwrite?(other, :alias)
+          raise ConfigError, "subclass cannot overwrite base class's definition"
```
tagomoris: ditto.

test/config/test_configurable.rb
```diff
@@ -164,10 +148,115 @@ class Example6
...
+      end
+    end
+
+    class DetailAddressDefault < Base
+      config_section :detail do
+        config_param :address, :string, default: "y"
+      end
+    end
+
+    class AddParam < Base
```
tagomoris: It's ideal to have another class definition only to add `config_param`, without changing default value of `address`.

lib/fluent/config/configure_proxy.rb
```diff
@@ -101,12 +116,19 @@ def merge_for_finalized(other)
...
+        end
+
+        options = {}
+        options[:param_name] = other.param_name
+        options[:required] = @required.nil? ? other.required : self.required
+        options[:multi] = @multi.nil? ? other.multi : self.multi
+        options[:alias] = @alias.nil? ? other.alias : self.alias
+        options[:final]  = true
+
+        other.final = true
```
tagomoris: Is it needed? This code changes subclass definition itself, not `merged` one.

test/config/test_configurable.rb
```diff
@@ -604,93 +683,140 @@ def e(name, arg = '', attrs = {}, elements = [])
...
+          test 'subclass can overwrite appendix.name, appendix.age w/ level 2' do
+            finalized = ConfigurableSpec::Final::Finalized.new
+            finalized.configure(CONF)
+            expected = { name: "y", age: 10 }
+            actual = { name: finalized.appendix.name, age: finalized.appendix.age }
+            assert_equal(expected, actual)
+          end
+
+          test 'subclass cannot overwrite appendix.name, appendix.age w/ level 3' do
+            level3 = ConfigurableSpec::Final::InheritsFinalized.new
```
tagomoris: Is it intentional not to raise exceptions for overwriting required/multi/alias onto finalized subsections, besides to raise exceptions for overwriting these onto non-finalized subsections?

lib/fluent/config/configure_proxy.rb
```diff
@@ -63,21 +63,36 @@ def multi?
...
+          raise ConfigError, "subclass cannot overwrite base class's definition"
+        end
+        if overwrite?(other, :alias)
+          raise ConfigError, "subclass cannot overwrite base class's definition"
+        end
+
+        options = {}
+        # param_name is used not to ovewrite plugin's instance
+        # varible, so this should be able to be overwritten
+        options[:param_name] = other.param_name
```
tagomoris: :+1:

lib/fluent/config/configure_proxy.rb
tagomoris: And, could you add header to show that it's a bug of plugin, not users' misconfigurations (e.g., "BUG:")

lib/fluent/config/configure_proxy.rb
```diff
@@ -101,12 +116,19 @@ def merge_for_finalized(other)
...
         #  * append params/defaults/sections which are missing in superclass
 
-        options = {
-          param_name: other.param_name,
-          required: (self.required.nil? ? other.required : self.required),
-          multi: (self.multi.nil? ? other.multi : self.multi),
-          final: true,
-        }
+        if other.final == false && overwrite?(other, :final)
+          raise ConfigError, "subclass cannot overwrite finalized base class's definition"
```
tagomoris: Same with required, multi and alias.

tagomoris:
Great fix!
I added some comments about some questions and some things nice to have.

lib/fluent/config/configure_proxy.rb
okkez: OK, I will fix.

lib/fluent/config/configure_proxy.rb
okkez: No, this should be removed in 5884b47

lib/fluent/config/configure_proxy.rb
okkez: I will fix exception message.

test/config/test_configurable.rb
okkez: I got it. I will raise exception when overwrite default value in subclass.

lib/fluent/config/configure_proxy.rb
tagomoris: I cannot find such diff.
https://github.com/fluent/fluentd/pull/844/commits/5884b476593a2d6d47dba89a63913ec74670ebba

okkez: I forgot to remove `other.final = true` in https://github.com/fluent/fluentd/pull/844/commits/5884b476593a2d6d47dba89a63913ec74670ebba

test/config/test_configurable.rb
okkez: in_tcp and in_udp overwrites default value of base class.

* https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tcp.rb#L25
* https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_udp.rb#L23

Do you want to raise exception in such case?

test/config/test_configurable.rb
okkez: No this is bug for my implementation.
I will add test and fix the bug.

test/config/test_configurable.rb
tagomoris: Oh, I failed to tell what I meant correctly. I didn't mention about any code change at here.
I just want more test cases:
* current `AddParam` changes default value of `address`, and add a parameter `phone_no`
* I want one more test pattern class `AddParam2` to add `phone_no`, without definition about `address`

okkez: Sorry for misunderstanding :bow: 
I've added test pattern without definition about `address`.

tagomoris: :+1:

okkez:
I've fixed and pushed changes for all comments.

tagomoris:
OK, Tests are green (after retries).
@okkez Could you squash changes?

okkez:
@tagomoris Ok, I will squash and force push later.

okkez:
Squashed.

okkez:
Travis CI failed, but it is not related to my changes.

tagomoris:
I requested rebuild and now it's green.

tagomoris:
Merged. Thank you!

PR #846, opened 2016-03-11
Rakefile: Suppress warnings from ruby, enabled by rake 11

cosmo0920:
In Rake 11.0.1 or later, `t.warning = true` is by default. This behavior had been introduced in https://github.com/ruby/rake/pull/97.
Should we suppress warnings from the new rake?
Or should we resolve warnings in Fluentd side?

IMO, concealing warnings is not good for Fluend. It reports potential problem which should be fixed in the future.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,8 +14,8 @@ Rake::TestTask.new(:base_test) do |t|
   #  $ bundle exec rake base_test TEST=test/test_*.rb
   t.libs << "test"
   t.test_files = Dir["test/**/test_*.rb"].sort
-  t.verbose = true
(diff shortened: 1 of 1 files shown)


tagomoris:
Warnings should be resolved. I think so too.
But resolving all warnings needs long time, hard work and many changes on our code. It should be done after big changes for next major release.

I created issue for that. https://github.com/fluent/fluentd/issues/848
So I want to merge this for now to show CI results clearly.

tagomoris:
Merged.

cosmo0920:
Thanks!

PR #847, opened 2016-03-11
Add buffered output plugins for tests

cosmo0920:
Revised: #518.

And I've added tests for added buffered output plugins for testing.

## Remaining Tasks

- [ ] Rewrite with new TestDriver and new v0.14 API (Not Yet)

### DIFF
--- a/lib/fluent/plugin/out_buffered_stdout.rb
+++ b/lib/fluent/plugin/out_buffered_stdout.rb
@@ -14,26 +14,47 @@
 #    limitations under the License.
 #
 
-require 'fluent/output'
+require 'fluent/plugin/output'
 
-module Fluent
-  class BufferedStdoutOutput < ObjectBufferedOutput
-    Plugin.register_output('buffered_stdout', self)
+module Fluent::Plugin
+  class BufferedStdoutOutput < Output
+    Fluent::Plugin.register_output('buffered_stdout', self)
 
     desc 'Output format.(json,hash)'
     config_param :output_type, default: 'json'
+    config_section :buffer do
+      config_set_default :chunk_keys, ['tag']
+      config_set_default :flush_at_shutdown, true
+      config_set_default :chunk_limit_size, 10 * 1024
+    end
+
+    attr_accessor :delayed
+
+    def initialize
+      super
+      @delayed = false
+    end
+
+    def prefer_delayed_commit
+      @delayed
+    end
 
     def configure(conf)
       super
-      @formatter = Plugin.new_formatter(@output_type)
+      @formatter = Fluent::Plugin.new_formatter(@output_type, parent: self)
       @formatter.configure(conf)
     end
 
-    def write_objects(tag, chunk)
-      chunk.msgpack_each {|time,record|
-        log.write "#{Time.at(time).localtime} #{tag}: #{@formatter.format(tag, time, record).chomp}\n"
-      }
-      log.flush
+    def write(chunk)
+      chunk.write_to($log)
+    end
+
+    def try_write(chunk)
+      chunk.write_to($log)
+    end
+
+    def format(tag, time, record)
+      "#{Time.at(time).localtime} #{tag}: #{@formatter.format(tag, time, record).chomp}\n"
     end
   end
 end
--- a/lib/fluent/plugin/out_buffered_stdout.rb
+++ b/lib/fluent/plugin/out_buffered_stdout.rb
@@ -28,6 +28,10 @@ module Fluent
       @formatter.configure(conf)
     end
 
+    def format(tag, time, record)
+      @formatter.format(tag, time, record)
+    end
+
     def write(chunk)
       chunk.msgpack_each do |time, tag, record|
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/out_buffered_stdout.rb
```diff
@@ -0,0 +1,38 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/output'
+
+module Fluent
+  class BufferedStdoutOutput < BufferedOutput
```
tagomoris: Is this really subclass of `BufferedOutput`? not `ObjectBufferedOutput`?
There seems no `format` method to be called.

cosmo0920: Thanks for the comment.
I think that `ObjectBufferedOutput` is not enough for this plugin because `ObjectBufferedOutput` does not implement format itself.
It seems to need to implement `format` in this plugin.

tagomoris: Strictly speaking:
* There's no need to implement `format` method if the plugin is a subclass of `ObjectBufferedOutput`
  * `ObjectBufferedOutput` plugins do NOT call `format` in its `emit`, and write msgpack binaries into buffer directly
  * https://github.com/fluent/fluentd/blob/master/lib/fluent/output.rb#L414
  * its chunks can be iterated using `chunk.msgpack_each` in `write_objects`, just like `out_forward` is doing
  * https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward.rb#L197
* It's needed to implement `format` method if a plugin is a subclass of `BufferedOutput`
  * BufferedOutput plugins write return values of `format` into chunks
  * Return values of `format` method MUST be msgpack binaries if you want to do `msgpack_each` for chunks

cosmo0920: Oh, I see.
Sorry for my misunderstanding.... :bow: 
I use `ObjectBufferedOutput` instead of `BufferedOutput`.

lib/fluent/plugin/out_buffered_stdout.rb
```diff
@@ -0,0 +1,41 @@
...
+    desc 'Output format.(json,hash)'
+    config_param :output_type, default: 'json'
+
+    def configure(conf)
+      super
+      @formatter = Plugin.new_formatter(@output_type)
+      @formatter.configure(conf)
+    end
+
+    def emit(tag, es, chain)
```
tagomoris: Along with `ObjectBufferedOutput` way, it should be `def write_objects(tag, chunk)`.
There's no need to handle `chain` in `write_object`.

test/plugin/test_out_buffered_stdout.rb
```diff
@@ -0,0 +1,33 @@
...
+    assert_equal 'json', d.instance.output_type
+
+    d = create_driver(CONFIG + "\noutput_type hash")
+    assert_equal 'hash', d.instance.output_type
+
+    assert_raise(Fluent::ConfigError) do
+      d = create_driver(CONFIG + "\noutput_type foo")
+    end
+  end
+end
```
tagomoris: Are there any reasons not to add tests to emit events actually?

cosmo0920: One reason is: There is no way to obtain STDOUT immediately in tests.
My understanding about buffered output plugin is wrong?

tagomoris: In tests, logger is configured with `Fluent::Test::DummyLogDevice`. It stores logs in an array, and these logs can be accessed by `driver.instance.log.out.logs`.
https://github.com/fluent/fluentd/blob/master/lib/fluent/test/base.rb#L95

I'm doing it in a pull-request diff.
https://github.com/fluent/fluentd/pull/843/files#diff-a741a1b1e75fe157fb8890bdd1a7040dR180

cosmo0920: Thanks!
I'll rebase this PR and try to use it. :smile:

cosmo0920: Umm..., there is no suitable test driver to test `ObjectBufferedOutput`.... :confounded:

tagomoris: Oh, I missed that there's no good test driver for ObjectBufferedOutput.

tagomoris:
@cosmo0920 Can you leave this pull-request as is for a while?
I think we can add these only for v0.14, and that version will have another (well-organized) test drivers.

cosmo0920:
Yup, I would like to leave as is for now.

tagomoris:
I'm sorry but I implemented `out_buffered_output` in #972 because it's very good for a testbed of test drivers.

tagomoris:
Hmm, a build on AppVeyor failed, but there are no details.

tagomoris:
Thank you!

PR #852, opened 2016-03-15
Suppress a lot of warnings

okkez:
This is the first step to wipe out warnings.

### DIFF
--- a/lib/fluent/mixin.rb
+++ b/lib/fluent/mixin.rb
@@ -133,13 +133,18 @@ module Fluent
     attr_accessor :remove_tag_prefix, :remove_tag_suffix, :add_tag_prefix, :add_tag_suffix
     def configure(conf)
       super
-      if remove_tag_prefix = conf['remove_tag_prefix']
-        @remove_tag_prefix = Regexp.new('^' + Regexp.escape(remove_tag_prefix))
-      end
 
-      if remove_tag_suffix = conf['remove_tag_suffix']
-        @remove_tag_suffix = Regexp.new(Regexp.escape(remove_tag_suffix) + '$')
-      end
+      @remove_tag_prefix = if conf.has_key?('remove_tag_prefix')
+                             Regexp.new('^' + Regexp.escape(conf['remove_tag_prefix']))
+                           else
+                             nil
+                           end
+
+      @remove_tag_suffix = if conf.has_key?('remove_tag_suffix')
+                             Regexp.new(Regexp.escape(conf['remove_tag_suffix']) + '$')
+                           else
+                             nil
+                           end
 
       @add_tag_prefix = conf['add_tag_prefix']
       @add_tag_suffix = conf['add_tag_suffix']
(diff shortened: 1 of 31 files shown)


okkez:
Some warnings may conflict with other PRs, so I leave these warnings.

lib/fluent/parser.rb
```diff
@@ -155,18 +155,10 @@ def parse(text)
             when @time_key
               time = @mutex.synchronize { @time_parser.parse(value) }
               if @keep_time_key
-                record[name] = if @type_converters.nil?
-                                 value
-                               else
-                                 convert_type(name, value)
-                               end
+                record[name] = convert_type(name, value)
```
repeatedly: Calling `convert_type` heavily is high cost.
So we should not call `convert_type` when user doesn't specify `types` parameter.

lib/fluent/mixin.rb
```diff
@@ -133,10 +133,12 @@ module HandleTagNameMixin
     attr_accessor :remove_tag_prefix, :remove_tag_suffix, :add_tag_prefix, :add_tag_suffix
     def configure(conf)
       super
+      @remove_tag_prefix = nil
```
repeatedly: How about `@remove_tag_prefix = conf.has_key?('remove_tag_prefix') ? Regexp.new('^' + Regexp.escape(conf['remove_tag_prefix'])) : nil`?

lib/fluent/mixin.rb
```diff
@@ -133,10 +133,12 @@ module HandleTagNameMixin
...
       if remove_tag_prefix = conf['remove_tag_prefix']
         @remove_tag_prefix = Regexp.new('^' + Regexp.escape(remove_tag_prefix))
       end
 
+      @remove_tag_suffix = nil
```
repeatedly: ditto

okkez:
I have fixed and pushed.
How about this?

tagomoris:
LGTM. (Squash required after all reviewer comments.)

sonots:
Is it possible to integrate ruby warning check into CI? Otherwise, I think same things happen again in the future.

sonots:
Good job anyway, this allows plugin developers to use ruby -w for their own plugins. Before, fluentd itself showed lots of warnings, so ruby -w for plugins did not work effectively.

tagomoris:
@sonots what is "ruby warning check" you mentioned?

sonots:
Checking whether ruby -w tells warnings or not

tagomoris:
It's impossible to remove all warnings immediately (because there are some/many warnings from libraries which fluentd depends on).
IMO, checking `ruby -w` outputs looks good idea at a point in future, but we cannot add it right now.

okkez:
@tagomoris squashed!

tagomoris:
@okkez Thanks! I'll merge this after Travis green (AppVeyor looks in trouble now).

tagomoris:
Merged. Thank you!

PR #853, opened 2016-03-15
socket_util: Add missing require

cosmo0920:
Otherwise, I get `NameError: uninitialized constant Fluent::Input` error when I execute `bundle exec ruby test/plugin/test_out_forward.rb`.


cosmo0920:
Thanks!

PR #854, opened 2016-03-15
Override default variables with SystemConfig

cosmo0920:
I've implemented to override default variables with system config mechanism.

This PR is related to this developer task: make process global constants obsoleve in fluent/env

ref: https://github.com/fluent/fluentd/wiki/Developer-notes-for-fixes-on-v0.14#plugin-miscellaneous

## Remaining Task(s)

- [x] Write test code to confirm override feature
  - [x] test code for overriding port
  - [x] test code for overriding file permission
  - [x] test code for overriding dir permission
      - [x] out_file
      - [x] buf_file
  - [x] Omit permission related tests when running on Windows
  - [x] remove DEFAULT_* from env.rb & add default values of ports or permissions to each plugins
  - [x] remove  xxx_override methods and use ` perm = @perm_param_of_plugin || system_config.file_permissions || PLUGIN_DEFAULT_PERM` in plugin
  - [x] squash puzzling commits
  - [x] Tidying up tests
  - [x] Remove port number overriding mechanism. Because it is useless.

## Question(s)

- ~~Where is suitable place to put into obsoleted variables? Or, leave them as-is is enough?~~ Solved.

### DIFF
--- a/lib/fluent/system_config.rb
+++ b/lib/fluent/system_config.rb
@@ -46,6 +46,12 @@ module Fluent
     config_param :rpc_endpoint, :string, default: nil
     config_param :enable_get_dump, :bool, default: nil
     config_param :process_name, default: nil
+    config_param :file_permission, default: nil do |v|
+      v.to_i(8)
+    end
+    config_param :dir_permission, default: nil do |v|
+      v.to_i(8)
+    end
 
     def self.create(conf)
       systems = conf.elements.select { |e|
@@ -77,6 +83,8 @@ module Fluent
       s.rpc_endpoint = @rpc_endpoint
       s.enable_get_dump = @enable_get_dump
       s.process_name = @process_name
+      s.file_permission = @file_permission
+      s.dir_permission = @dir_permission
 
       s
     end
@@ -92,6 +100,8 @@ module Fluent
         @rpc_endpoint = system.rpc_endpoint unless system.rpc_endpoint.nil?
         @enable_get_dump = system.enable_get_dump unless system.enable_get_dump.nil?
         @process_name = system.process_name unless system.process_name.nil?
+        @file_permission = system.file_permission unless system.file_permission.nil?
+        @dir_permission = system.dir_permission unless system.dir_permission.nil?
       }
     end
   end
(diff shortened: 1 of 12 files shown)


cosmo0920:
Oh, NTFS does not support UNIX like permissions!
It should do more work for Windows.... :<

lib/fluent/system_config.rb
```diff
@@ -31,6 +31,33 @@ def system_config_override(opts={})
         @_system_config.send(:"#{key.to_s}=", value)
       end
     end
+
+    def listen_port_override(port=DEFAULT_LISTEN_PORT)
```
tagomoris: `system_config_override` is to override instance variables for each plugin instances. Once it was overridden, following `system_config` method refers instance variable, not global system config instance.
(In other words, plugin just refer `system_config` even in tests after `system_config_override` is called in `setup` or somewhere else.)

So, there's no need to add `xxx_override` methods.

tagomoris:
Default values of port, permission and others are for each plugins, not for Fluentd global.
The best way to fix it is:
* remove `DEFAULT_*` from `env.rb`
* add default values of ports or permissions to each plugins

Especially for file/directory permissions, it's better to provide system config parameters because there's not so variations about it (in many cases, permission control policy is system-wide).

```ruby
  perm = @perm_param_of_plugin || system_config.file_permissions || PLUGIN_DEFAULT_PERM
```

lib/fluent/system_config.rb
```diff
@@ -46,6 +46,13 @@ class SystemConfig
     config_param :rpc_endpoint, :string, default: nil
     config_param :enable_get_dump, :bool, default: nil
     config_param :process_name, default: nil
+    config_param :listen_port, :integer, default: nil
```
tagomoris: Could you remove this line?
There's no need to configure system-wide `listen_port` because this configuration cannot be shared by anything.
Removing this line also makes `port_override` methods useless.

cosmo0920: OK. I'll do it.

repeatedly:
@tagomoris How about this? Can we merge this?

tagomoris:
LGTM. I'll merge this right now.

cosmo0920:
Thanks!

PR #859, opened 2016-03-18
Add missing requires

cosmo0920:
* `Filter` constant is used in `event_router`.
* `Output` class should know filter registry with `require 'filter'`.


cosmo0920:
Thanks!

Issue #801, opened 2016-02-12
in_forward loses messages when restarting td-agent


static-max:
I'm using current td-agent with fluentd-0.12.19 on Ubuntu 14.04 LTS using offical td-agent apt repository.

I have a very simple setup: fluentd receives messages via in_forward and writes them to a file. I emit messages using fluent-logger-java at a rate of 10 messages/second, but the error also shows when the rate is higher. **When I restart fluentd while the java sender is still emitting, messages are dropped/lost at fluentd**. I debugged the java sender and the message that's dropped while restarting fluentd get re-emitted by the sender using its buffer. So the error must be on fluentd side.
How can this happen? I thought a simple restart should by handled by fluentd?! What if fluentd crashes hard in production? Or is the in_forward plugin the wrong way to go?

```
<source>
  type forward
</source>

<match mytags.**>
  @type file
  path /tmp/testfile
</match>
```

repeatedly:
@komamitsu Could you check this?

komamitsu:
Okay. I'll check it later.

komamitsu:
@static-max 
I guess the followings occurred at that time.

1. Fluentd closed the accepted socket when the process was going to finish while fluent-logger-java kept sending messages
2. fluent-logger-java failed to send messages and stored them on the buffer (default 8KB) quietly
3. fluent-logger-java sent the buffered messages to Fluentd when it re-started

In this case, there are two possible causes.

a) In step 1, some messages sent from fluent-logger-java were stored in the OS's TCP buffer on Fluentd side. Calling fluent-logger-java's RawSocketSender#emit() succeeded since it received ACK from the TCP stack of Fluentd. But when the Fluentd process was going to finish, the messages were still buffered in the TCP buffer. As a result, the messages could be lost on Fluentd side.

b) In step 2, fluent-logger-java's send buffer could be full. As a result, some messages could be lost on fluent-logger-java side.

For a), Fluentd can do nothing against this problem without ACK response mode that guarantees at-least-once delivery. Most fluent-loggers don't support ACK response mode while https://github.com/komamitsu/fluency supports the mode. So you can resolve the issue with `fluency` (also it supports Fluentd failover.) But ACK response mode can lead duplicated message since it doesn't guarantees at-most-once.

For b), RawSocketSender#emit() returns false if the buffer is full and you can detect it. But the throughput is only 10rps and it's unlikely buffer was full at that time.

static-max:
@komamitsu
[...]

PR #861, opened 2016-03-18
in_forward: Fixed a problem where connections are not closed on fluentd's shutdown.

aiueo4u:
At shutdown, `in_forward` plugin doesn't close connections accepted by Coolio::TCPServer, so clients continue to sending requests which main process cannot recognize and are to be lost until the process exits. (I think https://github.com/fluent/fluentd/issues/801 arises from the same cause as this)

I have fixed this problem by keeping accepted connections and closing them at shutdown.

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -57,6 +57,7 @@ module Fluent
 
       @lsock = listen
       @loop.attach(@lsock)
+      @connections = []
 
       @usock = SocketUtil.create_udp_socket(@bind)
       @usock.bind(@bind, @port)
@@ -82,11 +83,15 @@ module Fluent
       @usock.close
       @thread.join
       @lsock.close
+      @connections.each { |c| c.close }
     end
 
     def listen
       log.info "listening fluent socket on #{@bind}:#{@port}"
-      s = Coolio::TCPServer.new(@bind, @port, Handler, @linger_timeout, log, method(:on_message))
+      s = Coolio::TCPServer.new(@bind, @port, Handler, @linger_timeout, log, method(:on_message)) do |connection|
+        @connections.reject! { |conn| conn.closed? }
+        @connections << connection
+      end
       s.listen(@backlog) unless @backlog.nil?
       s
     end


lib/fluent/plugin/in_forward.rb
```diff
@@ -82,11 +83,15 @@ def shutdown
...
       @thread.join
       @lsock.close
+      @connections.each { |c| c.close }
     end
 
     def listen
       log.info "listening fluent socket on #{@bind}:#{@port}"
-      s = Coolio::TCPServer.new(@bind, @port, Handler, @linger_timeout, log, method(:on_message))
+      s = Coolio::TCPServer.new(@bind, @port, Handler, @linger_timeout, log, method(:on_message)) do |connection|
+        @connections.reject! { |conn| conn.closed? }
```
repeatedly: Should this brock guard using Mutex?

aiueo4u: I think it doesn't need to be guarded.
(I would like to confirm just in case; does `this block` mean `do |connection| ... end`, not `{ |conn| conn.closed? }`, right?)

`@connections` is used for closing accepted connections at shutdown, so already closed connections don't need to be included in. I think it is safe that several processes or threads come in this block at the same time.

tagomoris: It looks needed for me too to guard `@connections.reject!` using Mutex.
I mean that:
```ruby
@mutex.synchronize do
  @connections.reject!{ |conn| conn.closed? }
  @connections << connection
end
```

Connection callback may run twice immediately if many connection arrived at a time.

aiueo4u: Thank you for your comments.

Surely, I think I should guard it by mutex, but please let me ask you a question.
Can this callback be run simultaneously by threads?
It seems that only one thread(`@thread`) handles waiting connections, communicating with accepted connections and replying heartbeat request. So only one thread can run this callback at a time, cannot it?

lib/fluent/plugin/in_forward.rb
```diff
@@ -82,11 +83,15 @@ def shutdown
       @usock.close
       @thread.join
       @lsock.close
```
aiueo4u: By the way, I have another question about `#shutdown`. Should watchers be detached after `{@usock,@lsock}.close` in `#shutdown`? Otherwise, requests sent from clients between `@loop.watchers.each { |w| w.detach if w.attached? }` and `@lsock.close` would be lost.

tagomoris:
This issue about closing sockets appropriately will be re-checked after porting `in_forward` into v0.14 APIs and socket/server plugin helpers. Pending.

tagomoris:
I created #1401 to solve this problem at v0.14 server plugin helper based code.
Thank you to inform this problem to us!

Issue #858, opened 2016-03-16
Counter API


tagomoris:
Fluentd v0.14 provides Counter API for purposes below:
- to provide counters with increment/decrement methods which works efficiently
- to support to persistent counting data using plugin storages
- to provide shared counter between processes

It's not necessary to make it pluggable, because it only has some variant:
- visible from a process
- shared with multi processes

This API is very useful to implement counter plugins (e.g., flowcounter, datacounter, ...) and also to implement to get metrics of Fluentd itself. Persistency can be delegated into Plugin Storage implementations.

PR #864, opened 2016-03-23
Introduce plugin storage

tagomoris:
This patch is to add a new feature "plugin storages" to store key-value data to pluggable storages.
It is useful for many purposes:
* re-implementation of pos file of file output plugin
* store intermediate state of counter plugins (datacounter, flowcounter, ...) and make it resumable
* make configurations dynamic and readable from external storages

This feature requires many more additions not included in this change for actual use cases (will be added in following changes):
* simple storage path configuration with plugin id via system config
* multi process support

### DIFF
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -6,8 +6,8 @@ require 'yajl'
 
 module Fluent
   module Plugin
-    class JSONStorage < Storage
-      Fluent::Plugin.register_storage('json', self)
+    class LocalStorage < Storage
+      Fluent::Plugin.register_storage('local', self)
 
       DEFAULT_DIR_MODE = 0755
       DEFAULT_FILE_MODE = 0644
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -30,7 +30,7 @@ module Fluent
           if @autosave || @persistent
             raise Fluent::ConfigError, "Plugin @id or path for <storage> required to save data"
           else
-            log.info "Both of Plugin @id and path for <storage> are not specified. Using on-memory store."
+            log.info "both of Plugin @id and path for <storage> are not specified. Using on-memory store."
             @on_memory = true
           end
         elsif @path
@@ -49,7 +49,7 @@ module Fluent
               data = Yajl::Parser.parse(open(@path, 'r:utf-8'){ |io| io.read })
               raise Fluent::ConfigError, "Invalid contents (not object) in plugin storage file: '#{@path}'" unless data.is_a?(Hash)
             rescue => e
-              log.error "Failed to read data fron plugin storage file", path: @path, error_class: e.class, error: e
+              log.error "failed to read data from plugin storage file", path: @path, error_class: e.class, error: e
               raise Fluent::ConfigError, "Unexpected error: failed to read data from plugin storage file: '#{@path}'"
             end
           else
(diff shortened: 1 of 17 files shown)


tagomoris:

repeatedly:
@sonots Could you review this first? I'm now busy so I don't have enough time to review large patch...

sonots:
Comments

* `JSONStorage` sounds like that we have to create new storage just for different formats. It looked `FileStorage` with `json` **format** is better design
  * Are you not assuming an extendability like `S3Storage`? Or, extendability is only for formats?
* I want `incr` methods which allows us to atomically increment for counter plugins.

tagomoris:
@sonots File is not good abstraction, because S3 (and many other file systems) has files.
I considered some other naming options like `LocalFileStorage`, `LocalStorage` or others, but couldn't find something good enough. (Because `json` storage also works as on-memory storage if local disk path is not configured.)
I don't have any plan to make formats configurable for now.

tagomoris:
`incr` or such like operation will be supported by Counter API #858

sonots:
> LocalStorage

LocalStorage sounds good for me because it expresses it would store into local memory or local file, but not S3. At least, it should be better than `JSONStorage` because it sounds like it would store into S3 (we can not know the storage location from its name)

> incr

Okay, thanks.

tagomoris:
I changed default storage name from "json" to "local".

sonots:
LGTM!

lib/fluent/plugin/storage_local.rb
```diff
@@ -0,0 +1,111 @@
...
+        if !@on_memory
+          dir = File.dirname(@path)
+          FileUtils.mkdir_p(dir, mode: @dir_mode) unless File.exist?(dir)
+          if File.exist?(@path)
+            raise Fluent::ConfigError, "Plugin storage path '#{@path}' is not readable/writable" unless File.readable?(@path) && File.writable?(@path)
+            begin
+              data = Yajl::Parser.parse(open(@path, 'r:utf-8'){ |io| io.read })
+              raise Fluent::ConfigError, "Invalid contents (not object) in plugin storage file: '#{@path}'" unless data.is_a?(Hash)
+            rescue => e
+              log.error "Failed to read data fron plugin storage file", path: @path, error_class: e.class, error: e
```
repeatedly: type: `fron`

lib/fluent/plugin/storage_local.rb
```diff
@@ -0,0 +1,111 @@
...
+        @store = {}
+      end
+
+      def configure(conf, plugin)
+        super
+
+        @on_memory = false
+        if !@path && !@_plugin_id_configured
+          if @autosave || @persistent
+            raise Fluent::ConfigError, "Plugin @id or path for <storage> required to save data"
```
repeatedly: In fluentd, log message starts lower-case by default.

lib/fluent/plugin_helper/storage.rb
```diff
@@ -0,0 +1,306 @@
...
+          return s.storage
+        elsif !s
+          unless type
+            raise ArgumentError, "BUG: type not specified without configuration"
+          end
+          storage = Plugin.new_storage(type)
+          storage.owner = self
+          config = if conf && conf.is_a?(Fluent::Config::Element)
+                     conf
+                   elsif conf && conf.is_a?(Hash)
```
repeatedly: `conf` is double-checked. `conf.is_a?(Hash)` should be moved to above condition?

tagomoris:
I pushed fixes for comments.

repeatedly:
LTGM

tagomoris:
Okay.

PR #866, opened 2016-03-24
Introduce New Plugin Classes

tagomoris:
New base classes for v0.14 new plugin APIs.
These are introduced to make these new features enabled:
* Plugin Helpers
* Plugin Storages
* Async/Parallel buffer flushing & delayed commit
* Buffer chunking with various data including time and tag (merging forest plugin)
* New test drivers

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -19,7 +19,8 @@ require 'fluent/configurable'
 module Fluent
   module Config
     class ConfigureProxy
-      attr_accessor :name, :final, :param_name, :init, :required, :multi, :alias, :argument, :params, :defaults, :descriptions, :sections
+      attr_accessor :name, :final, :param_name, :init, :required, :multi, :alias, :configured_in_section
+      attr_accessor :argument, :params, :defaults, :descriptions, :sections
       # config_param :desc, :string, :default => '....'
       # config_set_default :buffer_type, :memory
       #
@@ -49,6 +50,11 @@ module Fluent
         @alias = opts[:alias]
 
         raise "init and required are exclusive" if @init && @required
+
+        # specify section name for viewpoint of owner(parent) plugin
+        # for buffer plugins: all params are in <buffer> section of owner
+        # others: <storage>, <format> (formatter/parser), ...
+        @configured_in_section = nil
 
         @argument = nil # nil: ignore argument
         @params = {}
@@ -89,6 +95,9 @@ module Fluent
         if overwrite?(other, :alias)
           raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: alias"
         end
+        if overwrite?(other, :configured_in_section)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: configured_in"
+        end
 
         options = {}
         # param_name is used not to ovewrite plugin's instance
@@ -102,6 +111,9 @@ module Fluent
         options[:final] = @final || other.final
 
         merged = self.class.new(other.name, options)
+
+        # configured_in MUST be kept
+        merged.configured_in_section = self.configured_in_section
 
         merged.argument = other.argument || self.argument
         merged.params = other.params.merge(self.params)
@@ -144,6 +156,9 @@ module Fluent
         if overwrite?(other, :alias)
           raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: alias"
         end
+        if overwrite?(other, :configured_in_section)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: configured_in"
+        end
 
         options = {}
         options[:param_name] = other.param_name
@@ -154,6 +169,8 @@ module Fluent
         options[:final]  = true
 
         merged = self.class.new(other.name, options)
+
+        merged.configured_in_section = self.configured_in_section
 
         merged.argument = self.argument || other.argument
         merged.params = other.params.merge(self.params)
@@ -173,6 +190,15 @@ module Fluent
         end
 
         merged
+      end
+
+      def overwrite_defaults(other) # other is owner plugin's corresponding proxy
+        self.defaults = self.defaults.merge(other.defaults)
+        self.sections.keys.each do |section_key|
+          if other.sections.has_key?(section_key)
+            self.sections[section_key].overwrite_defaults(other.sections[section_key])
+          end
+        end
       end
 
       def parameter_configuration(name, *args, &block)
@@ -211,6 +237,13 @@ module Fluent
         end
 
         [name, block, opts]
+      end
+
+      def configured_in(section_name)
+        if @configured_in_section
+          raise ArgumentError, "#{self.name}: configured_in called twice"
+        end
+        @configured_in_section = section_name.to_sym
       end
 
       def config_argument(name, *args, &block)
--- a/lib/fluent/plugin.rb
+++ b/lib/fluent/plugin.rb
@@ -25,9 +25,12 @@ module Fluent
     # plugins for fluentd plugins: fluent/plugin/type/NAME.rb
     #   ex: storage, buffer chunk, ...
 
+    # first class plugins (instantiated by Engine)
     INPUT_REGISTRY     = Registry.new(:input,     'fluent/plugin/in_')
     OUTPUT_REGISTRY    = Registry.new(:output,    'fluent/plugin/out_')
     FILTER_REGISTRY    = Registry.new(:filter,    'fluent/plugin/filter_')
+
+    # feature plugin: second class plugins (instanciated by Plugins or Helpers)
     BUFFER_REGISTRY    = Registry.new(:buffer,    'fluent/plugin/buf_')
     PARSER_REGISTRY    = Registry.new(:parser,    'fluent/plugin/parser_')
     FORMATTER_REGISTRY = Registry.new(:formatter, 'fluent/plugin/formatter_')
@@ -105,11 +108,11 @@ module Fluent
       new_impl('filter', FILTER_REGISTRY, type)
     end
 
-    def self.new_buffer(type)
-      new_impl('buffer', BUFFER_REGISTRY, type)
+    def self.new_buffer(type, parent: nil)
+      new_impl('buffer', BUFFER_REGISTRY, type, parent)
     end
 
-    def self.new_parser(type)
+    def self.new_parser(type, parent: nil)
       require 'fluent/parser'
 
       if type[0] == '/' && type[-1] == '/'
@@ -117,16 +120,16 @@ module Fluent
         require 'fluent/parser'
         Fluent::TextParser.lookup(type)
       else
-        new_impl('parser', PARSER_REGISTRY, type)
+        new_impl('parser', PARSER_REGISTRY, type, parent)
       end
     end
 
-    def self.new_formatter(type)
-      new_impl('formatter', FORMATTER_REGISTRY, type)
+    def self.new_formatter(type, parent: nil)
+      new_impl('formatter', FORMATTER_REGISTRY, type, parent)
     end
 
-    def self.new_storage(type)
-      new_impl('storage', STORAGE_REGISTRY, type)
+    def self.new_storage(type, parent: nil)
+      new_impl('storage', STORAGE_REGISTRY, type, parent)
     end
 
     def self.register_impl(kind, registry, type, value)
@@ -138,17 +141,21 @@ module Fluent
       nil
     end
 
-    def self.new_impl(kind, registry, type)
+    def self.new_impl(kind, registry, type, parent=nil)
       # "'type' not found" is handled by registry
       obj = registry.lookup(type)
-      case
-      when obj.is_a?(Class)
-        obj.new
-      when obj.respond_to?(:call) && obj.arity == 0
-        obj.call
-      else
-        raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable (without arguments)."
+      impl = case
+             when obj.is_a?(Class)
+               obj.new
+             when obj.respond_to?(:call) && obj.arity == 0
+               obj.call
+             else
+               raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable (without arguments)."
+             end
+      if parent && impl.respond_to?("owner=")
+        impl.owner = parent
       end
+      impl
     end
   end
 end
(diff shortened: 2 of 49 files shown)


tagomoris:
I've almost done for:
* new base classes for input/output/buffer
* new implementation for memory/file buffer plugin

I've not done yet for:
* tests for new base classes and buffer plugins
* optimization for many code
* compatibility layers for old plugin
* test drivers for new plugins

sonots:
`Feature` is too generic, and I could not easily figure out what it means. It is better to think of better name, I think.

sonots:
`Metadata` looks nice since I've ever implemented a code to parse tsuffix to get chunk key time in fluent-plugin-s3. It was ugly.  😄

tagomoris:
@sonots Please give me another options :(

sonots:
What about `StateModel`? I first felt `StateMachine` would be good, but it looks it is not exactly `StateMachine`.

tagomoris:
I added a commit to solve `Feature` naming in other way.
https://github.com/fluent/fluentd/pull/866/commits/0594ec6cfcb4b3d96a7a3dd17566373b46e65bbd

tagomoris:
I cannot understand the difference between `StateModel` and `StateMachine`. And it's hidden thing even for plugin developers (plugin helper is `retry_state` and `retry_state_create`). So I want not to take long time for it...

sonots:
I was not sure what you wanted to express with the word "Feature" because there was no explanation of it, but just "Base" looks fine for me.

lib/fluent/config/configure_proxy.rb
```diff
@@ -89,6 +95,9 @@ def merge(other) # self is base class, other is subclass
         if overwrite?(other, :alias)
           raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: alias"
         end
+        if overwrite?(other, :configured_in_section)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: configured_in"
```
repeatedly: `configured_in_section` instead of `configured_in`?

lib/fluent/config/configure_proxy.rb
```diff
@@ -103,6 +112,9 @@ def merge(other) # self is base class, other is subclass
 
         merged = self.class.new(other.name, options)
 
+        # configured_in MUST be kept
+        merged.configured_in_section = self.configured_in_section
+
```
repeatedly: This space is needed?

lib/fluent/config/configure_proxy.rb
```diff
@@ -155,6 +170,8 @@ def merge_for_finalized(other)
 
         merged = self.class.new(other.name, options)
 
+        merged.configured_in_section = self.configured_in_section
+
```
repeatedly: ditto

lib/fluent/config/configure_proxy.rb
```diff
@@ -144,6 +156,9 @@ def merge_for_finalized(other)
         if overwrite?(other, :alias)
           raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: alias"
         end
+        if overwrite?(other, :configured_in_section)
+          raise ConfigError, "BUG: subclass cannot overwrite base class's config_section: configured_in"
```
repeatedly: ditto

lib/fluent/plugin.rb
```diff
@@ -138,17 +141,21 @@ def self.register_impl(kind, registry, type, value)
...
-        raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable (without arguments)."
+      impl = case
+             when obj.is_a?(Class)
+               obj.new
+             when obj.respond_to?(:call) && obj.arity == 0
+               obj.call
+             else
+               raise Fluent::ConfigError, "#{kind} plugin '#{type}' is not a Class nor callable (without arguments)."
+             end
+      if parent && impl.respond_to?(:"owner=")
```
repeatedly: `:owner=` is enough? `"` is needed?

lib/fluent/plugin/buf_file2.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      ##TODO: Buffer plugin cannot handle symlinks because new API @stage has many writing buffer chunks
+      ##      re-implement this feature on out_file, w/ enqueue_chunk(or generate_chunk) hook + chunk.path
+      # attr_accessor :symlink_path
+
+      @@buffer_paths = {}
+
+      def initialize
+        super
+        @uri_parser = URI::Parser.new
```
repeatedly: Is this needed?

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,351 @@
...
+
+      def new_metadata(timekey: nil, tag: nil, variables: nil)
+        Metadata.new(timekey, tag, variables)
+      end
+
+      # TODO: metadata_list cleaner
+      def add_metadata(metadata)
+        synchronize do
+          if i = @metadata_list.index(metadata)
+            @metadata_list[i]
```
repeatedly: return value is needed?

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,351 @@
...
+        # the case whole data can be stored in staged chunk: almost all emits will success
+        chunk = synchronize { @stage[metadata] ||= generate_chunk(metadata) }
+        chunk.synchronize do
+          begin
+            chunk.append(data)
+            if !size_over?(chunk) || force
+              chunk.commit
+              stored = true
+            end
+          ensure
```
repeatedly: `rescue`?

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,351 @@
...
+          end
+        end
+      end
+
+      def enqueue_chunk(metadata)
+        synchronize do
+          chunk = @stage.delete(metadata)
+          if chunk
+            chunk.synchronize do
+              if chunk.empty?
```
repeatedly: When empty chunk is generated?

lib/fluent/plugin/buffer.rb
```diff
@@ -0,0 +1,351 @@
...
+              enqueue_chunk(metadata)
+            end
+          end
+        end
+      end
+
+      def dequeue_chunk
+        return nil if @queue.empty?
+        synchronize do
+          chunk = @queue.shift
```
repeatedly: Is this for double-check?
`@queue.empty?` is called in the first line.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -0,0 +1,126 @@
...
+        end
+
+        def write_to(io)
+          open do |i|
+            FileUtils.copy_stream(i, io)
+          end
+        end
+
+        def msgpack_each(&block)
+          require 'fluent/engine'
```
repeatedly: msgpack_each is called frequently so calling `require` here is bad for performacne.

lib/fluent/plugin/output.rb
```diff
@@ -0,0 +1,643 @@
...
+
+      CHUNK_KEY_PATTERN = /^[-_.@a-zA-Z0-9]+$/
+      CHUNK_KEY_PLACEHOLDER_PATTERN = /\$\{[-_.@a-zA-Z0-9]+\}/
+
+      # `<buffer>` and `<secondary>` sections are available only when '#format' and '#write' are implemented
+      config_section :buffer, param_name: :buffer_config, init: true, required: false, multi: false, final: true do
+        config_argument :chunk_keys, :array, value_type: :string, default: []
+        config_param :@type, :string, default: 'memory'
+
+        config_param :timekey_range, :time, default: nil # range size to be used: `time.to_i / @timekey_range`
```
repeatedly: fluentd uses `time_key` and `time_key_xxx` for parameters so
`time_key` is better than `timekey`.

lib/fluent/plugin/output.rb
```diff
@@ -0,0 +1,643 @@
...
+        config_param :flush_interval, :time, default: 60, desc: 'The interval between buffer chunk flushes.'
+
+        config_param :flush_threads, :integer, default: 1, desc: 'The number of threads to flush the buffer.'
+
+        config_param :flush_tread_interval, :float, default: 1.0, desc: 'Seconds to sleep between checks for buffer flushes in flush threads.'
+        config_param :flush_burst_interval, :float, default: 1.0, desc: 'Seconds to sleep between flushes when many buffer chunks are queued.'
+
+        config_param :delayed_commit_timeout, :time, default: 60, desc: 'Seconds of timeout for buffer chunks to be committed by plugins later.'
+
+        config_param :retry_forever, :bool, default: false, desc: 'If true, plugin will ignore retry_limit_* options and retry flushing forever.'
```
repeatedly: There are no `retry_limit_*`. Is this typo?

lib/fluent/plugin/output.rb
```diff
@@ -0,0 +1,643 @@
...
+        config_param :retry_timeout, :time, default: 72 * 60 * 60, desc: 'The maximum seconds to retry to flush while failing, until plugin discards buffer chunks.'
+        # 72hours == 17 times with exponential backoff (not to change default behavior)
+        config_param :retry_max_times, :integer, default: nil, desc: 'The maximum number of times to retry to flush while failing.'
+
+        config_param :retry_secondary_threshold, :integer, default: 80, desc: 'Percentage of retry_limit_* to switch to use secondary while failing.'
```
repeatedly: ditto

lib/fluent/plugin/output.rb
```diff
@@ -0,0 +1,643 @@
...
+        desc 'If true, plugin will try to flush buffer just before shutdown.'
+        config_param :flush_at_shutdown, :bool, default: nil # change default by buffer_plugin.persistent?
+
+        desc 'How to enqueue chunks to be flushed. "fast" flushes per flush_interval, "immediate" flushes just after event arrival.'
+        config_param :flush_mode, :enum, list: [:none, :fast, :immediate], default: :fast
+        config_param :flush_interval, :time, default: 60, desc: 'The interval between buffer chunk flushes.'
+
+        config_param :flush_threads, :integer, default: 1, desc: 'The number of threads to flush the buffer.'
+
+        config_param :flush_tread_interval, :float, default: 1.0, desc: 'Seconds to sleep between checks for buffer flushes in flush threads.'
```
repeatedly: thread

repeatedly:
Could you write a rough image of buffer structure like this? http://docs.fluentd.org/images/buffer-internal-and-parameters.png

lib/fluent/config/configure_proxy.rb
tagomoris: No. This message is for authors of plugins, and plugins use `configured_in` method to specify this value.

lib/fluent/config/configure_proxy.rb
tagomoris: Because it is not for plugin itself, but for internal handling about overwriting default values. It should be separated from argument/params/defaults/sections.

lib/fluent/plugin.rb
tagomoris: Using string is ok.

lib/fluent/plugin/buf_file2.rb
tagomoris: No. I missed to delete it.

lib/fluent/plugin/buffer.rb
tagomoris: I confirmed that following code is wrong. Return value is needed to use exact a object for same metadata.

lib/fluent/plugin/buffer.rb
tagomoris: NO. `ensure`. This code always do `rollback` to reset internal state perfectly.
With `rescue`, this code should be like this:
```ruby
        chunk.synchronize do
          begin
            chunk.append(data)
            if !size_over?(chunk) || force
              chunk.commit
              stored = true
            else
              chunk.rollback
            end
          rescue
            chunk.rollback
            raise
          end
        end
```
This code is a bit efficient, but more complex.

lib/fluent/plugin/buffer.rb
tagomoris: When empty strings are appended.

lib/fluent/plugin/buffer.rb
tagomoris: The first line check is early return. It's out of critical session (`synchronize`).
Other threads might dequeue chunks from this queue between first line and second line.

lib/fluent/plugin/buffer/chunk.rb
tagomoris: Ah, it's actually nonsense requirement because it already depends on `fluent/msgpack_factory`.

lib/fluent/plugin/output.rb
tagomoris: The newer API uses different value from older `time_key`.
* `time_key` is a string, representation of time with `time_format` and strftime
* `timekey` is an integer, representing the first second of a time range

`config_param` name is directly a variable name for plugins. so we should make it different from older one for understandability of plugin authors.

lib/fluent/plugin/output.rb
tagomoris: It's older name in development. Will fix.

tagomoris:
@repeatedly for documentation? If so, I'll do it in later step.

repeatedly:
@tagomoris Mainly for documentation and good for reviwing your code.

tagomoris:
I drew an overview of v0.14 Plugin API method/data flow
https://dl.dropboxusercontent.com/u/62744056/Fluentd%20v0.14%20plugin%20API%20overview.png

tagomoris:
I've completed to write all tests. But on Travis, some errors are occurring. I'll fix it later.
Next, I'll do clean commits for each files, separate commits for code and for tests into different pull-requests, and the merge these.

tagomoris:
These are not done in this branch:
* optimization for many code
* compatibility layers for old plugin
* test drivers for new plugins

tagomoris:
I cleaned up commits, and now ready to merge this (If CI is green).
@repeatedly @sonots Any thoughts?

tagomoris:
I'll merge this branch after CI green confirmation.

cosmo0920:
Travis CI reports green. But AppVeyor complains about dozens of errors and failures.

sonots:
Just a question, are we going to publish with names such as `file2` or `memory2`? Or, will they be changed?

tagomoris:
They will be changed to `file` and `memory`. Current names are just temporal one.

tagomoris:
I confirmed that CIs are now green.

tagomoris:
Merged!

Issue #522, opened 2014-12-18
Change serialization syntax of Fluent::Config::Element


tagomoris:
Configuration dump should be done as default (v1) syntax on v0.14.

okkez:
Could you close this issue? @tagomoris

tagomoris:
@okkez Thank you for your work!

PR #867, opened 2016-03-25
Change serialization syntax

okkez:
string as following

In previous version:
   string_param1 string_value
   string_param2 string
value

In this version:
   string_param1 "string_value"
   string_param2 "string\nvalue"

I don't change serialization format for other types.

See #522

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -111,7 +111,17 @@ module Fluent
           if secret_param?(k)
             out << "#{nindent}#{k} xxxxxx\n"
           else
-            out << "#{nindent}#{k} #{v}\n"
+            case param_type(k)
+            when :string
+              out << "#{nindent}#{k} \"#{self.class.unescape_parameter(v)}\"\n"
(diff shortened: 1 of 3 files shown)


lib/fluent/config/element.rb
```diff
@@ -144,6 +142,42 @@ def secret_param?(key)
...
+        param_key = key.to_sym
+        proxy = @corresponding_proxies.detect do |_proxy|
+          _proxy.params.has_key?(param_key)
+        end
+        return nil unless proxy
+        _block, opts = proxy.params[param_key]
+        opts[:type]
+      end
+
+      def dump_value(k, v, indent, nindent)
```
tagomoris: Is there any reason to assign `out = ""` at once and then do `out << "..."`?
It looks good enough just to return string value by `if` expression.

okkez: Exactly.
I will fix this.

okkez:
I've fixed and pushed.

tagomoris:
Merged! Thank you so much.

PR #871, opened 2016-03-29
Separate msgpack_factory from Engine to a file

tagomoris:
* to make it available from many classes without loading whole Engine
* Many classes/plugins requires accessibility for msgpack_factory to handle Fluent::EventTime
  * but Engine is too big to load every time (especially in tests)
* This patch make it possible to use msgpack_factory from anywhere using Mixin
  * and fetched msgpack_factory is the same instance for Engine after Engine.init
  * to make it possible to use other ext types registered via Engine
* There're another methods to create factory instances without Engine effects
  * it should be required especially for tests

### DIFF
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -16,12 +16,12 @@
 
 require 'socket'
 
-require 'msgpack'
 require 'cool.io'
 
 require 'fluent/config'
 require 'fluent/event'
 require 'fluent/event_router'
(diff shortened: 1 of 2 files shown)


tagomoris:
@repeatedly @sonots any comments? I'll merge this few hours later if not.

sonots:
:+1:

tagomoris:
Merged.

PR #874, opened 2016-03-31
Add "string_list" type to get list of strings

tagomoris:
* it's very common case for fluentd plugins
* "array" is useful, but it can get various types for values
* this type should support 2 kind of formats at once:
  * comman-separated string values to make input data simple
  * JSON-format arrays for values with comma

This type definition make it possible to reduce # of self-implementation of `str.split`... by 3rd party plugins

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -94,6 +94,17 @@ module Fluent
       end
       param
     }
+    STRING_LIST_TYPE = Proc.new { |val, opts|
+      param = if val.is_a?(String)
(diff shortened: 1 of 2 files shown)


tagomoris:
Another proposal is to make `:array` to support comma-separated values with `type` option to support lists of various kind of types.

sonots:
I am :+1: for `:array` with `type` because increasing number of types makes things be messy

tagomoris:
I added exclusive proposal as #875

repeatedly:
Almost maintainers prefer #875 so closed.

PR #875, opened 2016-03-31
Add simplified syntax for hash and array

tagomoris:
* to specify values with very simple comma (and colon for key-value separator)
 * to re-format values because comma-separated values doesn't have type information

It looks to satisfy various use-cases of 3rd party plugins, and this change is just additional
change to allow format which is currently denied as ConfigError.

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -81,7 +81,7 @@ module Fluent
     BOOL_TYPE = Proc.new { |val, opts| Config.bool_value(val) }
     TIME_TYPE = Proc.new { |val, opts| Config.time_value(val) }
 
-    REFORMAT_VALUES = ->(type, value) {
+    REFORMAT_VALUE = ->(type, value) {
       if value.nil?
         value
       else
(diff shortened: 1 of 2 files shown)


tagomoris:
This change is exclusive with #874

sonots:
I prefer this than #874

test/config/test_types.rb
```diff
@@ -124,10 +124,33 @@ class TestConfigTypes < ::Test::Unit::TestCase
 
     test 'hash' do
       assert_equal({"x"=>"v","k"=>1}, Config::HASH_TYPE.call('{"x":"v","k":1}', {}))
+      assert_equal({"x"=>"v","k"=>"1"}, Config::HASH_TYPE.call('x:v,k:1', {}))
```
sonots: Could you add a test a hash with , surrounded by spaces?

tagomoris: done

lib/fluent/config/types.rb
```diff
@@ -80,19 +80,58 @@ def self.bool_value(str)
     SIZE_TYPE = Proc.new { |val, opts| Config.size_value(val) }
     BOOL_TYPE = Proc.new { |val, opts| Config.bool_value(val) }
     TIME_TYPE = Proc.new { |val, opts| Config.time_value(val) }
+
+    REFORMAT_VALUES = ->(type, value) {
```
repeatedly: Why `REFORMAT_VALUES`? `REFORMAT_VALUE`?

tagomoris: Fixed to `REFORMAT_VALUE`

tagomoris:
Merged.

PR #876, opened 2016-03-31
Add delimiter option to in_udp plugin

pokehanai:
According to the documentation http://docs.fluentd.org/articles/in_udp
`in_udp` should have `deliimiter` option but it currently does not.

I noticed it when my code sent multi log entries in a transmittion.

I made this patch while I am not sure whether it is a documentation bug
or not; hopefully it is not - to support bulk operation.

FYI: Current `in_syslog` looks fine with multiline entries.
Maybe its parser handles well.

### DIFF
--- a/lib/fluent/plugin/in_udp.rb
+++ b/lib/fluent/plugin/in_udp.rb
@@ -28,15 +28,27 @@ module Fluent
       log.info "listening udp socket on #{@bind}:#{@port}"
       @usock = SocketUtil.create_udp_socket(@bind)
       @usock.bind(@bind, @port)
+      @has_delimiter = @delimiter != ''
       SocketUtil::UdpHandler.new(@usock, log, @body_size_limit, callback)
     end
 
     def on_message(msg, addr)
-      if @delimiter == ''
-        super(msg, addr)
-      else
-        msg.split(@delimiter).each do |m|
+      return super(msg, addr) unless @has_delimiter
+
+      pos = msg.index(@delimiter)
+      return super(msg, addr) if pos.nil?
+
+      start = 0
+      while true
+        m = msg[start, pos - start]
+        super(m, addr)
+
+        start = pos + 1
+        pos = msg.index(@delimiter, start)
+        if pos.nil?
+          m = msg[start, msg.length - start]
           super(m, addr)
+          break
         end
       end
     end
--- a/lib/fluent/plugin/in_udp.rb
+++ b/lib/fluent/plugin/in_udp.rb
@@ -22,12 +22,19 @@ module Fluent
 
     config_set_default :port, 5160
     config_param :body_size_limit, :size, default: 4096
+    config_param :delimiter, :string, :default => "\n"
 
     def listen(callback)
       log.info "listening udp socket on #{@bind}:#{@port}"
       @usock = SocketUtil.create_udp_socket(@bind)
       @usock.bind(@bind, @port)
       SocketUtil::UdpHandler.new(@usock, log, @body_size_limit, callback)
+    end
+
+    def on_message(msg, addr)
+      msg.split(@delimiter).each do |m|
+        super(m, addr)
+      end
     end
   end
 end
--- a/lib/fluent/plugin/in_udp.rb
+++ b/lib/fluent/plugin/in_udp.rb
@@ -22,7 +22,7 @@ module Fluent
 
     config_set_default :port, 5160
     config_param :body_size_limit, :size, default: 4096
-    config_param :delimiter, :string, :default => "\n"
+    config_param :delimiter, :string, :default => ""
 
     def listen(callback)
       log.info "listening udp socket on #{@bind}:#{@port}"
@@ -32,8 +32,12 @@ module Fluent
     end
 
     def on_message(msg, addr)
-      msg.split(@delimiter).each do |m|
-        super(m, addr)
+      if @delimiter == ''
+        super(msg, addr)
+      else
+        msg.split(@delimiter).each do |m|
+          super(m, addr)
+        end
       end
     end
   end
--- a/test/plugin/test_in_udp.rb
+++ b/test/plugin/test_in_udp.rb
@@ -103,4 +103,42 @@ class UdpInputTest < Test::Unit::TestCase
       assert(emits[i][1].is_a?(Fluent::EventTime))
     }
   end
+
+  {
+    'none' => [
+      {'msg' => "tcptest3\ntcptest4\n", 'expected' => ['tcptest3', 'tcptest4']},
+    ],
+    'json' => [
+      {'msg' => {'k' => 10, 'message' => 100}.to_json + "\n" +
+                {'k' => 20, 'message' => 200}.to_json + "\n", 'expected' => [100, 200]},
+    ],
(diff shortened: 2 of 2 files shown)


repeatedly:
It seems a bug of document.

> I noticed it when my code sent multi log entries in a transmittion.

Looks like this?

```
{"time":time1,"k":"v1"}\n{"time":time2,"k":"v2"}
```

> FYI: Current in_syslog looks fine with multiline entries.

in_syslog doesn't care `\n` in UDP handler so multiline message body are treated as on message.
This is a different from this case.

hanai-tohru:
> Looks like this?
> 
> ```
> {"time":time1,"k":"v1"}\n{"time":time2,"k":"v2"}
>```
Yes.

> It seems a bug of document.

Why `in_udp` should not handle that?

repeatedly:
> Why in_udp should not handle that?

Personally speaking, it seems no problem but splitting all message is bad.
Some users send multiline content via UDP.
This patch breaks such usecases so we should keep backward compatibility.

test/plugin/test_in_udp.rb
```diff
@@ -103,4 +103,42 @@ def compare_test_result(emits, tests)
...
+      {'msg' => "tcptest3\ntcptest4\n", 'expected' => ['tcptest3', 'tcptest4']},
+    ],
+    'json' => [
+      {'msg' => {'k' => 10, 'message' => 100}.to_json + "\n" +
+                {'k' => 20, 'message' => 200}.to_json + "\n", 'expected' => [100, 200]},
+    ],
+    '/^\\[(?<time>[^\\]]*)\\] (?<message>.*)/' => [
+      {'msg' => "[Sep 10 00:00:00] message1\n[Sep 10 00:00:00] message2\n", 'expected' => ['message1', 'message2']},
+    ]
+  }.each { |format, test_cases|
```
repeatedly: Use test-unit's data method instead of `each` method definition.

pokehanai:
Okay, I understand.

Then, what about to have `delimiter` option with default value `""` (empty string); do not split in default.
Is it acceptable?

repeatedly:
> Is it acceptable?

Not acceptable because calling split with long string takes time and CPU resource.
By default, in_udp should not call split.

pokehanai:
Hmm.

Then, replacing with String#slice in a loop would reduce some CPU and memory impact.
What about that way?

Since sending from clients and receiving by the agent multiple udp messages consume more totaling CPU cycle, network bytes and time than comparison to a combined single message. So adding some cost like above would still be reasonable.

repeatedly:
> Then, replacing with String#slice in a loop would reduce some CPU and memory impact.
What about that way?

What does this sentence mean? Using `String#slice` instead of `String#split`?

> Since sending from clients and receiving by the agent multiple udp messages consume more totaling CPU cycle, network bytes and time than comparison to a combined single message.

Right, but we can't decrease the default performance because it breaks exisiting user environment.

pokehanai:
> What does this sentence mean? Using String#slice instead of String#split?

I have updated the pull request code.
Please have a look. Thank you.

lib/fluent/plugin/in_udp.rb
```diff
@@ -22,12 +22,35 @@ class UdpInput < SocketUtil::BaseInput
...
     def listen(callback)
       log.info "listening udp socket on #{@bind}:#{@port}"
       @usock = SocketUtil.create_udp_socket(@bind)
       @usock.bind(@bind, @port)
+      @has_delimiter = @delimiter != ''
       SocketUtil::UdpHandler.new(@usock, log, @body_size_limit, callback)
     end
+
+    def on_message(msg, addr)
+      return super(msg, addr) unless @has_delimiter
```
repeatedly: In this approach, `delimiter` should be `nil`.
`@has_delimier` is not needed.

repeatedly:
I confirmed newer version reduces memory allocation.
It is nice 👍

lib/fluent/plugin/in_udp.rb
pokehanai: Hmm?
I think `delimiter` should be treated only as string for the facts below. 
- `delimiter` is typed as string
- developer can overwrite it with `""` in the config file
- on the other hand there is no way (or at least I cannot find out) to set `nil` from the config file

repeatedly: In in_udp, delimiter should not be used by default. delimiter is enabled when user set in configuration. In this case,

- nil: delimiter is not set
- string like "\n": delimiter is set by configuration

This is very simple. If you use `""` for default value, you can't judge default value or user set value.

pokehanai: Basically the task of default value is equivalent to initial value, one of user-settable values.
So it should not be a special value. 

Some developers tend to write down default values demonstratively in config files for some reasons; as documentation, or any from their painful experiences, etc.

I don't think there is enough reason to take away their chance to do.

repeatedly: > Basically the task of default value is equivalent to initial value,

I agree and it is a `nil` in this case because default is delimiter is not set.

pokehanai: Is there any way to set `nil` in config file?

repeatedly: Currently, no way. I'm not sure we will add this feature.
Default `nil` means this parameter is not set in the configuration. `delimiter nil` is not realistic.

pokehanai: Then having default value with `nil` violates the design:
> Basically the task of default value is equivalent to initial value, one of user-settable values.

`nil` is not a user-settable value, so it is a special value.

> Some developers tend to write down default values demonstratively in config files for some reasons; as documentation, or any from their painful experiences, etc.

An empty string can be for them.

```
<source>
  @type udp
  delimiter ""
  ...
</source>
```

So an empty string should be the default value.

repeatedly: > Then having default value with nil violates the design:

What design?

> nil is not a user-settable value, so it is a special value.

Yes. config_param allows nil in the default setting to represent "this value is not set in the configuration so feature is disabled".
This is why config_param doesn't validate default value with typed parameters.

pokehanai: > Some developers tend to write down default values demonstratively in config files for some reasons; as documentation, or any from their painful experiences, etc.

```
<source>
  @type udp
  delimiter ""
</source>
```
If the delimiter set to an empty string, how `in_udp` should perform? Which behaviour will be approved for developers most?
1. no splitting
2. split by each character

repeatedly: From ruby's split result, `1.` is better.
But I think empty string has no real usecases, we should raises an error in configure method.

pokehanai: > Some developers tend to write down default values demonstratively in config files for some reasons; as documentation, or any from their painful experiences, etc.
> I don't think there is enough reason to take away their chance to do.

Then does they can populate the default values explicitly in the config?

repeatedly:
Temporary closed.
If you are still interested in updating patch,
please re-open this PR.

PR #877, opened 2016-03-31
Add `config_section` option `init` to initialize section objects

### LINKED ISSUE #850: Initialize config_section even if the section is not specified
## Summary
Like `config_param`, `config_section` would be initialized with default values if the section is not specified in the runtime config.

Related issue: https://github.com/awslabs/aws-fluent-plugin-kinesis/issues/49

## Detail
I use `config_section` for my plugin and its `config_param` has default value, like below:

        config_section :foo do
          config_param :a, :string, default: 'bar'
          config_param :b, :string, default: 'baz'
        end

In my plugin, I'd like to use this section as a kind of configuration object with some default values. If users specify the section, it works.

    <match **>
      type myplugin
      ...other configs
      <foo>
        a aaa
      </foo>
    </match>

    # in the code
    @foo #=> {a: "aaa", b: "baz"}

But if users omit the section to use the default values, it doesn't because the instance variable isn't initialized.

    <match **>
      type myplugin
      ...other configs
      # no sections
    </match>

    # in the code
    @foo #=> nil # actual
    @foo #=> {a: "bar", b: "baz"}   # ideal

https://github.com/fluent/fluentd/blob/master/lib/fluent/config/section.rb#L80 returns `nil`.

I hope it should be added some parameter to behave like this, such as `:initialize`:

    config_section :foo, initialize: true do
    end

## Workaround
Make the section `required` so that users must specify the section like:

    <match **>
      type myplugin
      ...other configs
      <foo>
      </foo>
    </match>

    # in the code
    @foo #=> {a: "bar", b: "baz"}

tagomoris:
if not specified: fixes #850

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -19,7 +19,7 @@ require 'fluent/configurable'
 module Fluent
   module Config
     class ConfigureProxy
-      attr_accessor :name, :final, :param_name, :required, :multi, :alias, :argument, :params, :defaults, :descriptions, :sections
+      attr_accessor :name, :final, :param_name, :init, :required, :multi, :alias, :argument, :params, :defaults, :descriptions, :sections
       # config_param :desc, :string, :default => '....'
       # config_set_default :buffer_type, :memory
       #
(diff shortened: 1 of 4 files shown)


repeatedly:
LGTM

PR #880, opened 2016-04-04
Add ServerEngine

naritta:
https://github.com/fluent/fluentd/wiki/Fluentd-running-on-ServerEngine

implement
・Logging
・Signal handler
・Daemonize
・Worker processes


repeatedly:
Could you add remaining list on the description?

naritta:
I got it.

naritta:
remaining is:
・create cache for loading config
・fix rpc_server signal handler
・add tests

cosmo0920:
Here is GitHub. 
We can use task list in descriptions with GFM: https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

naritta:
Now, the main signal handler in fluentd and serverengine is like below.

||fluentd|serverengine(default)|
|:--:|:--:|:--:|
|INT/TERM|graceful shutdown|graceful shutdown|
|HUP| graceful restart| graceful restart(not when worker_type is "process")|
|USR1|flush buffer|graceful restart|

Then I fixed
・to use INT/TERM and HUP handler on serverengine for graceful shutdown and restart.
・let USR1 handler on fluentd as it is for flush buffer. 

As a result, the signal handler will be same as ordinal fluentd like below.

||fluentd|
|:--:|:--:|
|INT/TERM|graceful shutdown|
|HUP|graceful restart|
|USR1|flush buffer|

naritta:
@repeatedly 
fluentd signal handler(here USR1) is prior to serverengine signal handler, so we need to concern about USR1 restarting function. flush buffer is executed if it got USR1.
But if you concern about it and we should set other signal (like WINCH?) as restarting function on serverengine just in case, please let me know.

naritta:
[MEMO] The remaining problem is
・cause error in binding UDP handle after closing and creating again.
・can't use fiddle in Ruby21-x64 and Ruby22-x64.

lib/fluent/command/debug.rb
```diff
@@ -66,7 +66,11 @@
 
 include Fluent::SystemConfig::Mixin
 
-$log = Fluent::Log.new(STDERR, Fluent::Log::LEVEL_TRACE)
+dl_opts = {}
+dl_opts[:log_level] = Fluent::Log::LEVEL_TRACE
+logdev = Fluent::Test::DummyLogDevice.new
```
tagomoris: `Fluent::Test::DummyLogDevice` is just for tests. Use `STDERR` as before.

lib/fluent/daemon.rb
```diff
@@ -0,0 +1,15 @@
...
+here = File.dirname(__FILE__)
+$LOAD_PATH << File.expand_path(File.join(here, '..'))
+
+require 'serverengine'
+require 'fluent/supervisor'
+
+server_module = Fluent.const_get(ARGV[0])
+worker_module = Fluent.const_get(ARGV[1])
+# it doesn't call ARGV in block because when reloading config, params will be initialized and then it can't use previous config.
+path = ARGV[2]
```
tagomoris: `config_path`?

lib/fluent/load.rb
```diff
@@ -17,6 +17,7 @@
 end
 require 'cool.io'
 
+require 'serverengine'
```
tagomoris: Please don't add `require` anymore here. Add it on each files really required.

lib/fluent/log.rb
```diff
@@ -53,9 +53,15 @@ def self.str_to_level(log_level_str)
       end
     end
 
-    def initialize(out=STDERR, level=LEVEL_TRACE, opts={})
-      @out = out
-      @level = level
+    def initialize(logger, opts={})
+      # overwrites logger.level= so that config reloading resets level of Fluentd::Log
+      set_logger_orig = logger.class.public_instance_method(:level=).bind(logger)
```
tagomoris: Is it not logger, but setter of log level of `logger`, right? If so, use names to mean it.

lib/fluent/log.rb
```diff
@@ -53,9 +53,15 @@ def self.str_to_level(log_level_str)
...
+      me = self
+      logger.define_singleton_method(:level=) {|level| set_logger_orig.call(level); me.level = self.level + 1 }
```
tagomoris: It's better to have comments to describe why `me.level = self.level +1`.

lib/fluent/plugin/in_forward.rb
```diff
@@ -57,11 +57,16 @@ def configure(conf)
     def start
       @loop = Coolio::Loop.new
 
-      @lsock = listen
+      socket_manager_path = ENV['SERVERENGINE_SOCKETMANAGER_PATH']
+      if Fluent.windows?
+        socket_manager_path = socket_manager_path.to_i
+      end
+      client = ServerEngine::SocketManager::Client.new(socket_manager_path)
```
tagomoris: It's better to have a utility method for these several lines (it looks to appear very often), but I have a plan of a new plugin helper for tcp/udp/tls sockets. So it's my task following this pull-request.

lib/fluent/supervisor.rb
```diff
@@ -24,8 +24,10 @@
 require 'fluent/plugin'
 require 'fluent/rpc'
 require 'fluent/system_config'
+# require 'serverengine'
+require_relative '../../../serverengine/lib/serverengine'
```
tagomoris: What's this?

lib/fluent/supervisor.rb
```diff
@@ -24,8 +24,10 @@
...
 
-if Fluent.windows?
+if ServerEngine.windows?
```
tagomoris: Don't use `ServerEngine.windows?` directly.
If it provides better way to know in windows or not, update `Fluent.windows?` method in fluent/env.rb to return result of `ServerEngine.windows?`.

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
-    def self.get_etc_passwd(user)
-      if user.to_i.to_s == user
-        Etc.getpwuid(user.to_i)
-      else
-        Etc.getpwnam(user)
-      end
+  module ServerModule
+    def before_run
+      log_opts = {suppress_repeated_stacktrace: config[:suppress_repeated_stacktrace]}
+      log = Supervisor::LoggerInitializer.new(config[:log_path], config[:log_level], config[:chuser], config[:chgroup], log_opts)
```
tagomoris: It is not `$log`, but `$log` is set to `@logger` at L45.
Is it intentional?

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+
+    def after_start
+      install_supervisor_signal_handlers
+    end
+
+    def install_supervisor_signal_handlers
+      trap :USR1 do
+        $log.debug "fluentd supervisor process get SIGUSR1"
+        supervisor_sigusr1_handler
+      end unless ServerEngine.windows?
```
tagomoris: `Fluent.windows?`

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+    end
+
+    def supervisor_sigusr1_handler
+      # @log.reopen!
```
tagomoris: Is it needed?

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+      if pid = @pm.pid
+        Process.kill(:USR1, pid)
+        # don't resuce Erro::ESRSH here (invalid status)
```
tagomoris: "rescue" (it's near from "reuse", so it's better to correct even in easy comment)

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
       end
     end
+  end
+
+  def self.load_config(path, params = {})
```
tagomoris: Class method for `Fluent` (top-level of Fluentd) namespace is not good practice. That namespace is for methods which is used from anywhere in Fluentd core and plugins.
This `load_config` method looks only for supervisor. So we should use its namespace (e.g., Fluent::Supervisor.load_config).

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+    config_mtime = File.mtime(path)
+    # reuse previous config if last load time is within 5 seconds and mtime of the config file is not changed
+    if Time.now - Time.at(params['pre_loadtime'].to_i) < 5 and config_mtime == params['pre_config_mtime']
```
tagomoris: `if params.has_key?('pre_loadtime') && Time.now - Time.at(...) ...`
I can understand that `Time.at(params['pre_loadtime'].to_i)` means `Time.at(0)` if it's missing, but not so explicit for all. We should write it in more explicit way.

tagomoris: And same about `pre_config_mtime`.

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+    inline_config = params['inline_config']
+    if inline_config == '-'
+      config_data << "\n" << STDIN.read
+    elsif inline_config
+      config_data << "\n" << inline_config.gsub("\\n","\n")
+    end
+    fluentd_conf = Fluent::Config.parse(config_data, config_fname, config_basedir, params['use_v1_config'])
+    system_config = SystemConfig.create(fluentd_conf)
+
+    log_level = system_config.log_level || params['log_level']
```
tagomoris: Command line specification must be prior than system config specification.

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+    suppress_repeated_stacktrace = system_config.suppress_repeated_stacktrace || params['suppress_repeated_stacktrace']
```
tagomoris: ditto.

lib/fluent/supervisor.rb
```diff
@@ -35,23 +37,109 @@
...
+                                path,
+                                JSON.dump(params)],
+       fluentd: fluentd_conf,
+       main_cmd: main_cmd,
+    }
+    pre_params = params.dup
+    params['pre_loadtime'] = Time.now.to_i
+    params['pre_config_mtime'] = config_mtime
+    params['pre_conf'] = se_config
+    # prevent pre_conf from being too big by reloading many times.
```
tagomoris: Very good comment!

lib/fluent/supervisor.rb
```diff
@@ -158,39 +242,11 @@ def start
       set_system_config
 
       dry_run if @dry_run
-      start_daemonize if @daemonize
       setup_rpc_server if @rpc_endpoint
```
tagomoris: Following 3 instructions should be in one `if` block.
```
if @rpc_endpoint
  setup_rpc_server
  setup_rpc_get_dump if @enable_get_dump
  run_rpc_server
end
```
Or, these 3 methods might be just one method (if it looks better than 3 methods).

lib/fluent/supervisor.rb
```diff
@@ -203,14 +259,33 @@ def options
       }
     end
 
+    def main_run
```
tagomoris: Currently, `start` is for supervised configuration, and `main_run` is for execution without supervisor. But these names don't express it.
Can you consider better names for these methods?

lib/fluent/supervisor.rb
```diff
@@ -243,82 +318,33 @@ def show_plugin_config
...
-        STDIN.reopen("/dev/null")
-        STDOUT.reopen("/dev/null", "w")
-        STDERR.reopen("/dev/null", "w")
-        @wait_daemonize_pipe_w.write @supervisor_pid.to_s
-        @wait_daemonize_pipe_w.close
-        @wait_daemonize_pipe_w = nil
-      end
-    end
-
+    # TODO: move RPC Server in serverengine side
```
tagomoris: Is it needed? What for?

lib/fluent/supervisor.rb
```diff
@@ -346,64 +372,73 @@ def stop_rpc_server
...
+        $log.debug "fluentd main process get SIGINT"
+        unless @finished
+          @finished = true
+          $log.debug "getting start to shutdown main process"
+          Fluent::Engine.stop
+        end
       end
 
-      $log.info "process finished", code: ecode
+      trap :USR1 do
```
tagomoris: We want to support force flushing of buffers even in Windows by using RPC server calls, but with this implementation, it can't be done.
This actual procedure should be written in a method, and signal handler (and supervisor-worker RPC handler or so for windows) should call it.

lib/fluent/supervisor.rb
```diff
@@ -243,82 +318,33 @@ def show_plugin_config
...
         $log.debug "fluentd RPC got /api/processes.killWorkers request"
-        supervisor_sigterm_handler
+        Process.kill :TERM, $$
         nil
       }
       @rpc_server.mount_proc('/api/plugins.flushBuffers') { |req, res|
         $log.debug "fluentd RPC got /api/plugins.flushBuffers request"
-        supervisor_sigusr1_handler
+        unless ServerEngine.windows?
+          Process.kill :USR1, $$
```
tagomoris: If ServerEngine provides any supervisor-worker communication path, we should use it instead of signaling SIGUSR1 (because it's not supported by Windows, but we want to provide buffer force flushing even in Windows if possible).

lib/fluent/supervisor.rb
```diff
@@ -243,82 +318,33 @@ def show_plugin_config
...
+        end
         nil
       }
       @rpc_server.mount_proc('/api/config.reload') { |req, res|
         $log.debug "fluentd RPC got /api/config.reload request"
-        $log.info "restarting"
-        supervisor_sighup_handler
+        unless ServerEngine.windows?
```
tagomoris: Same with SIGUSR1.

lib/fluent/supervisor.rb
```diff
@@ -346,64 +372,73 @@ def stop_rpc_server
...
-        @wait_daemonize_pipe_w.close
-        @wait_daemonize_pipe_w = nil
+        fluentd_spawn_cmd = $0 + ' '
       end
 
-      Process.waitpid(@main_pid)
-      @main_pid = nil
-      ecode = $?.to_i
+      $fluentdargv.each{|a|
+        fluentd_spawn_cmd << (a + " ")
```
tagomoris: Is this cmd string executed vis shell?
If so, these should be escaped with ShellWords. (I don't know well about Windows though.)

lib/fluent/supervisor.rb
```diff
@@ -346,64 +372,73 @@ def stop_rpc_server
...
       end
 
-      $log.info "process finished", code: ecode
+      trap :USR1 do
+        $log.debug "fluentd main process get SIGUSR1"
+        $log.info "force flushing buffered events"
+        @log.reopen!
 
-      if !@finished && Time.now - start_time < 1
-        $log.warn "process died within 1 second. exit."
```
tagomoris: Don't these check & warning exist anywhere?

lib/fluent/test/base.rb
```diff
@@ -119,7 +119,11 @@ def close
     class TestLogger < Fluent::PluginLogger
       def initialize
         @logdev = DummyLogDevice.new
-        super(Fluent::Log.new(@logdev))
+        dl_opts = {}
+        dl_opts[:log_level] = Fluent::Log::LEVEL_INFO
+        logger = ServerEngine::DaemonLogger.new(@logdev, dl_opts)
```
tagomoris: This file looks to need `require 'serverengine'`.

test/plugin/test_in_udp.rb
```diff
@@ -65,37 +77,6 @@ def test_time_format
...
-        tests.each { |test|
-          u.send(test['msg'], 0)
-        }
-        sleep 1
-      end
-
-      compare_test_result(d.emits, tests)
-    end
-  }
-
```
tagomoris: Why are these tests removed?

test/test_config.rb
```diff
@@ -155,16 +155,5 @@ def write_config(path, data)
...
-  def test_inline
-    prepare_config 
-    opts = {
-      config_path: "#{TMP_DIR}/config_test_1.conf",
-      inline_config: "<source>\n  type http\n  port 2222\n </source>"
-    }
-    assert_nothing_raised do
-      Fluent::Supervisor.new(opts)
-    end  
-  end
```
tagomoris: ditto.

tagomoris:
@naritta I added comments for code. Please check it out.

naritta:
Thank you very much.
I will check it.

lib/fluent/supervisor.rb
naritta: ```$log``` is initialized in ```log.init``` and pass this ```$log``` to serverengine.
If it cause confusion, I'll change the name.

lib/fluent/supervisor.rb
naritta: I think we can reflect new setting by reloading config file when dynamic reloading.
If it's wrong, please tell me.

lib/fluent/supervisor.rb
naritta: About reloading config by rpc, I fixed to kill worker and restart by auto restarting function on windows, as temporary plan.

About flushing buffer by rpc, I think we have to implement supervisor-worker communication path is serverengine.
If you have any better idea, please tell me.

naritta:
@tagomoris 
I fixed and added some replying comments.
Could you review when you have time?

lib/fluent/log.rb
```diff
@@ -53,9 +53,17 @@ def self.str_to_level(log_level_str)
...
 
-    def initialize(out=STDERR, level=LEVEL_TRACE, opts={})
-      @out = out
-      @level = level
+    def initialize(logger, opts={})
+      # overwrites logger.level= so that config reloading resets level of Fluentd::Log
+      orig_logger_level_setter = logger.class.public_instance_method(:level=).bind(logger)
+      me = self
+      # DaemonLogger's log level setting has difference of one level from that of this class.
+      # Then level in this class will be self.level + 1.
```
tagomoris: Does DaemonLogger have same # of levels with Fluentd's? (It's just difference about origin 0 or 1?)
If so, good to note it.

lib/fluent/supervisor.rb
tagomoris: Did you remove these lines?

naritta: I moved these lines into load_config for when reloading. Sorry for not telling.

tagomoris: Ah, I found it. Thanks.

lib/fluent/supervisor.rb
```diff
@@ -35,21 +37,187 @@
...
+      log_level = system_config.log_level || params['log_level']
+      suppress_repeated_stacktrace = system_config.suppress_repeated_stacktrace || params['suppress_repeated_stacktrace']
+      log_path = params['log_path']
+      chuser = params['chuser']
+      chgroup = params['chgroup']
+      rpc_endpoint = system_config.rpc_endpoint
+      enable_get_dump = system_config.enable_get_dump
+
+      log_opts = {suppress_repeated_stacktrace: suppress_repeated_stacktrace}
+      log = Supervisor::LoggerInitializer.new(log_path, log_level, chuser, chgroup, log_opts)
```
tagomoris: Finally I thought that variable name `log` is bad name. There're many `log` used as logger (and almost same about `$log`), but this `log` is not logger, but initializer.
So this name should be another one to show "this is initializer, not logger itself", and it's better to have comment to show that `.init` sets initialized logger to `$log`.

test/plugin/test_in_udp.rb
```diff
@@ -63,23 +80,33 @@ def test_time_format
...
+          {'msg' => {'k' => 123, 'message' => 'tcptest1'}.to_json + "\n", 'expected' => 'tcptest1'},
+          {'msg' => {'k' => 'tcptest2', 'message' => 456}.to_json + "\n", 'expected' => 456},
+      ],
+      '/^\\[(?<time>[^\\]]*)\\] (?<message>.*)/' => [
+          {'msg' => '[Sep 10 00:00:00] localhost: ' + 'x' * 100 + "\n", 'expected' => 'localhost: ' + 'x' * 100},
+          {'msg' => '[Sep 10 00:00:00] localhost: ' + 'x' * 1024 + "\n", 'expected' => 'localhost: ' + 'x' * 1024},
+      ]
   }.each { |format, test_cases|
     define_method("test_msg_size_#{format[0] == '/' ? 'regexp' : format}") do
+      if Fluent.windows?
```
tagomoris: Doesn't `ServerEngine::SocketManager::Server` support Windows environment?

tagomoris:
I added some more review comments right now.
@naritta Could you comment how you will fix (or why you won't fix these) per review comments? It's a bit hard to follow how they are fixed from entire diff, and becomes much harder if code are moved to another place.

lib/fluent/supervisor.rb
naritta: I fixed here.
https://github.com/fluent/fluentd/pull/880/commits/6ee3397f90066ea332b63cb38dd05ca8dcf22034

test/plugin/test_in_udp.rb
naritta: I thought it's safer that we use original unused_port method in fluentd.
I fixed to use serverengine's api here. https://github.com/fluent/fluentd/pull/880/commits/75a07d24881a8087d6d0ea070b774869f562baf4

lib/fluent/log.rb
naritta: The comment is not written in serverengine side, because serverengine is used for excepting fluentd also.
If this answer is not your hoping, please tell me.

naritta:
@tagomoris 
Thank you for reviewing and sorry for inconvenience.
I fixed and added comments.
And I also added test for supervisor.

lib/fluent/log.rb
tagomoris: I had a doubt that:
* `level` of Fluentd logger is a number, between 0-5
* the `+1` means ServerEngine's logger level is also a number, but between 1-6.
* is it "really" collect?

Fluentd engineer might be unfamiliar to ServerEngine (because these are different product :)), so we should add comments about ServerEngine's log levels especially for the case like doing `+1` on Fluentd's log level.

test/plugin/test_in_forward.rb
```diff
@@ -7,6 +7,22 @@
 require 'fluent/plugin/in_forward'
 
 class ForwardInputTest < Test::Unit::TestCase
+  class << self
+    def startup
+      if Fluent.windows?
+        socket_manager_path = ServerEngine::SocketManager::Server.generate_path
+      else
+        socket_manager_path = ServerEngine::SocketManager::Server.generate_path + __method__.to_s
+      end
```
tagomoris: Could you reply here about what the difference between windows and else?
In other words, I want to know the difference between `Server.generate_path` and `Server.generate_path + __method__.to_s` and why (to write plugin helper for these sockets).

naritta: serverengine's API generates
・in unix
generates Unix domain socket and path is time + pid
・in windows
generates TCPServer and path is unused port.

Then, in unix path will be same when the api is called at same time and in same process.
(In windows, it will be not same bacause it will find in each time when it's called)
So I added ``` __method__.to_s``` not to be same in unix.

But It seems to be confusing and I realized we need to create only one socket manager server before tests begin. (we don't have to create for each one test)
Then I fixed it now. Sorry for taking your time.

lib/fluent/log.rb
naritta: I added the comment.
https://github.com/fluent/fluentd/pull/880/commits/3cb29cb9912052e715c0972701923906a4ee973a

This is for that Serverengine logger ```level=``` overwrites fluentd's logger level. (Not fluentd logger ```level=``` overwrites serverengine)

tagomoris:
This looks ready to merge.

tagomoris:
Done!

naritta:
Thank you!

PR #882, opened 2016-04-05
Add 'Access-Control-Allow-Origin' to in_http plugin

HatsuneMiku3939:
According to the document
https://en.wikipedia.org/wiki/Cross-origin_resource_sharing
The server should respond with 'Access-Control-Allow-Origin' header
in its response indicating which origin sites are allowed.


repeatedly:
Thx!

HatsuneMiku3939:
Thx!

PR #883, opened 2016-04-06
Enforce to use `router.emit`

okkez:


tagomoris:
LGTM

riywo:
In the code of v0.14.0, there is still `Engine.emit_stream`.
https://github.com/fluent/fluentd/blob/v0.14.0/lib/fluent/process.rb#L172

Because of this code, `detach_process` doesn't work for fluent-plugin-kinesis. I know using `detach_process` in output plugin is not a good idea, but could you provide a backward compatibility? If it is difficult, I'd like to try to [get rid of this feature ASAP...](https://github.com/awslabs/aws-fluent-plugin-kinesis/issues/71)

tagomoris:
It looks pretty large & complex patch to support router with `detach_process` feature.
* introduce `router` (and `log`) from DetachProcessMixin to DetachProcessImpl
* use introduced `router` in DetachProcessImpl

The biggest problem is that there are no tests about this feature. So it can't be fixed so quickly.
Please open another issue for this if it's serious for you.

PR #884, opened 2016-04-06
Add Vagrantfile to run Fluentd in Linux box very easily

tagomoris:
especially for OSX users

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,17 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+  config.vm.network "private_network", type: "dhcp"
+  config.vm.provision "shell", inline: <<-SHELL
+    sudo apt-get install -y software-properties-common
+    sudo apt-add-repository ppa:brightbox/ruby-ng
+    sudo apt-get update
+    sudo apt-get install -y build-essential git ruby2.3 ruby2.3-dev
+    sudo apt-get install -y ruby-switch
+    sudo ruby-switch --set ruby2.3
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -0,0 +1,19 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+rake_target = ENV['TEST'] ? "test TEST=#{ENV['TEST']}" : "test"
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+  config.vm.network "private_network", type: "dhcp"
+  config.vm.provision "shell", inline: <<-SHELL
+    sudo apt-get install software-properties-common
```
cosmo0920: Should we use  `apt-get install -y` in this line?
This line works for `ubuntu/trusty64`, though.

Vagrantfile
```diff
@@ -0,0 +1,19 @@
...
+    sudo apt-add-repository ppa:brightbox/ruby-ng
+    sudo apt-get update
+    sudo apt-get install -y git ruby2.3 ruby2.3-dev
+    sudo apt-get install ruby-switch
+    sudo ruby-switch --set ruby2.3
+    gem install bundler
+    (cd /vagrant; bundle install; bundle exec rake #{rake_target})
```
cosmo0920: It seems that this installation process does not install gcc.
Is it enough for testing?

Vagrantfile
tagomoris: Right.

Vagrantfile
tagomoris: It sounds good to add these.

tagomoris: Adding `build-essential` only looks to work.

PR #887, opened 2016-04-07
fix not to overwrite optional headers/attrs of global logger

tagomoris:
PluginLoggerMixin did overwrite `optional_header` and `optional_attrs` of `@log`, and `@log`(PluginLogger instance) did overwrite `optional_*` of `@logger`.
But `@logger` is always equal to `$log`. So this code overwrites `optional_*` of global default logger always. It's a bug.

This patch will fix this problem by cloning Logger object always to create PluginLogger.


cosmo0920:
Looks good. 👍

PR #889, opened 2016-04-10
Add "encoding" parameter to in_tail plugin

miyakawataku:
Currently, ``in_tail`` plugin emits strings with ASCII-8BIT encoding, which is resulted from IO#readpartial. This behavior entails inconvenience when filtering records. For example:

- [What do we write regex at filter-grep in Japanease? / Fluentd Google Group](https://groups.google.com/forum/#!topic/fluentd/0AxzJu2SMS4)

This PR adds ``encoding`` parameter to in_tail plugin, leaving ASCII-8BIT as default for compatibility.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -66,7 +66,7 @@ module Fluent
       begin
         Encoding.find(encoding_name)
       rescue ArgumentError => e
-        raise ConfigError.new(e.message)
+        raise ConfigError, e.message
       end
     end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -61,6 +61,14 @@ module Fluent
     config_param :multiline_flush_interval, :time, default: nil
     desc 'Enable the additional watch timer.'
     config_param :enable_watch_timer, :bool, default: true
+    desc 'The encoding of the input.'
+    config_param :encoding, default: nil do |encoding_name|
+      begin
+        Encoding.find(encoding_name)
+      rescue ArgumentError => e
+        raise ConfigError.new(e.message)
+      end
+    end
 
     attr_reader :paths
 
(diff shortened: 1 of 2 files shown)


test/plugin/test_in_tail.rb
```diff
@@ -370,6 +381,42 @@ def test_whitespace
     assert_equal({"message" => "tab	"}, emits[5][2])
   end
 
+  def test_default_encoding
+    File.open("#{TMP_DIR}/tail.txt", "wb") {|f| }
```
repeatedly: Setting `read_from_head true` is better than creating a file here?

test/plugin/test_in_tail.rb
```diff
@@ -370,6 +381,42 @@ def test_whitespace
...
+        f.puts "test"
+      }
+      sleep 1
+    end
+
+    emits = d.emits
+    assert_equal(Encoding::ASCII_8BIT, emits[0][2]['message'].encoding)
+  end
+
+  def test_explicit_valid_encoding
```
repeatedly: Code is same as above. Please use `data` to avoid copy and paste test.

miyakawataku:
@repeatedly I have refined the tests.

It seems the failure on Travis CI (https://travis-ci.org/fluent/fluentd/jobs/122757775) is not related with this PR. Could you restart the job?

repeatedly:
It seems ok for me.

@tagomoris @sonots How about this?

lib/fluent/plugin/in_tail.rb
```diff
@@ -61,6 +61,14 @@ def initialize
     config_param :multiline_flush_interval, :time, default: nil
     desc 'Enable the additional watch timer.'
     config_param :enable_watch_timer, :bool, default: true
+    desc 'The encoding of the input.'
+    config_param :encoding, default: nil do |encoding_name|
+      begin
+        Encoding.find(encoding_name)
+      rescue ArgumentError => e
+        raise ConfigError.new(e.message)
```
tagomoris: Newing error looks a bit curious for me. I like: `raise ConfigError, e.message`

sonots:
LGTM!!

tagomoris:
Totally LGTM.

repeatedly:
Thanks!

PR #891, opened 2016-04-11
fluent-debug: Fix broken fluent-debug command

cosmo0920:
`Fluent::Engine.init` should be passed `system_config` in `Fluent::SystemConfig::Mixin`.


cosmo0920:
I've noticed that this PR's patch should backport into v0.12 branch. Because `system_config` refactoring breaks `fluent-debug` command. :'(

tagomoris:
Good catch!

cosmo0920:
Thanks!

Issue #848, opened 2016-03-11
Resolve all warnings in CI


tagomoris:
Currently, Fluentd code reports many warnings if it is enabled.
It shows potential errors and troubles in code. So warnings should be resolved basically.

This ticket is blocker for `v0.14.0` release.
1. revert https://github.com/fluent/fluentd/pull/846
2. remove all warnings from fluentd's code
3. remove warnings from dependencies as much as possible

okkez:
See also https://bugs.ruby-lang.org/issues/12026

okkez:
NOTE:

* `lib/fluent/plugin/socket_util.rb:61: warning: instance variable @timeout not initialized`
* https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/socket_util.rb#L61
* https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_stream.rb#L120

I think we cannot set `@timeout` from outside of the class.
(I think we can replace `@timeout` with `0` or initialize `@timeout` with `0`.)

----

```
lib/fluent/test/input_test.rb:112: warning: method redefined; discarding old emit_stream
lib/fluent/test/input_test.rb:112: warning: previous definition of emit_stream was here
lib/fluent/test/input_test.rb:115: warning: method redefined; discarding old emit_stream
lib/fluent/test/input_test.rb:115: warning: previous definition of emit_stream was here
lib/fluent/test/input_test.rb:112: warning: method redefined; discarding old emit_stream
```

* https://github.com/fluent/fluentd/blob/master/lib/fluent/test/input_test.rb#L110
  * Invoke `Fluent::Test::InputTestDriver#run` multiple times in a process
  * `Fluent::Engine` and `instance.router` is singleton object

tagomoris:
#966 will solve this problem.

tagomoris:
After #966, there are no warning shown in running tests. There are still some warnings in loading scripts, but I think it's not serious blocker for the project.
(For other committers: If it's wrong, please reopen this.)

PR #892, opened 2016-04-11
Remove redundant `require`

okkez:
This change suppresses following warnings(about 130 lines warnings):

lib/fluent/config/configure_proxy.rb:17: warning: loading in progress, circular require considered harmful
lib/fluent/config/types.rb:20: warning: loading in progress, circular require considered harmful
lib/fluent/event.rb:17: warning: loading in progress, circular require considered harmful
lib/fluent/filter.rb:19: warning: loading in progress, circular require considered harmful
lib/fluent/agent.rb:18: warning: loading in progress, circular require considered harmful
lib/fluent/output.rb:21: warning: loading in progress, circular require considered harmful
lib/fluent/root_agent.rb:21: warning: loading in progress, circular require considered harmful

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -28,7 +28,7 @@ module Fluent
   class Agent
     include Configurable
 
-    def initialize(opts = {})
+    def initialize(log:)
       super()
 
       @context = nil
@@ -37,7 +37,7 @@ module Fluent
       @started_outputs = []
       @started_filters = []
 
-      @log = Engine.log
+      @log = log
       @event_router = EventRouter.new(NoMatchMatch.new(log), self)
       @error_collector = nil
     end
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -15,7 +15,6 @@
 #
 
 require 'fluent/configurable'
-require 'fluent/engine'
 require 'fluent/plugin'
 require 'fluent/output'
 
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -14,8 +14,6 @@
 #    limitations under the License.
 #
 
-require 'fluent/configurable'
-
 module Fluent
   module Config
     class ConfigureProxy
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -46,6 +46,7 @@ module Fluent
         @required = opts[:required]
         @multi = opts[:multi]
         @alias = opts[:alias]
+        @type_lookup = opts[:type_lookup]
 
         raise "init and required are exclusive" if @init && @required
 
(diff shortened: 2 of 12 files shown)


okkez:
See also #848

lib/fluent/agent.rb
```diff
@@ -15,7 +15,6 @@
 #
 
 require 'fluent/configurable'
-require 'fluent/engine'
```
tagomoris: Removing this `require` looks good, but `Engine` is referred below to get logger.
I think we need to fix `Agent` to get logger in constructor, and to pass it from Engine when RootAgent is created in `engine.rb`.

lib/fluent/root_agent.rb
```diff
@@ -18,7 +18,6 @@
 
 require 'fluent/config/error'
 require 'fluent/agent'
-require 'fluent/engine'
```
tagomoris: `#handle_emits_error` refers `Engine.now`.
It should be replaced with `Fluent::EventTime.now` (with `require 'fluent/time'`)

lib/fluent/event.rb
```diff
@@ -14,8 +14,6 @@
 #    limitations under the License.
 #
 
-require 'fluent/engine'
```
tagomoris: EventStream classes requires `msgpack_factory`. Please add requirement for `fluent/msgpack_factory` and mixin `Fluent::MessagePackFactory::Mixin`, then use `msgpack_factory` (or `msgpack_packer`) method instead of `Engine.msgpack_factory`.

lib/fluent/config/types.rb
```diff
@@ -17,7 +17,6 @@
 require 'json'
 
 require 'fluent/config/error'
-require 'fluent/configurable'
```
tagomoris: It's better to:
* remove this requirement
* move `Configurable.register_type` calls to `configurable.rb`
* move `require 'fluent/config/types'` at the end of `configurable.rb` to head of that file

lib/fluent/config/configure_proxy.rb
```diff
@@ -14,8 +14,6 @@
 #    limitations under the License.
 #
 
-require 'fluent/configurable'
```
tagomoris: `ConfigureProxy` need to lookup types from registry.
Could you add a constructor argument of lambda to lookup types, and feed it to `ConfigureProxy.new` by `Configure::ClassMethods#configure_proxy`.
```
# configure.rb
type_lookup = ->(type){ Fluent::Configurable.lookup_type(type) }
proxy = Fluent::Config::ConfigureProxy.new(mod_name, type_lookup, required: true, multi: false)

# configure_proxy.rb
type = :string if type.nil?
block ||= @type_lookup.call(type)
```

tagomoris:
@okkez I added some comments. Could you add some more fixes?

okkez:
:ok: I will push fixes in this week.

okkez:
I've pushed fixes for all comments.

okkez:
Hmm, CI failed.

okkez:
CI failed against cc73ce2.
Please retry, I think this failure is not related to my changes.

tagomoris:
Thank you very much! LGTM.

PR #893, opened 2016-04-11
Add missing assertion

okkez:
This should be added in b885cead.


okkez:
See also #848

tagomoris:
LGTM. Thanks!

PR #895, opened 2016-04-12
Shutdown OutputThread in parallel

komamitsu:
I noticed BufferedOutput was slow when I ran 32 threads in it. BufferedOutput#shutdown stops and joins their threads sequentially. So I modified it to shutdown their threads in parallel.

As a result, it becomes several times faster in my case.

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -82,13 +82,21 @@ module Fluent
       @thread = Thread.new(&method(:run))
     end
 
-    def shutdown
+    def signal
       @finish = true
       @mutex.synchronize {
         @cond.signal
       }
       Thread.pass
+    end
+
+    def shutdown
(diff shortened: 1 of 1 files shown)


repeatedly:
It seems good for v0.14 for me.

tagomoris:
Fluentd v0.14 will use thread plugin helper for flush threads of output plugins. So this patch will be outdated very soon.
It's good only for v0.12 backport.

komamitsu:
@repeatedly @tagomoris Kindly ping.
Porting to v0.12 sounds reasonable to me.

repeatedly:
@komamitsu Could you send a patch to v0.12, not master?

komamitsu:
Okay. Created a new one https://github.com/fluent/fluentd/pull/938

PR #899, opened 2016-04-15
Logger dumps error with classname

tagomoris:
In these days, many plugin authors (and core developers) writes logging code like below many times.
```ruby
log.error "my message", error_class: e.class, error: e.to_s, other: attributes
```
Fairly speaking, this is waste of time and key typing.

Fluentd will log errors like below with this change by just specifying `error: e`:
```ruby
log.error "my message", error: e
# [error]: my message error=MyErrorClass[message of e]
```

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -292,7 +292,7 @@ module Fluent
       }
 
       map.each_pair {|k,v|
-        if k == "error".freeze
+        if k == "error".freeze && v.is_a?(Exception) && !map.has_key?("error_class")
           message << " error_class=#{v.class.to_s} error=#{v.to_s.inspect}"
         else
           message << " #{k}=#{v.inspect}"
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -293,7 +293,7 @@ module Fluent
 
       map.each_pair {|k,v|
         if k == "error".freeze
-          message << " error=#{v.class}[#{v.message}]"
+          message << " error_class=#{v.class.to_s} error=#{v.to_s.inspect}"
         else
           message << " #{k}=#{v.inspect}"
         end
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -292,7 +292,11 @@ module Fluent
       }
 
       map.each_pair {|k,v|
-        message << " #{k}=#{v.inspect}"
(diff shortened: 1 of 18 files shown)


tagomoris:
@repeatedly @sonots How do you think about this change?

sonots:
I am positive for specifying only an error object, but somebody might be watching logs of fluentd by a log monitoring system for alerting. Changing log format would break compatibilities for their system, so I think we should keep compatibilities of log format.

tagomoris:
@sonots Do you mean that we should print errors in format below?
```
error_class=MyErrorClass error="message of e"
```

sonots:
@tagomoris yes!

tagomoris:
@sonots Hmm, it looks a bit verbose for me, and v0.14 may be a good time to do that kind of change.
@repeatedly How do you think?

repeatedly:
Keeping compatibility is good for existing users. Maybe, asking it on mailing list is better.

> v0.14 may be a good time to do that kind of change.

This change affects v0.12 because we will backport several changes to v0.12 and skipping this patch is hard...
If you want to change error message format, we should suspend this patch until just before v0.14 release.

tagomoris:
Okay, I'll change format of errors.

tagomoris:
Done.

lib/fluent/log.rb
```diff
@@ -292,7 +292,11 @@ def event(level, args)
       }
 
       map.each_pair {|k,v|
-        message << " #{k}=#{v.inspect}"
+        if k == "error".freeze
+          message << " error_class=#{v.class.to_s} error=#{v.to_s.inspect}"
```
sonots: `v` would be a string like https://github.com/fluent/fluentd/blob/88054a0b4299845c6696335876cba03bb33ea6e3/lib/fluent/output.rb#L127, so need to add `and v.is_a?(Exception)`.

sonots: Need to add another condition `!map.has_key?('error_class')`, otherwirse, ` error_class=` appears twice for old codes such as:

```
log.error 'message', error_class: e.class, error: e
```

tagomoris: That's right. I fixed so.

sonots:
LGTM!

tagomoris:
Merged.

PR #904, opened 2016-04-19
Serialize Fluent::EventTime as Integer in JSON. fixes #897

### LINKED ISSUE #897: Fluent::EventTime is serialized as String for JSON
```ruby
["tag", Fluent::EventTime.now, {"key" => "value"}].to_json #=> '["tag","1460544188",{"key":"value"}]'
```

This may cause some problems in output plugins.

mururu:


tagomoris:
@mururu Great patch!
Could you add just one more test case for `JSON.dump`?

mururu:
Done.

tagomoris:
Thank you!

PR #905, opened 2016-04-19
Add new base class for Filter plugins

tagomoris:


tagomoris:
I'm going to merge this one: just different namespace (`Fluent::Plugin::Filter`) and v0.14 style plugin design.

PR #906, opened 2016-04-20
Add Fluent::Plugin::Formatter, Fluent::Plugin::Parser and formatter/parser plugin files

tagomoris:
* with compatibility layer for v0.12 style reference

### DIFF
--- a/lib/fluent/compat/formatter.rb
+++ b/lib/fluent/compat/formatter.rb
@@ -1,0 +1,109 @@
+#
+# Fluentd
+#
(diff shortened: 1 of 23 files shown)


tagomoris:
@repeatedly @sonots This change is not completed yet, but I'll do same thing for Parser. And I'll also create compat module/classes for Input/Output/Filter classes.
Please check this way is ok for you too or not.

tagomoris:
Done for all parser/formatter plugins.

tagomoris:
Are there no problems? If so, I'll merge this later.

PR #907, opened 2016-04-20
Fix logger on Fluent::StringUtil to use $log.

toyama0919:
Hello.

fix bug. related #657

> I found that log is not defined in Fluent::StringUtil.match_regexp. So this rescue block will raise NameError every time.


tagomoris:
LGTM, especially for adding tests!

toyama0919:
@tagomoris m(_ _)m

PR #908, opened 2016-04-20
Fix typo

yuuki1224:
Found typo in config exception message. Plugin name should start with `fluent-plugin` rather than `fluentd-plugin`?


repeatedly:
Thx!

PR #909, opened 2016-04-21
Add test for Fluent::Log

okkez:


tagomoris:
@okkez Can you add more tests for PluginLogger and PluginLoggerMixin?
I think it should have dummy plugin instances, and I think we can use `Fluent::Plugin::TestBase` for it.

okkez:
OK, I will try to add tests for PluginLogger and PluginLoggerMixin.

okkez:
@tagomoris I've added tests for PluginLogger and PluginLoggerMixin.

tagomoris:
@okkez Thank you very much!

PR #912, opened 2016-04-22
Compatibility layers for v0.12 plugins

tagomoris:
This change provides compatibility layers for v0.12 style plugins of:
* Input
* Output
* Filter

And this change also replaces default buffer plugins to v0.14 implementations.


tagomoris:
I've done for almost all tasks for v0.14 plugin APIs.
The rest thing is:
* remove `Fluent::Buffer` and its subclasses `fluent/plugin/buf_memory.rb` and `buf_file`, and these tests
* rename `buf_memory2.rb` and `buf_file2.rb` to `buf_memory.rb` and `buf_file`
* update few default values of `memory2`, `file2`

tagomoris:
I think startup/shutdown sequence need this hack: https://gist.github.com/unak/f36def728e8c5411eff8

tagomoris:
@repeatedly @sonots Anyway, could you review this change?

lib/fluent/plugin/out_file.rb
```diff
@@ -87,7 +102,13 @@ def configure(conf)
       @formatter = Plugin.new_formatter(@format)
       @formatter.configure(conf)
 
-      @buffer.symlink_path = @symlink_path if @symlink_path
+      if @symlink_path && @buffer.respond_to?(:path)
+        (class << @buffer; self; end).module_eval do
+          prepend SymlinkBufferMixin
```
repeatedly: `prepend` doesn't work on Ruby 1.9 so this code should be backported to v0.12, right?

tagomoris: I don't add label `v0.12` on this pull-request... and this hack is required because Buffer plugin of v0.14 API doesn't have symbolic link feature.
There's no reason to backport this change to v0.12.

repeatedly: My concern is this plugin is also used in v0.12.
So if someone writes the patch to v0.12 out_file, v0.14 changes may conflict.

tagomoris: Conflicts may occur if someone want to change L90 (in previous version), but any other changes will not conflict this change.

lib/fluent/compat/output.rb
```diff
@@ -0,0 +1,548 @@
...
+        ["flush_at_shutdown",  "flush_at_shutdown"],
+      ]
+
+      def configure(conf)
+        bufconf = CompatOutputUtils.buffer_section(conf)
+        config_style = (bufconf ? :v1 : :v0)
+        if config_style == :v0
+          buf_params = {
+            "flush_mode" => "fast",
+            "retry_type" => "expbackoff",
```
repeatedly: 'expbackoff' is too short for me. 'exponential_backoff' is better.
Maybe, almost users don't change this parameter so verbose is better.

lib/fluent/compat/output.rb
```diff
@@ -0,0 +1,548 @@
...
+      desc 'The size of each buffer chunk.'
+      config_param :buffer_chunk_limit, :size, default: 8*1024*1024
+      desc 'The length limit of the chunk queue.'
+      config_param :buffer_queue_limit, :integer, default: 256
+      desc 'The action when the size of buffer queue exceeds the buffer_queue_limit.'
+      config_param :buffer_queue_full_action, :enum, list: [:exception, :block], default: :exception
+
+      config_param :flush_at_shutdown, :bool, default: true
+
+      PARAMS_MAP = [
```
repeatedly: Why use Array instead of Hash?
`{"@type" => "buffer_type", ...}` is better than `[["buffer_type", "@type"], ...]` for me.

lib/fluent/compat/output.rb
```diff
@@ -0,0 +1,548 @@
...
+            buf_params[newer] = conf[older] if conf.has_key?(older)
+          end
+
+          bufconf = Fluent::Config::Element.new('buffer', '', buf_params, [])
+
+          conf.elements << bufconf
+
+          secconf = CompatOutputUtils.secondary_section(conf)
+          if secconf
+            if secconf['type'] && !secconf['@type']
```
repeatedly: These lines could be moved to secondary_section method?
And I want to log warning like `'type' is deprecated. 'type' parameter will be ignored since v1. Use @type instead`.

lib/fluent/compat/output.rb
```diff
@@ -0,0 +1,548 @@
...
+                            end
+          buf_params["timekey_range"] = @_timekey_range
+
+          bufconf = Fluent::Config::Element.new('buffer', 'time', buf_params, [])
+
+          conf.elements << bufconf
+
+          secconf = CompatOutputUtils.secondary_section(conf)
+          if secconf
+            if secconf['type'] && !secconf['@type']
```
repeatedly: ditto

lib/fluent/compat/output.rb
```diff
@@ -0,0 +1,548 @@
...
+      attr_accessor :localtime
+
+      # config_section :buffer, param_name: :buffer_config, init: true, required: false, multi: false, final: true do
+      #   config_set_default :@type, 'file2'
+      # end
+      config_section :buffer, param_name: :buffer_config do
+        config_set_default :@type, 'file2'
+      end
+
+      PARAMS_MAP = [
```
repeatedly: ditto

lib/fluent/compat/output.rb
tagomoris: Because it needs just iterating, not fetching by keys. Array is container for just iterating by `#each`, and it looks better for me rather than using Hash only with `#each_pair`

lib/fluent/compat/output.rb
tagomoris: IMO changing `conf` object should be done in `configure` method explicitly, not in `CompatOutputUtils.secondary_section` (in implicit way), for readability.
Showing `type` deprecation looks good idea.

lib/fluent/compat/output.rb
tagomoris: OK, I'll fix it as longer.

lib/fluent/compat/output.rb
okkez: We can iterate Hash as following:

```
>> {a: 1, b: 2}.each{|k,v| p [k,v]}
[:a, 1]
[:b, 2]
```

tagomoris: Yeah, but what i want to mean is Hash is not needed just for iteration.

lib/fluent/compat/output.rb
repeatedly: From my view, this modification is for backward compatibility, not actual plugin's behaviour.
So helper method can hide such dirty code from plugin code.
I think this is readable for non-core developers.

tagomoris: Yeah, it should be readable only for core-developers, and I think it's easy to understand for me.

repeatedly: I don't think understand for me.

@sonots What do you think?

lib/fluent/compat/output.rb
repeatedly: Array is good for just iterating, but an array of array is a bit different.
Hash is just for representing key => value mapping data so hash fits this case.

tagomoris: Okay, I've done.

lib/fluent/compat/output.rb
sonots: Not so strong opinion, but if I were you, I put them into `secondary_section` method.

tagomoris: I added an another method for obsoleted `type` handling.

tagomoris:
I will merge this change after CI green unless further comments exist.

PR #913, opened 2016-04-22
fix assigning an object with unconsistent API

tagomoris:


tagomoris:
@sonots could you check this please?

sonots:
No problem

PR #914, opened 2016-04-22
Add standard chunking format

tagomoris:
This change is to add formatting for buffer chunks. Output without explicit `#format` definition will use this standard format to write event streams to buffer chunks, and actually, it's same with `EventStream.to_msgpack_stream`.
At the same time, this change moves a method to iterate events from chunks to `event.rb` as mixin. It's about event streams, not chunk itself.

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -51,7 +51,6 @@ module Fluent
     end
   end
 
-
   class OneEventStream < EventStream
     def initialize(time, record)
       @time = time
@@ -175,7 +174,6 @@ module Fluent
     end
 
     def each(&block)
-      # TODO format check
       msgpack_unpacker.feed_each(@data, &block)
       nil
     end
@@ -184,5 +182,17 @@ module Fluent
       @data
     end
   end
-end
 
+  module ChunkMessagePackEventStreamer
+    include MessagePackFactory::Mixin
+    # chunk.extend(ChunkEventStreamer)
+    #  => chunk.each{|time, record| ... }
+    def each(&block)
+      open do |io|
+        msgpack_unpacker(io).each(&block)
+      end
+      nil
+    end
+    alias :msgpack_each :each
+  end
+end
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -194,5 +194,9 @@ module Fluent
       nil
     end
     alias :msgpack_each :each
+
+    def to_msgpack_stream
+      read
+    end
   end
 end
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -21,6 +21,10 @@ module Fluent
     include Enumerable
     include MessagePackFactory::Mixin
 
+    def records
+      raise NotImplementedError, "DO NOT USE THIS CLASS directly."
+    end
+
     def repeatable?
       false
     end
@@ -29,7 +33,8 @@ module Fluent
       raise NotImplementedError, "DO NOT USE THIS CLASS directly."
     end
 
-    def to_msgpack_stream
+    def to_msgpack_stream(time_int: false)
+      return to_msgpack_stream_forced_integer if time_int
       out = msgpack_packer
       each {|time,record|
         out.write([time,record])
@@ -57,6 +62,10 @@ module Fluent
       OneEventStream.new(@time, @record.dup)
     end
 
+    def records
+      1
+    end
+
     def repeatable?
       true
     end
@@ -81,6 +90,10 @@ module Fluent
       ArrayEventStream.new(entries)
     end
 
+    def records
+      @entries.size
+    end
+
     def repeatable?
       true
     end
@@ -102,7 +115,7 @@ module Fluent
   #
   # Use this class as below, in loop of data-enumeration:
   #  1. initialize blank stream:
-  #     streams[tag] ||= MultiEventStream
+  #     streams[tag] ||= MultiEventStream.new
   #  2. add events
   #     stream[tag].add(time, record)
   class MultiEventStream < EventStream
@@ -117,6 +130,10 @@ module Fluent
         es.add(time, record.dup)
       }
       es
+    end
+
+    def records
+      @time_array.size
     end
 
     def add(time, record)
@@ -144,8 +161,13 @@ module Fluent
 
   class MessagePackEventStream < EventStream
     # Keep cached_unpacker argument for existence plugins
-    def initialize(data, cached_unpacker = nil)
+    def initialize(data, records = 0, cached_unpacker = nil)
       @data = data
+      @records = records
+    end
+
+    def records
+      @records
     end
 
     def repeatable?
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -180,7 +180,7 @@ module Fluent
         chunk.synchronize do
           begin
             chunk.append(data)
-            if !size_over?(chunk) || force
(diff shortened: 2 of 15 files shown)


tagomoris:
This change is basis for many features in future.
* Very simple buffered plugin only with `#write` method
  * It can share many implementations with non-buffered procedure `#process` because chunk can be iterated
* Event streams now know # of records in it
  * It leads detailed monitoring how many records are processed in Fluentd process or by each plugins

tagomoris:

lib/fluent/event.rb
```diff
@@ -21,6 +21,10 @@ class EventStream
     include Enumerable
     include MessagePackFactory::Mixin
 
+    def records
```
repeatedly: The name `records` seems to represent `returns the event entries`.
`num_records` or `size`, rubish, is better for me.

lib/fluent/event.rb
```diff
@@ -29,7 +33,8 @@ def each(&block)
       raise NotImplementedError, "DO NOT USE THIS CLASS directly."
     end
 
-    def to_msgpack_stream
+    def to_msgpack_stream(time_int: false)
+      return to_msgpack_stream_forced_integer if time_int
```
repeatedly: Direct `to_msgpack_stream_forced_integer` call is deprecated?

lib/fluent/event.rb
```diff
@@ -162,5 +182,21 @@ def to_msgpack_stream
...
+    #  => chunk.each{|time, record| ... }
+    def each(&block)
+      open do |io|
+        msgpack_unpacker(io).each(&block)
+      end
+      nil
+    end
+    alias :msgpack_each :each
+
+    def to_msgpack_stream
```
repeatedly: `time_int: false` argument is not needed?

lib/fluent/event.rb
```diff
@@ -144,16 +160,20 @@ def each(&block)
 
   class MessagePackEventStream < EventStream
     # Keep cached_unpacker argument for existence plugins
-    def initialize(data, cached_unpacker = nil)
+    def initialize(data, records = 0, cached_unpacker = nil)
```
repeatedly: Some 3rd party plugins use `MessagePackEventStream.new(data, @cached_unpadker)` in their code.
Should we break these code?

lib/fluent/plugin/buffer.rb
```diff
@@ -198,6 +198,49 @@ def emit(metadata, data, force: false)
         emit_step_by_step(metadata, data)
       end
 
+      def emit_bulk(metadata, bulk, records)
+        return if bulk.nil? || bulk.empty?
```
repeatedly: When emit_bulk receives `nil` bulk data?
Should we care nil case?

lib/fluent/event.rb
tagomoris: `size` looks bad, because it sometimes means bytes of a data, and a type of EventStream (MessagePackEventStream) is actually a String object.
How about `events`?

lib/fluent/event.rb
tagomoris: Ah, it's my mistake. I'll fix it.

lib/fluent/event.rb
tagomoris: This mixin is used only for MessagePackEventStream, and it's already packed/encoded from ruby object to msgpack.
There's nothing to do about time encoding here.

lib/fluent/plugin/buffer.rb
tagomoris: Just for safe. And this method is not for busy loop.

lib/fluent/event.rb
repeatedly: `events` is better for me.

@sonots How about this? This method will be used in flowcounter_simple or similar gems.

lib/fluent/event.rb
repeatedly: Okay.

lib/fluent/event.rb
tagomoris: IMO we can remove it right now (because it's a v0.14 new feature), and it's better to be done.
But I'm not sure and this pull-request is not about that purpose.

lib/fluent/event.rb
```diff
@@ -162,5 +182,21 @@ def to_msgpack_stream
       @data
     end
   end
-end
 
+  module ChunkMessagePackEventStreamer
+    include MessagePackFactory::Mixin
+    # chunk.extend(ChunkEventStreamer)
+    #  => chunk.each{|time, record| ... }
+    def each(&block)
```
repeatedly: Original implementaion resuces IOError from MessagePack::Unpacker.
Is removing `rescue` safe?

```rb
def msgpack_each(&block)
  open do |io|
    u = msgpack_factory.unpacker(io)
    begin
      u.each(&block)
    rescue EOFError
    end
  end
end
```

tagomoris: MessagePack::Unpacker raises EOFError if data underflow occurs when unpacking data. It's caused by broken/unflushed data on disk.
For broken data, this code (without rescue) might raise failure (and it will not be solved forever), but it should be because it's content data of users.
For unflushed data, this code will raise error at once, but retry will succeed if it's after successful flush.

So, this code shouldn't do `rescue` and ignore.

lib/fluent/plugin/output.rb
```diff
@@ -534,6 +552,35 @@ def handle_stream(tag, es)
         meta_and_data.keys
       end
 
+      def handle_stream_with_standard_format(tag, es)
+        meta_and_data = {}
+        records = 0
```
repeatedly: Can we use `records = es.records(will be events)` instead?

tagomoris: No.
If input event stream is an instance of `MessagePackEventStream` and it was instantiated without # of events, `es.events` will return 0. It may occur in case when input event stream is from `in_forward`, and incoming data(opt) doesn't have additional information about # of events.

repeatedly: I see.

For optimization phase, we should separate the code into using `es.events` and each count mode.
Or we should calculate # of events in the initialization phase of MessagePackEventStream.
This is the out scope of this patch.

lib/fluent/event.rb
sonots: It looks this method returns an integer, not an array. So, both `records` and `events` look bad for me. 
What about `num_events` or just `length`?

tagomoris: OK, I'll use `length`.

lib/fluent/plugin/output.rb
tagomoris: Or, it might be better to use `es.length` if it's larger than 0. Else, do incrementation.

lib/fluent/event.rb
tagomoris: Once I would use `length`, but I re-found that `Chunk#records` already exists, and I made `EventStream#records` as same name with it.
`Chunk` also has `Chunk#size` (bytesize), and `#size != #length` is extraordinary confusing in ruby way. Hmm.

tagomoris: I think it's best to:
* change `Chunk#size` to return # of events, and make `#length` as alias of `#size`
* add `Chunk#bytesize` to return bytes of chunk content

But this changes the meaning of `chunk.size`, and it will break compatibility of plugins.

tagomoris: I'll create another pull-request to change `Chunk#records` and `EventStream#records` to `#num_events` later.
Let me merge this right now.

PR #917, opened 2016-04-26
Rename number of events

tagomoris:
in EventStream and Chunk.

But I leave a parameter name of `chunk_records_limit` because I think it's understandable for users, and `queued_records` and some other names of implementation / internal APIs.

In v0.12, `Chunk#size` means bytes of chunk content. So I'll add a mixin to extend chunks to return bytes of chunk content by `#size`, and apply it to chunks on compatibility layers.

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -21,9 +21,10 @@ module Fluent
     include Enumerable
     include MessagePackFactory::Mixin
 
-    def records
+    def size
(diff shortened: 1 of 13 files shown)


tagomoris:
@repeatedly @sonots check please.

sonots:
> I'll add a mixin to extend chunks to return bytes of chunk content by #size, and apply it to chunks on compatibility layers.

Okay, looks good to me.

PR #918, opened 2016-04-26
Use keyword arguments for `config_param` and `config_section`

okkez:
I've changed internal API related to `config_param` and `config_section`.


tagomoris:
LGTM

okkez:
Thanks!

PR #919, opened 2016-04-26
in_tail plugin's mysterious behavior if use wildcard pattern in windows

kazeula:
`in_tail` plugin allows to write wildcard pattern in `path` option.
If we use backslash path separation and wildcard pattern simultaneously in windows, the plugin doesn't work.
If we use backslash path separation and don't use wildcard pattern in windows, the plugin works.

This PR enables to be used backslash and wildcard simultaneously in windows.

Although I did some tests in my hand manually, I was not able to write tests in code because of lack of my knowledge of ruby.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -146,7 +146,7 @@ module Fluent
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
-          paths += Dir.glob(path).select { |p|
+          paths += Dir.glob(File.expand_path(path)).select { |p|
             if File.readable?(p)
               true
             else


tagomoris:
Without any tests, we cannot understand how/why this code fixes your problem :(

repeatedly:
> test

One idea is changing separate configuration into *nix and windows versions.

```ruby
  if Fluent.windows?
    EX_CONFIG = %[
      path test/plugin/*/%Y/%m/%Y%m%d-%H%M%S.log,test\plugin\data\log\**\*.log # use \ for * path
    ]
  else
    EX_CONFIG = # keep old configuration
  end
```

After applied above change, we check patch is good on AppVeyor test.

repeatedly:
BTW, expand_path expands relative path to absolute path so other tests are failed.
Maybe, calling expand_path to exclude_paths is also needed.

repeatedly:
I confirmed `expand_path` changes backslash to slash.

nurse:
This doesn't work if you have a file which includes glob meta characters.
```
irb(main):002:0> Dir.glob('a{}.txt')
=> []
irb(main):003:0> Dir.glob('a\{\}.txt')
=> ["a{}.txt"]
irb(main):004:0> File.expand_path('a\{\}.txt')
=> "C:/MinGW/msys/1.0/home/naruse/Software/source/td/a/{/}.txt"
irb(main):005:0> Dir.glob(File.expand_path('a\{\}.txt'))
=> []
```

repeatedly:
Verbosed other way is adding `no_meta_characters_in_path` option to guarantee `tr!('\', '/')` is safe by user.

nurse:
> Verbosed other way is adding no_meta_characters_in_path option to guarantee tr!('\', '/') is safe by user.

If adding fixed_string_path or something to specify paths which doesn't include wildcard, users can reduce glob call and doesn't need escaping glob meta characters also on Unix.

kazeula:
Thank you for your advice.
I have tried to find proper implementation of this PR, but it seems that I cannot find it.
Please feel free to do whatever you want to this PR.

repeatedly:
From nurse comment, we can't accept current patch so we should find another way.
One is adding option. Does someone have a better idea?

tagomoris:
I created an issue for this problem #1138, so let me close this.

PR #926, opened 2016-04-28
Fix logger on Fluent::StringUtil to use $log.

okkez:
backport #907


repeatedly:
Test failed due to scrub not found...

okkez:
Hmm... fluentd uses string-scrub gem.
I will try to fix this issue later.

okkez:
#933 fixes string-scrub problem.
I've rebased and force pushed.

okkez:
This PR may need #937

PR #933, opened 2016-05-04
Fix string-scrub dependency on ruby > 2

errm:
string-scrub 0.1.0 dropped today, and the strict
ruby version constraint stops fluent from being
installed on a modern ruby.


repeatedly:
I see. Thanks!

rohit01:
Awesome.. Thanks 👍

repeatedly:
I'm now working on releasing v0.12.23 with this change.

PR #934, opened 2016-05-04
Add support for drop_oldest_chunk when buffer queue is full

varunkumta:
- Drop oldest chunk from queue without writing to output when buffer
queue is full
- Add test

### DIFF
--- a/lib/fluent/buffer.rb
+++ b/lib/fluent/buffer.rb
@@ -150,7 +150,7 @@ module Fluent
     desc 'The length limit of the chunk queue.'
     config_param :buffer_queue_limit, :integer, default: 256
     desc 'The action when the size of buffer queue exceeds the buffer_queue_limit.'
-    config_param :buffer_queue_full_action, :enum, list: [:exception, :block], default: :exception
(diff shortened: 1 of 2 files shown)


repeatedly:
I will check it later.
Supporting `drop_oldest_chunk` itself is no problem for me.

repeatedly:
@varunkumta Could you write your use case?

agup006:
Hi @repeatedly, Use case is for when output plugin relies on time sensitive data and cannot be reached. E.g. connection on FluentD node is blocked or if service is down - continue to rotate data and switch newest data in for oldest data.

varunkumta:
@repeatedly , @agup006 and I work on the Linux omsagent (https://github.com/Microsoft/Build-OMS-Agent-for-Linux) - and the use case he mentioned is one we are adding for handling network outages better, since the data we collect like perf counters is more useful the more recent it is. So we drop older data in favor of newer data.

repeatedly:
@varunkumta @agup006 I understood the situation and I don't have a counter opinion.

@sonots @tagomoris If there is no concern, I will merge this patch.

tagomoris:
LGTM.

PR #937, opened 2016-05-09
Fix string scrub dependency

okkez:
We don't need to install string-scrub gem Ruby 2.1 or later.
Because `String#scrub` has been added since Ruby 2.1.

We need explicit `require "string-scrub"` to use `String#scrub` with Ruby 2.0 and Ruby1.9.3.


repeatedly:
Thx!

PR #938, opened 2016-05-09
Shutdown OutputThread in parallel

komamitsu:


lib/fluent/output.rb
```diff
@@ -119,15 +119,23 @@ def start
       @thread = Thread.new(&method(:run))
     end
 
-    def shutdown
+    def signal
```
repeatedly: `signal` seems general name.
How about `signal_shutdown`, `start_shutdown` or similar?

repeatedly:
No request from other users. Closed.

PR #941, opened 2016-05-09
backport: Fix test for TimeSlicedOutput

okkez:
This backports #916

### DIFF
--- a/lib/fluent/test/output_test.rb
+++ b/lib/fluent/test/output_test.rb
@@ -145,17 +145,19 @@ module Fluent
             assert_equal(@expected_buffer, buffer)
           end
 
-          chunks = @instance.instance_eval {
-            @buffer.instance_eval {
-              chunks = []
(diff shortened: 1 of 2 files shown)


repeatedly:
If backport this change to v0.12, do existing TimeSlicedOutput plugin tests become broken?

okkez:
No. We don't need to fix out_exec plugin (existing TimeSlicedOutput plugin) in this PR.
We have fixed out_file plugin, because out_file plugin's test uses BufferedOutputTestDriver.

tagomoris:
If 3rd party time sliced output plugins uses BufferedOutputTestDriver, then it should be broken.

tagomoris:
@repeatedly Don't you think to merge this change?

repeatedly:
@okkez Could you rebase the patch for current v0.12?

okkez:
Rebased.

repeatedly:
Thx!

PR #942, opened 2016-05-09
backport: Fix missing error class and test for out_forward

okkez:
This backports #922

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -451,11 +451,7 @@ module Fluent
     end
 
     def format_stream(tag, es)
-      if @time_as_integer
-        es.to_msgpack_stream_forced_integer
-      else
-        es.to_msgpack_stream
-      end
+      es.to_msgpack_stream
     end
 
     def emit(tag, es, chain)
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -450,9 +450,17 @@ module Fluent
       super
     end
 
+    def format_stream(tag, es)
+      if @time_as_integer
+        es.to_msgpack_stream_forced_integer
+      else
+        es.to_msgpack_stream
+      end
+    end
+
     def emit(tag, es, chain)
       @emit_count += 1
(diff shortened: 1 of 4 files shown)


lib/fluent/output.rb
```diff
@@ -450,9 +450,17 @@ def initialize
       super
     end
 
+    def format_stream(tag, es)
+      if @time_as_integer
```
repeatedly: `time_as_integer` doesn't have this option so this commit should not be backported to v0.12

okkez: I've removed `@time_as_integer`.

repeatedly:
@tagomoris Can we merge this?

lib/fluent/output.rb
```diff
@@ -450,9 +450,13 @@ def initialize
...
     end
 
+    def format_stream(tag, es)
+      es.to_msgpack_stream
+    end
+
     def emit(tag, es, chain)
       @emit_count += 1
-      data = es.to_msgpack_stream
+      data = format_stream(tag, es)
```
repeatedly: Extracting routine to `format_stream` is for `@time_as_integer` so this diff itself is not needed for this patch?
We want to reduce the change without no reason.

tagomoris:
Honestly speaking, I'm not sure this patch is needed for v0.12 branch. This change is important to add further fixes on these code/plugins, but `v0.12` branch is not development branch.

Code and merging this itself is ok for me.

repeatedly:
`ForwardOutputConnectionClosedError` definition is needed for v0.12, right?

tagomoris:
@repeatedly Right. But it's just a bug, and its fix should be separated from test-driver fixes ideally, especially in stable branch.

repeatedly:
@tagomoris okay.

@okkez Could you create a new PR?

okkez:
@repeatedly OK. Do we need 2 new PRs?

* Just add `ForwardOutputConnectionClosedError`
* Except above change in this PR

PR #943, opened 2016-05-09
Support `overflow_action` in v14 api

tagomoris:
I found that implementing `buffer_queue_full_action` in v0.14 requires change of internal API of Buffer.
This pull-request does:
* merge `Buffer#emit` and `Buffer#emit_bulk` to `Buffer#write`
* fix `Buffer#write` to get a set of metadata and its data (as a hash) to control commit/rollback as a transaction

This change is based on #912


tagomoris:
This fix changes more code than expected: details are in commit comment here:
https://github.com/fluent/fluentd/pull/943/commits/c3489d798415d736349134a0b9cf3527a8cbe9fd

tagomoris:
I confirmed that CI is green now.

PR #944, opened 2016-05-11
Fix fluent-cat to assign port directly

naritta:
https://github.com/fluent/fluentd/pull/854 miss fluent-cat fix.


tagomoris:
Nice catch!

naritta:
Thanks!

PR #946, opened 2016-05-12
Suppress warnings

okkez:
In previous version about 2500 lines of warnings.
In this version about 1200 lines of warnings.

Remained warinigs:

* From ServerEngine about 400 lines
  * See https://github.com/fluent/serverengine/pull/47
* `Encoding.default_external`, `Encoding.default_internal` about 800 lines
* Some other warnings about 50 lines

### DIFF
--- a/lib/fluent/plugin_id.rb
+++ b/lib/fluent/plugin_id.rb
@@ -47,11 +47,17 @@ module Fluent
     end
 
     def plugin_id_configured?
-      @_id_configured
+      if instance_variable_defined?("@_id_configured")
+        @_id_configured
+      end
     end
 
     def plugin_id
-      @id ? @id : "object:#{object_id.to_s(16)}"
+      if instance_variable_defined?("@id")
+        @id || "object:#{object_id.to_s(16)}"
+      else
+        "object:#{object_id.to_s(16)}"
+      end
     end
   end
 end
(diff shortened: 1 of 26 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -561,8 +561,8 @@ def handle_stream_with_standard_format(tag, es)
           meta_and_data[meta].add(time, record)
           records += 1
         end
-        meta_and_data.each_pair do |meta, es|
```
tagomoris: Please remove these change: I'm fixing these on other branches.

lib/fluent/plugin_id.rb
```diff
@@ -47,11 +47,15 @@ def plugin_id_for_test?
...
     def plugin_id_configured?
-      @_id_configured
+      if instance_variable_defined?("@_id_configured")
+        @_id_configured
+      end
     end
 
     def plugin_id
-      @id ? @id : "object:#{object_id.to_s(16)}"
+      if instance_variable_defined?("@id")
```
tagomoris: This method should return `"object:#{object_id.to_s(16)}"` if instance_variable `@id` is not defined.

okkez:
I've pushed commits and rebased for all comments.

tagomoris:
Great!

okkez:
Thanks!

PR #948, opened 2016-05-13
Add optional arguments to #elements to select specified child elements

tagomoris:
There are too many `conf.elements.select{|e| e.name == ... }`.


tagomoris:
@repeatedly @sonots I'll merge this soon.

Issue #360, opened 2014-06-20
Add support placeholder for include directive


y-ken:
Hi @repeatedly 

How about to support hostname or short hostname placeholder in include directive?
This  feature would bring us more flexible configuration management!

## specification

* '/etc/td-agent' directory files are same in all servers.
* to select load file on startup fluentd, it uses hostname

## use case

It is a way to switch behavior even though these are same configuration.

* [use case 1] fetch configuration from provisioning http server on startup.
* [use case 2] keep all server configuration as static file at '/etc/td-agent' directory from git repository.

#### sample configuration of use case 1

```
$ cat /etc/td-agent/td-agent.conf

include http://my-provisioning-server:8080/fluent/#{hostname}.conf
```

OR

```
$ cat /etc/td-agent/td-agent.rb

include {
  http://my-provisioning-server:8080/fluent/#{Socket.gethostname}.conf
}
```

#### sample configuration of use case 2

```
$ ls /etc/td-agent/
td-agent.conf
vms3021.conf
vms3022.conf
vms3023.conf
vms3024.conf
...snip...
```

```
$ cat /etc/td-agent/td-agent.conf

include /etc/td-agent/#{short_hostname}.conf
```

OR

```
$ cat /etc/td-agent/td-agent.rb

include {
  /etc/td-agent/#{Socket.gethostname.split('.')[0]}.conf
}
```

Thanks.

tagomoris:
Configuration DSL now supports `include` at master.

PR #949, opened 2016-05-14
[#360] Add include support for Ruby DSL

llby:
Oh 1 Failure at the windows.
But mac passed tests.
I should try debug...

### DIFF
--- a/lib/fluent/config/dsl.rb
+++ b/lib/fluent/config/dsl.rb
@@ -35,12 +35,17 @@ module Fluent
       end
 
       class Proxy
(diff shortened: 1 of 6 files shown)


llby:
In the Windows. `URI.parse` do this

```
uri = "C:/projects/fluentd/test/config/tmp/config_dsl/config_test_1.conf"
u = URI.parse(uri)
u.schema #=> "c"
u.path #=> "/projects/fluentd/test/config/tmp/config_dsl/config_test_1.conf"
```

PR #950, opened 2016-05-15
[#360] Add include support for Ruby DSL

llby:
This PR add include support for Ruby DSL.
And Fix bug parser include at the Windows.
Like this conditions
- First conf include absolute path
- Second(child) conf include relative path

Because `URI.parse` return `u.scheme = c`.
When Windows absolute path is `C:/projects/fluentd/test/tmp/config`.
Then used `open-uri` and basepath change `/`.

But after merge this PR.
There is a bug.
When It has some drives In the Windows.
Like a C: and D:
And upper case.
I think it is rare case.
If you need fix this bug.
Please make issue.

### DIFF
--- a/lib/fluent/config/dsl.rb
+++ b/lib/fluent/config/dsl.rb
@@ -106,7 +106,7 @@ module Fluent
           if args.first =~ /\.rb$/
             path = File.expand_path(args.first)
             data = File.read(path)
-            self.instance_eval(data, @proxy.include_basepath)
+            self.instance_eval(data, path)
           else
             ss = StringScanner.new('')
             Config::V1Parser.new(ss, @proxy.include_basepath, '', nil).eval_include(@attrs, @elements, args.first)
--- a/lib/fluent/config/dsl.rb
+++ b/lib/fluent/config/dsl.rb
@@ -35,12 +35,17 @@ module Fluent
       end
 
       class Proxy
-        def initialize(name, arg)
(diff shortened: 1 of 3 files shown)


lib/fluent/config/dsl.rb
```diff
@@ -96,6 +101,18 @@ def method_missing(name, *args, &block)
           self
         end
 
+        def include(*args)
+          ::Kernel.raise ::ArgumentError, "#{name} block requires arguments for include path" if args.nil? || args.size != 1
+          if args.first =~ /\.rb$/
+            path = File.expand_path(args.first)
+            data = File.read(path)
+            self.instance_eval(data, @proxy.include_basepath)
```
tagomoris: The 2nd argument of `instance_eval` should be the filename (`path`), not base path.

tagomoris:
Almost looks good to me.

nagata-logly:
@tagomoris 
Thank you your review
I got the `instance_eval` 2nd args

nagata-logly:
@tagomoris dsl code is cool!!

llby:
@tagomoris 
Thank you your review
I got the `instance_eval` 2nd args

and ... dsl code is cool!!

why test is failed??

tagomoris:
Merged. Thank you for contribution!

tagomoris:
This patch disables (overwrites) `Kernel.include`, but I think there are no users who wants to use mixin in configuration.

Issue #945, opened 2016-05-11
Merge in_tail_path into in_tail?


repeatedly:
I saw some users want to add tailed path into event record for identifying original file.
Time to merge in_tail_path?

sonots:
:+1:

repeatedly:
@sonots Could you write a patch?

sonots:
okay, i will do it later

PR #951, opened 2016-05-16
Add tail_path option to in_tail plugin

sonots:
Fix https://github.com/fluent/fluentd/issues/945

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -71,6 +71,8 @@ module Fluent
         raise ConfigError, e.message
       end
     end
+    desc 'Add the log path being tailed to records. Specify the field name to be used.'
+    config_param :path_key, :string, default: nil
 
     attr_reader :paths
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -252,6 +254,7 @@ def flush_buffer(tw)
                   else
                     @tag
                   end
+            record[@path_key] ||= tw.path unless @path_key.nil?
```
repeatedly: Using `||=` is for safety?

sonots: yes

sonots: I mean that `path_key` is not overwritten if record already has it.
This was the specification of the original implementation (https://github.com/fluent/fluentd/pull/281), but we can change the spec now. Do you prefer to overwrite?

repeatedly: Either is fine for me.
User can check path_key value is not contained or not easily so
additional check is not needed for me.

sonots: Okay, let us keep as is.

repeatedly: I follow your opinion

repeatedly:
Thx!

tagomoris:
Tests added by this change in `test_in_tail` fails very frequently.
For example: https://travis-ci.org/fluent/fluentd/jobs/130508869
@sonots Could you fix it?

sonots:
@tagomoris sorry, fixed the unstable test.

PR #955, opened 2016-05-16
Suppress encoding related warnings

okkez:
This change suppresses warning messages about 800 lines on CI.

```
lib/fluent/engine.rb:61: warning: setting Encoding.default_internal
lib/fluent/engine.rb:62: warning: setting Encoding.default_external
(about 400 times repeat)
lib/fluent/engine.rb:61: warning: setting Encoding.default_internal
lib/fluent/engine.rb:62: warning: setting Encoding.default_external
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -16,6 +16,7 @@ Rake::TestTask.new(:base_test) do |t|
   t.test_files = Dir["test/**/test_*.rb"].sort
   t.verbose = false
   t.warning = false
+  t.ruby_opts = ["-Eascii-8bit:ascii-8bit"]
 end
 
 task :parallel_test do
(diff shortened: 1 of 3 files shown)


okkez:
We can treat data as binary when read it or open the file.
The data from Socket is always binary.

We are always careful with data from outside of Fluentd after merge this PR.

tagomoris:
I think just removing assignments of `default_internal/external` is breaking change for some cases, and it should NOT be done for compatibility.
IMO we have two options:
* assigning encodings only when `Encoding.default_* != Encoding::ASCII_8BIT`
* removing assignment for default encodings, and add specifying `-E` to command line arguments of spawn in `supervisor.rb`, and enable same thing on `rake test`

Second options is better for me. @repeatedly any thoughts?

okkez:
In my local, I've tested second option. It works well on Linux and Windows.

tagomoris:
LGTM!

tagomoris:
Merged. Thank you!

PR #957, opened 2016-05-16
fix unstable test_in_tail

sonots:
Fix https://github.com/fluent/fluentd/pull/951#issuecomment-219389325

### DIFF
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -919,11 +919,9 @@ class TailInputTest < Test::Unit::TestCase
 
       emits = d.emits
       assert(emits.length == 4)
-      assert_equal("#{TMP_DIR}/tail1.txt", emits[0][2]["path"])
-      assert_equal("#{TMP_DIR}/tail2.txt", emits[1][2]["path"])
+      assert_equal(files, [emits[0][2]["path"], emits[1][2]["path"]].sort)
       # "test4" events are here because these events are flushed at shutdown phase
-      assert_equal("#{TMP_DIR}/tail1.txt", emits[2][2]["path"])
-      assert_equal("#{TMP_DIR}/tail2.txt", emits[3][2]["path"])
+      assert_equal(files, [emits[2][2]["path"], emits[3][2]["path"]].sort)
     end
   end
 end


tagomoris:
Is this test the problem correctly?
I think that it should do these:
* `tail1.txt` appears twice
* `tail2.txt` appears twice
* (optional?) lines from same file should 0-1 & 2-3.

Tests should confirm the sources of events.

sonots:
Not directly writing, but this does test both tail1.txt and tail2.txt appears twice.

tagomoris:
I meant that it should check that the sources of these records are `tail1.txt` and `tail2.txt` explicitly.

sonots:
like this?

```diff
diff --git a/test/plugin/test_in_tail.rb b/test/plugin/test_in_tail.rb
index a0c1b6a..80818d1 100644
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -908,10 +908,10 @@ class TailInputTest < Test::Unit::TestCase
         files.each do |file|
           File.open(file, 'ab') { |f|
             f.puts "f #{file} line should be ignored"
-            f.puts "s test1"
+            f.puts "s #{file}"
             f.puts "f test2"
             f.puts "f test3"
-            f.puts "s test4"
+            f.puts "s #{file}"
           }
         end
         sleep 1
@@ -920,8 +920,12 @@ class TailInputTest < Test::Unit::TestCase
       emits = d.emits
       assert(emits.length == 4)
       assert_equal(files, [emits[0][2]["path"], emits[1][2]["path"]].sort)
-      # "test4" events are here because these events are flushed at shutdown phase
+      assert_equal(emits[0][2]["path"], emits[0][2]["message"].lines.first.chomp)
+      assert_equal(emits[1][2]["path"], emits[1][2]["message"].lines.first.chomp)
+      # last line events are here because these events are flushed at shutdown phase
       assert_equal(files, [emits[2][2]["path"], emits[3][2]["path"]].sort)
+      assert_equal(emits[2][2]["path"], emits[2][2]["message"].lines.first.chomp)
+      assert_equal(emits[3][2]["path"], emits[3][2]["message"].lines.first.chomp)
     end
   end
 end
```

tagomoris:
No. It's incorrect because these paths might be different one from unknown files which are used in other tests and not cleaned up.
This tests should be like this:
```ruby
paths = emits.map{|e| e[2]["path"] }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail1.txt" }.size == 2 }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail2.txt" }.size == 2 }
```

sonots:
Well, I want to test last lines (which are flushed on shutdown https://github.com/fluent/fluentd/blob/0ce19ec716c6cb831b9e1d825ce269e38b64f66f/lib/fluent/plugin/in_tail.rb#L257) and non-last lines (which are emitted when next first lines are appeared https://github.com/fluent/fluentd/blob/0ce19ec716c6cb831b9e1d825ce269e38b64f66f/lib/fluent/plugin/in_tail.rb#L301) separately since they are treated in different lines of codes although this is just an optional.

```
assert_equal(files, [emits[0][2]["path"], emits[1][2]["path"]].sort)
assert_equal(files, [emits[2][2]["path"], emits[3][2]["path"]].sort)
```

expresses almost same tests with your tests in addition to the option.

sonots:
I mean what want to do is like below, but this is almost same with the current test cases

```ruby
paths = emits[0..1].map{|e| e[2]["path"] }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail1.txt" }.size == 1 }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail2.txt" }.size == 1 }
paths = emits[2..3].map{|e| e[2]["path"] }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail1.txt" }.size == 1 }
assert{ paths.select{|path| path == "#{TMP_DIR}/tail2.txt" }.size == 1 }
```

tagomoris:
Ah, okay, got it > "records flushed at shutdown"
I think it's better to separate test case for normal case and records flushed at shutdown though.

PR #962, opened 2016-05-17
Support overflow action in v14 api redo

tagomoris:
This is same with #943, which was wrongly merged as squashed commit.

### DIFF
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -51,12 +51,15 @@ module Fluent
           @unique_id = generate_unique_id
           @metadata = metadata
 
+          # state: staged/queued/closed
+          @state = :staged
+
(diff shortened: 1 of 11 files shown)


lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -89,15 +92,28 @@ def empty?
...
+        def queued?
+          @state == :queued
+        end
+
+        def closed?
+          @state == :closed
+        end
+
+        def enqueued!
+          @state = :queued
```
repeatedly: I just notified `=` is lacked.

tagomoris: not lacked. it's assignment.

repeatedly: Sorry, I missed `!`...

PR #963, opened 2016-05-17
Add support for emit_config to in_monitor_agent

varunkumta:
We want to monitor the health of our output plugins. To do so, we want to compare the current buffer_queue_length with the configured buffer_queue_limit. With the current implementation, in_monitor_agent only emits the first. Adding support for the second.

I could not find any tests for this plugin.


repeatedly:
It seems no problem. Thanks!

PR #968, opened 2016-05-19
Fix tests of in_object_space which fails in CI environment

tagomoris:


tagomoris:
I ran this test 3 time on Travis-CI ruby 2.3 + Linux, and all flushed in success.
This seems to fix will solve the problem about it.

PR #972, opened 2016-05-20
Introduce v14 test drivers

tagomoris:
This change introduces plugin test drivers for v0.14 plugin APIs, and includes some port of plugins for Input, Output (non-buffered and buffered [tag, time]).
Some fixes which were found via implementation of plugins/test drivers are also included.

This change depends on #981, #969 .


tagomoris:
@repeatedly @sonots I'll merge this later too. This (and earlier) changes will be included in RC2.

tagomoris:
Tests with ruby 2.2.4 on osx is still yellow (waiting), but others are green, and that is because CI service is in troubles.

Issue #973, opened 2016-05-20
tests of fluent-plugin-config-expander fails with v0.14 (master) branch (uninitialized constant Fluent::Input/Fluent::Output)


sonots:
## HOW TO REPRODUCE

```
git clone https://github.com/tagomoris/fluent-plugin-config-expander
cd fluent-plugin-config-expander
# fix Gemfile
# + gem 'fluentd', git: 'https://github.com/fluent/fluentd'
bundle update
bundle exec rake
```

## ERROR1. uninitialized constant Fluent::Input (NameError)

```
lib/fluent/plugin/in_config_expander.rb:3:in `<top (required)>': uninitialized constant Fluent::Input (NameError)
```

**REASON**

It looks this is caused because `fluent/test` now does not require 'fluent/input`

Adding `require 'fluent/input'` into test/helper.rb of the plugin made tests succeed.

## ERROR2. Fluent::ConfigError: '<store>' sections are required

```
Error: test_emit(ConfigExpanderOutputTest): Fluent::ConfigError: '<store>' sections are required
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/config/section.rb:160:in `block in generate'
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/config/section.rb:146:in `each'
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/config/section.rb:146:in `generate'
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/configurable.rb:68:in `configure'
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/plugin/base.rb:39:in `configure'
/Users/seo.naotoshi/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/fluentd-0b6e5048da40/lib/fluent/plugin_id.rb:34:in `configure'
[...]

PR #974, opened 2016-05-21
add Fluent::Plugin::BareOutput to support v0.12 MultiOutput plugins

tagomoris:


tagomoris:
@sonots Please check this change

sonots:
For the problem of ERROR2 of https://github.com/fluent/fluentd/issues/973, I got another error for config-expander as

```
Error: test_emit(ConfigExpanderOutputTest): RuntimeError: BUG: use router.emit instead of Engine.emit
```

I also have one question: what does "bare" infer? How plugin developers choose a parent class, `Output` or `BareOutput` to create an output plugin?

tagomoris:
It's intentional to prohibit `Engine.emit`... it will not be fixed so far.
`BareOutput` is very limited and skinny one, and it's not for almost plugin authors. it's just for meta plugins like forest, config-expander or something like that.

sonots:
Okay, so compatibility layer for `Engine.emit` is not provided. 
I understand that we have to tell about it to plugin developers. 

Code itself looks good to me.

tagomoris:
Once I thought about `Engine.emit`, but that is just under `Fluent::`, and no space to override for compatibilities. And that change is of v0.12 originally (we should have shown "deprecated" message in v0.12 versions).

tagomoris:
I added some comment to tell not to use this for normal output plugins.

PR #978, opened 2016-05-23
Extract utilities to create client/server using ServerEngine

okkez:
This is useful to write 3rd party plugins using TCP/UDP socket.


okkez:
Remains similar code here: https://github.com/fluent/fluentd/blob/master/lib/fluent/supervisor.rb#L51..L53

tagomoris:
Similar feature is planned to be implemented as a plugin helper, which is integrated with multi process support.
I'm sorry but we cannot merge this change.

okkez:
Ok, I'm looking forward to new feature!

PR #981, opened 2016-05-23
Fix overwriting config_section definitions:

tagomoris:
* "name" and "param_name" is strongly connected with plugin's code (even in base classes)
   so these MUST be not changed by subclasses (subclass MUST use same names)
 * separate "param_name" from "name" to detect overwriting "param_name" by subclasses to raise errors
   and add "variable_name" method as internal utility
 * make it possible to change default values in finalized sections
   because it's very popular that 3rd party plugin prefer different default value for built-in predefined parameters
 * make it possible to change "init" specification of finalized sections
   because newly added/changed default values might be able to init sections

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -41,7 +41,7 @@ module Fluent
         @name = name.to_sym
         @final = final
 
-        @param_name = (param_name || @name).to_sym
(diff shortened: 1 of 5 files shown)


tagomoris:
@repeatedly @sonots I'll merge this later if there's no objection.

Issue #930, opened 2016-05-03
buffer_chunk_limit exceeded when using detach_process


cbroglie:
Using a file buffer output plugin with `detach_process` results in chunk sizes >> `buffer_chunk_limit` when sending events at high speed. My settings specify 1MB chunk sizes, but it is easy to generate chunk sizes >50MB by writing 500k record (~200 bytes per record) in 4 seconds. My testing was done using a c4.2xl instance running Amazon Linux 2016.03 with td-agent2. 

#### Config
```xml
<match kinesis.**>
  @type kinesis_producer
  log_level trace
  region us-west-2
  stream_name cbroglie-fluentd-test
  aws_key_id XXX
  aws_sec_key XXX
  buffer_chunk_limit 1m
  buffer_queue_limit 100
  flush_interval 1
  try_flush_interval 0.1
  queued_chunk_flush_interval 0.01
  num_threads 10
  detach_process 1
  buffer_type file
  buffer_path /var/log/td-agent/ztrack.*.buffer
  <secondary>
    @type file
    path /var/log/td-agent/ztrack.failed
  </secondary>
</match>
```
#### Benchmark driver
```
$ time ./fluent-bench -count 500000
2016/05/03 22:19:31 processed=500000 p50=14.89µs p99=31.768µs
2016/05/03 22:19:31 succeeded=500000
2016/05/03 22:19:31 failed=0

real	0m4.322s
user	0m11.632s
sys	0m3.628s
$
```

#### Resulting buffer files
```
$ ls -lh /var/log/td-agent/
total 90M
-rw-r--r-- 1 td-agent td-agent 4.3K May  3 22:19 td-agent.log
-rw-r--r-- 1 td-agent td-agent  54M May  3 22:19 ztrack..b531f781175491342.buffer
-rw-r--r-- 1 td-agent td-agent 5.9M May  3 22:19 ztrack..q531f78093ddcb2ad.buffer
-rw-r--r-- 1 td-agent td-agent  15M May  3 22:19 ztrack..q531f780a3027f5f1.buffer
-rw-r--r-- 1 td-agent td-agent  16M May  3 22:19 ztrack..q531f780d3171f187.buffer
$
```

#### Full background
[...]

PR #982, opened 2016-05-23
Make the detach process forward interval configurable

cbroglie:
Setting this to a value less than the magic number of 0.2 (like 0.0) fixes the issue in #930

### DIFF
--- a/lib/fluent/process.rb
+++ b/lib/fluent/process.rb
@@ -440,7 +440,7 @@ module Fluent
       end
 
       if forward_interval = conf['detach_process_forward_interval']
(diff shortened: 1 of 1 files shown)


lib/fluent/process.rb
```diff
@@ -438,11 +438,17 @@ def configure(conf)
           @detach_process = false
         end
       end
+
+      if forward_interval = conf['detach_process_forward_interval']
+        @forward_interval = forward_interval.to_f
```
repeatedly: Use `Config.time_value` instead of `to_f`.

repeatedly:
Thanks!

PR #983, opened 2016-05-24
Rename v14 parameters

tagomoris:
I got reviewed for parameters newly added or renamed, and applied changes to rename better names.
This change will be included in v0.14.0 RC2.

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -214,7 +214,13 @@ module Fluent
           }
           PARAMS_MAP.each do |older, newer|
             next unless newer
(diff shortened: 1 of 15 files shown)


repeatedly:
In this PR, adding several plugins like BufferedNullOutput are intended?

tagomoris:
This change is based on #972, and changes of plugins are done in it.

repeatedly:
Okay, I understood.

tagomoris:
Tests with ruby 2.2.4 on osx is cancelled by Travis trouble, but others are green.

Issue #985, opened 2016-05-24
LoadError: cannot load such file -- string-scrub with ruby < 2.1


sonots:
## HOW TO REPRODUCE

```
$ ruby -v
ruby 2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin15.2.0]
$ gem install fluentd
$ touch foo.conf
$ fluentd -c foo.conf
2016-05-24 21:45:55 +0900 [info]: reading config file path="foo.conf"
2016-05-24 21:45:55 +0900 [info]: starting fluentd-0.12.24
/Users/seo.naotoshi/.rbenv/versions/2.0.0-p648/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- string-scrub (LoadError)
        from /Users/seo.naotoshi/.rbenv/versions/2.0.0-p648/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
```

## CAUSE

With this PR https://github.com/fluent/fluentd/pull/937, string-scrub was removed from runtime-dependency. See https://rubygems.org/gems/fluentd/versions/0.12.24.

This change

```
  unless "".respond_to?(:scrub)
    gem.add_runtime_dependency("string-scrub", [">= 0.1.1"])
  end
```

means that do `gem.add_runtime_dependency` if ruby < 2.1 (which does not have scrub methods) on **release** timing, not on `gem install`. Because fluentd v0.12.24 was released with ruby >= 2.1 (I guess so, only repeatedly knows), the dependency was removed.

sonots:
## HOW TO FIX

Reverting gemspec to below should fix the problem:

```
gem.add_runtime_dependency("string-scrub", [">= 0.1.1"])
```

if we want to skip building string-scrub on ruby >= 2.1, fixing string-scrub should be needed in addition.

sonots:
cc: @okkez

okkez:
:bow: Sorry for my fault about gem dependency.
#986 fixes all problems about gem dependency?

sonots:
Yes, I think so

PR #986, opened 2016-05-24
Revert: Fix string-scrub dependency #937 (only gemspec)

sonots:
Fix https://github.com/fluent/fluentd/issues/985

See also discussion with hsbt (Japanese)
https://twitter.com/sonots/status/735097429251366917


sonots:
@repeatedly I think we should release v0.12.25 just for this as v0.12.x supports ruby 2.0.

Issue #990, opened 2016-05-25
fluent-plugin-sql fails with fluentd v0.14.0.rc.3 (error="undefined method `key')


kakoni:
Testing RC3 with fluent-plugin-sql.
out-sql fails because
```
2016-05-25 11:10:32 +0300 [warn]: failed to flush the buffer. plugin_id="object:3fc15cdfc728" retry_time=0 next_retry=2016-05-25 11:10:33 +0300 chunk="533a63190cdc0d19f6dd96b
a60e49317" error_class=NoMethodError error="undefined method `key' for #<Fluent::Plugin::Buffer::MemoryChunk:0x007f82be13ab30>"
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/dev/fluentd/lib/fluent/plugin/out_sql.rb:224:in `block (2 levels) in write'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/dev/fluentd/lib/fluent/plugin/out_sql.rb:223:in `each'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/dev/fluentd/lib/fluent/plugin/out_sql.rb:223:in `block in write'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/.gem/ruby/2.3.0/gems/activerecord-4.2.6/lib/active_record/connection_adapters/abstract/connection_pool.rb:292:in `with_conne
ction'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/dev/fluentd/lib/fluent/plugin/out_sql.rb:221:in `write'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/.gem/ruby/2.3.0/bundler/gems/fluentd-69df4fe63215/lib/fluent/compat/output.rb:109:in `write'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/.gem/ruby/2.3.0/bundler/gems/fluentd-69df4fe63215/lib/fluent/plugin/output.rb:778:in `try_flush'
  2016-05-25 11:10:33 +0300 [warn]: /Users/kakoni/.gem/ruby/2.3.0/bundler/gems/fluentd-69df4fe63215/lib/fluent/plugin/output.rb:953:in `flush_thread_run'
[...]

PR #992, opened 2016-05-26
Provide `chunk.key` for v0.12 plugins which specifies `key` in overriding `#emit`

tagomoris:
3rd party BufferedOutput plugins sometimes overrides `#emit` to control chunking keys.
This change is to provide compatibility for such plugins by using metadata.variables to pass specified chunking key to `#write`.

This fix solves #990.

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -299,10 +299,14 @@ module Fluent
         if @overrides_emit
           current_emit_count = @emit_count
           size = es.size
-          emit(tag, es, NULL_OUTPUT_CHAIN)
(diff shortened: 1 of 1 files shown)


tagomoris:
I checked this change with fluent-plugin-sql and configuration below:
```
<source>
  type dummy
  tag test
</source>

<match test>
  @type sql
  host localhost
  port 5432
  database fluentd_test
  adapter postgresql
  username fluentd
  password fluentd
  <table>
    table mytest
    column_mapping message:message
  </table>
  <table test>
    table mytest
    column_mapping message:message
  </table>
</match>
```

Issue #1004, opened 2016-05-30
fluent-plugin-geoip output version does not work with fluentd v0.14.0.rc.3


okkez:
### How to reproduce

```
$ git clone  git@github.com:y-ken/fluent-plugin-geoip.git
$ cd fluent-plugin-geoip
$ bundle install
$ bundle exec fluentd -c fluent.conf
```

It does not work with Fluentd master, too. (Works well with Fluentd v0.12 using same configuration)
Output version config:

```
  <source>
    @type dummy
    tag "raw.dummy"
    dummy {"host":"66.102.9.80","message":"test"}
  </source>
  <match raw.**>
    @type geoip
    add_tag_prefix debug.
    geoip_lookup_key "host"
    flush_interval 1s
    <record>
      city ${city["host"]}
      latitude ${latitude["host"]}
      longitude ${longitude["host"]}
      country_code3 ${country_code3["host"]}
      country ${country_code["host"]}
      country_name ${country_name["host"]}
      dma ${dma_code["host"]}
      area ${area_code["host"]}
      region ${region["host"]}
    </record>
    <buffer>
      flush_mode interval
      retry_type exponential_backoff
      flush_interval 1s
    </buffer>
  </match>
  <match debug.**>
    @type stdout
  </match>
```

On the other hand, filter version works well.
Filter version config:

```
<source>
  type dummy
  tag raw.dummy
  dummy {"host":"66.102.9.80", "message":"test\ntest"}
</source>

<filter raw.**>
  type geoip
  #remove_tag_suffix .dummy
  add_tag_prefix debug.
  #tag raw.geoip
  geoip_lookup_key  host
  #geoip_database    "/path/to/your/GeoIPCity.dat"
  <record>
    city            ${city["host"]}
    latitude        ${latitude["host"]}
    longitude       ${longitude["host"]}
    country_code3   ${country_code3["host"]}
    country         ${country_code["host"]}
    country_name    ${country_name["host"]}
    dma             ${dma_code["host"]}
    area            ${area_code["host"]}
    region          ${region["host"]}
  </record>
  flush_interval    1s
</filter>

<match raw.**>
  type stdout
</match>
```

okkez:
Fixed by #1005

PR #1005, opened 2016-05-30
Fix to call #format_stream of plugin themselves when RecordFilter mixin included

tagomoris:
* because RecordFilter depends on #format_stream implemented by that mixin itself
*   it is same with the case that plugins re-implements #format_stream


tagomoris:
This change fixes #1004.

okkez:
It works! LGTM.

tagomoris:
Merged.

PR #1015, opened 2016-06-01
Fix to overwrite configure_proxy name only for root sections

### LINKED ISSUE #1013: `merged_configure_proxy` overwrites name of proxy
The name of proxy object which is returned by `Fluent::Configurable#merged_configure_proxy` is `Fluent::Plugin::Base`. I expected real plugin name such as `Fluent::Plugin::DummyInput`.
This is not useful for debug.

https://github.com/fluent/fluentd/blob/master/lib/fluent/configurable.rb#L159..L167

```ruby
configurables = ancestors.reverse.select{ |a| a.respond_to?(:configure_proxy) } # => [Fluent::PluginBase, ..., Fluent::Plugin::DummyInput]
configurables.map{ |a| a.configure_proxy(a.name || a.object_id.to_s) }.reduce(:merge)
```

In Fluentd v0.14.0:

```ruby
merged_configure_proxy.name # => "Fluent::Plugin::Base"
```

I expected:

```ruby
# For example, when we use in_dummy plugin
merged_configure_proxy.name # => "Fluent::Plugin::DummyInput"
```

tagomoris:
To make debug easier.
This change fixes #1013.

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -37,9 +37,15 @@ module Fluent
       #   end
       # end
 
-      def initialize(name, param_name: nil, final: nil, init: nil, required: nil, multi: nil, alias: nil, type_lookup:)
+      def initialize(name, root: false, param_name: nil, final: nil, init: nil, required: nil, multi: nil, alias: nil, type_lookup:)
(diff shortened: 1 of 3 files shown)


okkez:
Looks good to me. Thanks!

PR #1023, opened 2016-06-03
Add plugin helpers for parser and formatter

tagomoris:


tagomoris:
@repeatedly please review this change.

lib/fluent/plugin_helper/formatter.rb
```diff
@@ -0,0 +1,141 @@
...
+            block.call(formatter) if block_given?
+          rescue => e
+            log.error "unexpected error while #{method_name}", usage: usage, formatter: formatter, error: e
+          end
+        end
+      end
+
+      def stop
+        super
+        # timer stops automatically in super
```
repeatedly: What does this comment mean?

lib/fluent/plugin_helper/formatter.rb
```diff
@@ -0,0 +1,141 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'monitor'
+require 'forwardable'
```
repeatedly: Are these `require needed?

lib/fluent/plugin_helper/parser.rb
```diff
@@ -0,0 +1,141 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'monitor'
+require 'forwardable'
```
repeatedly: ditto

lib/fluent/plugin_helper/formatter.rb
```diff
@@ -0,0 +1,141 @@
...
+module Fluent
+  module PluginHelper
+    module Formatter
+      def formatter_create(usage: '', type: nil, conf: nil)
+        formatter = @_formatters[usage]
+        return formatter if formatter
+
+        if !type
+          raise ArgumentError, "BUG: both type and conf are not specified" unless conf
+          raise Fluent::ConfigError, "@type is not specified for <format>" unless conf['@type']
```
repeatedly: For me, `@type is required in <format>` is clear than this message but not critical.

lib/fluent/plugin_helper/formatter.rb
```diff
@@ -0,0 +1,141 @@
...
+          end
+          formatter = Plugin.new_formatter(section[:@type], parent: self)
+          formatter.configure(section.corresponding_config_element)
+          @_formatters[section.usage] = formatter
+        end
+      end
+
+      def start
+        super
+        @_formatters_started = true
```
repeatedly: This line should be moved to end of this fucntion?
Formatters don't start yet in this position.

lib/fluent/plugin_helper/formatter.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+
+      def close
+        formatter_operate(:close)
+        super
+      end
+
+      def terminate
+        formatter_operate(:terminate)
+        @_formatters = {}
```
repeatedly: Is `@_formatters_started = false` needed for complete clean up?

lib/fluent/plugin_helper/formatter.rb
tagomoris: Ah, I missed to remove these.

lib/fluent/plugin_helper/formatter.rb
tagomoris: It's understandable for me too.

lib/fluent/plugin_helper/formatter.rb
tagomoris: No. It's needed to set the variable to true here.
If any formatter calls `formatter_create`, that instance will not be started if we set it true at the end of this method.

lib/fluent/plugin_helper/formatter.rb
tagomoris: Oops, it should be removed.

lib/fluent/plugin_helper/formatter.rb
tagomoris: Right.

tagomoris:
I pushed a commit to fix for review comments.

PR #1028, opened 2016-06-05
Fix `DEFAULT_PLUGIN_PATH`

yui-knk:
Before this commit `DEFAULT_PLUGIN_PATH` pointed
`lib/fluent/registry.rb/plugin`.
I feel this is not intended and change to point
`lib/fluent/plugin`.


tagomoris:
LGTM.

PR #1031, opened 2016-06-05
Add `@type` to secondary section

yui-knk:
Without `@type` developers can not getting started
as web site says.

http://docs.fluentd.org/articles/install-by-gem

### DIFF
--- a/fluent.conf
+++ b/fluent.conf
@@ -90,6 +90,7 @@
     host 192.168.0.11
   </server>
   <secondary>
+    @type forward
     <server>
       host 192.168.0.12
     </server>


repeatedly:
Thx!

yui-knk:
😍

tagomoris:
If secondary plugin type is same with primary's one, `@type` in `<secondary>` section can be skipped.
It's default behavior of `v0.12` version.

tagomoris:
I reverted this patch at #1034.

yui-knk:
OK, I see. Thank you for your fix commit about secondary type :)

PR #1032, opened 2016-06-06
fix bug not to use primary plugin type in default

tagomoris:
If secondary type is not specified explicitly.
It's the same behavior with v0.12.


tagomoris:

repeatedly:
@tagomoris All tests failed. Maybe log result is changed.

tagomoris:
Pushed a fix for test :P

Issue #1027, opened 2016-06-04
error when using --no-supervisor option


m-mizutani:
fluentd fails when using --no-supervisor option at UNIXSocket.new in `serverengine/socket_manager_unix.rb`. This error may be caused by plugin/in_forward.rb because environment variable `SERVERENGINE_SOCKETMANAGER_PATH` is empty when no supervisor mode.

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_forward.rb#L62

Steps to Reproduce
---------------

Just start fluentd with --no-supervisor with below configuration.

fluentd or td-agent version.
---------------

- fluentd 0.14.0

Environment information, e.g. OS.
----------------

- OS: OS X 10.11.5
- Ruby: ruby 2.2.0

Configuration and Logs
-----------------

```
% fluentd --no-supervisor -c test/fluentd.conf
2016-06-04 20:13:03 +0900 [info]: reading config file path="test/fluentd.conf"
2016-06-04 20:13:03 +0900 [info]: starting fluentd-0.14.0 without supervision
2016-06-04 20:13:03 +0900 [info]: gem 'fluentd' version '0.14.0'
2016-06-04 20:13:03 +0900 [info]: adding match pattern="test.**" type="stdout"
2016-06-04 20:13:03 +0900 [info]: adding source type="forward"
2016-06-04 20:13:03 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type forward
    port 24224
  </source>
  <match test.**>
    @type stdout
  </match>
</ROOT>
2016-06-04 20:13:03 +0900 [info]: listening fluent socket on 0.0.0.0:24224

2016-06-04 20:13:03 +0900 [error]: unexpected error error_class=TypeError error="no implicit conversion of nil into String"
  2016-06-04 20:13:03 +0900 [error]: /Users/mizutani/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/serverengine-1.6.4/lib/serverengine/socket_manager_unix.rb:27:in `initialize'
  2016-06-04 20:13:03 +0900 [error]: /Users/mizutani/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/serverengine-1.6.4/lib/serverengine/socket_manager_unix.rb:27:in `new'
[...]

PR #1033, opened 2016-06-06
Add --run-worker to distinguish it from running without supervisor

naritta:
This fixes https://github.com/fluent/fluentd/issues/1027

・We have to distinguish running only worker without supervisor from running worker under supervisor, because when using only worker, we have to create socket manager in worker.

・If we fix not to use socket manager when --no-supervisor, we have to fix plugin side, then the point to fix will increase. So I fixed supervisor side.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -60,12 +60,13 @@ op.on('-d', '--daemon PIDFILE', "daemonize fluent process") {|s|
   opts[:daemonize] = s
 }
 
-op.on('--under-supervisor', "run fluent worker under supervisor (this option is NOT for users)") { opts[:supervise] = false
-  opts[:worker_with_supervisor] = true
+op.on('--under-supervisor', "run fluent worker under supervisor (this option is NOT for users)") {
+  opts[:supervise] = false
 }
 
 op.on('--no-supervisor', "run fluent worker without supervisor") {
   opts[:supervise] = false
+  opts[:standalone_worker] = true
 }
 
 op.on('--user USER', "change user") {|s|
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -60,12 +60,11 @@ op.on('-d', '--daemon PIDFILE', "daemonize fluent process") {|s|
   opts[:daemonize] = s
 }
 
-op.on('--run-worker', "run fluent worker under supervisor") {
-  opts[:supervise] = false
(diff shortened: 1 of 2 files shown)


tagomoris:
I want just to correct my idea:
* normally, Fluentd's worker process will work with supervisor including ServerEngine instance, and it (supervisor) provides SocketManager
* now Fluentd supervisor process executes workers with `--no-supervisor` option
* plugins always use SocketManager API, and it's required for multi-process support (in future)
* If users apply `--no-supervisor` option by themselves, Fluentd doesn't setup SocketManager in worker processes. It breaks plugin's behavior.

Your change will do:
* add another option `--run-worker` to use it from supervisor process to run workers without setup of SocketManager
* fix worker behavior with `--no-supervisor` to setup SocketManager to pass it to plugins

Is it right?

naritta:
Thank you for summarizing. It's right.

tagomoris:
Great.
I think these situation is a bit complex and difficult to understand, so we need to add these details as code comment in supervisor.rb. @naritta could you add them?

In addition to it, we should take care about naming of command line options and these descriptions.
There are 3 mode to launch Fluentd:
* launch Fluentd supervisor: without any options
* launch Fluentd worker only, without supervisor: `--no-supervisor`
* launch Fluentd workers under supervisor (launched by supervisor): `--run-worker`?, `--under-supervisor`?

IMO it's more understandable to name the new option as `--xxx-supervisor` by comparing existing `--no-supervisor`.
Moreover, it's better to describe explicitly that "this option is NOT for users" on that option to help users not to try using them.

tagomoris:
@naritta ping? May I add fixes on this change?

naritta:
@tagomoris Sorry, I'll fix now. Could you wait a little?

naritta:
Sorry to be late. Could you review when you have time?

tagomoris:
IMO `worker_with_supervisor` still looks curious. In default, worker runs under supervisor, but `worker_with_supervisor` 's default value is `false`.

I think that variable (state) should be a something to be turned into true when `--no-supervisor` is specified.
I propose that:
* rename it as `standalone_worker` with default value `false`
* turn it true only when `--no-supervisor` specified
* call `create_socket_manager` if `@standalone_worker` is true

@naritta How do you think about this?

naritta:
@tagomoris Thank you for reviewing. I think so, too.
I fixed it and could you review when you have time?

tagomoris:
Code looks good to me. I'll merge this after confirmed CI green.

tagomoris:
Merged.

naritta:
Thank you very much for reviewing.

PR #1047, opened 2016-06-10
Test utility methods and additional test drivers

tagomoris:
When we want to write tests of 3rd party plugins, we need some more code for:
* generating Fluent::EventTime easily
* generating Fluent::Config::Element easily (required for owned plugins - parser, formatter, storage, buffer and others)
* test drivers for owned plugins to support to
  * instantiate plugins
  * configure plugins
  * set logger, plugin id and override system configs

### DIFF
--- a/lib/fluent/test/driver/base_owned.rb
+++ b/lib/fluent/test/driver/base_owned.rb
@@ -1,0 +1,104 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


tagomoris:
Code of test drivers are not proven because there are not tests for each parsers/formatters.
I'll fix these drivers when I'll add tests for these.

lib/fluent/test/driver/base_owned.rb
```diff
@@ -0,0 +1,104 @@
...
+          @section_name = ''
+        end
+
+        attr_reader :instance, :logs
+
+        def configure(conf, syntax: :v1)
+          if conf.is_a?(Fluent::Config::Element)
+            @config = conf
+          elsif conf.is_a?(Hash)
+            @config = Fluent::Config::Element.new(@section_name, "", Hash[conf.map{|k,v| [k.to_s, v]}])
```
cosmo0920: It seems that last `[]` is needed.

```diff
-@config = Fluent::Config::Element.new(@section_name, "", Hash[conf.map{|k,v| [k.to_s, v]}])
+@config = Fluent::Config::Element.new(@section_name, "", Hash[conf.map{|k,v| [k.to_s, v]}], [])
```

tagomoris: Correct. I'll fix it.

tagomoris: done

tagomoris:
Merged.

PR #1048, opened 2016-06-13
Migrate out_exec_filter to v0.14 API

cosmo0920:
I've tried to migrate `out_exec_filter` to v0.14 API.
Should we provide asynchronous buffered mode in this plugin?

### DIFF
--- a/lib/fluent/plugin/out_exec_filter.rb
+++ b/lib/fluent/plugin/out_exec_filter.rb
@@ -23,9 +23,9 @@ require 'fluent/timezone'
 require 'fluent/plugin/exec_util'
 require 'fluent/config/error'
 
-module Fluent
-  class ExecFilterOutput < BufferedOutput
-    Plugin.register_output('exec_filter', self)
+module Fluent::Plugin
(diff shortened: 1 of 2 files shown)


tagomoris:
I don't think so > async
Because writing data into pipes is atomic operation, and there are no way to confirm whether it succeeded or not.

cosmo0920:
> I don't think so > async
> Because writing data into pipes is atomic operation, and there are no way to confirm whether it succeeded or not.

I see.

tagomoris:
I missed to comment on this change: can you use `child_process` plugin helper for external process?

tagomoris:
Hmm, I found there are some difficult points with `child_process` plugin helper. So I'll add following fixes for it.

cosmo0920:
> some difficult points

I've found that several difficult points for now:

* `child_process` helper does not have a functionality to distribute output into number of N processes
* `child_process` helper does not have a functionality to handle resident processes

tagomoris:
#1297 does what this pull-request did. Let me close this now.

PR #1049, opened 2016-06-14
Migrate out_relabel to v0.14 API

cosmo0920:
Also I've added test for `out_relabel`.


tagomoris:
Thank you for your contribution!

PR #1050, opened 2016-06-14
Extract parser tests into each file

cosmo0920:
- [x] Extract monolithic parser tests into each file
- [x] Inherits `Fluent::Plugin::Parser` in `Fluent::Plugin::Parser::TimeParser`
- [x] Use `Fluent::Test::Driver::Parser` in tests
  - [x] ~~plugin/test_parser~~ Move test_parser tests which are related to `Compat::Parser` into test/comat
  - [x] plugin/test_parser_time


lib/fluent/test/helpers.rb
```diff
@@ -45,6 +45,14 @@ def msgpack(type)
           raise ArgumentError, "unknown msgpack object type '#{type}'"
         end
       end
+
+      def str2time(str_time, format = nil)
```
tagomoris: Could you add keyword argument `format` to `event_time` instead of adding another method `str2time`?

lib/fluent/test/helpers.rb
```diff
@@ -25,9 +25,13 @@ def config_element(name = 'test', argument = '', params = {}, elements = [])
         Fluent::Config::Element.new(name, argument, params, elements)
       end
 
-      def event_time(str=nil)
+      def event_time(str=nil, format=nil)
```
tagomoris: Can you use keyword argument `event_time(str=nil, format: nil)` instead of optional argument?

cosmo0920: I've fixed it.

tagomoris:
LGTM. Thank you!

PR #1051, opened 2016-06-15
add support of X-Forwarded-For header

HatsuneMiku3939:
In the case of running fluentd behind HTTP proxy or load balancer,
a source address of TCP/IP packet will be replaced by load balancer.

This commit will be helpful in these case.


repeatedly:
It seems good. Thanks!

HatsuneMiku3939:
Thanks!

PR #1044, opened 2016-06-09
Extract RecordFilter Mixin into compat

cosmo0920:
Related to https://github.com/fluent/fluentd/wiki/Changes-proposed-for-v0.14.x#misc-development-tasks.

PR #1052, opened 2016-06-15
Extract more mixin module to compat

cosmo0920:
Follows up https://github.com/fluent/fluentd/pull/1044.
I've extracted modules which are defined in `mixin.rb` into compat entirely.

### DIFF
--- a/lib/fluent/compat/set_time_key_mixin.rb
+++ b/lib/fluent/compat/set_time_key_mixin.rb
@@ -14,7 +14,9 @@
 #    limitations under the License.
 #
 
+require 'fluent/config/error'
 require 'fluent/compat/record_filter_mixin'
+require 'fluent/compat/time_formatter'
 
 module Fluent
   module Compat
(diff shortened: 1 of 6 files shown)


lib/fluent/compat/set_time_key_mixin.rb
```diff
@@ -0,0 +1,69 @@
...
+           raise Fluent::ConfigError, "Invalid boolean expression '#{s}' for include_time_key parameter" if include_time_key.nil?
+
+           @include_time_key = include_time_key
+         end
+
+         if @include_time_key
+           @time_key     = conf['time_key'] || 'time'
+           @time_format  = conf['time_format']
+
+           if    conf['localtime']
```
tagomoris: Additional spaces

tagomoris:
Almost code looks good.

lib/fluent/compat/set_time_key_mixin.rb
cosmo0920: I guess these additional spaces to clarify getting localtime/utc information from conf object.
Former is to get `localtime`, latter is to get `utc` from conf object.

tagomoris: Ah... Okay, understood.

tagomoris:
I'll merge this later.

tagomoris:
Merged. Thank you!

PR #1053, opened 2016-06-15
Relax tests sometimes fail in CI environment

tagomoris:


tagomoris:
I'll merge this branch to master, including code to raise error when something happens to make tests to fail.
This is serious timing problem, and cannot reproduce it by my own wish.

PR #1054, opened 2016-06-16
Warn too many staged chunks at configure

tagomoris:
Output plugin with many chunk keys will generate too many staged chunks.
It doesn't cause crashes directly, but make performance worse and may cause any problems with file buffers.


tagomoris:
This change also includes a commit to separate basic calls/classes about testing from `lib/fluent/test/base`, because that file is about deprecated `Fluent::Test::TestDriver`.

PR #1058, opened 2016-06-17
Migrate filter_stdout plugin to v0.14 API

okkez:


lib/fluent/plugin/filter_stdout.rb
```diff
@@ -31,7 +30,7 @@ class StdoutFilter < Filter
     def configure(conf)
       super
 
-      @formatter = Plugin.new_formatter(@format)
+      @formatter = Fluent::Plugin.new_formatter(@format)
```
tagomoris: This plugin should use `:formatter` plugin helper to instantiate/configure Formatter instance.
But currently, there are no way to keep compatibility of configurations between styles of v0.12 (flat parameters of filter and formatter) and v0.14 (`<format>` subsection).
I'll add that method and document. Please wait for a while.

lib/fluent/test/driver/filter.rb
```diff
@@ -25,9 +25,31 @@ module Driver
...
           super
           raise ArgumentError, "plugin is not an instance of Fluent::Plugin::Filter" unless @instance.is_a? Fluent::Plugin::Filter
+          @filtered = []
+        end
+
+        def filtered_records
+          @filtered.map {|_time, record| record }
+        end
+
+        def instance_hook_after_started
```
tagomoris: Great addition!

test/plugin/test_filter_stdout.rb
```diff
@@ -1,84 +1,89 @@
...
+    }
+    d.filtered_records
   end
 
   def test_through_record
     d = create_driver
     time = Time.now
-    filtered = emit(d, {'test' => 'test'}, Fluent::EventTime.from_time(time))
-    assert_equal({'test' => 'test'}, filtered)
+    filtered = filter(d, Fluent::EventTime.from_time(time), {'test' => 'test'})
```
tagomoris: Just `event_time()` test helper method available to get current EventTime instance instead of `time = Time.now; Fluent::EventTime.from_time(time)`.
(Many same code exist below).

tagomoris:
This change depends on #1079

okkez:
Use helpers `compat_parameters` and `inject`.
@tagomoris Could you review again?

lib/fluent/plugin/filter_stdout.rb
```diff
@@ -14,31 +14,39 @@
...
 
     # for tests
     attr_reader :formatter
 
-    desc 'The format of the output.'
-    config_param :format, :string, default: 'stdout'
-    # config_param :output_type, :string, :default => 'json' (StdoutFormatter defines this)
-
     def configure(conf)
+      conf['format'] ||= DEFAULT_FORMAT_TYPE
```
tagomoris: Don't do this operations.
Please use `default_type` keyword argument of `formatter_create` method instead.

tagomoris:
Using `format_config` and `inject_config` is not so bad, but plugin test drivers for v0.14 plugins can receive Element instances for `#configure` method.
I think it's easier and smarter to build Element object than concatenating configuration strings.

okkez:
OK, I will rewrite tests.

tagomoris:
Great. Thank you for contribution!

PR #1059, opened 2016-06-20
Migrate in_tail to v0.14 API

okkez:


okkez:
At 20c8403, one test errors on Windows. I will try to fix the error tomorrow.

okkez:
Ruby 2.1.10 failed on Travis CI. But in my local, I could not reproduce failure.

cosmo0920:
We should use `require 'fluent/test/driver/input'` instead of `require 'fluent/test'` in test.

lib/fluent/plugin/in_tail.rb
```diff
@@ -28,11 +28,13 @@
   Fluent::FileWrapper = File
 end
 
-module Fluent
-  class TailInput < Input
-    include SystemConfig::Mixin
+module Fluent::Plugin
+  class TailInput < Fluent::Plugin::Input
+    include Fluent::SystemConfig::Mixin
```
tagomoris: This mixin is included in `Fluent::Plugin::Base`, the base class of all plugins.

lib/fluent/plugin/in_tail.rb
```diff
@@ -103,7 +105,7 @@ def configure(conf)
     end
 
     def configure_parser(conf)
-      @parser = Plugin.new_parser(conf['format'])
+      @parser = Fluent::Plugin.new_parser(conf['format'])
```
tagomoris: Use `parser` plugin helper to instantiate it with owner's attributes (e.g., plugin id and logger)

lib/fluent/plugin/in_tail.rb
```diff
@@ -187,8 +181,8 @@ def refresh_watchers
 
     def setup_watcher(path, pe)
       line_buffer_timer_flusher = (@multiline_mode && @multiline_flush_interval) ? TailWatcher::LineBufferTimerFlusher.new(log, @multiline_flush_interval, &method(:flush_buffer)) : nil
-      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher,  &method(:receive_lines))
-      tw.attach(@loop)
+      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher, &method(:receive_lines))
+      tw.attach(@_event_loop)
```
tagomoris: Call `helpers :event_loop` plugin helper and use its method `event_loop_attach` instead of using `@_event_loop` directly. It can't be aware in test drivers.

lib/fluent/plugin/in_tail.rb
```diff
@@ -802,6 +771,4 @@ def read_inode
       end
     end
   end
-
-  NewTailInput = TailInput # for backward compatibility
```
tagomoris: I'm not sure whether we can remove it right now or not.
@repeatedly ?

okkez:
I've pushed changes for comments.

:memo:  
* [ ] test using parse section.

okkez:
@tagomoris Could you review again? :bow:
I've pushed changes using `compat_parameters` and `parser` plugin helpers.

tagomoris:
OK, will do. Let me spend some time.

tagomoris:
@okkez All tests are failing either way :(

lib/fluent/plugin/in_tail.rb
repeatedly: We can remove it.
in_tail_ex and in_tail_multiline are deprecated 2+ years ago.
New user, v0.14 user, never need these plugin support.

okkez:
Some tests fails, but I cannot reproduce in my local machine.
Hmm ...

tagomoris:
@okkez It looks a test which fails sometimes... I think it's better to make it sure NOT to fail always.

lib/fluent/plugin/in_tail.rb
```diff
@@ -78,22 +77,27 @@ def initialize
...
       unless @pos_file
         $log.warn "'pos_file PATH' parameter is not set to a 'tail' source."
         $log.warn "this parameter is highly recommended to save the position to resume tailing."
       end
 
-      configure_parser(conf)
       configure_tag
 
-      @multiline_mode = conf['format'] =~ /multiline/
+      parser_config = conf.elements('parse').first
```
tagomoris: `parser_config` is already assigned at L81.

tagomoris:
In v0.12, `format` is hidden required parameter (`configure_parser` raises error if missing).
IMO we need to raise ConfigError explicitly if `<parse> @type ... </parse>` is missing.

lib/fluent/plugin/in_tail.rb
```diff
@@ -120,26 +119,20 @@ def configure_tag
     def start
       super
 
+      @parser = parser_create(conf: @config.elements('parse').first)
```
tagomoris: Are there any reason not to use `parser_config` for conf?

lib/fluent/plugin/in_tail.rb
```diff
@@ -120,26 +119,20 @@ def configure_tag
...
         @pf = PositionFile.parse(@pf_file)
       end
 
-      @loop = Coolio::Loop.new
       refresh_watchers
-
-      @refresh_trigger = TailWatcher::TimerWatcher.new(@refresh_interval, true, log, &method(:refresh_watchers))
-      @refresh_trigger.attach(@loop)
-      @thread = Thread.new(&method(:run))
+      timer_execute(:refresh_watchers, @refresh_interval, &method(:refresh_watchers))
```
tagomoris: Time title is global. So title with `tail_` prefix is better than one without it.

lib/fluent/plugin/in_tail.rb
```diff
@@ -187,8 +179,11 @@ def refresh_watchers
 
     def setup_watcher(path, pe)
       line_buffer_timer_flusher = (@multiline_mode && @multiline_flush_interval) ? TailWatcher::LineBufferTimerFlusher.new(log, @multiline_flush_interval, &method(:flush_buffer)) : nil
-      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher,  &method(:receive_lines))
-      tw.attach(@loop)
+      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher, &method(:receive_lines))
+      tw.attach do |watcher|
+        watcher.timer_trigger = timer_execute(:timer_trigger, 1, &watcher.method(:on_notify)) if watcher.enable_watch_timer
```
tagomoris: There are no reason to assign `timer_trigger` to watcher, right?

okkez: We need to assign `timer_trigger` to watcher.
Detach `timer_trigger` at L390.

tagomoris: All attached event handlers will be detached automatically by shutdown sequence of event loop plugin helper.

lib/fluent/plugin/in_tail.rb
okkez: Here is in `start`, `parser_config` is defined in `configure`.

okkez:
I've pushed changes for all comments.

lib/fluent/plugin/in_tail.rb
tagomoris: Ah, I misread it.

tagomoris:
LGTM. A potential change is to use storage plugin instead of pos file, but it should be done in another pull request.

tagomoris:
@repeatedly please check once more - this plugin is very important for many users.

lib/fluent/plugin/in_tail.rb
```diff
@@ -78,22 +77,37 @@ def initialize
...
+      parser_config = conf.elements('parse').first
+      if parser_config
+        (1..Fluent::Plugin::MultilineParser::FORMAT_MAX_NUM).each do |n|
+          parser_config["format#{n}"] = conf["format#{n}"] if conf["format#{n}"]
+        end
+      end
+
       super
 
+      unless parser_config
```
repeatedly: Why this check is here?
This check should be moved to before `if parser_config` at line 82?

lib/fluent/plugin/in_tail.rb
repeatedly: By this change, configuration error happens at start phase, right?

lib/fluent/plugin/in_tail.rb
okkez: Ah. I misunderstand that `super` at L88 may add `<parse>` section.

lib/fluent/plugin/in_tail.rb
```diff
@@ -244,8 +247,9 @@ def close_watcher(tw, close_io = true)
     end
 
     def close_watcher_after_rotate_wait(tw)
-      closer = TailWatcher::Closer.new(@rotate_wait, tw, log, &method(:close_watcher))
-      closer.attach(@loop)
+      timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
```
repeatedly: `Closer` calls `detach` inside `ensure`.
Does this code detach one-shot timer correctly?

lib/fluent/plugin/in_tail.rb
okkez: Yes.
See also https://github.com/fluent/fluentd-docs/blob/master/docs/v0.14/api-plugin-parser.txt#L78-L79

lib/fluent/plugin/in_tail.rb
okkez: Yes, this code calls `close_watcher(tw)` at once.
Because `repeat: false` is specified, see also ev(3).

lib/fluent/plugin/in_tail.rb
repeatedly: Hmm... I see. This API design seems not good for users.
configuration error should be found in `configure` phase.
So I recommend to change Parser API, maybe also Formatter, into 2 layers.

```
# configure
@parser = parser_create(conf: @config.elements('parse').first)

# start
@parser.start
```

lib/fluent/plugin/in_tail.rb
repeatedly: Executing event is only once but don't call `detach`.
So a watcher is still attached to event loop. It is resource leak.

okkez: I've sent PR for this resource leak. https://github.com/fluent/fluentd/pull/1178

lib/fluent/plugin/in_tail.rb
okkez: @tagomoris How do you think?

okkez:
Rebased.

okkez:
Ruby 2.1.10 on Mac failed CI. But others succeeded.

```
Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnermediator.rb:49:in `ensure in run': On subject Time, (RR::Errors::DoubleNotFoundError)
unexpected method invocation:
  now()
expected invocations:
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnermediator.rb:50:in `run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunner.rb:40:in `start_mediator'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunner.rb:25:in `start'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnerutilities.rb:24:in `run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/autorunner.rb:438:in `block in run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/autorunner.rb:494:in `change_work_directory'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/autorunner.rb:437:in `run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit/autorunner.rb:62:in `run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.1.0/gems/test-unit-3.1.9/lib/test/unit.rb:502:in `block (2 levels) in <top (required)>'

rake aborted!
```

lib/fluent/plugin/in_tail.rb
tagomoris: I'm sorry that I missed to point about it... `create_parser` should be called in `configure` method.

Parser plugin helper will creates instance of `parser`, and then will call `#start` of created parser automatically. No need to call `@parser.start` by plugins themselves.
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin_helper/parser.rb#L95

okkez:
AppVeyor failed. But this is not related to my changes.

```
F
===============================================================================
Failure: test: value of auto increment key is not suspended after stop-and-start(DummyTest::doesn't suspend internal counters in default)
C:/projects/fluentd/test/plugin/test_in_dummy.rb:129:in `block (2 levels) in <class:DummyTest>'
     126:       end
     127: 
     128:       first_id2 = d2.events.first[2]['id']
  => 129:       assert_equal 0, first_id2
     130: 
     131:       assert !File.exist?(File.join(TEST_PLUGIN_STORAGE_PATH, 'json', 'test-01.json'))
     132:     end
<0> expected but was
<1>
===============================================================================
```

tagomoris:
LGTM.

tagomoris:
@okkez Could you rebase and resolve conflicts once again?
I want to merge this change for v0.14.5.

okkez:
Sure! I will rebase later.

okkez:
Rebased!

tagomoris:
Thank you for great patch!

PR #1060, opened 2016-06-20
Migrate in_gc_stat to v0.14 API

okkez:


tagomoris:
Could you try to use `:timer` plugin helper to run callback per fixed interval?
You don't have to write any code about timer and event loops in v0.14.

okkez:
Thank you for review.
I've pushed changes using `:timer` plugin helper.

tagomoris:
LGTM!

PR #1062, opened 2016-06-20
Split events emitted at once to multi chunks

tagomoris:
This change make it possible to emit large event stream into some chunks.
It can't be done at v0.10/v0.12, but it should be done to remove warnings about chunks larger than chunk limit size.

When I was writing this change, I found some problems about locking chunks and buffer global lock.
I also improved it because the buffer with this change will operate much more chunks than ever.

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -44,6 +44,11 @@ module Fluent
       false
     end
 
+    # for test
+    def ==(other)
+      self.to_msgpack_stream == other.to_msgpack_stream
+    end
+
     def slice(index, num)
       raise NotImplementedError, "DO NOT USE THIS CLASS directly."
     end
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -44,11 +44,6 @@ module Fluent
       false
     end
 
-    # for test
-    def ==(other)
-      self.to_msgpack_stream == other.to_msgpack_stream
-    end
-
     def slice(index, num)
       raise NotImplementedError, "DO NOT USE THIS CLASS directly."
     end
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -44,6 +44,10 @@ module Fluent
       false
     end
 
+    def slice(index, num)
+      raise NotImplementedError, "DO NOT USE THIS CLASS directly."
+    end
+
     def each(&block)
       raise NotImplementedError, "DO NOT USE THIS CLASS directly."
     end
@@ -124,6 +128,10 @@ module Fluent
       @entries.empty?
     end
 
+    def slice(index, num)
+      ArrayEventStream.new(@entries.slice(index, num))
+    end
+
     def each(&block)
       @entries.each(&block)
       nil
@@ -167,6 +175,10 @@ module Fluent
       @time_array.empty?
     end
 
+    def slice(index, num)
+      MultiEventStream.new(@time_array.slice(index, num), @record_array.slice(index, num))
+    end
+
     def each(&block)
       time_array = @time_array
       record_array = @record_array
@@ -178,23 +190,32 @@ module Fluent
   end
 
   class MessagePackEventStream < EventStream
-    # Keep cached_unpacker argument for existence plugins
-    def initialize(data, cached_unpacker = nil, size = 0)
+    # https://github.com/msgpack/msgpack-ruby/issues/119
+
+    # Keep cached_unpacker argument for existing plugins
+    def initialize(data, cached_unpacker = nil, size = 0, unpacked_times: nil, unpacked_records: nil)
       @data = data
       @size = size
+      @unpacked_times = unpacked_times
+      @unpacked_records = unpacked_records
     end
 
     def empty?
-      # This is not correct, but actual number of records will be shown after iteration, and
-      # "size" argument is always 0 currently (because forward protocol doesn't tell it to destination)
-      false
+      @data.empty?
     end
 
     def dup
-      MessagePackEventStream.new(@data.dup, @size)
+      if @unpacked_times
+        MessagePackEventStream.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
+      else
+        MessagePackEventStream.new(@data.dup, nil, @size)
+      end
     end
 
     def size
+      # @size is unbelievable always when @size == 0
+      # If the number of events is really zero, unpacking events takes very short time.
+      ensure_unpacked! if @size == 0
       @size
     end
 
@@ -202,8 +223,41 @@ module Fluent
       true
     end
 
+    def ensure_unpacked!
+      return if @unpacked_times && @unpacked_records
+      @unpacked_times = []
+      @unpacked_records = []
+      msgpack_unpacker.feed_each(@data) do |time, record|
+        @unpacked_times << time
+        @unpacked_records << record
+      end
+      # @size should be updated always right after unpack.
+      # The real size of unpacked objects are correct, rather than given size.
+      @size = @unpacked_times.size
+    end
+
+    # This method returns MultiEventStream, because there are no reason
+    # to surve binary serialized by msgpack.
+    def slice(index, num)
+      ensure_unpacked!
+      MultiEventStream.new(@unpacked_times.slice(index, num), @unpacked_records.slice(index, num))
+    end
+
     def each(&block)
-      msgpack_unpacker.feed_each(@data, &block)
+      if @unpacked_times
+        @unpacked_times.each_with_index do |time, i|
+          block.call(time, @unpacked_records[i])
+        end
+      else
+        @unpacked_times = []
+        @unpacked_records = []
+        msgpack_unpacker.feed_each(@data) do |time, record|
+          @unpacked_times << time
+          @unpacked_records << record
+          block.call(time, record)
+        end
+        @size = @unpacked_times.size
+      end
       nil
     end
 
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -180,17 +180,27 @@ module Fluent
 
         staged_bytesize = 0
         operated_chunks = []
+        unstaged_chunks = {} # metadata => [chunk, chunk, ...]
+        enqueued_chunks = []
 
         begin
           metadata_and_data.each do |metadata, data|
             write_once(metadata, data, format: format) do |chunk, adding_bytesize|
               chunk.mon_enter # add lock to prevent to be committed/rollbacked from other threads
               operated_chunks << chunk
+              if chunk.unstaged?
+                unstaged_chunks[metadata] ||= []
+                unstaged_chunks[metadata] << chunk
+              end
+              # this value includes bytes of unstaged chunks, but it will be removed by enqueue_unstaged_chunk
               staged_bytesize += adding_bytesize
             end
           end
 
           return if operated_chunks.empty?
+
+          # Now, this thread acquires many locks of chunks... getting buffer-global lock causes dead lock.
+          # Any operations needs buffer-global lock (including enqueueing) should be done after releasing locks.
 
           first_chunk = operated_chunks.shift
           # Following commits for other chunks also can finish successfully if the first commit operation
@@ -199,7 +209,9 @@ module Fluent
           # permission errors, disk failures and other permanent(fatal) errors.
           begin
             first_chunk.commit
-            enqueue_chunk(first_chunk.metadata) if enqueue || chunk_size_full?(first_chunk)
+            if enqueue || first_chunk.unstaged? || chunk_size_full?(first_chunk)
+              enqueued_chunks << first_chunk
+            end
             first_chunk.mon_exit
           rescue
             operated_chunks.unshift(first_chunk)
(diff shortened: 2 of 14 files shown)


tagomoris:
@repeatedly Please review this change!

repeatedly:
@sonots Could you review this change first?

repeatedly:
BTW, tests are filed on Mac environment. Is this known issue?

lib/fluent/event.rb
```diff
@@ -70,6 +88,10 @@ def repeatable?
       true
     end
 
+    def slice(index, num)
+      self.dup
```
repeatedly: If `slice(1, 1)` is called, `slice` should return same object, not empty steram?

tagomoris:
@repeatedly That's a know one, I'm investigating it continuously.

lib/fluent/event.rb
tagomoris: I can't understand what you mean.
But, I think that `OneEventStream#slice` should return empty stream for first argument `>=1`.

lib/fluent/plugin/buffer.rb
```diff
@@ -400,74 +464,119 @@ def write_once(metadata, data, bulk: false, &block)
...
+      # 1. split event streams into many (10 -> 100 -> 1000 -> ...) chunks
+      # 2. append splits into the staged chunks as much as possible
+      # 3. create unstaged chunk and append rest splits -> repeat it for all splits
+
+      def write_step_by_step(metadata, data, format, splits_count, &block)
+        splits = []
+        if splits_count > data.size
+          splits_count = data.size
+        end
+        slice_size = if splits_count > data.size
```
tagomoris: The first condition is nonsense.

tagomoris:
I pushed some commits to fixes along with some comments on this thread, and rebased on current master HEAD.

tagomoris:
AppVeyor's one task is still failing (maybe file handle leak or something else), but I'll merge this later.
@repeatedly Any other review comments?

repeatedly:
I need more time to review multi-threaded part and benchmark.
For safety v0.14.1 release, merging this patch is better in v0.14.2.

tagomoris:
Memo: I found that this change miss to update `Fluent::Compat::BufferedOutput#handle_stream_simple`.

tagomoris:
I rebased changes on master HEAD, and added a fix for compat layer.

tagomoris:
@repeatedly Please review the changes. I'll take a look for test failures in osx environments.

lib/fluent/event.rb
```diff
@@ -178,32 +195,74 @@ def each(&block)
...
+      return if @unpacked_times && @unpacked_records
+      @unpacked_times = []
+      @unpacked_records = []
+      msgpack_unpacker.feed_each(@data) do |time, record|
+        @unpacked_times << time
+        @unpacked_records << record
+      end
+      # @size should be updated always right after unpack.
+      # The real size of unpacked objects are correct, rather than given size.
+      @size = @unpacked_times.size
```
repeatedly: After unpacked, `@data` should be set to `nil` for GC?

repeatedly: Hmm... `@data` is used in `size`.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -74,7 +74,7 @@ def append(data)
         end
 
         def concat(bulk, bulk_size)
-          raise "BUG: appending to non-staged chunk, now '#{self.state}'" unless self.staged?
+          raise "BUG: appending to unwritable chunk, now '#{self.state}'" unless self.writable?
```
repeatedly: `appending` -> `concatenating` ?

lib/fluent/plugin/buffer/memory_chunk.rb
```diff
@@ -29,7 +29,7 @@ def initialize(metadata)
         end
 
         def append(data)
-          raise "BUG: appending to non-staged chunk, now '#{self.state}'" unless self.staged?
+          raise "BUG: appending to non-staged chunk, now '#{self.state}'" unless self.writable?
```
repeatedly: `non-staged` -> `unwritable`?

lib/fluent/plugin/buffer.rb
```diff
@@ -172,34 +172,46 @@ def metadata(timekey: nil, tag: nil, variables: nil)
       end
 
       # metadata MUST have consistent object_id for each variation
-      # data MUST be Array of serialized events
+      # data MUST be Array of serialized events, or EventStream
       # metadata_and_data MUST be a hash of { metadata => data }
-      def write(metadata_and_data, bulk: false, enqueue: false)
+      def write(metadata_and_data, format: nil, size: nil, enqueue: false)
         return if metadata_and_data.size < 1
         raise BufferOverflowError, "buffer space has too many data" unless storable?
```
repeatedly: `storable?` should care data size of `metadata_and_data`?

lib/fluent/plugin/buffer.rb
```diff
@@ -271,6 +312,18 @@ def enqueue_chunk(metadata)
         nil
       end
 
+      def enqueue_unstaged_chunk(chunk)
+        synchronize do
```
repeatedly: This method is called in `synchronize` in `write`.
This `synchronize` is needed?

lib/fluent/plugin/buffer.rb
```diff
@@ -172,34 +172,46 @@ def metadata(timekey: nil, tag: nil, variables: nil)
...
           metadata_and_data.each do |metadata, data|
-            write_once(metadata, data, bulk: bulk) do |chunk, adding_bytesize|
+            write_once(metadata, data, format: format, size: size) do |chunk, adding_bytesize|
               chunk.mon_enter # add lock to prevent to be committed/rollbacked from other threads
               operated_chunks << chunk
+              if chunk.unstaged?
+                unstaged_chunks[metadata] ||= []
+                unstaged_chunks[metadata] << chunk
+              end
+              # this value includes bytes of unstaged chunks, but it will be removed by enqueue_unstaged_chunk
```
repeatedly: `enqueue_unstaged_chunk` seems to not touch `@stage_size`.
What does "it will be removed" mean?

lib/fluent/plugin/buffer.rb
```diff
@@ -172,34 +172,46 @@ def metadata(timekey: nil, tag: nil, variables: nil)
...
       # metadata_and_data MUST be a hash of { metadata => data }
-      def write(metadata_and_data, bulk: false, enqueue: false)
+      def write(metadata_and_data, format: nil, size: nil, enqueue: false)
         return if metadata_and_data.size < 1
         raise BufferOverflowError, "buffer space has too many data" unless storable?
 
         staged_bytesize = 0
         operated_chunks = []
+        unstaged_chunks = {} # metadata => [chunk, chunk, ...]
+        enqueued_chunks = []
```
repeatedly: `enqueued_chunks` is misleading name because chunks in `enqueued_chunks` is not queued yet.

lib/fluent/plugin/buffer.rb
```diff
@@ -211,17 +223,43 @@ def write(metadata_and_data, bulk: false, enqueue: false)
...
+            # At here, staged chunks may be enqueued by other threads.
+            @stage_size += staged_bytesize
+
+            enqueued_chunks.each do |c|
+              if c.staged? && (enqueue || chunk_size_full?(c))
+                m = c.metadata
+                enqueue_chunk(m)
+                if unstaged_chunks[m]
+                  u = unstaged_chunks[m].pop
+                  if u.unstaged? && !chunk_size_full?(u)
```
repeatedly: Can reuse above `chunk_size_full?` result?

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -74,7 +74,7 @@ def append(data)
         end
 
         def concat(bulk, bulk_size)
```
repeatedly: Off topic. Argument name should follow base class. `Chunk#concat` signature is `def concat(bulk, records)`.

lib/fluent/event.rb
tagomoris: `@data` should be kept as is, because `#to_msgpack_stream` returns `@data` itself.

lib/fluent/plugin/buffer.rb
tagomoris: It can't, because content of `metadata_and_data` are not formatted yet at this line.

lib/fluent/plugin/buffer.rb
tagomoris: All calls to operate `@queue` and `@stage` must be safe for any situation, including methods added in future. Otherwise, someday someone makes deadlock.

lib/fluent/plugin/buffer.rb
tagomoris: `chunk_size_full?(c)` at L244 is about another chunk `c`, and one at L249 is about `u`.

repeatedly: Ah... it's my mistake.

lib/fluent/plugin/buffer.rb
repeatedly: Okay.

lib/fluent/plugin/buffer/file_chunk.rb
tagomoris: update message at d002e1df85dbd9d39e170ddebbdcf7a468982d5c

lib/fluent/plugin/buffer/memory_chunk.rb
tagomoris: update message at d002e1df85dbd9d39e170ddebbdcf7a468982d5c

lib/fluent/plugin/buffer.rb
tagomoris: That comment was outdated, and I found a bug around it.
I fixed it at this commit. a51b96f24c9782c53b0793275c61faa25fe52118

lib/fluent/plugin/buffer.rb
tagomoris: I pushed a commit to rename it e0e3658e984b636a80e2eb5c2e84f18477b0a80d

tagomoris:
@repeatedly Please check this change with latest commits.

tagomoris:
@repeatedly ping?

lib/fluent/plugin/buffer.rb
repeatedly: Hmm... let me clarify. When BufferOverflow is thrown if "incoming data + exising staged and queued size" is larger than buffer limit size?
From the code, BufferOverflowError is thrown at only this point.

lib/fluent/plugin/buffer.rb
```diff
@@ -400,74 +465,117 @@ def write_once(metadata, data, bulk: false, &block)
...
-            # but #enqueue_chunk does nothing in such case
-            enqueue_list << metadata
-            raise ShouldRetry
           end
         end
 
         unless stored
           # try step-by-step appending if data can't be stored into existing a chunk in non-bulk mode
-          write_step_by_step(metadata, data, data.size / 3, &block)
+          write_step_by_step(metadata, data, format, 10, &block)
```
repeatedly: This `10` is heuristic value?
If so, comment is good for why we choose 10.

tagomoris: I added a comment for it.

tagomoris:
@repeatedly I added a commit to add code comment. Is that all for your review comment?

lib/fluent/plugin/buffer.rb
tagomoris: Incoming data size is out of consideration here.
Buffer plugin will raise BufferOverflowError if its bytesize is larger than configured limit when output plugin will write event streams into buffer chunks.
It is almost same behavior of v0.12 Fluentd buffer plugin.

The only thing to be considered is the case that buffer chunk size limit is enough large in comparison with total size limit. In that case, soft limiting for buffer total size doesn't work well, and buffer may use much larger size than configured total limit.
I think it's better to add a configuration check for total/chunk bytesize configuration to help such case.

repeatedly:
I tested split case CPU usage on 2 c3.8xlarge instances.
Here is the result with 100k / sec.
From the result, if the incoming events are larger than `buffer_chunk_limit`, CPU usage is high.

##  fluentd-benchmark one_forward based configuration

flush_interval 0s. No split chunks.

- Forwarder CPU: 78%
- Aggregator CPU: 15%

#### Baseline: v0.12.27

- Fowarder CPU: 69%
- Aggregator CPU: 10.5%

## Use tdlog instead of flowcounter_simple for buffer test

### Forwarder (buffer_chunk_limit 32m)

- CPU: 72%

### Aggregator (tdlog + buffer_chunk_limit 65m)

No split chunks because aggregator's buffer_chunk_limit is larger than forwarder's chunk.

- CPU: 45% - 70%

### Aggregator (tdlog + buffer_chunk_limit 8m)

Split 32mb chunks into smaller chunks.

- CPU: 65% - 85% (average 75%)

tagomoris:
It is in my design - we can't help to consume CPU usage for splitting a chunk into chunks. We get safer chunking instead of additional CPU usage.
@repeatedly How do you think about it?

repeatedly:
@tagomoris Yeah, hard to reduce CPU usage for it.
But we should mention CPU usage issue in the announcement and fluentd-docs because
it may degrade the pipeline performance on aggregator node.

tagomoris:
I agree about it.

PR #1063, opened 2016-06-22
Plugin helper tag time hostname

tagomoris:
This plugin helper is to inject tag, time and hostname into records.
This plugin helper will replace SetTagKeyMixin/SetTimeKeyMixin, by explicit method calls instead of overridden `filter_record`.

`hostname_key` is also added to solve feature requests to inject it into a field.


tagomoris:
This change is based on #1062.

tagomoris:
@repeatedly This change is required for v0.14.1 because this change introduces a plugin helper instead of deprecated SetTimeKeyMixin/SetTagKeyMixin, and changes how to write plugins.

repeatedly:
Okay, let me check. Please wait.

lib/fluent/plugin_helper/inject.rb
```diff
@@ -0,0 +1,112 @@
...
+#    limitations under the License.
+#
+
+require 'fluent/event'
+require 'time'
+
+module Fluent
+  module PluginHelper
+    module Inject
+      def inject_record(tag, time, record)
```
repeatedly: `inject_record` is misleading name.
`inject` is used like `inject xxx into yyy` so this name seems to mean `inject record into yyy`.
So we should use other name, `inject_value_to_record` or something.

lib/fluent/plugin_helper/inject.rb
```diff
@@ -0,0 +1,112 @@
...
+          r[@_inject_tag_key] = tag
+        end
+        if @_inject_time_key
+          r[@_inject_time_key] = @_inject_time_formatter.call(time)
+        end
+
+        r
+      end
+
+      def inject_event_stream(tag, es)
```
repeatedly: ditto

lib/fluent/plugin_helper/inject.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      def self.included(mod)
+        mod.instance_eval do
+          config_section :inject, required: false, multi: false, param_name: :inject_config do
+            config_param :hostname_key, :string, default: nil
+            config_param :hostname, :string, default: nil
+            config_param :tag_key, :string, default: nil
+            config_param :time_key, :string, default: nil
+            config_param :time_type, :enum, list: [:float, :unixtime, :string], default: :float
```
repeatedly: Is `unixtime` typo?

test/plugin_helper/test_inject.rb
```diff
@@ -0,0 +1,385 @@
...
+    assert_equal :string, @d.instance_eval{ @inject_config.time_type }
+    assert_not_nil @d.instance_eval{ @_inject_time_formatter }
+  end
+
+  sub_test_case 'using inject_record' do
+    test 'injects hostname automatically detected' do
+      detected_hostname = `hostname`.chomp
+      @d.configure(config_inject_section("hostname_key" => "host"))
+      logs = @d.log.out.logs
+      assert{ logs.first.include?('[info]: using hostname for specified field host_key="host" host_name="SATOSHI-no-MacBook-Air.local"') }
```
repeatedly: Does testing with `SATOSHI-no-MacBook-Air.local` work on other environment?

test/test_event.rb
```diff
@@ -20,6 +46,21 @@ def setup
       dupped = @es.dup
       assert_kind_of OneEventStream, dupped
       assert_not_equal @es.object_id, dupped.object_id
+      assert_duplicated_records @es, dupped
+    end
+
+    test 'slice' do
+      assert_equal 0, @es.slice(1,1).size
```
repeatedly: Use `(1, 1)` to Ruby style.

lib/fluent/plugin_helper/inject.rb
tagomoris: Plugin Helper methods should be named as `${helper_name}_anything`.
So I'll change these names as `inject_values_to_record` and `inject_values_to_event_stream`.

test/plugin_helper/test_inject.rb
tagomoris: I think some commits dropped when i cherry-picked commits from #1062. will fix.

tagomoris:
I pushed some commits to fix (mainly) tests.

tagomoris:
Now CI is green, finally.

Issue #1042, opened 2016-06-07
supervisor spawn does not carry forward any ruby options passed 


larsf:
In fluentd 0.12 it simply forked itself, and so it was a clone of the process starting up:

@main_pid = fork do
        main_process(&block)
      end

In fluentd 0.14 it now builds its own exec command and spawns a new process using server engine:
  fluentd_spawn_cmd = ServerEngine.ruby_bin_path + " -Eascii-8bit:ascii-8bit "
        fluentd_spawn_cmd << $0.shellescape + ' '
        $fluentdargv.each{|a|
          fluentd_spawn_cmd << (a.shellescape + " ")
        }
      end

      fluentd_spawn_cmd << ("--no-supervisor")
      $log.info "spawn command to main: " + fluentd_spawn_cmd

      params = {}
      params['main_cmd'] = fluentd_spawn_cmd

Any ruby parameters are lost this way:

If you used bundler with fluentd, the first process that started would get started using:
ruby -r bundler/setup <path to fluentd>

the -r bundler/setup does not get passed for the command line for the server engine spawn, and the second fluentd process no longer finds the gems it is bundled with, and you will see a lot of this during startup:

016-06-07 15:41:19 -0700 [info]: reading config file path="../etc/fluentd/fluentd.conf"
2016-06-07 15:41:19 -0700 [info]: starting fluentd-0.14.0
2016-06-07 15:41:19 -0700 [info]: spawn command to main: /opt/mapr/fluentd/fluentd-0.14.00/lib/fluentd-0.14.00-linux-x86_64/lib/ruby/bin/ruby -Eascii-8bit:ascii-8bit /opt/mapr/fluentd/fluentd-0.14.00/bin/../lib/fluentd-0.14.00-linux-x86_64/lib/app/bin/fluentd -c ../etc/fluentd/fluentd.conf --no-supervisor
/opt/mapr/fluentd/fluentd-0.14.00/lib/fluentd-0.14.00-linux-x86_64/lib/ruby/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- yajl (LoadError)
	from /opt/mapr/fluentd/fluentd-0.14.00/lib/fluentd-0.14.00-linux-x86_64/lib/ruby/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #1066, opened 2016-06-23
Fix supervisor to pass RUBYOPT to worker process

naritta:
Fix https://github.com/fluent/fluentd/issues/1042

I did't use shellescape in unix for setting various values directly.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -439,14 +439,17 @@ module Fluent
     def supervise
       $log.info "starting fluentd-#{Fluent::VERSION}"
 
+      rubyopt = ENV["RUBYOPT"]
       if Fluent.windows?
         fluentd_spawn_cmd = ServerEngine.ruby_bin_path + " -Eascii-8bit:ascii-8bit "
+        fluentd_spawn_cmd << ' "' + rubyopt.gsub('"', '""') + '" ' if rubyopt
         fluentd_spawn_cmd << ' "' + $0.gsub('"', '""') + '" '
         $fluentdargv.each{|a|
           fluentd_spawn_cmd << ('"' + a.gsub('"', '""') + '" ')
         }
       else
         fluentd_spawn_cmd = ServerEngine.ruby_bin_path + " -Eascii-8bit:ascii-8bit "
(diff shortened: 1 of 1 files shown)


naritta:
MEMO:
In unix, we can set RUBYOPT in either way like:
``` RUBYOPT="-r bundler/setup" ./bin/fluentd -c fluent.conf ```
``` ruby -r bundler/setup ./bin/fluentd -c fluent.conf ```

In windows, we can't specify RUBYOPT in command line, then we have to set like:
``` ruby -r bundler/setup ./bin/fluentd -c fluent.conf ```

File: lib/fluent/supervisor.rb [hunk 1]
```diff
@@ -439,14 +439,17 @@ def show_plugin_config
     def supervise
       $log.info "starting fluentd-#{Fluent::VERSION}"
 
+      rubyopt = ENV["RUBYOPT"]
       if Fluent.windows?
         fluentd_spawn_cmd = ServerEngine.ruby_bin_path + " -Eascii-8bit:ascii-8bit "
+        fluentd_spawn_cmd << ' "' + rubyopt.gsub('"', '""') + '" ' if rubyopt
```
repeatedly: @nurse Is this `gsub` safe?

nurse: Not safe. Maybe it should use [Shellwords](http://docs.ruby-lang.org/ja/2.3.0/class/Shellwords.html).

repeatedly: @naritta Could you fix?

naritta: Shellwords did't work on windows, then I found this 
http://stackoverflow.com/questions/5636601/is-there-an-equivalent-to-the-ruby-shellwords-module-for-the-windows-shell

If I'm misunderstanding, please tell me.

nurse: Ah, sure. It sounds good.

tagomoris: I cannot understand the reason why this code is like that.
Please add code comment to describe it.

repeatedly:
If there is no concern, I will merge this patch after added the comment.

repeatedly:
Thansk!

naritta:
Thank you very much for reviewing.

PR #1067, opened 2016-06-23
Add from_encoding parameter to in_tail plugin

panage:
Encoding parameter was added in_tail on v.0.12.24 and I used it.
When I reading logs encoded in CP932+LF, fluentd output warn log like a below.
```
2016-06-01 05:14:40 +0000 [warn]: ... error="invalid byte sequence in UTF-8"
```
```
<source>
  @type tail
  path /path/to/log/foo.log
  pos_file /path/to/log/foo.log.pos
  tag foo.var
  encoding utf-8
  format /^ something to read $/
</source>
<match **>
  @type stdout
</match>
```

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -115,19 +115,18 @@ module Fluent
     end
 
     def configure_encoding
-      if @encoding
-        @encoding = parse_encoding_param(@encoding)
+      unless @encoding
         if @from_encoding
-          @from_encoding = parse_encoding_param(@from_encoding)
-        end
-      else
-        if @from_encoding
-          $log.warn "'from_encoding' parameter must be specified in conjunction with 'encoding' paramter."
-          $log.warn "these parameter set to nil."
+          log.warn "'from_encoding' parameter must be specified with 'encoding' parameter."
+          log.warn "'from_encoding' is ignored"
           @encoding = nil
           @from_encoding = nil
+          return
         end
       end
+
+      @encoding = parse_encoding_param(@encoding) if @encoding
+      @from_encoding = parse_encoding_param(@from_encoding) if @from_encoding
     end
 
     def parse_encoding_param(encoding_name)
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -117,11 +117,7 @@ module Fluent
     def configure_encoding
       unless @encoding
         if @from_encoding
-          log.warn "'from_encoding' parameter must be specified with 'encoding' parameter."
-          log.warn "'from_encoding' is ignored"
-          @encoding = nil
-          @from_encoding = nil
-          return
+          raise ConfigError, "tail: 'from_encoding' parameter must be specified with 'encoding' parameter."
         end
       end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -65,9 +65,9 @@ module Fluent
     desc 'Enable the additional watch timer.'
     config_param :enable_watch_timer, :bool, default: true
     desc 'The encoding after conversion of the input.'
-    config_param :encoding, default: nil
+    config_param :encoding, :string, default: nil
     desc 'The encoding of the input.'
-    config_param :from_encoding, default: nil
+    config_param :from_encoding, :string, default: nil
     desc 'Add the log path being tailed to records. Specify the field name to be used.'
     config_param :path_key, :string, default: nil
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -64,14 +64,10 @@ module Fluent
     config_param :multiline_flush_interval, :time, default: nil
     desc 'Enable the additional watch timer.'
     config_param :enable_watch_timer, :bool, default: true
+    desc 'The encoding after conversion of the input.'
+    config_param :encoding, default: nil
     desc 'The encoding of the input.'
-    config_param :encoding, default: nil do |encoding_name|
-      begin
-        Encoding.find(encoding_name)
-      rescue ArgumentError => e
-        raise ConfigError, e.message
-      end
-    end
+    config_param :from_encoding, default: nil
     desc 'Add the log path being tailed to records. Specify the field name to be used.'
     config_param :path_key, :string, default: nil
 
@@ -92,6 +88,8 @@ module Fluent
 
       configure_parser(conf)
       configure_tag
+      @encoding = parse_encoding_param(conf['encoding'])
+      @from_encoding = parse_encoding_param(conf['from_encoding'])
 
       @multiline_mode = conf['format'] =~ /multiline/
       @receive_handler = if @multiline_mode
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -64,14 +64,10 @@ def initialize
     config_param :multiline_flush_interval, :time, default: nil
     desc 'Enable the additional watch timer.'
     config_param :enable_watch_timer, :bool, default: true
+    desc 'The encoding after conversion of the input.'
+    config_param :encoding, default: nil
```
tagomoris: Specify type of this parameter.

lib/fluent/plugin/in_tail.rb
```diff
@@ -64,14 +64,10 @@ def initialize
...
     desc 'The encoding of the input.'
-    config_param :encoding, default: nil do |encoding_name|
-      begin
-        Encoding.find(encoding_name)
-      rescue ArgumentError => e
-        raise ConfigError, e.message
-      end
-    end
+    config_param :from_encoding, default: nil
```
tagomoris: ditto.

lib/fluent/plugin/in_tail.rb
```diff
@@ -92,6 +88,8 @@ def configure(conf)
 
       configure_parser(conf)
       configure_tag
+      @encoding = parse_encoding_param(conf['encoding'])
+      @from_encoding = parse_encoding_param(conf['from_encoding'])
```
tagomoris: It's bad practice to get values by `conf[]` reference, because it's not affected from type conversion and default value configurations. It might be done by subclasses (3rd party plugin which inherits `in_tail`).

tagomoris: `@from_encoding` looks to require that `@encoding` is configured.
It should be checked at here.

panage:
@tagomoris 
Thank you for reviewing. I would like to fix them.

lib/fluent/plugin/in_tail.rb
```diff
@@ -117,6 +114,30 @@ def configure_tag
...
 
+    def configure_encoding
+      if @encoding
+        @encoding = parse_encoding_param(@encoding)
+        if @from_encoding
+          @from_encoding = parse_encoding_param(@from_encoding)
+        end
+      else
+        if @from_encoding
+          $log.warn "'from_encoding' parameter must be specified in conjunction with 'encoding' paramter."
```
tagomoris: Use `log` instead of `$log` to enable plugin logger (`log_level` for each plugin sections).
And I think it's ok to say just `must be specified with "encoding" parameter`.

lib/fluent/plugin/in_tail.rb
```diff
@@ -117,6 +114,30 @@ def configure_tag
...
+          $log.warn "these parameter set to nil."
```
tagomoris: It looks better to show `"from_encoding" is ignored` if it is specified without `encoding`.
"set to nil" explains nothing for users (users doesn't know anything about implementation).

tagomoris: Or, it may be good to raise ConfigError for such configuration.

lib/fluent/plugin/in_tail.rb
```diff
@@ -117,6 +114,30 @@ def configure_tag
...
+          @encoding = nil
+          @from_encoding = nil
+        end
+      end
+    end
+
+    def parse_encoding_param(encoding_name)
+      begin
+        Encoding.find(encoding_name) if encoding_name
```
tagomoris: It seems that it's invalid to call this method with `nil` argument, and there are no code to do such things.
So `if` in postposition looks nonsense.

tagomoris:
LGTM!

tagomoris:
@repeatedly Please check just once more.

lib/fluent/plugin/in_tail.rb
```diff
@@ -117,6 +114,29 @@ def configure_tag
       end
     end
 
+    def configure_encoding
+      unless @encoding
+        if @from_encoding
+          log.warn "'from_encoding' parameter must be specified with 'encoding' parameter."
+          log.warn "'from_encoding' is ignored"
```
repeatedly: This log is confused.
If `from_encoding` must be specified with `encoding`, we should raise an ConfigError.

In addition, no need two line logs. 2nd line should be merged into 1st.

lib/fluent/plugin/in_tail.rb
```diff
@@ -251,7 +267,11 @@ def close_watcher_after_rotate_wait(tw)
     def flush_buffer(tw)
       if lb = tw.line_buffer
         lb.chomp!
-        lb.force_encoding(@encoding) if @encoding
+        if @encoding && @from_encoding
```
repeatedly: Here is microbenchmark result. I think default should be fast. So `f2` implementation is better for almost users.

```
% ruby b_enc_perf.rb
Rehearsal --------------------------------------------------------
f1 with nil / nil      0.840000   0.000000   0.840000 (  0.842853)
f2 with nil / nil      0.760000   0.000000   0.760000 (  0.769146)
f1 with true / true    0.870000   0.000000   0.870000 (  0.871896)
f2 with true / true    0.920000   0.000000   0.920000 (  0.917469)
f1 with true / false   0.990000   0.000000   0.990000 (  0.989552)
f2 with true / false   0.920000   0.010000   0.930000 (  0.926458)
----------------------------------------------- total: 5.310000sec

f1 with nil / nil      0.890000   0.000000   0.890000 (  0.890333)
f2 with nil / nil      0.820000   0.000000   0.820000 (  0.822715)
f1 with true / true    0.890000   0.000000   0.890000 (  0.885648)
f2 with true / true    0.940000   0.000000   0.940000 (  0.938831)
f1 with true / false   1.000000   0.000000   1.000000 (  1.003493)
f2 with true / false   0.950000   0.000000   0.950000 (  0.956331)
```

Code:

```rb
require 'benchmark'

class Foo
  def initialize(enc, from_enc)
    @enc = enc
    @from_enc = from_enc
  end

  def f1(num)
    num.times {
      if @enc && @from_enc
        1
      elsif @enc
        2
      end
    }
  end

  def f2(num)
    num.times {
      if @enc
        if @from_enc
          1
        else
          2
        end
      end
    }
  end
end

n = 20000000

Benchmark.bmbm do |x|
  x.report('f1 with nil / nil') {
    foo = Foo.new(nil, nil)
    foo.f1(n)
  }
  x.report('f2 with nil / nil') {
    foo = Foo.new(nil, nil)
    foo.f2(n)
  }

  x.report('f1 with true / true') {
    foo = Foo.new(true, true)
    foo.f1(n)
  }
  x.report('f2 with true / true') {
    foo = Foo.new(true, true)
    foo.f2(n)
  }

  x.report('f1 with true / false') {
    foo = Foo.new(true, false)
    foo.f1(n)
  }
  x.report('f2 with true / false') {
    foo = Foo.new(true, false)
    foo.f2(n)
  }
end
```

panage: Thank you for microbenchmark result!

tagomoris:
Merged. Thank you for contribution!

PR #1069, opened 2016-06-30
Fix TestDriver to let plugin instance always to have its class name

sonots:
Noticed via https://github.com/fluent/fluent-plugin-s3/pull/154

v0.14 TestLogger currently outputs as:

```
2011-01-03 13:07:36 +0000 [warn]: [] out_s3: delayed events were put
```

when a TestDriver is created with a block.

v0.14 TestLogger tries to add optional_header like `[#{self.class.name}]`

```ruby
From: fluent/fluentd/lib/fluent/log.rb @ line 426 Fluent::PluginLoggerMixin#configure:

    422: def configure(conf)
    423:   super

    428:   if level = conf['@log_level']
    429:     unless @log.is_a?(PluginLogger)
    430:       @log = PluginLogger.new($log.dup)
    431:     end
    432:     @log.level = level
    433:     @log.optional_header = "[#{self.class.name}#{plugin_id_configured? ? "(" + @id + ")" : ""}] "
    434:     @log.optional_attrs = {}
    435:   end
    436: end
```

but, TestDriver creates plugin instance with Class.new if block is given, therefore, it does not have class name. In such case we get the extra `[]`.

With this fix, we will get `[Fluent::S3Output]` as we expect.

### DIFF
--- a/lib/fluent/test/base.rb
+++ b/lib/fluent/test/base.rb
@@ -30,7 +30,9 @@ module Fluent
           if block
             # Create new class for test w/ overwritten methods
             #   klass.dup is worse because its ancestors does NOT include original class name
+            klass_name = klass.name
             klass = Class.new(klass)
+            klass.define_singleton_method("name") { klass_name }
             klass.module_eval(&block)
           end
           @instance = klass.new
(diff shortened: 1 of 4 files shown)


tagomoris:
LGTM, but it's missing for v0.14 Test Drivers.

tagomoris:
@sonots ping?

sonots:
Modified similarly ParserTestDriver and FormatterTestDriver

tagomoris:
I meant that the same fix is needed for v0.14 plugin test drivers under `fluent/test/driver`.

sonots:
Sorry to be late. Fixed for test drivers under fluent/test/driver, too.

tagomoris:
LGTM.

PR #1073, opened 2016-07-01
Disable "process died within 1 second. exit." behavior

### LINKED ISSUE #1072: Supervisor doesn't restart server process if it couldn't listen on a port
# Overview

Supervisor doesn't restart server process if it couldn't start within 1 second with this message:

```
2016-07-01 10:39:53 -0700 [warn]: process died within 1 second. exit.
```

This happens also when the server failed to start due to "address already In use" error and the port is temporarily used by another process by accident (e.g. remaining fluentd process that is going to exit soon).

Especially in production environment that uses td-agent launched through /etc/init.d/td-agent script, or if process is started by in_multiprocess plugin, it's helpful if supervisor keeps restarting the server process (with some (exponential) sleep in between) until it succeeds. Repeating error logs is OK because it's actually a problematic situation that should be alerted.
# Problem

Actual problem happened with in_multiprocess plugin.
A minimal config was as following:

```
<source>
  @type multiprocess
  <process>
    cmdline -c child1.conf
  </process>
  <process>
    cmdline -c child2.conf
  </process>
</source>
```

child1.conf:

```
<source>
  @type forward
  port 24224
</source>
```

child2.conf:

```
<source>
  @type forward
  port 24225  # this local port was used by someone temporarily
</source>
```

Local port 24225 was used by another fluentd process by accident (it was another fluentd process running by accident) and child2 didn't start.
# Environment
- td-agent 2.3.1
- ubuntu 14.04

frsyuki:
Fixes #1072.
Discussion should be done on #1072.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -354,8 +354,8 @@ module Fluent
       $log.info "process finished", code: ecode
 
       if !@finished && Time.now - start_time < 1
-        $log.warn "process died within 1 second. exit."
-        exit ecode
+        $log.warn "process died within 1 second unexpectedly. restarting after 5 seconds."
+        sleep 5
       end
     end
 


repeatedly:
We should not change the default behaviour.
If someone checks the number of processes for startup,
this change breaks such monitoring system.

Adding command line option is better.

tagomoris:
This change breaks the behavior of current Fluentd.

PR #1076, opened 2016-07-04
Migrate in_object_space to v0.14 API

cosmo0920:


lib/fluent/plugin/in_object_space.rb
```diff
@@ -31,50 +33,20 @@ def initialize
...
-      def on_timer
-        @callback.call
-      rescue
-        # TODO log?
-        @log.error $!.to_s
-        @log.error_backtrace
-      end
-    end
-
     def configure(conf)
```
tagomoris: Could you remove this method definition? No need to leave it here (just calling `super`).

lib/fluent/plugin/in_object_space.rb
```diff
@@ -31,50 +33,20 @@ def initialize
...
       super
 
-      @loop = Coolio::Loop.new
-      @timer = TimerWatcher.new(@emit_interval, true, log, &method(:on_timer))
-      @loop.attach(@timer)
-      @thread = Thread.new(&method(:run))
+      timer_execute(:object_space_input, @emit_interval, &method(:on_timer))
     end
 
     def shutdown
```
tagomoris: Remove this method definition because it just calls `super`.

tagomoris:
LGTM. Thank you for contribution!

PR #1077, opened 2016-07-04
Add feature to get filtered records to Fluent::Test::Driver::Filter

okkez:


okkez:
This PR is extracted from #1058
We need this feature to migrate other filter plugins.

tagomoris:
LGTM.

PR #1078, opened 2016-07-04
Migrate in exec to v0.14

cosmo0920:


lib/fluent/plugin/in_exec.rb
```diff
@@ -65,15 +67,15 @@ def configure(conf)
...
+        raise Fleunt::ConfigError, "'tag' or 'tag_key' option is required on exec input"
       end
 
       if @time_key
         if @time_format
           f = @time_format
           @time_parse_proc =
             begin
-              strptime = Strptime.new(f)
+              strptime = ::Strptime.new(f)
```
okkez: Do we need full qualified class name?
I think this is not needed here.

lib/fluent/plugin/in_exec.rb
```diff
@@ -106,37 +108,15 @@ def start
...
-
-        begin
-          Process.kill(:KILL, @pid)
-        rescue #Errno::ECHILD, Errno::ESRCH, Errno::EPERM
-        end
-        @thread.join
-      end
-
       super
     end
```
okkez: We can remove `shutdown` definition.

lib/fluent/plugin/in_exec.rb
cosmo0920: Yes, we can remove this full qualifier.

okkez:
In `setup`, you can use `event_time("2011-01-02 13:14:15")` instead of `Fluent::EventTime.parse("2011-01-02 13:14:15")`.

tagomoris:
`child_process` plugin helper should this plugin code shorter.

tagomoris:
LGTM. Thank you for great change!

PR #1079, opened 2016-07-05
Add parser&filter support for compat_parameters plugin helper

tagomoris:
This pull-request adds some methods to convert v0.12 style parser/formatter parameters to v0.14 style. (flat "format" parameter -> `@type` in `<format>` section)

With this plugin helper, users can move to v0.14 without any modification on configuration file about formatting/parsing.

This change requires some changes listed below:
* add `default_type` keyword argument in methods of parser/formatter/storage plugin helpers to specify default type of these plugins even without any subsections for these plugins
* move code for injecting tag/time (HandleTagAndTimeMixin) from formatter to filter/output using inject plugin helper
* enable `inject` plugin helper to simulate HandleTagAndTimeMixin in compat filter/output plugins in default

Generally speaking, injecting tag/time into records are not of formatter, but filter and output plugins (there are many plugins to do it without formatter plugins). It's best to migrate to use just a code (inject plugin helper) from many similar code in many files.


tagomoris:
@repeatedly Hi, could you please review this?

lib/fluent/compat/parser_utils.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Compat
+    module ParserUtils
+      PARSER_PARAMS = Fluent::PluginHelper::CompatParameters::PARSER_PARAMS
+
+      def self.convert_parser_conf(conf)
+        return if conf.elements(name: 'parse').first
+
+        parser_params = {}
+        PARSER_PARAMS.each do |older, newer|
+          next unless newer
```
repeatedly: `PARSER_PARAMS` doesn't have `false`/`nil` value.
Is this `unless` for safety?

tagomoris: Yes. Someone might add another entry which has `nil` value.

lib/fluent/plugin/formatter_out_file.rb
```diff
@@ -32,9 +32,42 @@ class OutFileFormatter < Formatter
...
+          if conf.has_key?('localtime') && Fluent::Config.bool_value(conf['localtime'])
+            conf['localtime'] = true
+          elsif conf.has_key?('utc') && Fluent::Config.bool_value(conf['utc'])
+            conf['localtime'] = false
+            # Specifying "localtime false" means using UTC in TimeFormatter
+            # And specifying "utc" is different from specifying "timezone +0000"(it's not always UTC).
+            # There are difference between "Z" and "+0000" in timezone formatting.
+            # TODO: add kwargs to TimeFormatter to specify "using localtime", "using UTC" or "using specified timezone" in more explicit way
+          else
+            log.warn "both of localtime and utc are specified as false"
```
repeatedly: How about raising ConfigError?
Specifying `localtime` and `utc` is mistake.
Stopping fluentd prevents unexpected data.

lib/fluent/plugin_helper/compat_parameters.rb
```diff
@@ -40,59 +41,185 @@ module CompatParameters
...
+
+        conf
+      end
+
+      def compat_parameters_inject(conf)
+        return unless conf.elements('inject').empty?
+        return if INJECT_PARAMS.keys.all?{|k| !conf.has_key?(k) }
+
+        # TODO: warn obsolete parameters if these are deprecated
+        attr = compat_parameters_copy_to_subsection_attributes(conf, INJECT_PARAMS) do |compat_key, value|
```
repeatedly: This `do - end`is needed?
It seems same as default behaviour.

repeatedly:
Others are LGTM

lib/fluent/plugin/formatter_out_file.rb
tagomoris: It sounds good idea. I'll fix here to do so.

lib/fluent/plugin_helper/compat_parameters.rb
tagomoris: Ah, it's not needed... I think it remains after some trial&errors. I'll remove it.

tagomoris:
I added commits for review comments.

PR #1082, opened 2016-07-05
Use map instead of inject

okkez:


okkez:
In this case, `map` is enough and suitable.

tagomoris:
Good fix. Thanks!

PR #1085, opened 2016-07-05
Extract formatter plugin test using new test driver

okkez:


okkez:
https://github.com/fluent/fluentd/blob/master/test/test_formatter.rb includes test for old (v0.12) API.

tagomoris:
LGTM

Issue #1084, opened 2016-07-05
Stopping Fluentd service on Windows doesn't work well


tagomoris:
On Windows environment, Fluentd doesn't stop when Administrator operates "stop" for Fluentd Windows Services in Control Panel, which is registered via `fluentd --regwinsvc`.

When I launched Fluentd process via command prompt, it stops successfully by Ctrl-c.

- Fluentd v0.14.1 (via gem install)
- Windows Server 2012

Configuration file:
```
<source>
  @type forward
  @label @dump
</source>

<label @dump>
  <match **>
    @type file
    path C:/mydump.*.log
    time_slice_format %Y%m%d%H%M%S
    time_slice_wait 1s
  </match>
</label>
```

There were no logs when the service was stopped via Services panel, and two "Ruby interpreter" processes were still running.
```
2016-07-05 07:23:58 +0000 [info]: reading config file path="C:\\fluent_test.conf"
2016-07-05 07:23:58 +0000 [info]: starting fluentd-0.14.1
2016-07-05 07:23:58 +0000 [info]: spawn command to main: C:/Ruby22/bin/ruby.exe
-Eascii-8bit:ascii-8bit  "C:/Ruby22/bin/fluentd" "-c" "C:\fluent_test.conf" "-o" "C:\fluentd.log" "-x" "fluentdwinsvc" --under-supervisor
2016-07-05 07:23:59 +0000 [info]: reading config file path="C:\\fluent_test.conf"
2016-07-05 07:23:59 +0000 [info]: starting fluentd-0.14.1 without supervision
2016-07-05 07:23:59 +0000 [info]: gem 'fluent-plugin-td' version '0.10.28'
2016-07-05 07:23:59 +0000 [info]: gem 'fluentd' version '0.14.1'
2016-07-05 07:23:59 +0000 [info]: gem 'fluentd' version '0.14.0.pre.1'
2016-07-05 07:23:59 +0000 [info]: adding match in @dump pattern="**" type="file"
2016-07-05 07:23:59 +0000 [info]: adding source type="forward"
2016-07-05 07:23:59 +0000 [info]: using configuration file: <ROOT>
  <source>
    @type forward
    @label @dump
  </source>
  <label @dump>
    <match **>
      @type file
      path "C:/mydump.*.log"
      time_slice_format "%Y%m%d%H%M%S"
      time_slice_wait 1s
      buffer_path C:/mydump.*.log
      <buffer time>
        flush_mode lazy
[...]

PR #1088, opened 2016-07-06
[WIP] Fix stopping Windows Service Fluentd

tagomoris:
This change is to solve #1084. But there are still some problems.

* `Process.kill(:TERM, pid)` seems not to work as expected from the process as a Windows Service: Errno::INVALID raised (Invalid argument)
* `Process.kill(:INT, pid)` seems not to work as expected too: Errno::EBADF raised (Bad file descriptor)
* `Process.kill(:KILL, pid)` works as expected, but it kills Fluentd without shutdown sequence

Moreover, using `KILL` kills supervisor process only, without killing worker process. It is not what we want. Making `--no-supervisor` default helps this problem, but a big problem (shutdown sequence isn't called) still exists.

### DIFF
--- a/lib/fluent/winsvc.rb
+++ b/lib/fluent/winsvc.rb
@@ -26,22 +26,6 @@ require 'win32/event'
 # include Windows::Library
 # include Windows::Debug
 
-def read_fluentdopt
-  Win32::Registry::HKEY_LOCAL_MACHINE.open("SYSTEM\\CurrentControlSet\\Services\\fluentdwinsvc") do |reg|
-    reg.read("fluentdopt")[1]
-  end
-end
-
-def service_main_start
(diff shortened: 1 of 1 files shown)


tagomoris:
TODO:
* make RPC handlers free from signals, and work well in Windows environment
* make RPC server free from `supervisor.rb`, and can be launched in worker for `--no-supervisor` mode
* make `--no-supervisor` default when Fluentd is launched as Windows Service
* fix winsvc to use HTTP RPC to shutdown Fluentd process

tagomoris:
unak -san says in Twitter that `Process.kill(:TERM, pid)` is available only for sending it to `$$` (the process itself). Sending it to other process fails always with EINVAL.

tagomoris:
This is in progress in other branches.

PR #1089, opened 2016-07-06
Fix to stop Fluentd worker process by Ctrl-C when --no-supervisor specified

tagomoris:
Currently, if a user launches Fluentd on console with `--no-supervisor` option, that process cannot be shutdown by Ctrl-C.
This change fixes it.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -487,6 +487,13 @@ module Fluent
       # worker process SHOULD NOT do anything with SIGINT, SHOULD just ignore.
       trap :INT do
         $log.debug "fluentd main process get SIGINT"
+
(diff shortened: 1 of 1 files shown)


tagomoris:
@repeatedly Please check and merge this.

repeatedly:
Good catch!

tagomoris:
Latest ffi.gem is in trouble - CI on Windows is failing. Hmm.

cosmo0920:
ffi 1.9.13 has been released: https://rubygems.org/gems/ffi/versions/1.9.13
We should re-run CI with this version of ffi gem.

tagomoris:
I rebuilt that test, and now it's green.

PR #1094, opened 2016-07-11
Register Fluent::Plugin::RegexpParser as parser plugin

okkez:


okkez:
I want to use `Fluent::Plugin::RegexpParser` as a parser plugin.
I'm trying to migrate in_tail to v0.14 API.

We can write configuration as following, ffter megerd this PR.

```aconf
<source>
  @type tail
  <parse>
    @type regexp
    expression /pattern/
  </parse>
</source>
```

We can use `parser_create` helper method as following.

```ruby
@parser = parser_create(type: "regexp", conf: {"expression" => "/pattern/"})
```

tagomoris:
Good fix. But don't use Compat modules in v0.14 `lib/fluent/plugin` directory.
How about to create new regexp-based plugins like this?

```ruby
module Fluent::Plugin
  class NginxParser < RegexpParser
    Fluent::Plugin.register_parser('nginx', self)
    config_set_default :expression, "....."
  end
end
```

okkez:
I've pushed changes using `Fluent::Plugin::RegexpParser`.
But still remains Compat modules in parser_multiline.rb

tagomoris:
Are there any specific problems for parser_multiline.rb to be solved?

okkez:
For now, `Fluent::Plugin::RegexpParser` cannot accept `Regexp::MULTILINE` option or `Regexp` object.

okkez:
> For now, Fluent::Plugin::RegexpParser cannot accept Regexp::MULTILINE option or Regexp object.

I've added new `config_param` to `Fluent::Plugin::RegexpParser` and replace Compat module in parser_multiline.rb.
Could you revriew again? @tagomoris

test/plugin/test_parser_apache_error.rb
```diff
@@ -1,11 +1,11 @@
 require_relative '../helper'
 require 'fluent/test/driver/parser'
-require 'fluent/plugin/parser'
+require 'fluent/plugin/parser_apache_error'
 
 class ApacheErrorParserTest < ::Test::Unit::TestCase
   def setup
     Fluent::Test.setup
-    @parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin.new_parser('apache_error'))
+    @parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin::ApacheErrorParser.new)
```
tagomoris: This instance is not parser, but test driver instance. I know it's too late to notify about it, but I want not to leave it as-is because it's misleading for following developers.
@okkez Could you fix it?

test/plugin/test_parser_nginx.rb
```diff
@@ -1,11 +1,12 @@
 require_relative '../helper'
 require 'fluent/test/driver/parser'
-require 'fluent/plugin/parser'
+require 'fluent/plugin/parser_nginx'
```
tagomoris: ditto.

test/plugin/test_parser_regexp.rb
```diff
@@ -23,88 +23,255 @@ def internal_test_case(parser)
...
-  def test_parse_with_configure
-    # Specify conf by configure method instaed of intializer
-    regexp = Regexp.new(%q!^(?<host>[^ ]*) [^ ]* (?<user>[^ ]*) \[(?<time>[^\]]*)\] \[(?<date>[^\]]*)\] "(?<flag>\S+)(?: +(?<path>[^ ]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)$!)
-    parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin::RegexpParser.new(regexp))
-    parser.configure('time_format'=>"%d/%b/%Y:%H:%M:%S %z", 'types'=>'user:string,date:time:%d/%b/%Y:%H:%M:%S %z,flag:bool,path:array,code:float,size:integer')
-    internal_test_case(parser.instance)
-    assert_equal(regexp, parser.instance.patterns['format'])
-    assert_equal("%d/%b/%Y:%H:%M:%S %z", parser.instance.patterns['time_format'])
-  end
+    def test_parse_with_typed
```
tagomoris: Great test name change.

tagomoris: Ah, I missed the diff :(

tagomoris:
I added comments only for test code (and it originally came from other changes).
Others looks good to me.

okkez:
I've fixed and pushed.

tagomoris:
Thank you! I'll merge this later.

PR #1095, opened 2016-07-11
Fix a typo

okkez:
expected_records ->
expect_records


tagomoris:
Great catch. Thank you!

PR #1097, opened 2016-07-12
Shrink buffer api

tagomoris:
This change is needed to minimize points to be overridden for 3rd party plugins which want to interrupt writing operations.

### DIFF
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -63,9 +63,9 @@ module Fluent
 
         # data is array of formatted record string
         def append(data)
-          adding = ''.force_encoding(Encoding::ASCII_8BIT)
+          adding = ''.b
           data.each do |d|
-            adding << d.force_encoding(Encoding::ASCII_8BIT)
+            adding << d.b
           end
           concat(adding, data.size)
         end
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -63,7 +63,11 @@ def initialize(metadata)
 
         # data is array of formatted record string
         def append(data)
-          raise NotImplementedError, "Implement this method in child class"
+          adding = ''.force_encoding(Encoding::ASCII_8BIT)
+          data.each do |d|
```
tagomoris: Right.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -63,7 +63,11 @@ def initialize(metadata)
 
         # data is array of formatted record string
         def append(data)
-          raise NotImplementedError, "Implement this method in child class"
+          adding = ''.force_encoding(Encoding::ASCII_8BIT)
```
okkez: We can use `''.b` here.

repeatedly:
LGTM and either fine for me > `.b`

tagomoris:
I pushed a commit to use `String#b`, and will merge this later (after CI green).

tagomoris:
Merged.

PR #1100, opened 2016-07-13
Fix typos

okkez:
`@emit_streams` does not exist.


okkez:
In previous version, we cannot use `d.run(expect_emits: 1) do ...`.

tagomoris:
Oops. Nice catch.

PR #1101, opened 2016-07-13
Reduce test execution time in syslog

cosmo0920:
```log
# master
% bundle exec ruby test/plugin/test_in_syslog.rb
Loaded suite test/plugin/test_in_syslog
Started
.......

Finished in 10.643330534 seconds.
--------------------------------------------------------------------------------
7 tests, 34 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------
0.66 tests/s, 3.19 assertions/s

# This branch
% bundle exec ruby test/plugin/test_in_syslog.rb
Loaded suite test/plugin/test_in_syslog
Started
.......

Finished in 7.054952274 seconds.
--------------------------------------------------------------------------------
7 tests, 34 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------
0.99 tests/s, 4.82 assertions/s
```

### DIFF
--- a/test/plugin/test_in_syslog.rb
+++ b/test/plugin/test_in_syslog.rb
@@ -59,16 +59,12 @@ class SyslogInputTest < Test::Unit::TestCase
         {'msg' => '<6>Dec 11 00:00:00 localhost logger: foo', 'expected' => Fluent::EventTime.from_time(Time.strptime('Dec 11 00:00:00', '%b %d %H:%M:%S'))},
         {'msg' => '<6>Dec  1 00:00:00 localhost logger: foo', 'expected' => Fluent::EventTime.from_time(Time.strptime('Dec  1 00:00:00', '%b  %d %H:%M:%S'))},
       ]
-      d.end_if do
-        d.emit_count >= 2
-      end
       d.run(expect_emits: 2) do
         u = Fluent::SocketUtil.create_udp_socket(k)
         u.connect(k, PORT)
         tests.each {|test|
           u.send(test['msg'], 0)
         }
-        sleep(0.1) until d.stop?
       end
 
       events = d.events
@@ -82,16 +78,12 @@ class SyslogInputTest < Test::Unit::TestCase
     d = create_driver
     tests = create_test_case
 
-    d.end_if do
-      d.emit_count >= 2
-    end
     d.run(expect_emits: 2) do
       u = UDPSocket.new
       u.connect('127.0.0.1', PORT)
       tests.each {|test|
         u.send(test['msg'], 0)
       }
-      sleep(0.1) until d.stop?
     end
 
     compare_test_result(d.events, tests)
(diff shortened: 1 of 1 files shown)


test/plugin/test_in_syslog.rb
```diff
@@ -59,14 +59,16 @@ def test_time_format
         {'msg' => '<6>Dec 11 00:00:00 localhost logger: foo', 'expected' => Fluent::EventTime.from_time(Time.strptime('Dec 11 00:00:00', '%b %d %H:%M:%S'))},
         {'msg' => '<6>Dec  1 00:00:00 localhost logger: foo', 'expected' => Fluent::EventTime.from_time(Time.strptime('Dec  1 00:00:00', '%b  %d %H:%M:%S'))},
       ]
-
-      d.run do
+      d.end_if do
+        d.emit_count >= 2
+      end
```
tagomoris: This specification is just same with `d.run(expect_emits: 2)`. Why do you do this?

test/plugin/test_in_syslog.rb
```diff
@@ -59,14 +59,16 @@ def test_time_format
...
+      d.run(expect_emits: 2) do
         u = Fluent::SocketUtil.create_udp_socket(k)
         u.connect(k, PORT)
         tests.each {|test|
           u.send(test['msg'], 0)
         }
-        sleep 1
+        sleep(0.1) until d.stop?
```
tagomoris: Hmm, test driver should do this in default. How do you think @cosmo0920 ?

cosmo0920: Ideally, so.

Should test driver include below change?

```diff
diff --git a/lib/fluent/test/driver/base.rb b/lib/fluent/test/driver/base.rb
index 377bf2e..a82c628 100644
--- a/lib/fluent/test/driver/base.rb
+++ b/lib/fluent/test/driver/base.rb
@@ -231,6 +231,7 @@ module Fluent
             end
           else
             block.call
+            sleep 0.1 until stop?
           end
         end
```

test/plugin/test_in_syslog.rb
cosmo0920: I want to sleep until `stop?` is `false`, and awake after `stop?` is `true` here.

cosmo0920: Current, test driver implementation does not wait when block given.
So, in this case, `d.events` always returns empty events because `u.send` is heavy operation.

test/plugin/test_in_syslog.rb
tagomoris: Like that. I'm creating the change for it.

tagomoris:
I added #1117.

cosmo0920:
I've rebased and added mising assertions.

tagomoris:
LGTM. Thanks!

PR #1102, opened 2016-07-13
Use new test driver syntax in_exec test cases

cosmo0920:
We should use new test driver syntax and brush up more rigid test cases in test_in_exec.


tagomoris:
LGTM. Thanks!

PR #1103, opened 2016-07-13
test driver: Retrieve raw event_streams for test

okkez:


okkez:
After merge this, we can do following:

```ruby
d = create_driver
d.run(default_tag: "test") do
  d.feed(time, record1)
  d.feed(time, record2)
end

assert_equal(n_events, d.event_streams.size)
```

We can write complex condition in `d.run` block.

tagomoris:
I think that `d.events` has enough feature for such requirement.
Isn't it correct?

okkez:
I want to do something like [here](https://github.com/fluent/fluentd/blob/master/lib/fluent/test/driver/base.rb#L198)
I want to check `@emit_strems.size`  in `d.run do ...` for `in_tail` [test](https://github.com/fluent/fluentd/pull/1059/files#diff-bfcc447c6001db7b2bf25b5193e6c646R301).

tagomoris:
IMO it's better to add API to get the number of emits If it's what you want.
Supplying `@emit_streams` makes it impossible to change internal data structure of test driver.

okkez:
OK, I will add new API to retrieve `@emit_streams.size`.

tagomoris:
Cool. Thanks.

PR #1104, opened 2016-07-13
handle ERROR_PATH_NOT_FOUND

ichii386:
On Windows, WindowsFile.new (i.e., in_tail plugin) raises an unhandled error if the path is under non-existent directory.
This has been since FileWrapper.stat() changed not to use FileWrapper.io(), which returns ENOENT unless _open_osfhandle() returns valid fd.

I'm not sure this wouldn't cause any side effect but I report it anyway :)


repeatedly:
@nurse Could you check this?

nurse:
Looks good, thanks!

PR #1105, opened 2016-07-14
Use new test driver syntax in in_gc_stat test cases

cosmo0920:
We should use new test driver notations in test_in_gc_stat.


tagomoris:
LGTM. Great change.

cosmo0920:
Thanks!

Issue #902, opened 2016-04-17
using filter to grep out message from nested map value


jchauncey:
I have messages that look like the following:

``` json
{"log":"2016/04/17 21:48:57 INFO: Starting nginx...\n","stream":"stderr","docker":{"container_id":"3578603a1df45535ae076c3646f6dc45cadad0c4d54c614a599a78a3b38d7543"},"kubernetes":{"namespace_name":"deis","pod_id":"2f12a4ac-04e6-11e6-8ac1-42010af000ff","pod_name":"deis-router-sus7l","container_name":"deis-router","labels":{"app":"deis-router"},"host":"gke-jchauncey-42ec0f7e-node-h9uj"}}
```

I want to filter out all messages that come from certain containers so I have tried doing the following:

```
<filter **>
  @type grep
  exclude1 kubernetes.container_name deis-logger.*
</filter>
```

```
<filter kubernetes.**>
  @type grep
  exclude1 kubernetes.container_name deis-logger.*
</filter>
```

And other combinations but nothing seems to work.

So here is the configuraiton file:

```
<ROOT>
  <filter kubernetes.**>
    type kubernetes_metadata
    kubernetes_url https://10.167.240.1:443
    bearer_token_file /var/run/secrets/kubernetes.io/serviceaccount/token
    verify_ssl false
  </filter>
  <filter kubernetes.**>
    @type grep
    exclude1 kubernetes.container_name deis-logger.*
  </filter>
  <source>
    type tail
    path /var/log/containers/*.log
    pos_file /var/log/containers.log.pos
    time_format %Y-%m-%dT%H:%M:%S.%NZ
    tag kubernetes.*
    format json
    read_from_head true
  </source>
  <source>
    type tail
    format syslog
    path /var/log/startupscript.log
    pos_file /var/log/startupscript.log.pos
    tag startupscript
  </source>
  <source>
    type tail
    format /^time="(?<time>[^)]*)" level=(?<severity>[^ ]*) msg="(?<message>[^"]*)"( err="(?<error>[^"]*)")?( statusCode=($<status_code>\d+))?/
    time_format %Y-%m-%dT%H:%M:%S.%NZ
    path /var/log/docker.log
    pos_file /var/log/docker.log.pos
    tag docker
  </source>
  <source>
    type tail
    format none
    path /var/log/etcd.log
    pos_file /var/log/etcd.log.pos
    tag etcd
  </source>
  <source>
    type tail
    format multiline
[...]

Issue #649, opened 2015-08-05
record_transformer remove_keys: support structured data


svengau:
Hi, 

It would be nice that `remove_keys` support an array in record_transformer.

Example: remove set-cookie header from an NGinx response:

```
<filter **>
  type record_transformer
  remove_keys res.headers.set-cookie
</filter>
```

Thanks,
Sven

sonots:
remove_keys support an array like (comma separated):

```
remove_keys a,b,c
```

svengau:
Sorry for the misunderstand, I mean for this kind of log: 

```
res {
  headers {
    set-cookie: "xxxxxxxxxx",
    etag: "xxxxxxxx",
    ...
  }
}
```
having the possibility to remove structured data, like set-cookie and etag:
```
remove_keys res.headers.set-cookie,res.headers.etag

**Comment by sonots** [comment]
Let me take a note. I do not think this problem can be solved soon. 

1. Need to determine an official way to express a structured data. Using `.` as a separator means that we need to prohibit `.` for record keys so that we can differentiate `record["a.b"]` and `record["a"]["b"]`, for example. 
  * Prohibiting `.` for record keys breaks backward compatibility
2. Need to add features to transform record structured data (not only remove) side by side.

**Comment by repeatedly** [comment]
Does JSONPath use `.` for separator?

**Comment by llby** [comment]
@repeatedly @sonots 
How abount add `remove_json_keys` at config_param
and use like this

```
remove_json_keys {"message":[{"res":{"headers":{"set-cookie":0}}},{"res":{"headers":{"etag":0}}}]}
```

0 means delete
because json needs value

repeatedly:
Use https://github.com/fluent/fluentd/issues/1578 instead

PR #1107, opened 2016-07-17
[#649] record_transformer support remove keys structured data

llby:
record_transformer's `remove_keys` only support remove first hierarchy keys of record.
`remove_json_keys` remove structured data of record['message']. When It is json.
from issue #649

If it work fine. Then I'll try to #902.


tagomoris:
@llby Could you add what you want in description?

tagomoris:
For nested key-values, we don't have any common syntax to specify it in records.
I think we can't accept this change because this change uses JSON in unique way, and it can't be used for other plugins.

For nested key-values, we already have an item to make plugin helpers to handle it. But we need many discussions about what syntax/API is the best for it before implementing it.
https://github.com/fluent/fluentd/wiki/Changes-proposed-for-v0.14.x#misc-development-tasks

llby:
Hmm..
That's right.
I understand.

okkez:
#1800 resolves this PR.
Could you close this PR? @repeatedly

repeatedly:
Ah, I forgot to close this. Thanks for the notification.

Issue #1068, opened 2016-06-28
CSV parser doesn't have a parameter "delimiter"


tagomoris:
It's better to have it.

tagomoris:
#1108 already solved this issue.

PR #1108, opened 2016-07-17
[#1068] Add delimiter option to parser_csv

llby:
- some test dosen't run. because it has not arg 'param'.
- CSV.parse_line return nil from empty. so flag 'null_empty_string' do nothing.
- and same problem has test_parser_tsv.rb


tagomoris:
Do you mean that adding `col_sep` argument to `CSV.parse_line` changes default behavior for empty string?
I think it's not acceptable for many users. Are there any way to help it?

llby:
`col_sep` is not affected.
`CSV.parse_line` change `,,` to `nil,nil,nil` from the beginning.
`test_parse_with_null_value_pattern` is not work.
because lost `(param)`.
so I modify to test pass.

tagomoris:
Okay, I understood that `test-unit` has a bug for such case: https://github.com/test-unit/test-unit/issues/122

lib/fluent/plugin/parser_csv.rb
```diff
@@ -23,8 +23,10 @@ module Plugin
     class CSVParser < ValuesParser
       Plugin.register_parser('csv', self)
 
+      config_param :delimiter, :string, default: ','
+
       def parse(text)
-        yield values_map(CSV.parse_line(text))
+        yield values_map(CSV.parse_line(text,{col_sep: @delimiter}))
```
tagomoris: Use keyword argument style instead of specifying a hash object: `CSV.parse_line(text, col_sep: @delimiter)`

tagomoris:
I created a pull-request to fix bugs you pointed: https://github.com/fluent/fluentd/pull/1112
Could you rebase your change on master after merging it?

tagomoris:
I merged #1112 .

llby:
done.

tagomoris:
Great addition. Thanks!

PR #1111, opened 2016-07-18
Add an RPC API call to flush buffers and stop daemon gracefully

mansinahar:
**Why this PR?** 
Fluentd already supports SIGUSR1 to flush buffers and SIGINT/SIGTERM to
stop the daemon gracefully. It would be really good to at least support
an RPC API call that could do both i.e. flush the buffers and stop the
daemon gracefully.

**What this PR does?**
Adds support for an RPC API call that would flush the
buffers and stop the daemon gracefully.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -82,6 +82,14 @@ module Fluent
         end
         nil
       }
+      @rpc_server.mount_proc('/api/plugins.flushBuffersAndKillWorkers') { |req, res|
+        $log.debug "fluentd RPC got /api/plugins.flushBuffersAndKillWorkers request"
+        unless Fluent.windows?
+          Process.kill :USR1, $$
+          Process.kill :TERM, $$
+        end
+        nil
+      }
       @rpc_server.mount_proc('/api/config.reload') { |req, res|
         $log.debug "fluentd RPC got /api/config.reload request"
         if Fluent.windows?


tagomoris:
Is it enough good to use the last signal what we can use?
IMO adding RPC API is good, but it's not enough good to use USR2 for it.
@repeatedly How do you think about it?

repeatedly:
> it's not enough good to use USR2 for it.

I agreed this point. Singal slot is limited and this is done by two signal combination.
RPC has no limitation so adding feature to RPC seems no problem for me.

tagomoris:
@mansinahar Could you change this patch not to use SIGUSR2 if it's ok for you?
It's too precious to use for it in our opinion.

mansinahar:
@repeatedly and @tagomoris First of all sorry for the silence. Was a little busy the last 2 days. I understand your point and I will make the changes to just have an RPC API. Thanks for the feedback!

tagomoris:
To be honest, sending signals to self process is bad practice, and I will remove such code in these coming releases. Signals are not ensured for these arrival order.
Could you rewrite your code to call 1. flush buffers 2. shutdown instead of sending signals?

mansinahar:
@tagomoris So for shutting down currently we send the TERM signal to the supervisor process so do we wanna change that too and shut it via serverengine's stop method or something?

tagomoris:
I'm willing to change these finally. But these are just one action per one signal (RPC request), so it's not so prior.
But yours is to kick two handler, so it's prior not to use signals (because of inorderity of signal handling).

mansinahar:
@tagomoris Yeah, I understand what you're trying to say but I can flush the buffers without using a signal and then either use the signal to terminate the daemon (it will be only one signal as I am not using a signal to flush buffers) or maybe call the serverengine's stop method. So, would it be better to just do the later already?

tagomoris:
@mansinahar I think serverengine's method is better than using signals. When we're using signals even only for shutdown, we should consider about whether it works well on Windows environment. If not, we can just ignore it.
(But it's not so big problem if there's just one signal handling and well-ordered operations.)

mansinahar:
@tagomoris Hey, so I tried and figured out a way to use serverengine's stop method to stop the daemon but I am still using signal to flush the buffer and even in this case the order is not maintained because it sends the signal to flush buffer and continues to terminate the daemon (without waiting for the flush buffer to complete). I was, therefore, looking at how the RPC call for flush buffer behaves in case of windows like environment where these signals aren't supported but it seems that currently, it does nothing in case of windows. So, I wanted to ask do we support a way to flush buffers without sending a signal (even on windows)?

tagomoris:
Ah, I missed that it's still under construction to deliver messages to control worker processes from supervisor on issue #1088. I'm sorry for my misleading comment.
I think using signals for this purpose is ok for now, and we'll update the code after #1088 be fixed.

@mansinahar the only one more comment: could you add warning to show that the feature is disabled and RPC call will be ignored on Windows environment?

tagomoris:
This feature was merged at #1134. Thank you for proposing this!

PR #1117, opened 2016-07-20
Test code block will block until specified conditions are archieved

tagomoris:
The main purpose of this change is to change `d.run` to wait all specified conditions turns into `true`.
Without this change, `d.run(expect_emits: 2){ 1; }` completes immediately. This behavior is different from test authors.
With this change, `d.run(expect_emits: 2){ ... }` blocks until plugins emit twice.

This commit does the change above: https://github.com/fluent/fluentd/commit/12c7b622ac200978ae2eab83544d74efde4f4902

Other commits merges test drivers for normal plugins (owners) and owned plugins. Before this change, test drivers for owned plugins does NOT provide some features like `d.end_if`.
It's highly confusing, so I created the base class for all test drivers, and merged features commonly provided.


tagomoris:
I'll merge this right after CI green.

cosmo0920:
:+1:

PR #1118, opened 2016-07-20
Use Fluent::Plugin::Filter instead Fluent::Filter to cover all filter classes

repeatedly:
`is_a?(Filter)` can't detect `Fluent::Plugin::Filter` and it causes non compat filter is registered as a Output plugin.

I changed test filter classes for testing this case but I'm not sure this is good or not...

### DIFF
--- a/lib/fluent/event_router.rb
+++ b/lib/fluent/event_router.rb
@@ -159,7 +159,7 @@ module Fluent
       pipeline = nil
       @match_rules.each_with_index { |rule, i|
         if rule.match?(tag)
-          if rule.collector.is_a?(Filter)
+          if rule.collector.is_a?(Plugin::Filter)
             pipeline ||= Pipeline.new
             pipeline.add_filter(rule.collector)
           else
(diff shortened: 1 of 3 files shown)


repeatedly:
Currently, `record_transformer` doesn't work ;p

lib/fluent/event_router.rb
```diff
@@ -159,7 +159,7 @@ def find(tag)
       pipeline = nil
       @match_rules.each_with_index { |rule, i|
         if rule.match?(tag)
-          if rule.collector.is_a?(Filter)
+          if rule.collector.is_a?(Plugin::Filter)
```
tagomoris: Use `is_a?(Fluent::Plugin::Filter)` to detect an object is a filter or not.

tagomoris: I misread this line :P

tagomoris:
I cannot understand why changes on tests are needed. Could you explain it?

repeatedly:
Because current test doesn't check Fluent::Plugin::Filter case.
So I added Fluent::Plugin::Filter and Fluent::Filter case to Pipeline test.

tagomoris:
I got it.
LGTM.

PR #1119, opened 2016-07-20
Fix not to overwrite something unexpectedly

tagomoris:
Originally, some plugin helpers (e.g., storage, formatter, parser, ...) defines `config_section` for sets of parameters by `included` hook. It calls `config_section` with plugin class as receivers.
But, if plugin class calls `config_section` to override default value, that **overwrites** the section definition defined by plugin helpers.
It makes many unexpected behaviors (including breaking instance variables).

This change is to prevent overwriting section definitions by plugins. By inserting a mixin module, `config_section` definition in plugin classes overrides definitions in plugin helpers, instead of overwriting it.

### DIFF
--- a/lib/fluent/configurable.rb
+++ b/lib/fluent/configurable.rb
@@ -144,9 +144,10 @@ module Fluent
       end
 
       def config_section(name, **kwargs, &block)
+        section_already_exists = !!merged_configure_proxy.sections[name]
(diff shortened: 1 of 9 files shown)


tagomoris:

repeatedly:
okay. I will review it later.

repeatedly:
LGTM 👍

PR #1130, opened 2016-07-28
LoggerInitializer updated to create directory when it doesn't exists.

bmpandrade:
The intent of this change is to allow fluentd to create a directory that is passed by --log and don't exists on the system.

Cheers

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -268,7 +268,8 @@ module Fluent
 
       def init
         if @path && @path != "-"
-          @io = File.open(@path, "a")
+          Dir.mkdir(File.dirname(@path)) unless File.exists?(@path)
+	  @io = File.open(@path, "a")
           if @chuser || @chgroup
             chuid = @chuser ? ServerEngine::Daemon.get_etc_passwd(@chuser).uid : nil
             chgid = @chgroup ? ServerEngine::Daemon.get_etc_group(@chgroup).gid : nil


lib/fluent/supervisor.rb
```diff
@@ -268,7 +268,8 @@ def initialize(path, level, chuser, chgroup, opts)
 
       def init
         if @path && @path != "-"
-          @io = File.open(@path, "a")
+          Dir.mkdir(File.dirname(@path)) unless File.exists?(@path)
```
tagomoris: It fails when the file (`@path`) doesn't exist and directory (`File.dirname(@path)`) already exists, and when the parent directories of directory (`File.dirname(File.dirname(@path))`) doesn't exist.
It's better to use `FileUtils.mkdir_p(File.dirname(@path)) unless Dir.exist?(File.dirname(@path))`

tagomoris: `File.exists?` (and `Dir.exists?`) are deprecated methods. Use `File.exist?` instead.

lib/fluent/supervisor.rb
```diff
@@ -268,7 +268,8 @@ def initialize(path, level, chuser, chgroup, opts)
...
+	  @io = File.open(@path, "a")
```
tagomoris: Don't make diff to change indentation.

tagomoris:
Please check tests and CI results.

bmandrade:
Thanks @tagomoris for the feedback. I'll give it a look and make the proper updates 👍

tagomoris:
@bmandrade ping?

bmandrade:
Sorry @tagomoris , I'll try to take on this as soon as possible

ganmacs:
closed by https://github.com/fluent/fluentd/pull/2732

Issue #1131, opened 2016-07-29
v0.14 doesn't load plugins from directory specified by "-p" option


tagomoris:
Fluentd v0.14 launched with `-p DIR` option doesn't load plugins like `DIR/out_yay.rb`.
Plugins like `DIR/fluent/plugin/out_yay.rb` are loaded. So it's a bug of Registry & Plugin which are newly introduced in v0.14.

tagomoris:
#1133 should solve this issue.

PR #1133, opened 2016-08-01
Use separate search prefix for gems and additional plugin path

HatsuneMiku3939:
Allow to search plugin from multiple prefix directory. 

Related with issue #1131 (v0.14 doesn't load plugins from directory specified by "-p" option)

### DIFF
--- a/lib/fluent/plugin.rb
+++ b/lib/fluent/plugin.rb
@@ -26,15 +26,15 @@ module Fluent
     #   ex: storage, buffer chunk, ...
 
     # first class plugins (instantiated by Engine)
-    INPUT_REGISTRY     = Registry.new(:input,     'fluent/plugin/in_')
-    OUTPUT_REGISTRY    = Registry.new(:output,    'fluent/plugin/out_')
-    FILTER_REGISTRY    = Registry.new(:filter,    'fluent/plugin/filter_')
+    INPUT_REGISTRY     = Registry.new(:input,     'fluent/plugin/in_',         dir_search_prefix: 'in_')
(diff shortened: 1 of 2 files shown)


tagomoris:
With this change, Fluentd seems to load all `in_*.rb` files from anywhere possible. Is it correct and intentional?
Fluentd should load plugins from directory specified by `-p DIR` (primal), or gems which have paths like `fluent/plugin/in_*.rb`.

HatsuneMiku3939:
No. It's not intentional. Thanks for mention it.

It seems to we need two rule of plugin search prefix. 
* for specified directory (not have paths like `fluent/plugin/in_*.rb`)
* for gems

HatsuneMiku3939:
This is working as we expect. but not looks good to me. 
Do you have an idea to improve this?

tagomoris:
Current your implementation is breaking code compatibility of `Registry` class. It might be used by any 3rd party plugins. It can't be merged.

tagomoris:
My idea is to add optional keyword argument (like `file_prefix` or some better name) to be used only for `@paths`, earlier than `@search_prefix`.

HatsuneMiku3939:
@tagomoris 
Sorry for breaking change. actually, I didn't realize `Registry` is opened to 3rd party plugins. Thanks for mention it.

How about this one? https://github.com/HatsuneMiku3939/fluentd/commit/d57576af199b3171d190263caa67ef3a3c9c821d

tagomoris:
@HatsuneMiku3939 it looks good to me!

HatsuneMiku3939:
@tagomoris 
Thanks!

repeatedly:
LGTM

PR #1134, opened 2016-08-01
add a RPC call to flush buffers and stop workers

tagomoris:
revised version of #1111.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -75,6 +75,16 @@ module Fluent
         Process.kill :TERM, $$
         nil
       }
+      @rpc_server.mount_proc('/api/processes.flushBuffersAndKillWorkers') { |req, res|
+        $log.debug "fluentd RPC got /api/processes.flushBuffersAndKillWorkers request"
+        if Fluent.windows?
+          $log.warn "operation 'flushBuffersAndKillWorkers' is not supported on Windows now."
+        else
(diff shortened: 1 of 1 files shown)


lib/fluent/supervisor.rb
```diff
@@ -82,6 +82,16 @@ def run_rpc_server
         end
         nil
       }
+      @rpc_server.mount_proc('/api/plugins.flushBuffersAndKillWorkers') { |req, res|
```
repeatedly: killWorkers operation is not related with plugins so adding it to `plugins` prefix seems not good.
processes is better.

tagomoris: Understood.

tagomoris:
@repeatedly could you review once again?

lib/fluent/supervisor.rb
```diff
@@ -75,6 +75,16 @@ def run_rpc_server
         Process.kill :TERM, $$
         nil
       }
+      @rpc_server.mount_proc('/api/processes.flushBuffersAndKillWorkers') { |req, res|
+        $log.debug "fluentd RPC got /api/plugins.flushBuffersAndKillWorkers request"
```
repeatedly: plugins -> processes

tagomoris: uguh, fixed :P

repeatedly:
LGTM

Issue #1135, opened 2016-08-01
Add size to tell the number of events in a message


ganmacs:
To tell the number of events, add `size` key to `option` in forward protocol.

I think it'll work by adding below task.
* add `size` key to `option` at `ForwardOutput#send_data` 
* pass the size value to constructor's argument of `EventStream` at `ForwardInput#on_message`

ref https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v0

tagomoris:
You mean "add `size` key to `option` at `ForwardOutput#send_data`", right?
Currently, `EventStream#initialize` doesn't have `size` argument. Will you add it?

ganmacs:
> You mean "add size key to option at ForwardOutput#send_data", right?

Yes,  i'll fix it.


> Currently, EventStream#initialize doesn't have size argument. Will you add it?

Sorry, i was wrong.
`MessagePackEventStream` and `MultiEventStream`  require adding size argument. 
And `MessagePackEventStream` has already added `size`  argument.
So i'll add `size` argument to `MultiEventStream#initialize` like below

```rb
def initialize(time_array = [], record_array = [], size= time_arrry.size)
 ...
end
```

tagomoris:
@ganmacs Okay, got it. Looks good to me.

PR #1137, opened 2016-08-03
Add size key to option in forward protocol

ganmacs:
ref issue is #1135
But this patch some differences form  #1135 proposed.
Because all we want to do is to suppress calling `MessagePackEventStream#ensure_unpacked!` , this patch did not add `size` to argument of MultiEventStream#initialize

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -111,6 +111,10 @@ module Fluent
       def size
         self.bytesize
       end
+
+      def size_of_events
+        @size + @adding_size
+      end
     end
 
     module BufferedChunkMixin
--- a/test/plugin/test_out_forward.rb
+++ b/test/plugin/test_out_forward.rb
@@ -120,6 +120,70 @@ class ForwardOutputTest < Test::Unit::TestCase
     assert_equal 2, d.instance.ack_response_timeout
   end
 
+  def test_sending_contains_with_ack
+    target_input_driver = create_target_input_driver(true)
+
+    d = create_driver(CONFIG + %[
+      ack_response_timeout 1s
+    ])
+
+    time = Time.parse("2011-01-02 13:14:15 UTC").to_i
+
+    records = [
+      {"a" => 1},
+      {"a" => 2}
+    ]
+    d.register_run_post_condition do
+      d.instance.responses.length == 1
+    end
+
+    target_input_driver.run do
+      d.run do
+        records.each do |record|
+          d.emit record, time
+        end
+      end
+    end
+
+    emits = target_input_driver.emits
+    assert_equal ['test', time, records[0]], emits[0]
+    assert_equal ['test', time, records[1]], emits[1]
+
+    assert_equal target_input_driver.instance.received_options[0]['size'], 2
+  end
+
+  def test_sending_contains_without_ack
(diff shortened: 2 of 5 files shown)


test/plugin/test_out_forward.rb
```diff
@@ -379,12 +443,18 @@ def create_target_input_driver(do_respond=false, disconnect=false, conf=TARGET_C
     DummyEngineDriver.new(Fluent::ForwardInput) {
       handler_class = Class.new(Fluent::ForwardInput::Handler) { |klass|
         attr_reader :chunk_counter # for checking if received data is successfully deserialized
+        attr_reader :storage
```
tagomoris: I found that this name (only for tests) is a bit misleading, because plugins in v0.14 may have "storage plugin" instances.
Could you rename this? Any names are ok... (e.g., `data`, `received`, `content_options`...)

test/plugin/test_out_forward.rb
```diff
@@ -457,9 +530,24 @@ def test_heartbeat_type_none
     assert_equal node.available, true
   end
 
+  module SuppressCallSuperMixin
```
tagomoris: Please add code comment why this module is needed.

lib/fluent/compat/output.rb
```diff
@@ -111,6 +111,10 @@ module ChunkSizeCompatMixin
       def size
         self.bytesize
       end
+
+      def contains_size
```
tagomoris: `contains_size` looks a method to return boolean for me.
How about `size_of_events` or so?

ganmacs: `size_of_events ` looks good.
fixed it.

https://github.com/fluent/fluentd/pull/1137/commits/8b3f97a2c90680a7d5a8f7820bb33a818c261d0f#diff-ff8e41ed5c655ada05e38acdcee75eb8R115

test/plugin/test_out_forward.rb
ganmacs: changed `received_options`
https://github.com/fluent/fluentd/pull/1137/commits/8b3f97a2c90680a7d5a8f7820bb33a818c261d0f#diff-b12bf2d976d381a19fba652823a95a39R446

test/plugin/test_out_forward.rb
ganmacs: added a comment.
https://github.com/fluent/fluentd/pull/1137/commits/f43c0842a1fd02aaab48a43c7a7a6f0cbc3f561d

tagomoris:
LGTM.
It might be better to add a configuration parameter to use traditional (v0) protocol, ... but I think it's not worth to do it (there are no variant implementation of `in_forward` as far as i know, especially something which cannot handle option).

tagomoris:
I went back to code, and found some points to be fixed. Sorry for late pointing.

tagomoris:
With this change, there are no reason to use `@extend_internal_protocol`. It's just same with `@require_ack_response` by your change. So I think it's better to remove `@extend_internal_protocol` and to use `@require_ack_response` instead of it.

ganmacs:
Removed `@extend_internal_protocol`and began to use `@require_ack_response`  instead of it.

tagomoris:
The code looks clean and straightforward! Great for merge.

Issue #1139, opened 2016-08-04
Add a method to return time and record on filter plugins (instead of #format, with pipeline)


ganmacs:
Add `filter_time_and_record` that returns `record` and `time` to `Fluent::Plugin::Filter` for plugins that updates `time`(e.g [fluent-plugin-parser](https://github.com/tagomoris/fluent-plugin-parser)).
(Existing filter plugins have `Fluent::Plugin::Filter#filter` method that returns `record` untill now.)

If you'll create a new filter plugin, you should implement either `filter` or `filter_time_and_record` method to your new filter plugin.
`filter_time_and_record` that you'll implement should return either `[time, record]` or `nil`.

ganmacs:
@tagomoris 
I think that a method name of new filter that returns `record` and `time`is  `filter_time_and_record`.

What do you think about this method name?

tagomoris:
That method is for plugins which updates time (like fluent-plugin-parser), right? If so, please update description to show use cases.

Naming is always big issue... I like shorter one. How about `filter_with_time`?
@repeatedly any thoughts?

ganmacs:
> That method is for plugins which updates time (like fluent-plugin-parser), right? If so, please update description to show use cases.

thanks. updated!.

repeatedly:
Hmm... this is hard thing.
`filter_with_time` looks like "filter records using time" but I don't have good name idea.

tagomoris:
IMO, `filter` method has arguments of `tag`, `time` and `record`.
So I expect plugin authors to understand that `filter_with_time` means "return value includes time".

If you don't have alternatives, I'll proceed this as is.

tagomoris:
@repeatedly https://github.com/fluent/fluentd/issues/1139#issuecomment-238142242

PR #1140, opened 2016-08-04
Add `filter_with_time` to `Fluent::Plugin::Filter`

ganmacs:
This is for plugins that updates times when filtering.

ref: https://github.com/fluent/fluentd/issues/1139

### DIFF
--- a/lib/fluent/plugin/filter.rb
+++ b/lib/fluent/plugin/filter.rb
@@ -45,27 +45,29 @@ module Fluent
 
       def filter_stream(tag, es)
         new_es = MultiEventStream.new
-        es.each do |time, record|
-          begin
-            filtered_record = do_filter(tag, time, record)
-            new_es.add(*filtered_record) if filtered_record
-          rescue => e
-            router.emit_error_event(tag, time, record, e)
+        if @has_filter_with_time
+          es.each do |time, record|
+            begin
+              filtered_time, filtered_record = filter_with_time(tag, time, record)
+              new_es.add(filtered_time, filtered_record) if filtered_time && filtered_record
+            rescue => e
+              router.emit_error_event(tag, time, record, e)
+            end
+          end
+        else
+          es.each do |time, record|
+            begin
+              filtered_record = filter(tag, time, record)
+              new_es.add(time, filtered_record) if filtered_record
+            rescue => e
+              router.emit_error_event(tag, time, record, e)
+            end
           end
         end
         new_es
       end
 
       private
-
-      def do_filter(tag, time, record)
-        if @has_filter_with_time
-          filter_with_time(tag, time, record)
-        else
-          record = filter(tag, time, record)
-          [time, record] if record
-        end
-      end
 
       def has_filter_with_time?
         implmented_methods = self.class.instance_methods(false)
--- a/lib/fluent/plugin/filter.rb
+++ b/lib/fluent/plugin/filter.rb
@@ -30,7 +30,16 @@ module Fluent
 
       helpers :event_emitter
 
+      def initialize
+        super
+        @has_filter_with_time = has_filter_with_time?
+      end
+
       def filter(tag, time, record)
+        raise NotImplementedError, "BUG: filter plugins MUST implement this method"
+      end
+
+      def filter_with_time(tag, time, record)
         raise NotImplementedError, "BUG: filter plugins MUST implement this method"
       end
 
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/filter.rb
```diff
@@ -28,24 +28,57 @@ class Filter < Base
...
+      def do_filter(tag, time, record)
+        if implemented_method_name == :filter
+          record = filter(tag, time, record)
+          [time, record] if record
+        else
+          filter_with_time(tag, time, record)
+        end
+      end
+
+      def implemented_method_name
```
tagomoris: This process (detecting implemented methods) should be done as early as possible. Lazy call (the first time when events arrived) makes data loss if the filter was implemented in invalid way.
IMO it's best to do it in `#initialize`. Raising error in `#initialize` makes fluentd to stop at the time just launched.

lib/fluent/plugin/filter.rb
```diff
@@ -28,24 +28,57 @@ class Filter < Base
...
           rescue => e
             router.emit_error_event(tag, time, record, e)
           end
         end
         new_es
       end
+
+      private
+
+      def do_filter(tag, time, record)
```
tagomoris: This method is calling `if` every time for `each` blocks.
Of course it makes a small performance overhead (and your code creates `[time, record]` instance every time, which makes one more overhead).
Please make `if` outside of `es.each` and write different code for each methods.
```
if @without_time
  es.each do |time, record|
    begin
      new_es.add(time, filter(tag, time, record))
    rescue => e
      # ..
    end
  end
else
  # ...
end
```

This code looks not so cool, but doesn't create additional objects and doesn't call additional `if` and other methods. It's very important for loops heavily called of data processing middleware.

lib/fluent/plugin/filter.rb
```diff
@@ -28,24 +28,57 @@ class Filter < Base
       include PluginLoggerMixin
       include PluginHelper::Mixin
 
+      class DuplicatedFilterMethodError < StandardError; end
```
tagomoris: Will this error be rescued with this name by any code?
If not, there are no need to create a specific name for each errors, especially for BUG of plugins. See code to raise errors with "BUG: ..." messages in Fluentd.

ganmacs: I fixed by adding "Bug:" prefix to a raised message .
https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R81

lib/fluent/plugin/filter.rb
ganmacs: Because we can control a timing of raising an error and it is useful for test etc ,I added a code that check which method(`filter` or `filter_with_time`) implemented at `#configure` 

https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R34

lib/fluent/plugin/filter.rb
ganmacs: Thanks for your comment. I didn't mind such point.
I fixed.
https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R48

lib/fluent/plugin/filter.rb
```diff
@@ -30,22 +30,64 @@ class Filter < Base
...
+        implmented_methods = self.class.instance_methods(false)
+        # Plugins that override `filter_stream` don't need check,
+        # because they may not call `filter` or `filter_with_time`
+        # for example fluentd/lib/fluent/plugin/filter_record_transformer.rb
+        return if implmented_methods.include?(:filter_stream)
+        case
+        when [:filter, :filter_with_time].all? { |e| implmented_methods.include?(e) }
+          raise "BUG: Filter plugins MUST be implemented either `filter` or `filter_with_time`"
+        when implmented_methods.include?(:filter)
+          :filter
```
repeatedly: This seems verbose. `true` / `false` is enough for me.

lib/fluent/plugin/filter.rb
```diff
@@ -30,22 +30,64 @@ class Filter < Base
...
+              new_es.add(time, filtered_record) if filtered_record
+            rescue => e
+              router.emit_error_event(tag, time, record, e)
+            end
+          end
+        else
+          es.each do |time, record|
+            begin
+              filtered_record_and_time = filter_with_time(tag, time, record)
+              new_es.add(*filtered_record_and_time) if filtered_record_and_time
```
repeatedly: Array expansion seems slower.

```
% ruby b_multi_arg.rb
Rehearsal --------------------------------------
f1   0.640000   0.000000   0.640000 (  0.647215)
f2   0.740000   0.010000   0.750000 (  0.773401)
----------------------------- total: 1.390000sec

f1   0.670000   0.010000   0.680000 (  0.709262)
f2   0.770000   0.000000   0.770000 (  0.815869)
```

```rb
require 'benchmark'

class Foo
  def initialize
    @value = [1, 2]
  end

  def f1(num)
    num.times {
      a, b = @value
      filter(a, b)
    }
  end

  def f2(num)
    num.times {
      as = @value
      filter(*as)
    }
  end

  def filter(a, b)
  end
end

n = 10000000

Benchmark.bmbm do |x|
  x.report('f1') {
    foo = Foo.new
    foo.f1(n)
  }
  x.report('f2') {
    foo = Foo.new
    foo.f2(n)
  }
end
```

lib/fluent/plugin/filter.rb
ganmacs: fixed. 
https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R82

lib/fluent/plugin/filter.rb
ganmacs: fixed.
https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R51

lib/fluent/plugin/filter.rb
```diff
@@ -30,22 +30,62 @@ class Filter < Base
...
       end
+
+      private
+
+      def has_filter_with_time?
+        implmented_methods = self.class.instance_methods(false)
+        # Plugins that override `filter_stream` don't need check,
+        # because they may not call `filter` or `filter_with_time`
+        # for example fluentd/lib/fluent/plugin/filter_record_transformer.rb
+        return if implmented_methods.include?(:filter_stream)
```
tagomoris: Returning without any value looks not good manner for this method... What is set into `@has_filter_with_time?`?
It's better to return `nil` or something else explicitly.
Code comment looks very good.

tagomoris:
@ganmacs Could you fix just one more point?

lib/fluent/plugin/filter.rb
ganmacs: fixed ! 
https://github.com/fluent/fluentd/pull/1140/files#diff-3bdc55bd92825d185fcb6fc41334cef3R77

tagomoris:
Great addition. Thank you!

PR #1141, opened 2016-08-04
Make CI faster for Windows environment

tagomoris:
Currently we are running 6 test jobs on AppVeyer per builds (ruby [2.1, 2.2, 2.3] * [x86, x64]). But it requires 40minutes even with 2 concurrent jobs.

My idea is:
* run full tests just 4 times (2.3-x86, 2.3-x64, 2.2-x64, 2.2-x64)
* create a very short test to confirm that minimal test cases run successfully

Some gems which fluentd depends on are binary gems, which contains some versions of binary for versions of ruby, and architectures of compute (x86/x64). These combinations should be tested.

But running full tests is too much for that purpose.
So I think we can reduce total time of CI tasks by replacing full tests with very short/simple tests for some combinations of environments.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,7 +13,11 @@ Rake::TestTask.new(:base_test) do |t|
   #  $ bundle exec rake base_test TEST=test/test_specified_path.rb
   #  $ bundle exec rake base_test TEST=test/test_*.rb
   t.libs << "test"
-  t.test_files = Dir["test/**/test_*.rb"].sort
+  t.test_files = if ENV["WIN_RAPID"]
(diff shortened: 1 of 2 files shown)


tagomoris:
Now I'm checking `test_script` in build matrix is valid specification or not.

tagomoris:
I confirmed that it's not impossible to overwrite `test_script` per environment matrix. hmm.

tagomoris:
I made it by running very limited tests on selected environment matrix items.

tagomoris:
CI complete in 23minutes
https://ci.appveyor.com/project/fluent/fluentd/build/1137

repeatedly:
LGTM

PR #1147, opened 2016-08-05
Ensure strict mode is used for Oj parser.

edevil:
If we don't use strict mode strings that start with ":" are converted to symbols:
```
irb(main):002:0> Oj.load('{"log":"::1 - - [04/Aug/2016:16:15:13 +0000] \"OPTIONS * HTTP/1.0\" 200 126 \"-\" \"Apache/2.4.10 (Debian) PHP/5.6.24 (internal dummy connection)\"\n","stream":"stdout","time":"2016-08-04T16:15:13.726932176Z"}')
=> {"log"=>:":1 - - [04/Aug/2016:16:15:13 +0000] \"OPTIONS * HTTP/1.0\" 200 126 \"-\" \"Apache/2.4.10 (Debian) PHP/5.6.24 (internal dummy connection)\"\n", "stream"=>"stdout", "time"=>"2016-08-04T16:15:13.726932176Z"}
```
Using strict mode this does not happen:
```
irb(main):003:0> Oj.default_options = {bigdecimal_load: :float, :mode => :strict}
=> {:bigdecimal_load=>:float, :mode=>:strict}
irb(main):004:0> Oj.load('{"log":"::1 - - [04/Aug/2016:16:15:13 +0000] \"OPTIONS * HTTP/1.0\" 200 126 \"-\" \"Apache/2.4.10 (Debian) PHP/5.6.24 (internal dummy connection)\"\n","stream":"stdout","time":"2016-08-04T16:15:13.726932176Z"}')
=> {"log"=>"::1 - - [04/Aug/2016:16:15:13 +0000] \"OPTIONS * HTTP/1.0\" 200 126 \"-\" \"Apache/2.4.10 (Debian) PHP/5.6.24 (internal dummy connection)\"\n", "stream"=>"stdout", "time"=>"2016-08-04T16:15:13.726932176Z"}
```
Obtaining symbols instead of strings causes havoc for some downstream plugins.

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -39,7 +39,7 @@ module Fluent
         begin
           raise LoadError unless @json_parser == 'oj'
           require 'oj'
-          Oj.default_options = {bigdecimal_load: :float}
+          Oj.default_options = {bigdecimal_load: :float, mode: :strict}
           @load_proc = Oj.method(:load)
           @error_class = Oj::ParseError
         rescue LoadError


repeatedly:
Wow, I first new this behaviour.
Good catch!

repeatedly:
Tests are passed so exisiting json parsing works.

tagomoris:
I found that this change changes the process global default behavior, including behavior of `formatter_json`.
Now `formatter_json` raises exception when it dumps values with symbols, with message "Failed to dump Symbol Object to JSON in strict mode.".

tagomoris:
I re-checked the modes of Oj, and found that:
* Oj's default is `object`, to process preceeding ':' characters (and does much more others): http://www.ohler.com/dev/oj_misc/encoding_format.html
* any modes wasn't set for `default_options` in `parser_json`
* `compat` mode was set for `default_options` in `formatter_json`

And, Oj only has process global `default_options`. Boo.
I think we must use `compat` mode, or remove Oj from entire Fluentd.

tagomoris:
I found this problem at here: https://gist.github.com/dimaqq/d65aea5d7e4f3fcb54fcc55747cf363e#file-gistfile1-txt-L159
from: https://github.com/fluent/fluentd/issues/1222#issuecomment-246299627

PR #1151, opened 2016-08-08
Migrate in_monitor_agent to v0.14 API

cosmo0920:
I've migrated in_monitor_agent plugin to v0.14 API.

### Remaining tasks

- [x] Write more test cases for in_monitor_agent

### Not for this PR tasks

* support more plugin classes in in_monitor_agent (e.g. parser and storage plugins support)

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -22,6 +22,7 @@ require 'cool.io'
 
 require 'fluent/plugin/input'
 require 'fluent/plugin/output'
+require 'fluent/plugin/multi_output'
 require 'fluent/plugin/filter'
 
 module Fluent::Plugin
@@ -281,7 +282,7 @@ module Fluent::Plugin
     # from the plugin `pe` recursively
     def self.collect_children(pe, array=[])
       array << pe
-      if pe.is_a?(Fluent::MultiOutput) && pe.respond_to?(:outputs)
+      if pe.is_a?(Fluent::Plugin::MultiOutput) && pe.respond_to?(:outputs)
         pe.outputs.each {|nop|
           collect_children(nop, array)
         }
@@ -295,7 +296,7 @@ module Fluent::Plugin
       matches = Fluent::Engine.root_agent.event_router.instance_variable_get(:@match_rules)
       matches.each { |rule|
         if rule.match?(tag)
-          if rule.collector.is_a?(Fluent::Output)
+          if rule.collector.is_a?(Fluent::Plugin::Output)
             return get_monitor_info(rule.collector, opts)
           end
         end
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -221,14 +221,14 @@ module Fluent::Plugin
       @srv.mount('/api/plugins.json', JSONMonitorServlet, self)
       @srv.mount('/api/config', LTSVConfigMonitorServlet, self)
       @srv.mount('/api/config.json', JSONConfigMonitorServlet, self)
-      thread_create :servlet do
+      thread_create :in_monitor_agent_servlet do
         @srv.start
       end
       if @tag
         log.debug "tag parameter is specified. Emit plugins info to '#{@tag}'"
 
         opts = {with_config: false}
-        timer_execute(:monitor_emit, @emit_interval, repeat: true) {
+        timer_execute(:in_monitor_agent_emit, @emit_interval, repeat: true) {
           es = Fluent::MultiEventStream.new
           now = Fluent::Engine.now
           plugins_info_all(opts).each { |record|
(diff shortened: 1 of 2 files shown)


tagomoris:
More test cases are very good for this change.
But I think it's not time to add new features... it should be done in another following pull-requests.

cosmo0920:
`MultiOutput` class should belong to `output` plugin category?
Or, leave it as-is for now?

tagomoris:
Leave it as-is. It's far different from normal output plugins.

cosmo0920:
I've tried to add test cases for in_monitor_agent.

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -341,10 +292,10 @@ def self.collect_children(pe, array=[])
     # try to match the tag and get the info from the matched output plugin
     # TODO: Support output in label
     def plugin_info_by_tag(tag, opts={})
-      matches = Engine.root_agent.event_router.instance_variable_get(:@match_rules)
+      matches = Fluent::Engine.root_agent.event_router.instance_variable_get(:@match_rules)
       matches.each { |rule|
         if rule.match?(tag)
-          if rule.collector.is_a?(Output)
+          if rule.collector.is_a?(Fluent::Output)
```
tagomoris: `Fluent::Plugin::Output` looks correct instead of `Fluent::Output`(this means v0.12 plugin instances only).

cosmo0920: Oh, I've missed it. 😰 
I'll fix it.

cosmo0920:
I've tried to reduce too strict assertions and add tests.
Please review this PR once again.

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -330,7 +282,7 @@ def all_plugins
     # from the plugin `pe` recursively
     def self.collect_children(pe, array=[])
       array << pe
-      if pe.is_a?(MultiOutput) && pe.respond_to?(:outputs)
```
tagomoris: It is required to collect data for old MultiOutput plugins.
Old (v0.12) MultiOutput plugin are not children of `Fluent::Plugin::MultiOutput`, but `Fluent::Compat::MultiOutput` (and it is child of `Fluent::Plugin::BareOutput`).
fluent-plugin-forest and other plugins are affected with it.

cosmo0920: Oh, I see....
I've added additional commit to resolve your commented problem: https://github.com/fluent/fluentd/pull/1151/commits/03cf62e8d1b15ca90a52b979b2d08ec4d06dd6aa

tagomoris: You should do `pe.is_a?(Fluent::Plugin::MultiOutput) || pe.is_a?(Fluent::MultiOutput)` (both respond to `#otuputs`).

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -248,50 +220,29 @@ def start
       @srv.mount('/api/plugins.json', JSONMonitorServlet, self)
       @srv.mount('/api/config', LTSVConfigMonitorServlet, self)
       @srv.mount('/api/config.json', JSONConfigMonitorServlet, self)
-      @thread = Thread.new {
+      thread_create :servlet do
```
tagomoris: Thread name is process global.
Use thread name with plugin name prefix.

lib/fluent/plugin/in_monitor_agent.rb
tagomoris: @cosmo0920 ping?

cosmo0920: pong. I've pushed a commit to fix this.

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -341,10 +293,10 @@ def self.collect_children(pe, array=[])
     # try to match the tag and get the info from the matched output plugin
     # TODO: Support output in label
     def plugin_info_by_tag(tag, opts={})
-      matches = Engine.root_agent.event_router.instance_variable_get(:@match_rules)
+      matches = Fluent::Engine.root_agent.event_router.instance_variable_get(:@match_rules)
       matches.each { |rule|
         if rule.match?(tag)
-          if rule.collector.is_a?(Output)
+          if rule.collector.is_a?(Fluent::Plugin::Output) || rule.collector.is_a?(Fluent::Output)
```
cosmo0920: I think that we should both `Fluent::Plugin::Output` and `Fluent::Output` as output plugins, right?

tagomoris: Right.

tagomoris:
LGTM.

cosmo0920:
Thanks! :wink:

PR #1152, opened 2016-08-08
Set 0 to umask for keeping v0.12 behaviour. fix #1019

### LINKED ISSUE #1019: unable to connect to unix:///var/run/td-agent/td-agent.sock (Permission denied)
Install td-agent2 and faced the following problem at sending data

- Version: td-agent2 (fluentd 0.14.0)
- OS: CentOS 6.8
- Configuration

```
<source>
  @type unix
  path  /var/run/td-agent/td-agent.sock
</source>
```

- step

1. Install td-agent with "install-redhat-td-agent2.sh"
2. Wrote "/etc/td-agent/td-agent.conf"
3. Update all gems "td-agent-gem update"
4. Apply #1017 (rename in_stream.rb to in_unix.rb)
5. Start process "service td-agent start"

- problem

it is not allowed to write for group and other

**td-agent1**
```
# ls -l /var/run/td-agent/td-agent.sock
srwxrwxrwx 1 td-agent td-agent 0  4月 14 17:52 2016 /var/run/td-agent/td-agent.sock
```
**td-agent2**
```
# ls -l /var/run/td-agent/td-agent.sock
srwxr-xr-x 1 td-agent td-agent 0  6月  3 12:30 2016 /var/run/td-agent/td-agent.sock
```

repeatedly:
Fluentd v0.12 sets `0` to umask but this code is missing in v0.14.

https://github.com/fluent/fluentd/blob/1f7cdffdb7bcae2560e536ec9cd796eff5777fc2/lib/fluent/supervisor.rb#L250

fluentd should pass chumask parameter to ServerEngine. This also fixes #1019 problem.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -240,6 +240,7 @@ module Fluent
           logger_initializer: logger_initializer,
           chuser: chuser,
           chgroup: chgroup,
+          chumask: 0,
           suppress_repeated_stacktrace: suppress_repeated_stacktrace,
           daemonize: daemonize,
           rpc_endpoint: rpc_endpoint,


repeatedly:
Re-build appveyor build.
If there is no concern, I will merge this patch.

PR #1154, opened 2016-08-10
Create new out file plugins only for secondary

### LINKED ISSUE #873: Special plugin only to dump buffer contents to file
Many users are using `out_file` to dump buffer contents with `<secondary>` section when buffered output plugins are failing flush.
But `out_file` is too complex and with too many features for such purpose. We cannot assure anything if users use complex configuration in `<secondary>` section.

I propose a new plugin only for `<secondary>` section just to dump buffer contents to a path on local host, with just few, minimum # of options/parameters.

To be considered:
* Is it enough to support local disk only? any others?
* Can we re-format these records in any specific format to make it possible to extract/read dumped files by a utility command?

ganmacs:
### Consider points

* Plugin's name 
This plugins's name is `out_secondary_file` now.
It may be too verbose.

* Supporting params 
I think supporting `compress` and `append` params.

### Not for this PR tasks

* Create a utility command that read dumped data

Close #873

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -55,7 +55,11 @@ module Fluent
       # if chunk size (or records) is 95% or more after #write, then that chunk will be enqueued
       config_param :chunk_full_threshold, :float, default: DEFAULT_CHUNK_FULL_THRESHOLD
 
-      Metadata = Struct.new(:timekey, :tag, :variables)
+      Metadata = Struct.new(:timekey, :tag, :variables) do
+        def nil?
+          timekey.nil? && tag.nil? && variables.nil?
+        end
+      end
 
       # for tests
       attr_accessor :stage_size, :queue_size
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -56,7 +56,7 @@ module Fluent
       config_param :chunk_full_threshold, :float, default: DEFAULT_CHUNK_FULL_THRESHOLD
 
       Metadata = Struct.new(:timekey, :tag, :variables) do
-        def nil?
+        def empty?
           timekey.nil? && tag.nil? && variables.nil?
         end
       end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -49,6 +49,7 @@ module Fluent
         def initialize(metadata)
           super()
           @unique_id = generate_unique_id
+          @chunk_id = Fluent::UniqueId.hex(@unique_id)
           @metadata = metadata
 
           # state: unstaged/staged/queued/closed
@@ -59,7 +60,7 @@ module Fluent
           @modified_at = Time.now
         end
 
-        attr_reader :unique_id, :metadata, :created_at, :modified_at, :state
+        attr_reader :unique_id, :chunk_id, :metadata, :created_at, :modified_at, :state
 
         # data is array of formatted record string
         def append(data)
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -49,7 +49,7 @@ module Fluent
         def initialize(metadata)
           super()
           @unique_id = generate_unique_id
-          @chunk_id = Fluent::UniqueId.hex(@unique_id)
+          @chunk_id = dump_unique_id_hex(@unique_id)
           @metadata = metadata
 
           # state: unstaged/staged/queued/closed
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -49,7 +49,6 @@ module Fluent
         def initialize(metadata)
           super()
           @unique_id = generate_unique_id
-          @chunk_id = dump_unique_id_hex(@unique_id)
           @metadata = metadata
 
           # state: unstaged/staged/queued/closed
@@ -60,7 +59,7 @@ module Fluent
           @modified_at = Time.now
         end
 
-        attr_reader :unique_id, :chunk_id, :metadata, :created_at, :modified_at, :state
+        attr_reader :unique_id, :metadata, :created_at, :modified_at, :state
 
         # data is array of formatted record string
         def append(data)
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -54,13 +54,7 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "'path' parameter is required on secondary file output"
       end
 
-      if pos = @path.index('*')
-        @path_prefix = @path[0,pos]
-        @path_suffix = @path[pos+1..-1]
-      else
-        @path_prefix = @path + "."
-        @path_suffix = ".log"
-      end
+      configure_path!
 
       test_path = generate_path(Time.now.strftime("%Y%m%d"))
       unless Fluent::FileUtil.writable_p?(test_path)
@@ -70,12 +64,11 @@ module Fluent::Plugin
       @dir_perm = system_config.dir_permission || DIR_PERMISSION
       @file_perm = system_config.file_permission || FILE_PERMISSION
 
-      # @tempalte = genereate_tempalte fix
       super
     end
 
     def write(chunk)
-      id = extract_placeholders('', chunk)
+      id = extract_placeholders(@path, chunk)
       path = generate_path(id)
       FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
@@ -107,25 +100,38 @@ module Fluent::Plugin
 
     private
 
-    def genereate_tempalte
-      rvalue = ""
-      if @chunk_key_tag
-        rvalue += "{tag}_"
+    def configure_path!
+      matched = @path.scan(/\${([\w\d.@-]+)}/).flat_map(&:itself)
+      p matched
+      if matched.empty?
+        if pos = @path.index('*')
+          @path_prefix = @path[0,pos]
+          @path_suffix = @path[pos+1..-1]
+        else
+          @path_prefix = @path + "."
+          @path_suffix = ".log"
+        end
+      else
+        validate_path_is_comptible_with_primary_buffer?(matched)
+        @path_prefix = ""
+        @path_suffix = ".log"
       end
+    end
 
-      if @chunk_key_time
-        rvalue += "%Y%m%d_"
+    def validate_path_is_comptible_with_primary_buffer?(matched)
+      raise "TimeFormat is not imcompatible with primary buffer's params" if !@chunk_key_time && @path != Time.now.strftime(@path)
+      matched.each do |e|
+        case
+        when @chunk_key_tag && e =~ /tag\w*/
+          # nothing
+        when !@chunk_key_tag && e =~ /tag\w*/
+          raise "#{e} is not imcompatible with primary buffer's params"
+        when @chunk_keys.include?(e)
+          # nothing
+        else
+          raise "#{e} is not imcompatible with primary buffer's params"
+        end
       end
-
-      @chunk_keys.each do |v|
-        rvalue += "${#{v}}_"
-      end
-
-      if rvalue.end_with?("_")
-        rvalue = rvalue[0..-2]
-      end
-
-      rvalue
     end
 
     def suffix
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -101,7 +101,7 @@ module Fluent::Plugin
     private
 
     def configure_path!
-      matched = @path.scan(/\${([\w.@-]+)}/).flat_map { |e| e } # for suporting 2.1 or less
+      matched = @path.scan(/\${([\w.@-]+(\[\d+\])?)}/).flat_map(&:first) # to trim suffix [\d+]
       if matched.empty? && !path_has_time_format?
         if pos = @path.index('*')
           @path_prefix = @path[0, pos]
@@ -121,9 +121,9 @@ module Fluent::Plugin
       raise "TimeFormat is not imcompatible with primary buffer's params" if !@chunk_key_time && path_has_time_format?
       matched.each do |e|
         case
-        when @chunk_key_tag && e =~ /tag\w*/
+        when @chunk_key_tag && e =~ /tag(\[\d+\])?/
           # ok
-        when !@chunk_key_tag && e =~ /tag\w*/
+        when !@chunk_key_tag && e =~ /tag(\[\d+\])?/
           raise "#{e} is not imcompatible with primary buffer's params"
         when @chunk_keys.include?(e.to_sym)
           # ok
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -40,6 +40,7 @@ module Fluent::Plugin
       end
 
       @placeholders = @path.scan(/\${([\w.@-]+(\[\d+\])?)}/).flat_map(&:first) # to trim suffix [\d+]
+      @unique_id = dump_unique_id_hex(generate_unique_id)
 
       validate_path_is_comptible_with_primary_buffer!
 
@@ -92,7 +93,7 @@ module Fluent::Plugin
 
     def generate_id(chunk)
       if chunk.metadata.empty?
-        dump_unique_id_hex(chunk.unique_id)
+        @append ? @unique_id : dump_unique_id_hex(chunk.unique_id)
       else
         extract_placeholders(@path, chunk.metadata)
       end
@@ -115,7 +116,6 @@ module Fluent::Plugin
         end
       end
     end
-
 
     def path_has_time_format?
       @path != Time.now.strftime(@path)
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -118,17 +118,17 @@ module Fluent::Plugin
     end
 
     def validate_path_is_comptible_with_primary_buffer?(matched)
-      raise "TimeFormat is not imcompatible with primary buffer's params" if !@chunk_key_time && path_has_time_format?
+      raise "BUG: file path has imcompatible placeholder: Time" if !@chunk_key_time && path_has_time_format?
       matched.each do |e|
         case
         when @chunk_key_tag && e =~ /tag(\[\d+\])?/
           # ok
         when !@chunk_key_tag && e =~ /tag(\[\d+\])?/
-          raise "#{e} is not imcompatible with primary buffer's params"
-        when @chunk_keys.include?(e.to_sym)
+          raise "BUG: file path has imcompatible placeholder: #{e}"
+        when @chunk_keys && @chunk_keys.include?(e.to_sym)
           # ok
         else
-          raise "#{e} is not imcompatible with primary buffer's params"
+          raise "BUG: file path has imcompatible placeholder: #{e}"
         end
       end
     end
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -29,7 +29,7 @@ module Fluent::Plugin
 
     desc "The directory path of the output file."
     config_param :directory, :string
-    desc "The baseanme of the output file."
+    desc "The basename of the output file."
     config_param :basename, :string, default: "dump.bin"
     desc "The flushed chunk is appended to existence file or not."
     config_param :append, :bool, default: false
@@ -42,6 +42,10 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
       end
 
+      if @basename.include?("/")
+        raise Fluent::ConfigError, "basename should not include `/`"
+      end
+
       @path_without_suffix = File.join(@directory, @basename)
       validate_compatible_with_primary_buffer!(@path_without_suffix)
 
@@ -52,7 +56,7 @@ module Fluent::Plugin
                   ".gz"
                 end
 
-      test_path = generate_path(File.join(@directory, Time.now.strftime("%Y%m%d")))
+      test_path = @path_without_suffix
       unless Fluent::FileUtil.writable_p?(test_path)
         raise Fluent::ConfigError, "out_secondary_file: `#{@directory}` should be writable"
       end
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -25,12 +25,15 @@ module Fluent::Plugin
 
     FILE_PERMISSION = 0644
     DIR_PERMISSION = 0755
+    PLACEHOLDER_REGEX = /\${(tag(\[\d+\])?|[\w.@-]+)}/
 
-    desc "The Path of the file."
-    config_param :path, :string
+    desc "The directory path of the output file."
+    config_param :directory, :string
+    desc "The baseanme of the output file."
+    config_param :basename, :string, default: "dump.bin"
     desc "The flushed chunk is appended to existence file or not."
     config_param :append, :bool, default: false
-    config_param :compress, :enum, list: [:text, :gz, :gzip], default: :text
+    config_param :compress, :enum, list: [:text, :gzip], default: :text
 
     def configure(conf)
       super
@@ -39,27 +42,19 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
       end
 
-      @placeholders = @path.scan(/\${([\w.@-]+(\[\d+\])?)}/).flat_map(&:first) # to trim suffix [\d+]
-      @unique_id = dump_unique_id_hex(generate_unique_id)
+      @path_without_suffix = File.join(@directory, @basename)
+      validate_compatible_with_primary_buffer!(@path_without_suffix)
 
-      validate_path_is_compatible_with_primary_buffer!
+      @suffix = case @compress
+                when :text
+                  ""
+                when :gzip
+                  ".gz"
+                end
 
-      if !@placeholders.empty? || path_has_time_format?
-        @path_prefix = ""
-        @path_suffix = ".log"
-      else
-        if pos = @path.index('*')
-          @path_prefix = @path[0, pos]
-          @path_suffix = @path[pos+1..-1]
-        else
-          @path_prefix = @path + "."
-          @path_suffix = ".log"
-        end
-      end
-
-      test_path = generate_path(Time.now.strftime("%Y%m%d"))
+      test_path = generate_path(File.join(@directory, Time.now.strftime("%Y%m%d")))
       unless Fluent::FileUtil.writable_p?(test_path)
-        raise Fluent::ConfigError, "out_secondary_file: `#{test_path}` is not writable"
+        raise Fluent::ConfigError, "out_secondary_file: `#{@directory}` should be writable"
       end
 
       @dir_perm = system_config.dir_permission || DIR_PERMISSION
@@ -67,8 +62,8 @@ module Fluent::Plugin
     end
 
     def write(chunk)
-      base = generate_basename(chunk)
-      path = generate_path(base)
+      path_without_suffix = extract_placeholders(@path_without_suffix, chunk.metadata)
+      path = generate_path(path_without_suffix)
       FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
       case @compress
@@ -77,7 +72,7 @@ module Fluent::Plugin
           f.flock(File::LOCK_EX)
           chunk.write_to(f)
         }
-      when :gz, :gzip
+      when :gzip
         File.open(path, "ab", @file_perm) {|f|
           f.flock(File::LOCK_EX)
           gz = Zlib::GzipWriter.new(f)
@@ -91,54 +86,37 @@ module Fluent::Plugin
 
     private
 
-    def generate_basename(chunk)
-      if chunk.metadata.empty?
-        @append ? @unique_id : dump_unique_id_hex(chunk.unique_id)
-      else
-        extract_placeholders(@path, chunk.metadata)
+    def validate_compatible_with_primary_buffer!(path_without_suffix)
+      placeholders = path_without_suffix.scan(PLACEHOLDER_REGEX).flat_map(&:first) # to trim suffix [\d+]
+
+      if !@chunk_key_time && has_time_format?(path_without_suffix)
+        raise Fluent::ConfigError, "out_secondary_file: basename or directory has an incompatible placeholder, remove time formats, like `%Y%m%d`, from basename or directory"
+      end
+
+      if !@chunk_key_tag && (ph = placeholders.find { |placeholder| placeholder.match(/tag(\[\d+\])?/) })
+        raise Fluent::ConfigError, "out_secondary_file: basename or directory has an incompatible placeholder #{ph}, remove tag placeholder, like `${tag}`, from basename or directory"
+      end
+
+      vars = placeholders.reject { |placeholder| placeholder.match(/tag(\[\d+\])?/) }
+
+      if ph = vars.find { |v| !@chunk_keys.include?(v) }
+        raise Fluent::ConfigError, "out_secondary_file: basename or directory has an incompatible placeholder #{ph}, remove variable placeholder, like `${varname}`, from basename or directory"
       end
     end
 
-    def validate_path_is_compatible_with_primary_buffer!
-      if !@chunk_key_time && path_has_time_format?
-        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder, add time formats, like `%Y%m%d`, to `path`"
-      end
-
-      if !@chunk_key_tag && (ph = @placeholders.find { |p| p.match(/tag(\[\d+\])?/) })
-        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder #{ph}, add tag placeholder, like `${tag}`, to `path`"
-      end
-
-      if @chunk_keys.empty?
-        vars = @placeholders.reject { |p| p.match(/tag(\[\d+\])?/) }
-
-        if ph = vars.find { |v| !@chunk_keys.include?(v) }
-          raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder #{ph}, add variable placeholder, like `${varname}`, to `path`"
-        end
-      end
+    def has_time_format?(str)
+      str != Time.now.strftime(str)
     end
 
-    def path_has_time_format?
-      @path != Time.now.strftime(@path)
-    end
-
-    def suffix
-      case @compress
-      when :text
-        ""
-      when :gz, :gzip
-        ".gz"
-      end
-    end
-
-    def generate_path(base)
+    def generate_path(path_without_suffix)
       if @append
-        "#{@path_prefix}#{base}#{@path_suffix}#{suffix}"
+        "#{path_without_suffix}#{@suffix}"
       else
         i = 0
         loop do
-          path = "#{@path_prefix}#{base}_#{i}#{@path_suffix}#{suffix}"
+          path = "#{path_without_suffix}.#{i}#{@suffix}"
           return path unless File.exist?(path)
-          i += 1;
+          i += 1
         end
       end
     end
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -101,11 +101,10 @@ module Fluent::Plugin
     private
 
     def configure_path!
-      matched = @path.scan(/\${([\w\d.@-]+)}/).flat_map(&:itself)
-      p matched
-      if matched.empty?
+      matched = @path.scan(/\${([\w.@-]+)}/).flat_map { |e| e } # for suporting 2.1 or less
+      if matched.empty? && !path_has_time_format?
         if pos = @path.index('*')
-          @path_prefix = @path[0,pos]
+          @path_prefix = @path[0, pos]
           @path_suffix = @path[pos+1..-1]
         else
           @path_prefix = @path + "."
@@ -119,19 +118,23 @@ module Fluent::Plugin
     end
 
     def validate_path_is_comptible_with_primary_buffer?(matched)
-      raise "TimeFormat is not imcompatible with primary buffer's params" if !@chunk_key_time && @path != Time.now.strftime(@path)
+      raise "TimeFormat is not imcompatible with primary buffer's params" if !@chunk_key_time && path_has_time_format?
       matched.each do |e|
         case
         when @chunk_key_tag && e =~ /tag\w*/
-          # nothing
+          # ok
         when !@chunk_key_tag && e =~ /tag\w*/
           raise "#{e} is not imcompatible with primary buffer's params"
-        when @chunk_keys.include?(e)
-          # nothing
+        when @chunk_keys.include?(e.to_sym)
+          # ok
         else
           raise "#{e} is not imcompatible with primary buffer's params"
         end
       end
+    end
+
+    def path_has_time_format?
+      @path != Time.now.strftime(@path)
     end
 
     def suffix
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -125,7 +125,7 @@ module Fluent::Plugin
           # ok
         when !@chunk_key_tag && e =~ /tag(\[\d+\])?/
           raise "BUG: file path has imcompatible placeholder: #{e}"
-        when @chunk_keys && @chunk_keys.include?(e.to_sym)
+        when @chunk_keys && @chunk_keys.include?(e)
           # ok
         else
           raise "BUG: file path has imcompatible placeholder: #{e}"
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -42,7 +42,7 @@ module Fluent::Plugin
       @placeholders = @path.scan(/\${([\w.@-]+(\[\d+\])?)}/).flat_map(&:first) # to trim suffix [\d+]
       @unique_id = dump_unique_id_hex(generate_unique_id)
 
-      validate_path_is_comptible_with_primary_buffer!
+      validate_path_is_compatible_with_primary_buffer!
 
       if !@placeholders.empty? || path_has_time_format?
         @path_prefix = ""
@@ -67,8 +67,8 @@ module Fluent::Plugin
     end
 
     def write(chunk)
-      id = generate_id(chunk)
-      path = generate_path(id)
+      base = generate_basename(chunk)
+      path = generate_path(base)
       FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
       case @compress
@@ -91,7 +91,7 @@ module Fluent::Plugin
 
     private
 
-    def generate_id(chunk)
+    def generate_basename(chunk)
       if chunk.metadata.empty?
         @append ? @unique_id : dump_unique_id_hex(chunk.unique_id)
       else
@@ -99,7 +99,7 @@ module Fluent::Plugin
       end
     end
 
-    def validate_path_is_comptible_with_primary_buffer!
+    def validate_path_is_compatible_with_primary_buffer!
       if !@chunk_key_time && path_has_time_format?
         raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder, add time formats, like `%Y%m%d`, to `path`"
       end
@@ -130,13 +130,13 @@ module Fluent::Plugin
       end
     end
 
-    def generate_path(id)
+    def generate_path(base)
       if @append
-        "#{@path_prefix}#{id}#{@path_suffix}#{suffix}"
+        "#{@path_prefix}#{base}#{@path_suffix}#{suffix}"
       else
         i = 0
         loop do
-          path = "#{@path_prefix}#{id}_#{i}#{@path_suffix}#{suffix}"
+          path = "#{@path_prefix}#{base}_#{i}#{@path_suffix}#{suffix}"
           return path unless File.exist?(path)
           i += 1;
         end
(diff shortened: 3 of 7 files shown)


lib/fluent/plugin/buffer.rb
```diff
@@ -55,7 +55,11 @@ class BufferChunkOverflowError < BufferError; end # A record size is larger than
       # if chunk size (or records) is 95% or more after #write, then that chunk will be enqueued
       config_param :chunk_full_threshold, :float, default: DEFAULT_CHUNK_FULL_THRESHOLD
 
-      Metadata = Struct.new(:timekey, :tag, :variables)
+      Metadata = Struct.new(:timekey, :tag, :variables) do
+        def nil?
```
tagomoris: Ruby's document says that "Only the object nil responds true to nil?".
This method breaks it. Change this method name to something else, like `empty?`.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -49,6 +49,7 @@ class Chunk
         def initialize(metadata)
           super()
           @unique_id = generate_unique_id
+          @chunk_id = Fluent::UniqueId.hex(@unique_id)
```
tagomoris: This class includes `UniqueId::Mixin`, so we can use `#dump_unique_id_hex` method here.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -59,7 +60,7 @@ def initialize(metadata)
           @modified_at = Time.now
         end
 
-        attr_reader :unique_id, :metadata, :created_at, :modified_at, :state
+        attr_reader :unique_id, :chunk_id, :metadata, :created_at, :modified_at, :state
```
tagomoris: What's the `chunk_id`?

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+    config_param :path, :string
+    desc "The flushed chunk is appended to existence file or not."
+    config_param :append, :bool, default: false
+    config_param :compress, default: nil do |val|
+      SUPPORTED_COMPRESS[val].tap do |e|
+        raise ConfigError, "Unsupported compression algorithm '#{val}'" unless e
+      end
+    end
+
+    def initialize
```
tagomoris: Remove useless implementation.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+      super
+    end
+
+    def configure(conf)
+      unless @as_secondary
+        raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
+      end
+
+      unless @path = conf["path"]
```
tagomoris: This check code is not required: `config_param` parameters without `default` argument are required parameters. Configuration error occurs if these are missing.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      test_path = generate_path(Time.now.strftime("%Y%m%d"))
+      unless Fluent::FileUtil.writable_p?(test_path)
+        raise Fluent::ConfigError, "out_file: `#{test_path}` is not writable"
+      end
+
+      @dir_perm = system_config.dir_permission || DIR_PERMISSION
+      @file_perm = system_config.file_permission || FILE_PERMISSION
+
+      super
```
tagomoris: `Fluent::Configurable#configure` does assignment values into `@parameter` instance variables.
So calling `super` should be done in early part of plugins' `#configure` method in normal.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      unless @path = conf["path"]
+        raise Fluent::ConfigError, "'path' parameter is required on secondary file output"
+      end
+
+      configure_path!
+
+      test_path = generate_path(Time.now.strftime("%Y%m%d"))
+      unless Fluent::FileUtil.writable_p?(test_path)
+        raise Fluent::ConfigError, "out_file: `#{test_path}` is not writable"
```
tagomoris: This plugin is not `out_file`.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    def write(chunk)
+      id = extract_placeholders(@path, chunk)
+      path = generate_path(id)
+      FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
+
+      case @compress
+      when nil
+        File.open(path, "ab", @file_perm) {|f|
+          f.flock(File::LOCK_EX)
```
tagomoris: I just wan to confirm: is this lock freed when `f` is closed?

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+          gz = Zlib::GzipWriter.new(f)
+          chunk.write_to(gz)
+          gz.close
+        }
+      end
+
+      path
+    end
+
+    def extract_placeholders(str, chunk)
```
tagomoris: This method does things that is different from `Output#extract_placeholders`. So method name must be different from it.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+      if chunk.metadata.nil?
+        chunk.chunk_id
+      else
+        super(str, chunk.metadata)
+      end
+    end
+
+    private
+
+    def configure_path!
```
tagomoris: What does this method do?

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    FILE_PERMISSION = 0644
+    DIR_PERMISSION = 0755
+
+    desc "The Path of the file."
+    config_param :path, :string
+    desc "The flushed chunk is appended to existence file or not."
+    config_param :append, :bool, default: false
+    config_param :compress, default: nil do |val|
```
tagomoris: use `:enum` type for fixed list of symbols instead of such code

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,162 @@
...
+      if matched.empty? && !path_has_time_format?
+        if pos = @path.index('*')
+          @path_prefix = @path[0, pos]
+          @path_suffix = @path[pos+1..-1]
+        else
+          @path_prefix = @path + "."
+          @path_suffix = ".log"
+        end
+      else
+        validate_path_is_comptible_with_primary_buffer?(matched)
```
tagomoris: Normally `?` suffixed method returns true/false, not for raising errors.

lib/fluent/plugin/output.rb
```diff
@@ -185,13 +185,26 @@ def initialize
...
         @chunk_keys = @chunk_key_time = @chunk_key_tag = nil
         @flush_mode = nil
+        @timekey_zone = nil
       end
 
       def acts_as_secondary(primary)
         @as_secondary = true
         @primary_instance = primary
+
+        @chunk_keys = @primary_instance.chunk_keys if @primary_instance.chunk_keys
```
tagomoris: `@chunk_keys` will not be initialized when `@primary_instance.chunk_keys` is not specified. It will be warned.

lib/fluent/plugin/output.rb
```diff
@@ -185,13 +185,26 @@ def initialize
...
+        @chunk_key_tag = @primary_instance.chunk_key_tag if @primary_instance.chunk_key_tag
```
tagomoris: ditto.

tagomoris:
@ganmacs I reviewed at once. There are some points which must be fixed.
Could you update this change? I'll review again in detail after that.

lib/fluent/plugin/output.rb
ganmacs: I changed this method name to `empty?`

https://github.com/fluent/fluentd/pull/1154/commits/3a341875c87c5c1bc110b70abf1bddddff0895a9

lib/fluent/plugin/buffer.rb
ganmacs: I changed this method name `empty?`
https://github.com/fluent/fluentd/pull/1154/files#diff-7d794f77abc9222413b9120533cff327R59

lib/fluent/plugin/buffer/chunk.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/commits/920b008d96be6e03992974cb8bb7c7139fc4dca0

lib/fluent/plugin/buffer/chunk.rb
ganmacs: It is added to use in `OutSecondary` ( https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R84 )

lib/fluent/plugin/out_secondary_file.rb
ganmacs: I changed calling `super`  in early of `#configure' method.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R36

lib/fluent/plugin/out_secondary_file.rb
ganmacs: I changed this method to `generate_id`.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R82

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.

https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R33

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.

https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R109

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.

https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R50

lib/fluent/plugin/output.rb
ganmacs: https://github.com/fluent/fluentd/pull/1154/commits/654b9b00d332a838e2dbb92045782f62ab861e14

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Thanks. I don't know that.
Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R35

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Yes, it is.
http://docs.ruby-lang.org/ja/2.3.0/method/File/i/flock.html (Sorry, This is Japanse article. I can't file English one)
It describes that free locked object by clos 

の明示的なアンロック以外に、ファイルのcloseやRubyインタプリタの終了 (プロセスの終了)によっても自動的にロック状態は解除されます。

lib/fluent/plugin/buffer/chunk.rb
tagomoris: Ah, i missed it... but it was used just once in newly added code.
So it seems not worth to add. If once it'll be added, it becomes a member of public API, and we must maintain it.
(In addition, adding stringified unique_id is bad idea because any internal code might change unique_id after instanciated. So it should be a `#chunk_id` method to return `dump_unique_id_hex(@unique_id)` at least.)

lib/fluent/plugin/out_secondary_file.rb
tagomoris: Okay, good.
http://docs.ruby-lang.org/en/2.3.0/File.html#method-i-flock

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,308 @@
...
+  class DummyOutput < Fluent::Plugin::Output
+    def write(chunk); end
+  end
+
+  def create_primary(buffer_cofig = config_element('buffer'))
+    DummyOutput.new.configure(config_element('ROOT','',{}, [buffer_cofig]))
+  end
+
+  def create_driver(conf = CONFIG, primary = create_primary)
+    c = Fluent::Test::OutputTestDriver.new(Fluent::Plugin::SecondaryFileOutput)
```
tagomoris: `Fluent::Test::OutputTestDriver` is for plugins on v0.12 (old) APIs.
Use `fluent/test/driver/output` instead.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+          raise "BUG: file path has imcompatible placeholder: #{e}"
+        when @chunk_keys && @chunk_keys.include?(e)
+        # ok
+        else
+          raise "BUG: file path has imcompatible placeholder: #{e}"
+        end
+      end
+    end
+
+    def path_has_tags(matched)
```
tagomoris: This method is used just once. Is it worth to add?

lib/fluent/plugin/out_secondary_file.rb
tagomoris: That method (renamed to `generate_id`) returns the result of `extract_placeholders`.
Is it an id? For what?

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+    config_param :compress, :enum, list: [:normal, :gz, :gzip], default: :normal
+
+    def configure(conf)
+      super
+
+      unless @as_secondary
+        raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
+      end
+
+      configure_path!
```
tagomoris: `configure_path!` doesn't show what this method does. It's bad practice to use such name.
(code below L44-47 is also about path, isn't it?)

Use descriptive method name, or write code here directly (if it's enough short).

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+      when :normal
+        ""
+      when :gz, :gzip
+        ".gz"
+      end
+    end
+
+    def generate_path(id)
+      if @append
+        "#{@path_prefix}#{id}#{@path_suffix}#{suffix}"
```
tagomoris: If `chunk.metadata` is empty, `id` becomes chunk unique id.
So this path is always unique, and never be appended any other chunk contents. Is it intentional?

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+          @path_suffix = @path[pos+1..-1]
+        else
+          @path_prefix = @path + "."
+          @path_suffix = ".log"
+        end
+      end
+    end
+
+    def validate_path_is_comptible_with_primary_buffer!(matched)
+      raise "BUG: file path has imcompatible placeholder: Time" if !@chunk_key_time && path_has_time_format?
```
tagomoris: It should be a kind of Configuration error, not BUG of code.

tagomoris: And configuration error message should guide users how to correct it, or show what's wrong in configuration clearly.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+      matched.each do |e|
```
tagomoris: Is it needed to use loop & case? In configuration stage, small loops (inefficient code) doesn't matter in any way.
It looks enough to do:
* `@chunk_key_tag` xor `placeholders.any?{|placeholder| placeholder.matches(/tag pattern/) }`
* `@chunk_key_time` xor `has_time_format?(path)`
* `@chunk_keys` == rest placeholders (excluding tags)

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,151 @@
...
+    Fluent::Plugin.register_output("secondary_file", self)
+
+    FILE_PERMISSION = 0644
+    DIR_PERMISSION = 0755
+
+    desc "The Path of the file."
+    config_param :path, :string
+    desc "The flushed chunk is appended to existence file or not."
+    config_param :append, :bool, default: false
+    config_param :compress, :enum, list: [:normal, :gz, :gzip], default: :normal
```
tagomoris: Refer & follow existing name for configuration value means "uncompressed":
https://github.com/fluent/fluent-plugin-webhdfs/blob/master/lib/fluent/plugin/out_webhdfs.rb#L88
https://github.com/fluent/fluent-plugin-s3/blob/master/lib/fluent/plugin/out_s3.rb#L416

lib/fluent/plugin/buffer/chunk.rb
ganmacs: Remove it. and use `dump_unique_id_hex` directly.
https://github.com/fluent/fluentd/pull/1154/commits/d7386e081fe0964dbac57f97d406a16275d7bc92

test/plugin/test_out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/commits/52b83bc66b45e9dae34b021ce42f44cb3087a608

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Remove this method.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R46

lib/fluent/plugin/out_secondary_file.rb
ganmacs: I wrote this method body here directly.

https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R46

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed error message.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R103

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R101

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/commits/258fec1d970ec026b8a22fdb855b8d7fce190719

lib/fluent/plugin/out_secondary_file.rb
ganmacs: No, I forgot it. So I changed logic as below.

if `chunk.metadata` is empty and `append` is true, use `unique_id` that secondary_file has to add `chunk`s in the same file,


https://github.com/fluent/fluentd/pull/1154/commits/1314d04eaa49638fcdeab7ce0a00c7d7e795564b

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,146 @@
...
+      super
+
+      unless @as_secondary
+        raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
+      end
+
+      @placeholders = @path.scan(/\${([\w.@-]+(\[\d+\])?)}/).flat_map(&:first) # to trim suffix [\d+]
+      @unique_id = dump_unique_id_hex(generate_unique_id)
+
+      validate_path_is_comptible_with_primary_buffer!
```
tagomoris: Typo: `comptible` -> `compatible`

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,146 @@
...
+      unless Fluent::FileUtil.writable_p?(test_path)
+        raise Fluent::ConfigError, "out_secondary_file: `#{test_path}` is not writable"
+      end
+
+      @dir_perm = system_config.dir_permission || DIR_PERMISSION
+      @file_perm = system_config.file_permission || FILE_PERMISSION
+    end
+
+    def write(chunk)
+      id = generate_id(chunk)
```
tagomoris: As I mentioned before, `generate_id` actually returns path with extracted placeholders.
https://github.com/fluent/fluentd/pull/1154#discussion_r75439633
It's not `id`.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,146 @@
...
+        if pos = @path.index('*')
+          @path_prefix = @path[0, pos]
+          @path_suffix = @path[pos+1..-1]
+        else
+          @path_prefix = @path + "."
+          @path_suffix = ".log"
+        end
+      end
+
+      test_path = generate_path(Time.now.strftime("%Y%m%d"))
```
tagomoris: Aren't there any problem if `@path` has placeholders?

lib/fluent/test/base.rb
```diff
@@ -37,7 +37,7 @@ def initialize(klass, &block)
         else
           @instance = klass
         end
-        @instance.router = Engine.root_agent.event_router
+        @instance.router = Engine.root_agent.event_router if @instance.respond_to?(:router=)
```
tagomoris: This change is not needed anymore (And v0.12 plugins always has `#router=`).

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.

https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R45

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Sorry, I overlooked it. I changed a name of this method to `generate_basename`
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R70

lib/fluent/plugin/out_secondary_file.rb
ganmacs: No, it just checks that directory is writable or not. So the contents of `@path` doesn't matter.
(the contents of `@path` are validated in [here]( https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R45) )

lib/fluent/test/base.rb
ganmacs: I removed it.
https://github.com/fluent/fluentd/pull/1154/commits/3877b3ac9ae200ae6a55725d6580bb806c380cd4#diff-11193d96f15fd472415e15b1c787938eL40

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+module Fluent::Plugin
+  class SecondaryFileOutput < Output
+    Fluent::Plugin.register_output("secondary_file", self)
+
+    FILE_PERMISSION = 0644
+    DIR_PERMISSION = 0755
+    PLACEHOLDER_REGEX = /\${(tag(\[\d+\])?|[\w.@-]+)}/
+
+    desc "The directory path of the output file."
+    config_param :directory, :string, default: "secondary/"
```
tagomoris: This default value doesn't work well for daemonized Fluentd process (`cd /` is basic technique to daemonize processes).

tagomoris: IMO it's good for this parameter not to have default values.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+      super
+
+      unless @as_secondary
+        raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
+      end
+
+      @placeholders = @basename.scan(PLACEHOLDER_REGEX).flat_map(&:first) # to trim suffix [\d+]
+      validate_basename_is_compatible_with_primary_buffer!
+
+      @directory += "/" if @directory.size != 0 && !@directory.end_with?("/")
```
tagomoris: Use `File.join` to concat directory and path for make it better to support Windows environment.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+      unless Fluent::FileUtil.writable_p?(test_path)
+        raise Fluent::ConfigError, "out_secondary_file: `#{test_path}` is not writable"
+      end
+
+      @dir_perm = system_config.dir_permission || DIR_PERMISSION
+      @file_perm = system_config.file_permission || FILE_PERMISSION
+    end
+
+    def write(chunk)
+      name = extract_placeholders(@basename, chunk.metadata)
```
tagomoris: With this code, users cannot use placeholders in directory. Is it intentional?
I think it's very useful to separate files per tags or days like this `/path/to/backup/${tag}/dump.bin`.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+          gz = Zlib::GzipWriter.new(f)
+          chunk.write_to(gz)
+          gz.close
+        }
+      end
+
+      path
+    end
+
+    # To avoid calling Fluent::Compat::TimeSlicedOutput#extract_placeholders if primary plugin is `out_file`
```
tagomoris: Very good comment!
Just a fix needed: "if primary plugin is `out_file`" -> "if primary plugin is a compat (v0.12) plugin".

tagomoris: But I think this method is not needed... I'll write the reason in review comment for `output.rb`.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    private
+
+    def validate_basename_is_compatible_with_primary_buffer!
+      if !@chunk_key_time && path_has_time_format?
+        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder, add time formats, like `%Y%m%d`, to `path`"
+      end
+
+      if !@chunk_key_tag && (ph = @placeholders.find { |p| p.match(/tag(\[\d+\])?/) })
```
tagomoris: Block parameter `p` is hiding `Kernel.p`. Use different name.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder #{ph}, add tag placeholder, like `${tag}`, to `path`"
```
tagomoris: Is this error message wrong?
This error will be raised when "tag" isn't in chunk keys and `${tag}` placeholders appear.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+    # To avoid calling Fluent::Compat::TimeSlicedOutput#extract_placeholders if primary plugin is `out_file`
+    def extract_placeholders(str, metadata)
+      super
+    end
+
+    private
+
+    def validate_basename_is_compatible_with_primary_buffer!
+      if !@chunk_key_time && path_has_time_format?
+        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder, add time formats, like `%Y%m%d`, to `path`"
```
tagomoris: Is this error message wrong? (... same with tag)

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder #{ph}, add tag placeholder, like `${tag}`, to `path`"
+      end
+
+      if @chunk_keys.empty?
+        vars = @placeholders.reject { |p| p.match(/tag(\[\d+\])?/) }
+
+        if ph = vars.find { |v| !@chunk_keys.include?(v) }
+          raise Fluent::ConfigError, "out_secondary_file: File path has an incompatible placeholder #{ph}, add variable placeholder, like `${varname}`, to `path`"
+        end
+      end
```
tagomoris: What is this `if` block checking?
If `@chunk_keys.empty?` is true, `!@chunk_keys.include?(v)` is always true.
Check both of code and test code.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def path_has_time_format?
```
tagomoris: This method name is incorrect now.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+      @basename != Time.now.strftime(@basename)
+    end
+
+    def suffix
```
tagomoris: This can be done in `#configure` or `#start` by assigning `@suffix` instance variable, right?

lib/fluent/plugin/output.rb
```diff
@@ -185,13 +185,25 @@ def initialize
...
+        @chunk_key_tag = @primary_instance.chunk_key_tag || false
+        if @primary_instance.chunk_key_time
+          @chunk_key_time = @primary_instance.chunk_key_time
+          @timekey_zone = @primary_instance.timekey_zone
+          @output_time_formatter_cache = {}
+        end
+
+        require_override = !self.class.instance_methods(false).include?(:extract_placeholders)
         (class << self; self; end).module_eval do
-          define_method(:extract_placeholders){ |str, metadata| @primary_instance.extract_placeholders(str, metadata) }
```
tagomoris: Overwriting this method is to use `@chunk_keys` and some other instance variables of primary plugin instances.
But your update brings these instance variable and these values into secondary plugin instances too.
So I think we can just remove this L194.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,347 @@
...
+class FileOutputSecondaryTest < Test::Unit::TestCase
+  include Fluent::UniqueId::Mixin
+
+  def setup
+    Fluent::Test.setup
+    FileUtils.rm_rf(TMP_DIR)
+    FileUtils.mkdir_p(TMP_DIR)
+  end
+
+  TMP_DIR = File.expand_path(File.dirname(__FILE__) + "/../tmp/out_file#{ENV['TEST_ENV_NUMBER']}")
```
tagomoris: Using `out_secondary_file` is better than `out_file`.

tagomoris:
@ganmacs I added review comments.
And please re-check test names. I found some wrong English, and names which doesn't explain what the test is checking good enough.
For example,  "test 'should be output unzip file'" doesn't explain "when?" "why?". In addition, "unzip file" is incorrect expression... "plain text / uncompressed file" is correct.

lib/fluent/plugin/out_secondary_file.rb
repeatedly: I agree. Required parameter is good for avoiding troubles.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,133 @@
...
+    DIR_PERMISSION = 0755
+    PLACEHOLDER_REGEX = /\${(tag(\[\d+\])?|[\w.@-]+)}/
+
+    desc "The directory path of the output file."
+    config_param :directory, :string, default: "secondary/"
+    desc "The baseanme of the output file."
+    config_param :basename, :string, default: "dump.bin"
+    desc "The flushed chunk is appended to existence file or not."
+    config_param :append, :bool, default: false
+    config_param :compress, :enum, list: [:text, :gz, :gzip], default: :text
```
repeatedly: I think `[:text, :gzip]` is enough.
S3 plugin provides `gzip` compress type with `gz` extension.

lib/fluent/plugin/out_secondary_file.rb
ganmacs: ok. Removed default value.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R31

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Thanks. Fixed it.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R45

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Yes. At first, I don't think directory can use placeholders.
But,  `/path/to/backup/${tag}/dump.bin.` looks very useful.
so, I support directory can use placeholders.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R45
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R65
https://github.com/fluent/fluentd/pull/1154/files#diff-39910dc690f56dd659cdccdca103e55aR376

lib/fluent/plugin/out_secondary_file.rb
ganmacs: I removed `output.rb`'s `define_method`( https://github.com/fluent/fluentd/pull/1154/files#diff-7a3f304568ba6143aaf2b7e78011d044L194 ) and remove this method.

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R96

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed(add -> remove).
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R93

lib/fluent/plugin/out_secondary_file.rb
ganmacs: I missed. So I removed it and added a test.
https://github.com/fluent/fluentd/pull/1154/files#diff-39910dc690f56dd659cdccdca103e55aR344

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R48

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R107

test/plugin/test_out_secondary_file.rb
ganmacs: Fixed.
https://github.com/fluent/fluentd/pull/1154/files#diff-39910dc690f56dd659cdccdca103e55aR20

lib/fluent/plugin/out_secondary_file.rb
ganmacs: Thanks. I fixed it.
https://github.com/fluent/fluentd/pull/1154/files#diff-68f04670661c50196427de0d60c06636R36

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+    c.configure(conf)
+  end
+
+  sub_test_case 'configture' do
+    test 'default configuration' do
+      d = create_driver %[directory #{TMP_DIR}]
+      assert_equal 'dump.bin', d.instance.basename
+      assert_equal TMP_DIR, d.instance.directory
+      assert_equal :text, d.instance.compress
+    end
```
tagomoris: Add assertion for `append`.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+
+    test 'should be configurable' do
+      d = create_driver
+      assert_equal 'out_file_test', d.instance.basename
+      assert_equal TMP_DIR, d.instance.directory
+      assert_equal :gzip, d.instance.compress
+    end
```
tagomoris: ditto.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+
+    test 'should only use in secondary' do
+      c = Fluent::Test::Driver::Output.new(Fluent::Plugin::SecondaryFileOutput)
+      assert_raise do
```
tagomoris: `assert_raise` must be used with error class (and message as far as possible).
This assertion successes if `c.configure` raises errors caused by bugs.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+        create_driver %[directory #{TMP_DIR}/test_dir/foo/bar/]
+      end
+
+      assert_nothing_raised do
+        FileUtils.mkdir_p("#{TMP_DIR}/test_dir")
+        File.chmod(0777, "#{TMP_DIR}/test_dir")
+        create_driver %[directory #{TMP_DIR}/test_dir/foo/bar/]
+      end
+
+      assert_raise Fluent::ConfigError do
```
tagomoris: Add message for assertion if possible.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,124 @@
...
+    config_param :compress, :enum, list: [:text, :gzip], default: :text
+
+    def configure(conf)
+      super
+
+      unless @as_secondary
+        raise Fluent::ConfigError, "This plugin can only be used in the <secondary> section"
+      end
+
+      @name = File.join(@directory, @basename)
```
tagomoris: What's `@name` of? It's too less descriptive.
This variable is for the data:
* a string
* to be extracted for placeholders (don't use this value directly)
* without suffix

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+
+      assert_nothing_raised do
+        create_driver %[directory #{TMP_DIR}/test_dir/foo/bar/]
+      end
+    end
+  end
+
+  def check_gzipped_result(path, expect)
+    # Zlib::GzipReader has a bug of concatenated file: https://bugs.ruby-lang.org/issues/9790
+    # Following code from https://www.ruby-forum.com/topic/971591#979520
```
tagomoris: Very good comment to show what this code is doing and why.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+        gzr.finish
+        break if unused.nil?
+        io.pos -= unused.length
+      end
+    }
+
+    assert_equal expect, result
+  end
+
+  class DummyMemoryChunk < Fluent::Plugin::Buffer::MemoryChunk; end
```
tagomoris: Use `primary.buffer.create_metadata` and `primary.buffer.generate_chunk` instead of methods to simulate these, because it's too dependent to Fluentd's internal code and plugins.
Plugins should NOT depend on internals of other plugins as far as possible, because other plugins are out of control.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+  def create_es_chunk(metadata, es)
+    DummyMemoryChunk.new(metadata).tap do |c|
+      c.concat(es.to_msgpack_stream, es.size) # to_msgpack_stream is standard_format
+      c.commit
+    end
+  end
+
+  sub_test_case 'write' do
+    setup do
+      @record = { key: "vlaue" }
```
tagomoris: Keys of records must be a string, not symbol.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+      @time = Time.parse("2011-01-02 13:14:15 UTC").to_i
```
tagomoris: Use `event_time` method to create time objects. It's actually `Fluent::EventTime` objects rather than `Time` in v0.14.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+    test 'should output compressed file with gzip option' do
+      d = create_driver
+      c = create_es_chunk(create_metadata, @es)
+      path = d.instance.write(c)
+
+      assert_equal "#{TMP_DIR}/out_file_test.0.gz", path
+      check_gzipped_result(path, @es.to_msgpack_stream.force_encoding('ASCII-8BIT'))
+    end
+
+    test 'should be output plain text without gzip option' do
```
tagomoris: nonsense "be"

tagomoris: without "compress" option?

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+      packed_value = @es.to_msgpack_stream.force_encoding('ASCII-8BIT')
+
+      5.times do |i|
+        path = d.instance.write(c)
+        assert_equal "#{TMP_DIR}/out_file_test.#{i}.gz", path
+        check_gzipped_result(path, packed_value)
+      end
+    end
+
+    test 'path should be the same with append option' do
```
tagomoris: Do you mean "path should be unchanged if append options is enabled"?
Original subject is nonsense.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+
+      [*1..5].each do |i|
+        path = d.instance.write(c)
+        assert_equal "#{TMP_DIR}/out_file_test.gz", path
+        check_gzipped_result(path, packed_value * i)
+      end
+    end
+  end
+
+  sub_test_case 'placeholder regex' do
```
tagomoris: What does "regex" mean?

tagomoris: These tests are about format and syntax of placeholders, right?

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+      variable7: '${tags}'
+    )
+    test 'matches with a valid placeholder' do |path|
+      assert Fluent::Plugin::SecondaryFileOutput::PLACEHOLDER_REGEX.match(path)
+    end
+
+    data(
+      invalid_tag: 'tag',
+      invalid_tag2: '{tag}',
+      invalid_tag3: '${tag',
```
tagomoris: How's about strings like `${tag${key}` ?
More patterns are needed.

test/plugin/test_out_secondary_file.rb
tagomoris: This comment (and next one) is also for `sub_test_case 'path'`.

test/plugin/test_out_secondary_file.rb
```diff
@@ -0,0 +1,393 @@
...
+    end
+
+    test 'normal path without gzip option' do
+      d = create_driver %[
+        directory #{TMP_DIR}
+        basename out_file_test
+      ]
+      path = d.instance.write(@c)
+      assert_equal "#{TMP_DIR}/out_file_test.0", path
+    end
```
tagomoris: More tests are needed for the case when `append true`.

tagomoris:
I added review comments.

test/plugin/test_out_secondary_file.rb
ganmacs: `Test::Unit::Assertions#assert_raise_message` can assert erorr message.
But, it can't assert both error class and error message.
I leave it as it is.

tagomoris: Use `assert_raise Error.new("message")` pattern.
https://github.com/fluent/fluentd/blob/master/test/plugin_helper/test_parser.rb#L86

ganmacs:
@tagomoris  
I updated.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -0,0 +1,124 @@
...
+      validate_compatible_with_primary_buffer!(@path_without_suffix)
+
+      @suffix = case @compress
+                when :text
+                  ""
+                when :gzip
+                  ".gz"
+                end
+
+      test_path = generate_path(File.join(@directory, Time.now.strftime("%Y%m%d")))
```
tagomoris: `test_path` loses `@basename`.

ganmacs: Yes. But `test_path` is used for a test that directory is writable or not. 
I think it is reasonable that `test_path` loses `@basename`.

tagomoris: Are there any positive reason not to use `@basename` for this test?
`@basename` might contain `/`... (users do everything possible).

ganmacs: > `@basename` might contain /... (users do everything possible).

I didn't care about it.
I added a test for it.
And I use  `basename` as `test_path`

https://github.com/fluent/fluentd/pull/1154/commits/eec980342f408384948a47d9d9c79d228ef14ad7

tagomoris:
I found that `Fluent::FileUtil.writable_p` returns false if the directory of specified path doesn't exist.
It means configuration like `directory /path/to/${tag}/%Y%m%d` always raises configuration errors.
Hmm.

ganmacs:
@tagomoris 

I wrote a test that directory has placeholders (  https://github.com/fluent/fluentd/pull/1154/files#diff-39910dc690f56dd659cdccdca103e55aR417 ) and this test passed.

If configuration is like `directory /path/to/${tag}/%Y%m%d`, `Fluent::FileUtil.writable_p`  check direcotry `/path/to/` is writable or not (https://github.com/fluent/fluentd/blob/master/lib/fluent/compat/file_util.rb#L39).

I think it is good enough to check.

tagomoris:
@ganmacs Ah, I misread the code... you're right.
LGTM.

tagomoris:
Merged! It'll be released as v0.14.5. Thank you!

PR #1155, opened 2016-08-10
Handle Windows event correctly when using winsvc.rb

unak:
When invoking fluentd as a Windows Service via winsvc.rb, winsvc.rb expects that fluetnd handle the Windows event passed by `-x` option.
But currently fluentd simply ignores it.
This patch implements to handle the event, and fixes a part of #1084.


tagomoris:
LGTM

PR #1159, opened 2016-08-11
Migrate out_file to v0.14 API

okkez:
TODO

* [x] Add test using sub sections (parse, inject, buffer)

### DIFF
--- a/lib/fluent/plugin/formatter_out_file.rb
+++ b/lib/fluent/plugin/formatter_out_file.rb
@@ -56,6 +56,8 @@ module Fluent
 
         super
 
+        Fluent::Timezone.validate!(@timezone) if @timezone
+
         @timef = case @time_type
                  when :float then ->(time){ time.to_r.to_f }
                  when :unixtime then ->(time){ time.to_i }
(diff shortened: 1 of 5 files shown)


repeatedly:
Any progress on this patch?

tagomoris:
I'm writing an internal feature for parser/formatter/others to provide TimeParser/TimeFormatter & timezone configurations.
This change will be updated after that code.

tagomoris:
I added the feature to handle TimeParser/TimeFormatter in parser/formatter plugins & inject/extract plugin helpers.
@okkez Could you update this patch on those?

okkez:
OK, I will work for this after RubyKaigi2016.
Maybe 9/13 or later :bow:

okkez:
@tagomoris I've pushed changes use new TimeFormatter.
Could you check latest 3 commits?

tagomoris:
`@formatter.time_formatter_create` in output plugin code looks bad :P 
I suppose we need parameters for plugins to control timezones...

tagomoris:
Mmm. I'll take this task, because there are some difficult points to be considered...

tagomoris:
I re-created the change as #1257. Thank you to show me problems and basic form to be done.

PR #1165, opened 2016-08-17
Create tool that load dumped event

### LINKED ISSUE #873: Special plugin only to dump buffer contents to file

ganmacs:
`fluent-binlog-reader` is a command line tool to read the data which is saved as `Message Pack` format.

### `head`

`head` is used to display the beginning of a text file.

```
$ fluent-binlog-reader head packed.log
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
```

You can specify the number of events to display.

```
$ fluent-binlog-reader head -n 1 packed.log
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
```

### `cat`

`cat` reads files sequentially, writing them to standard output.

```
$ fluent-binlog-reader cat packed.log
2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}

...

2016-08-12T17:24:18+09:00       packed.log      {"message":"dummy"}
```

### `formats`

`formats` displays plugins that you can use.

```
$ fluent-binlog-reader formats
out_file
stdout
json
hash
msgpack
ltsv
csv
single_value
```

### `--format`

You can specify the `formatter` as shown below.

```
$ fluent-binlog-reader cat packed.log --format=json
{"message":"dummy"}
...
{"message":"dummy"}
```

## Format params

`formatter` has configurable params.(e.g. `CsvFormatter` has `delimiter`, `force_quotes`, and `fields`)
You can pass these params by using `-e` option.

```
$ fluent-binlog-reader cat --formats=csv  -e fields=message packed.log
"dummy"
...
"dummy"
```

### `--plugin`

You can add new plugins path.

```
# $ touch my_plugins/fomatter_csv2.rb
# $ vi my_plugins/fomatter_csv2.rb

$ fluent-binlog-reader cat formats=csv2 --plugin=my_plugins
...
```

## TODO

- [x] Subcommands
  - [x] `head`
  - [x] `cat`
  - [x] `formats`
- [x] adding plugins path
- [x] specifed format
- [x] specifed format params
- [x] add tests

### DIFF
--- a/lib/fluent/command/binlog_reader.rb
+++ b/lib/fluent/command/binlog_reader.rb
@@ -168,7 +168,7 @@ module BinlogReaderCommand
         i = 1
         Fluent::MessagePackFactory.unpacker(io).each do |(time, record)|
           print @formatter.format(@path, time, record) # path is used for tag
-          break if i == @options[:count] && @options[:count] != -1
+          break if @options[:count] && i == @options[:count]
           i += 1
         end
       end
@@ -196,7 +196,7 @@ module BinlogReaderCommand
 
   class Cat < Head
     DEFAULT_CAT_OPTIONS = {
-      count: -1
+      count: nil                # Overwrite DEFAULT_HEAD_OPTIONS[:count]
     }
 
     def default_options
--- a/lib/fluent/command/binlog_reader.rb
+++ b/lib/fluent/command/binlog_reader.rb
@@ -71,7 +71,7 @@ module BinlogReaderCommand
       @opt_parser = OptionParser.new do |opt|
         opt.separator 'Options:'
 
-        opt.on('-p DIR', '--plugin', 'add library path') do |v|
+        opt.on('-p DIR', '--plugin', 'add library directory path') do |v|
           @options[:plugin] << v
         end
       end
@@ -92,11 +92,13 @@ module BinlogReaderCommand
     def parse_options!
       @opt_parser.parse!(@argv)
 
-      if !@options[:plugin].empty? && (dir = @options[:plugin].find { |d| !Dir.exist?(d) })
-        usage "Directory #{dir} doesn't exist"
-      elsif !@options[:plugin].empty?
-        @options[:plugin].each do |d|
-          Fluent::Plugin.add_plugin_dir(d)
+      unless @options[:plugin].empty?
+        if dir = @options[:plugin].find { |d| !Dir.exist?(d) }
+          usage "Directory #{dir} doesn't exist"
+        else
+          @options[:plugin].each do |d|
+            Fluent::Plugin.add_plugin_dir(d)
+          end
         end
       end
     rescue => e
@@ -120,7 +122,7 @@ module BinlogReaderCommand
     def configure_option_parser
       @options.merge!(config_params: {})
 
-      @opt_parser.banner = "Usage: fluent- #{self.class.to_s.split('::').last.downcase} [options] file"
+      @opt_parser.banner = "Usage: fluent-binlog-reader #{self.class.to_s.split('::').last.downcase} [options] file"
 
       @opt_parser.on('-f TYPE', '--format', 'configure output format') do |v|
         @options[:format] = v.to_sym
--- a/lib/fluent/command/binlog_reader.rb
+++ b/lib/fluent/command/binlog_reader.rb
@@ -22,17 +22,17 @@ require 'fluent/formatter'
 require 'fluent/plugin'
 require 'fluent/config/element'
 
-class FluentUnpacker
+class FluentBinlogReader
   SUBCOMMAND = %w(cat head formats)
   HELP_TEXT = <<HELP
-Usage: fluent-unpacker <command> [<args>]
+Usage: fluent-binlog-reader <command> [<args>]
 
-Commands of fluent-unpacker:
+Commands of fluent-binlog-reader:
    cat     :     Read files sequentially, writing them to standard output.
    head    :     Display the beginning of a text file.
    format  :     Display plugins that you can use.
 
-See 'fluent-unpacker <command> --help' for more information on a specific command.
+See 'fluent-binlog-reader <command> --help' for more information on a specific command.
 HELP
 
   def initialize(argv = ARGV)
@@ -40,7 +40,7 @@ HELP
   end
 
   def call
-    command_class = UnpackerCommand.const_get(command)
+    command_class = BinlogReaderCommand.const_get(command)
     command_class.new(@argv).call
   end
 
@@ -62,7 +62,7 @@ HELP
   end
 end
 
-module UnpackerCommand
+module BinlogReaderCommand
   class Base
     def initialize(argv = ARGV)
       @argv = argv
@@ -99,6 +99,8 @@ module UnpackerCommand
           Fluent::Plugin.add_plugin_dir(d)
         end
       end
+    rescue => e
+      usage e
     end
   end
 
@@ -118,7 +120,7 @@ module UnpackerCommand
     def configure_option_parser
       @options.merge!(config_params: {})
 
-      @opt_parser.banner = "Usage: fluent-unpacker #{self.class.to_s.split('::').last.downcase} [options] file"
+      @opt_parser.banner = "Usage: fluent- #{self.class.to_s.split('::').last.downcase} [options] file"
 
       @opt_parser.on('-f TYPE', '--format', 'configure output format') do |v|
         @options[:format] = v.to_sym
--- a/lib/fluent/command/unpacker.rb
+++ b/lib/fluent/command/unpacker.rb
@@ -79,13 +79,13 @@ module Command
         @argv = @argv[0...i]
       end
 
-      @options = DEFAULT_OPTIONS
+      @options = DEFAULT_OPTIONS.dup
       @opt_parser = OptionParser.new do |opt|
         opt.banner = "Usage: fluent-unpacker #{self.class.to_s.split('::').last.downcase} [options] file [-- <params>]"
 
         opt.separator 'Options:'
 
-        opt.on('-f TYPE', '--format', 'configure output format: ') do |v|
+        opt.on('-f TYPE', '--format', 'configure output format') do |v|
           @options[:format] = v.to_sym
         end
 
@@ -96,7 +96,7 @@ module Command
     end
 
     def call
-      raise NotImplementedError, "Must Implement this method"
+      raise NotImplementedError, 'BUG: command  MUST implement this method'
     end
 
     private
@@ -111,35 +111,37 @@ module Command
       exit 1
     end
 
-    def lookup_formatter(format, conf)
+    def lookup_formatter(format, params)
+      conf = Fluent::Config::Element.new('ROOT', '', params, [])
       formatter = Fluent::Plugin.new_formatter(format)
 
       if formatter.respond_to?(:configure)
         formatter.configure(conf)
       end
       formatter
+    rescue => e
+      usage e
     end
   end
 
   class Head < Base
     DEFAULT_HEAD_OPTIONS = {
-      number: 5
+      count: 5
     }
 
     def initialize(argv = ARGV)
       super
-      @options.merge(DEFAULT_HEAD_OPTIONS)
+      @options.merge!(DEFAULT_HEAD_OPTIONS)
       @path = configure_option_parser
     end
 
     def call
-      conf = Fluent::Config::Element.new('ROOT', '', @params, [])
-      @formatter = lookup_formatter(@options[:format], conf)
+      @formatter = lookup_formatter(@options[:format], @params)
 
       File.open(@path, 'r') do |io|
         i = 0
         unpacker(io).each do |(time, record)|
-          break if i == @options[:number]
+          break if i == @options[:count]
           i += 1
           puts @formatter.format(@path, time, record) # tag is use for tag
         end
@@ -149,14 +151,22 @@ module Command
     private
 
     def configure_option_parser
-      @opt_parser.on('-n COUNT', 'head like -n') do |v|
-        @options[:number] = v.to_i
+      @opt_parser.on('-n COUNT', 'Set the number of lines to display') do |v|
+        @options[:count] = v.to_i
       end
 
       path = @opt_parser.parse(@argv)
-      usage 'Path is required' if path.empty?
-      usage "#{path.first} is not found" unless File.exist?(path.first)
-      path.first
+
+      case
+      when path.empty?
+        usage 'Path is required'
+      when !File.exist?(path.first)
+        usage "#{path.first} is not found"
+      when @options[:count] < 1
+        usage "illegal line count -- #{@options[:count]}"
+      else
+        path.first
+      end
     end
   end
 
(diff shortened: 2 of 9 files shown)


lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+
+class FluentUnpacker
+  SUBCOMMAND = %w(cat head formats)
+  HELP_TEXT = <<HELP
+Usage: fluent-unpacker <command> [<args>]
+
+Commands of fluent-unpacker:
+   cat     :     Read files sequentially, writing them to standard output.
+   haed    :     Display the beginning of a text file.
+   foramts :     Display plugins that you can use.
```
tagomoris: typo

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+require 'fluent/config/element'
+
+class FluentUnpacker
+  SUBCOMMAND = %w(cat head formats)
+  HELP_TEXT = <<HELP
+Usage: fluent-unpacker <command> [<args>]
+
+Commands of fluent-unpacker:
+   cat     :     Read files sequentially, writing them to standard output.
+   haed    :     Display the beginning of a text file.
```
tagomoris: typo

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+  end
+
+  def usage(msg = nil)
+    puts HELP_TEXT
+    puts "Error: #{msg}" if msg
+    exit 1
+  end
+end
+
+module Command
```
tagomoris: Top level `Command` module looks very harmful.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    def initialize(argv = ARGV)
+      super
+
+      @options.merge!(DEFAULT_OPTIONS)
+      @params = {}
+      if i = @argv.index('--')
+        @params = @argv[i+1..-1].reduce({}) do |acc, e|
+          k, v = e[1..-1].split('=')
```
tagomoris: Why is `e[1..-1]` needed? What's the expression pattern expected for arguments after `--`?

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+      @options.merge!(DEFAULT_HEAD_OPTIONS)
+      @path = parse_options!
+    end
+
+    def call
+      @formatter = lookup_formatter(@options[:format], @params)
+
+      File.open(@path, 'r') do |io|
+        i = 0
+        MessagePack::Unpacker.new(io).each do |(time, record)|
```
tagomoris: Don't use raw `MessagePack::Unpacker`, because we need to unpack values of ext type `Fluent::EventType`.
Use `Fluent::MessagePackFactory.unpacker` instead.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+          break if i == @options[:count]
```
tagomoris: IMO breaking after `puts` looks usual code.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+    end
+
+    private
+
+    def parse_options!
+      @opt_parser.on('-n COUNT', 'Set the number of lines to display') do |v|
+        @options[:count] = v.to_i
+      end
+
+      path = super
```
tagomoris: The return value of super method looks the values of argv (items for options are excluded).
It's not path.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+    def initialize(argv = ARGV)
+      super
+      @path = parse_options!
+    end
+
+    def call
+      @formatter = lookup_formatter(@options[:format], @params)
+
+      File.open(@path, 'r') do |io|
+        MessagePack::Unpacker.new(io).each do |(time, record)|
```
tagomoris: ditto.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+
+    def call
+      @formatter = lookup_formatter(@options[:format], @params)
+
+      File.open(@path, 'r') do |io|
+        i = 0
+        MessagePack::Unpacker.new(io).each do |(time, record)|
+          break if i == @options[:count]
+          i += 1
+          puts @formatter.format(@path, time, record) # tag is use for tag
```
tagomoris: Formatters generates lines terminated with newlines.
Does this `puts` adds additional newline?

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+module Command
+  class Base
+    def initialize(argv = ARGV)
+      @argv = argv
+
+      @options = {}
+      @opt_parser = OptionParser.new do |opt|
+        opt.separator 'Options:'
+
+        opt.on('-p DIR', '--plugins', 'add library path') do |v|
```
tagomoris: `--plugin` is used in `fluent/command/fluentd.rb`.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+          @options[:plugins] = v
```
tagomoris: This option can't be specified twice or more. Is it intended?

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+  class Formats < Base
+    def initialize(argv = ARGV)
+      super
+      parse_options!
+    end
+
+    def call
+      prefix = Fluent::Plugin::FORMATTER_REGISTRY.dir_search_prefix || 'formatter_'
+
+      new_path = Fluent::Plugin::FORMATTER_REGISTRY.paths.last
```
tagomoris: Is this line for `-p DIR` option?
If so, that value can be fetched by `@options[:plugins]` simply, instead of fetching last item of registry paths.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,253 @@
...
+        require File.absolute_path(e)
+      end
+
+      $LOAD_PATH.map do |lp|
+        Dir.glob("#{lp}/#{prefix}*").each do |e|
+          require e
+        end
+      end
+
+      specs = Gem::Specification.flat_map { |spec| spec.lib_files }.select do |e|
```
tagomoris: This looks items of `lib_files`, not gemspec. Right?

tagomoris:
I added review comments.
And... `fluent-unpacker` looks the name too internal (users don't take care about the fact that dumping the content of buffer files is equal to unpacking msgpack event stream).
Do you have another idea about naming of this tool?

ganmacs:
I think it is better `standard-format-reader` or  `dumped-log-reader`.
How do you think? 
@tagomoris

tagomoris:
I prefer `dumped-log-reader`. But `fluent-` prefix is needed.
`fluent-dumped-log-reader`, `fluent-read-dump`, ...
@repeatedly Any thoughts?

lib/fluent/command/unpacker.rb
ganmacs: `e[1..-1]` is used for trimming prefix(`-`) in ` -fields=message`.
it is used like this.
 `$ fluent-unpacker cat --formats=csv  packed.log -- -fields=message`

But, I removed it .

lib/fluent/command/unpacker.rb
ganmacs: Are there any problem using `--plugins` in `fluent/command/fluentd.rb`?

ganmacs:
@tagomoris 
I updated without the name of this command.

tagomoris:
Your design to configure formatter plugins looks:
* (before) `fluent-unpacker -f out_file FILE_NAME -- -output_time=false -output_tag=false`
* (after) `fluent-unpacker -f out_file FILE_NAME -- output_time=false output_tag=false`

Is it correct?

If so, using `--` to separate configuration parameters for formatters and others looks very curious. Normally `--` is used to separate options and other arguments (may includes `-` as heading char).

@ganmacs How do you think about to use another command line option to specify formatter configuration parameters?
e.g.,) `fluent-unpacker -f out_file -e output_time_false -e output_tag=false FILE_NAME`

In this way, we can use `--` to specify `FILE_NAME` with heading `-`.

ganmacs:
@tagomoris 
I changed command line design like below.
`fluent-unpacker -f out_file -e output_time_false -e output_tag=false FILE_NAME`

tagomoris:
It's too late to point... but, `head` and `cat` can share code :P

ganmacs:
> `head` and `cat` can share code

Done.

lib/fluent/command/unpacker.rb
```diff
@@ -0,0 +1,234 @@
...
+      usage "#{@path} is not found" unless File.exist?(@path)
+    end
+  end
+
+  class Cat < Head
+    DEFAULT_CAT_OPTIONS = {
+      count: -1
+    }
+
+    def initialize(argv = ARGV)
```
tagomoris: nonsense def (just doing `super`).

test/command/test_unpacker.rb
```diff
@@ -0,0 +1,350 @@
...
+end
+
+class TestBaseCommand < ::Test::Unit::TestCase
+  TMP_DIR = File.expand_path(File.dirname(__FILE__) + "/../tmp/command/unpacker#{ENV['TEST_ENV_NUMBER']}")
+
+  def create_message_packed_file(path, times = [event_time], records = [{ 'message' => 'dummy' }])
+    es = Fluent::MultiEventStream.new(times, records)
+    v = es.to_msgpack_stream
+    File.open("#{TMP_DIR}/#{path}", 'w') do |f|
+      f.print(v)
```
tagomoris: Generally speaking, it's better practice to use `IO#write` to output binary data without any modification. `IO#print` may be affected by separator character (with multiple arguments).
It's just comment. No need to update here.

test/command/test_unpacker.rb
```diff
@@ -0,0 +1,350 @@
...
+  def setup
+    FileUtils.rm_rf(TMP_DIR)
+    FileUtils.mkdir_p(TMP_DIR)
+  end
+
+  def timezone(timezone = 'UTC')
+    old = ENV['TZ']
+    ENV['TZ'] = timezone
+    yield
+    ENV['TZ'] = old
```
tagomoris: This re-assignment should be done in `ensure`.

tagomoris:
I added some minor comments.
The last major problem is naming...
@repeatedly do you have any idea about naming?

repeatedly:
Hmm... if this tool focuses on binary format, fluent-binlog-reader is good for me.
`dump` seems too general because other text files, e.g. result of out_file, are also dumped.

tagomoris:
looks good > `fluent-binlog-reader`

ganmacs:
@tagomoris 
I changed a command name `unpacker` -> `binlog-reader`

@repeatedly Thank you for your help!

repeatedly:
Two directories for test formatter plugins are needed?
`test/scripts/fluent/plugin/formatter/{formatter_test1,formatter_test2}.rb` seems enough.

lib/fluent/command/binlog_reader.rb
```diff
@@ -0,0 +1,232 @@
...
+      puts "Error: #{msg}" if msg
+      exit 1
+    end
+
+    def parse_options!
+      @opt_parser.parse!(@argv)
+
+      if !@options[:plugin].empty? && (dir = @options[:plugin].find { |d| !Dir.exist?(d) })
+        usage "Directory #{dir} doesn't exist"
+      elsif !@options[:plugin].empty?
```
repeatedly: This check seems verbose.

```rb
if !@options[:plugin].empty?
  if dir = @options[:plugin].find { |d| !Dir.exist?(d) }
     usage "Directory #{dir} doesn't exist"
  else
    @options[:plugin].each do |d|
      Fluent::Plugin.add_plugin_dir(d)
    end
  end
end
```

test/command/test_binlog_reader.rb
```diff
@@ -0,0 +1,351 @@
+require_relative '../helper'
+
+require 'oj'
```
repeatedly: Should use `yajl`. `oj` is optional.

lib/fluent/command/binlog_reader.rb
```diff
@@ -0,0 +1,232 @@
...
+      @options.merge!(DEFAULT_OPTIONS)
+      configure_option_parser
+    end
+
+    private
+
+    def configure_option_parser
+      @options.merge!(config_params: {})
+
+      @opt_parser.banner = "Usage: fluent- #{self.class.to_s.split('::').last.downcase} [options] file"
```
repeatedly: > `fluent- `

This space is unneeded.

lib/fluent/command/binlog_reader.rb
```diff
@@ -0,0 +1,234 @@
...
+    end
+
+    def call
+      @formatter = lookup_formatter(@options[:format], @options[:config_params])
+
+      File.open(@path, 'r') do |io|
+        i = 1
+        Fluent::MessagePackFactory.unpacker(io).each do |(time, record)|
+          print @formatter.format(@path, time, record) # path is used for tag
+          break if i == @options[:count] && @options[:count] != -1
```
repeatedly: `@options[:count]` is validated in `parser_options`
This `!= -1` check is needed?

ganmacs:
@repeatedly 

There are two tests 
( https://github.com/fluent/fluentd/pull/1165/files#diff-5e97ad93fa2a92bc851090105261df21R329 
and  https://github.com/fluent/fluentd/pull/1165/files#diff-5e97ad93fa2a92bc851090105261df21R339) for `-p` option'.

The first one is a test when two directories are passed.
The second one is a test when a directory is passed.
So  I think it is need two directories for a test.

lib/fluent/command/binlog_reader.rb
ganmacs: Yes. It is used in `Cat` command (https://github.com/fluent/fluentd/pull/1165/files#diff-507344e380ffc9fa2093c57196f514d4R199)
`Cat` class inherits `Head` class.

repeatedly:
@ganmacs Ah, I see. Got it.

lib/fluent/command/binlog_reader.rb
repeatedly: Hmm.... if so, `nil` is better for default value?

ganmacs: You're right. I fixed it.

ganmacs:
@tagomoris @repeatedly 
I updated. Please review the code again.

tagomoris:
LGTM.

PR #1172, opened 2016-08-19
Support data compression in buffer plugins

### LINKED ISSUE #1169: Support a data compression in buffer plugins
Buffer plugins maintain data as raw `String` object.
This way consumes many memory resource and bandwidth of the network.
To reduce this wasteful consumption, I propose that support a data compression in buffer plugins.

### Tasks

* Add `compress true` in buffer plugin configuration.
* Compress a `chunk` by using zlib in `concat`
* Decompress a `chunk` in `write_to`

ganmacs:
### TODO
- [x] Add compress true in buffer plugin configuration.
- [x]  Compress a chunk  by `to_compressed_msgpack_stream` in `EventStream`
- [x] Decompress a chunk in `write_to`
- [x] Add some test

### Not for this PR tasks

* Support data compression in forward protocol

Close #1169

### DIFF
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -146,15 +146,18 @@ module Fluent
         end
 
         def read(**kwargs)
+          raise ArgumentError, '`compressed: gzip` can be used for Compressable module' if kwargs[:compressed] == :gzip
           raise NotImplementedError, "Implement this method in child class"
         end
 
         def open(**kwargs, &block)
+          raise ArgumentError, '`compressed: gzip` can be used for Compressable module' if kwargs[:compressed] == :gzip
           raise NotImplementedError, "Implement this method in child class"
         end
 
         def write_to(io, **kwargs)
-          open(**kwargs) do |i|
+          raise ArgumentError, '`compressed: gzip` can be used for Compressable module' if kwargs[:compressed] == :gzip
+          open do |i|
             IO.copy_stream(i, io)
           end
         end
@@ -188,12 +191,10 @@ module Fluent
           end
 
           def write_to(io, **kwargs)
-            if kwargs[:compressed] == :gzip
-              super
-            else
-              # To get IO that has compressed data
-              open(compressed: :gzip) do |chunk_io|
-                # Avoid to load large String object when file_chunk
+            open(compressed: :gzip) do |chunk_io|
+              if kwargs[:compressed] == :gzip
+                IO.copy_stream(chunk_io, io)
+              else
                 decompress(input_io: chunk_io, output_io: io)
               end
             end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -21,6 +21,7 @@ require 'fluent/event'
 
 require 'monitor'
 require 'tempfile'
+require 'zlib'
 
 module Fluent
   module Plugin
@@ -66,7 +67,8 @@ module Fluent
         attr_reader :unique_id, :metadata, :created_at, :modified_at, :state
 
         # data is array of formatted record string
-        def append(data)
+        def append(data, **kwargs)
+          raise ArgumentError, '`compress: gzip` can be used for Compressable module' if kwargs[:compress] == :gzip
           adding = ''.b
           data.each do |d|
             adding << d.b
@@ -164,6 +166,20 @@ module Fluent
 
         module Decompressable
           include Fluent::Plugin::Compressable
+
+          def append(data, **kwargs)
+            if kwargs[:compress] == :gzip
+              io = StringIO.new
+              Zlib::GzipWriter.wrap(io) do |gz|
+                data.each do |d|
+                  gz.write d
+                end
+              end
+              concat(io.string, data.size)
+            else
+              super
+            end
+          end
 
           def open(**kwargs, &block)
             if kwargs[:compressed] == :gzip
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -15,6 +15,7 @@
 #
 
 require 'fluent/plugin/buffer'
+require 'fluent/plugin/compressable'
 require 'fluent/unique_id'
 require 'fluent/event'
 
@@ -46,7 +47,7 @@ module Fluent
 
         # TODO: CompressedPackedMessage of forward protocol?
 
-        def initialize(metadata)
+        def initialize(metadata, compress: :text)
           super()
           @unique_id = generate_unique_id
           @metadata = metadata
@@ -57,6 +58,8 @@ module Fluent
           @size = 0
           @created_at = Time.now
           @modified_at = Time.now
+
+          extend Decompressable if compress == :gzip
         end
 
         attr_reader :unique_id, :metadata, :created_at, :modified_at, :state
@@ -149,9 +152,29 @@ module Fluent
           raise NotImplementedError, "Implement this method in child class"
         end
 
-        def write_to(io)
+        def write_to(io, options = {})
           open do |i|
             IO.copy_stream(i, io)
+          end
+        end
+
+        module Decompressable
+          include Fluent::Plugin::Compressable
+
+          def write_to(io, options = {})
+            unless options[:compress] == :gzip
+              c = decompress(read)
+              if @chunk.is_a?(String)
+                @chunk = c
+              else
+                # reset io(@chunk) to overwrite new chunk
+                @chunk.seek(0, IO::SEEK_SET)
+                @chunk.truncate(0)
+                @chunk.write c
+              end
+            end
+
+            super
           end
         end
       end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -20,6 +20,7 @@ require 'fluent/unique_id'
 require 'fluent/event'
 
 require 'monitor'
+require 'tempfile'
 
 module Fluent
   module Plugin
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -144,7 +144,7 @@ module Fluent
           self
         end
 
-        def read
+        def read(**kwargs)
           raise NotImplementedError, "Implement this method in child class"
         end
 
@@ -172,7 +172,18 @@ module Fluent
                 @chunk.truncate(0)
                 @chunk.write c
               end
+          def read(**kwargs)
+            return super if kwargs[:compress] == :gzip
+
+            # avoid creating duplicated IO
+            if @chunk.is_a?(IO)
+              # reset io(@chunk) to read
+              @chunk.seek(0, IO::SEEK_SET)
+              decompress('', io: @chunk) # not to call IO#read
+            else
+              decompress(super)
             end
+          end
 
             super
           end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -152,7 +152,7 @@ module Fluent
           raise NotImplementedError, "Implement this method in child class"
         end
 
-        def write_to(io, options = {})
+        def write_to(io, **kwargs)
           open do |i|
             IO.copy_stream(i, io)
           end
@@ -161,17 +161,6 @@ module Fluent
         module Decompressable
           include Fluent::Plugin::Compressable
 
-          def write_to(io, options = {})
-            unless options[:compress] == :gzip
-              c = decompress(read)
-              if @chunk.is_a?(String)
-                @chunk = c
-              else
-                # reset io(@chunk) to overwrite new chunk
-                @chunk.seek(0, IO::SEEK_SET)
-                @chunk.truncate(0)
-                @chunk.write c
-              end
           def read(**kwargs)
             return super if kwargs[:compress] == :gzip
 
@@ -185,7 +174,8 @@ module Fluent
             end
           end
 
-            super
+          def write_to(io, **kwargs)
+            io.write(read(kwargs))
           end
         end
       end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -166,8 +166,14 @@ module Fluent
               super
             else
               super(kwargs) do |chunk_io|
-                comressed_data = decompress(chunk_io.read)
-                yield StringIO.new(comressed_data)
+                output_io = if chunk_io.is_a?(StringIO)
+                              StringIO.new
+                            else
+                              Tempfile.new('decompressed-data')
+                            end
+                decompress(input_io: chunk_io, output_io: output_io)
+                output_io.seek(0, IO::SEEK_SET)
+                yield output_io
               end
             end
           end
@@ -185,9 +191,9 @@ module Fluent
               super
             else
               # To get IO that has compressed data
-              open(compressed: :gzip) do |io_chunk|
+              open(compressed: :gzip) do |chunk_io|
                 # Avoid to load large String object when file_chunk
-                decompress(io_chunk.read, output_io: io)
+                decompress(input_io: chunk_io, output_io: io)
               end
             end
           end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -162,20 +162,22 @@ module Fluent
           include Fluent::Plugin::Compressable
 
           def read(**kwargs)
-            return super if kwargs[:compress] == :gzip
-
-            # avoid creating duplicated IO
-            if @chunk.is_a?(IO)
-              # reset io(@chunk) to read
-              @chunk.seek(0, IO::SEEK_SET)
-              decompress('', io: @chunk) # not to call IO#read
+            if kwargs[:compressed] == :gzip
+              super
             else
               decompress(super)
             end
           end
 
           def write_to(io, **kwargs)
-            io.write(read(kwargs))
+            if kwargs[:compressed] == :gzip
+              super
+            else
+              open do |io_chunk|
+                # Avoid to load large String object when file_chunk
+                decompress(io_chunk.read, output_io: io)
+              end
+            end
           end
         end
       end
--- a/lib/fluent/plugin/buffer/chunk.rb
+++ b/lib/fluent/plugin/buffer/chunk.rb
@@ -148,18 +148,29 @@ module Fluent
           raise NotImplementedError, "Implement this method in child class"
         end
 
-        def open(&block)
+        def open(**kwargs, &block)
           raise NotImplementedError, "Implement this method in child class"
         end
 
         def write_to(io, **kwargs)
-          open do |i|
+          open(**kwargs) do |i|
             IO.copy_stream(i, io)
           end
         end
 
         module Decompressable
           include Fluent::Plugin::Compressable
+
+          def open(**kwargs, &block)
+            if kwargs[:compressed] == :gzip
+              super
+            else
+              super(kwargs) do |chunk_io|
+                comressed_data = decompress(chunk_io.read)
+                yield StringIO.new(comressed_data)
+              end
+            end
+          end
 
           def read(**kwargs)
             if kwargs[:compressed] == :gzip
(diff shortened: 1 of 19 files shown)


lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -149,11 +152,31 @@ def open(&block)
           raise NotImplementedError, "Implement this method in child class"
         end
 
-        def write_to(io)
+        def write_to(io, options = {})
```
tagomoris: Use `**kwargs` instead of tailing `option = {}`.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -149,11 +152,31 @@ def open(&block)
...
           open do |i|
             IO.copy_stream(i, io)
           end
         end
+
```
tagomoris: There are no care about `chunk.read`, so that call returns raw value of compressed data.
Is it intended?
I think `chunk.read` (without any optional arguments) should return decompressed data.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -149,11 +152,31 @@ def open(&block)
...
+        module Decompressable
+          include Fluent::Plugin::Compressable
+
+          def write_to(io, options = {})
+            unless options[:compress] == :gzip
+              c = decompress(read)
+              if @chunk.is_a?(String)
+                @chunk = c
```
tagomoris: Is it the hack for `MemoryChunk#write_to`, right?
IMO the implementation of `MemoryChunk#write_to` is wrong, and it should be `io.write self.read`.
If `self.read` returns decompressed value, there are no need to do such hack.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -149,11 +152,31 @@ def open(&block)
...
+              else
+                # reset io(@chunk) to overwrite new chunk
+                @chunk.seek(0, IO::SEEK_SET)
+                @chunk.truncate(0)
+                @chunk.write c
```
tagomoris: This occurs too heavy I/O operations, and this may cause data loss if `@chunk.write` raises any errors. Don't truncate/overwrite original file chunk.
Can you use StringIO or something others at the same time of decompress operation?

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,51 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'zlib'
+
+module Fluent
+  module Plugin
+    module Compressable
```
tagomoris: IMO these methods should be able to handle IO objects as output destination (by specifying it as optional keyword arguments).
It reduces overhead of StringIO -> String -> IO (file chunks).

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,51 @@
...
+        io.string
+      end
+
+      # compressed_data is String like `compress(data1) + compress(data2) + ... + compress(dataN)`
+      def decompress(compressed_data)
+        return compressed_data if compressed_data.empty?
+        io = StringIO.new(compressed_data)
+        ret = ''
+
+        loop do
```
tagomoris: Note where this hack came from.

tagomoris:
I added review comments for implementations.
@ganmacs Could you update your code? I'll back to here after that for code review and test review.

ganmacs:
@tagomoris
 I updated.

lib/fluent/plugin/compressable.rb
tagomoris: I wrote "as output destination". Of course, IO objects as input data source are useful.
On the other hand, `#decompress` still returns 2560MB string for file buffer chunks configured as `chunk_limit_size 256m`. It's a large overhead.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -141,19 +144,40 @@ def purge
...
+          include Fluent::Plugin::Compressable
+
+          def read(**kwargs)
+            return super if kwargs[:compress] == :gzip
+
+            # avoid creating duplicated IO
+            if @chunk.is_a?(IO)
+              # reset io(@chunk) to read
+              @chunk.seek(0, IO::SEEK_SET)
+              decompress('', io: @chunk) # not to call IO#read
```
tagomoris: Specifying blank string for unused argument looks bad API.
Is there a way to use default value for that parameter? Or any other solution? (At least, `nil` is better for unused parameter value.)

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -141,19 +144,40 @@ def purge
...
           open do |i|
             IO.copy_stream(i, io)
           end
         end
+
+        module Decompressable
+          include Fluent::Plugin::Compressable
+
+          def read(**kwargs)
+            return super if kwargs[:compress] == :gzip
```
tagomoris: This keyword argument name should be `compressed`, not `compress`.
Because caller of this method expects "compressed data", not expecting "compress data here (again)".

test/plugin/test_output.rb
```diff
@@ -553,4 +553,36 @@ def waiting(seconds)
       @i.stop; @i.before_shutdown; @i.shutdown; @i.after_shutdown; @i.close; @i.terminate
     end
   end
+
+  sub_test_case '#generate_format_proc' do
```
tagomoris: Great to add this!

ganmacs:
@tagomoris 
I updated.

tagomoris:
This implementation doesn't have code to decompress data in `Chunk#open`.
So if plugins does `chunk.open{|io| socket.write(io.read) }`, it reads (unexpected) compressed data.

ganmacs:
@tagomoris 
I updated. support `Chunk#open`.

ganmacs:
@tagomoris 
I updated this PR.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -141,19 +144,60 @@ def purge
...
+
+          def open(**kwargs, &block)
+            if kwargs[:compressed] == :gzip
+              super
+            else
+              super(kwargs) do |chunk_io|
+                output_io = if chunk_io.is_a?(StringIO)
+                              StringIO.new
+                            else
+                              Tempfile.new('decompressed-data')
```
tagomoris: Add `require 'tempfile'` in this file.

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,89 @@
...
+        Zlib::GzipWriter.wrap(io) do |gz|
+          gz.write data
+        end
+
+        output_io || io.string
+      end
+
+      # compressed_data is String like `compress(data1) + compress(data2) + ... + compress(dataN)`
+      # https://www.ruby-forum.com/topic/971591#979503
+      def decompress(compressed_data = '', **kwargs)
```
tagomoris: Use keyword argument instead of handling "kwargs" hash.
`decompress(compressed_data = nil, input_io: nil, output_io: nil)`

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,89 @@
...
+        output_io = kwargs[:output_io]
+        input_io = kwargs[:input_io]
+
+        case
+        when input_io && output_io
+          io_decompress(input_io, output_io)
+        when compressed_data.empty?
```
tagomoris: What is this check for? Is it a variant of `string_decompress`?

tagomoris: It's better to use `nil` to detect unspecified argument.

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,89 @@
...
+          compressed_data
+        when output_io
```
tagomoris: Check `compressed_data` is valid string.

test/plugin/test_buffer_file_chunk.rb
```diff
@@ -768,4 +771,64 @@ def gen_chunk_path(prefix, unique_id)
       end
     end
   end
+
+    test '#open passes decompressed data to block when compress is gzip' do
```
tagomoris: incorrect indent.

tagomoris: What `#open` passes (to block) is not data, but io object to provide decompressed data.

test/plugin/test_buffer_file_chunk.rb
```diff
@@ -768,4 +771,64 @@ def gen_chunk_path(prefix, unique_id)
...
+
+    decomressed_data = c.open do |io|
+      v = io.read
+      assert_equal str, v
+      v
+    end
+    assert_equal str, decomressed_data
+  end
+
+    test '#open with compress option passes decompressed data to block when compress is gzip' do
```
tagomoris: Invalid indent.

tagomoris: "#open with compressed option" ?
+ not data, but io object

tagomoris:
For safe of 3rd party plugin authors, methods of `Chunk` (default chunk, not compressable one) should raise ArgumentError if `compressed: :gzip` is specified.

test/plugin/test_buffer_file_chunk.rb
```diff
@@ -768,4 +771,64 @@ def gen_chunk_path(prefix, unique_id)
...
+
+    assert_equal str, c.read
+    assert_equal compressed_str, c.read(compressed: :gzip)
+
+    io = StringIO.new
+    c.write_to(io)
+    assert_equal str, io.string
+  end
+
+  test '#write_to with compress option writes compressed data when compress is gzip' do
```
tagomoris: "decompressed option"

test/plugin/test_buffer_memory_chunk.rb
```diff
@@ -262,4 +265,64 @@ class BufferMemoryChunkTest < Test::Unit::TestCase
     assert_equal d3.to_json + "\n", lines[2]
     assert_equal d4.to_json + "\n", lines[3]
   end
+
+  test '#open passes decompressed data to block when compress is gzip' do
```
tagomoris: same comments with FileChunk.

test/plugin/test_compressable.rb
```diff
@@ -0,0 +1,55 @@
...
+
+class CompressableTest < Test::Unit::TestCase
+  include Fluent::Plugin::Compressable
+
+  sub_test_case '#compress' do
+    test 'compress data' do
+      str = 'text data'
+      compressed_str = compress(str)
+      assert_not_equal compressed_str, str
+    end
```
tagomoris: Please use `assert_equal` with fixed compressed data for specified (fixed) data source.
And, use a bit more longer data source :)

test/plugin/test_compressable.rb
```diff
@@ -0,0 +1,55 @@
+require_relative '../helper'
+require 'fluent/plugin/compressable'
+
+class CompressableTest < Test::Unit::TestCase
+  include Fluent::Plugin::Compressable
+
+  sub_test_case '#compress' do
```
tagomoris: Fixed data source and compressed one can be prepared as instance variables like `@src` & `@gzip` in `setup`.

lib/fluent/plugin/compressable.rb
ganmacs: This code checks the size of compressed_data(string) is 0 or not.

becuase `Zlib::GzipReader.new` raises Error, if `Zlib::GzipReader.new` recieved  `io` like `StringIO.new('')` , 
https://github.com/fluent/fluentd/pull/1172/files#diff-a36e94f98cbad7c4380816a06392d343R57

ganmacs:
@tagomoris 
I Updated.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -141,19 +146,74 @@ def purge
...
+          raise ArgumentError, '`compressed: gzip` can be used for Compressable module' if kwargs[:compressed] == :gzip
           open do |i|
             IO.copy_stream(i, io)
           end
         end
+
+        module Decompressable
+          include Fluent::Plugin::Compressable
+
+          def bytesize(**kwargs)
```
tagomoris: Do you mean that `chunk.bytesize` should return the bytesize of decompressed data size? It's different from actual bytesize.
With this definition, buffer chunks doesn't grow up to configured byte size (by `chunk_limit_size`).

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -141,19 +146,74 @@ def purge
...
+            if kwargs[:compressed] == :gzip
+              super
+            else
+              return unless commit # reset adding_bytes forcely
```
tagomoris: Don't do `commit` as side-effect... it breaks chunks and transactions.

lib/fluent/plugin/compressable.rb
```diff
@@ -0,0 +1,87 @@
...
+        output_io || io.string
+      end
+
+      # compressed_data is String like `compress(data1) + compress(data2) + ... + compress(dataN)`
+      # https://www.ruby-forum.com/topic/971591#979503
+      def decompress(compressed_data = nil, output_io: nil, input_io: nil)
+        case
+        when input_io && output_io
+          io_decompress(input_io, output_io)
+        when compressed_data.empty? || compressed_data.nil?
```
tagomoris: This line must be `compressed_data.nil? || compressed_data.empty?`. `nil` doesn't have `#empty?`.
And, there are no tests for this case (`#compress` called without `compressed_data`).

test/plugin/test_buffer_file_chunk.rb
```diff
@@ -768,4 +771,83 @@ def gen_chunk_path(prefix, unique_id)
...
   end
+
+  sub_test_case 'compressed buffer' do
+    setup do
+      @src = 'text data'
+      @gzipped_src = compress(@src)
+    end
+
+    test '#bytesize returns a size of the compressed string' do
+      src = 'テキストデータ text data'
```
tagomoris: Please avoid to use Japanese characters. It makes hard to edit these test cases for people not using Japanese.

test/plugin/test_compressable.rb
```diff
@@ -0,0 +1,71 @@
...
+      @src = 'text data for compressing' * 5
+      @gzipped_src = compress(@src)
+    end
+
+    test 'compress data' do
+      str = "this is a sample text" * 10
+      compressed_str = "\x1F\x8B\b\x00Y;\xC6W\x00\x03+\xC9\xC8,V\x00\xA2D\x85\xE2\xC4\xDC\x82\x9CT\x85\x92\xD4\x8A\x92\x92\xA1,\b\x00\xEF}\x9C\xF1\xD2\x00\x00\x00"
+      assert_not_equal compressed_str, compress(str)
+      assert_not_equal @gzipped_src, @src
+    end
```
tagomoris: It's better to have assertion to compare the length of src and gzipped... because it's about "compression".

test/plugin/test_compressable.rb
```diff
@@ -0,0 +1,71 @@
...
+    test 'write compressed data to IO with output_io option' do
+      io = StringIO.new
+      compress(@src, output_io: io)
+      assert_equal @gzipped_src, io.string
+    end
+  end
+
+  sub_test_case '#decompress' do
+    setup do
+      @src = 'text data'
```
tagomoris: Use longer data.

tagomoris:
I added some more comments :)

ganmacs:
Updated.

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -165,6 +166,18 @@ def write_to(io, **kwargs)
         module Decompressable
           include Fluent::Plugin::Compressable
 
+          def bytesize(**kwargs)
+            if kwargs[:decompressed] == :text
```
tagomoris: `decompressed` is this intended?

lib/fluent/plugin/buffer/chunk.rb
```diff
@@ -165,6 +166,18 @@ def write_to(io, **kwargs)
         module Decompressable
           include Fluent::Plugin::Compressable
 
+          def bytesize(**kwargs)
```
tagomoris: IMO the computing cost to return bytesize of decompressed data is too high, and it's reasonable to omit this feature.
Are there any use case?

tagomoris: I understood the use case in `out_forward`... Hmm

lib/fluent/plugin/out_forward.rb
```diff
@@ -304,7 +304,7 @@ def send_data(node, tag, chunk)
         sock.write tag.to_msgpack  # tag
 
         # beginRaw(size)
-        sz = chunk.size
+        sz = chunk.size(decompressed: :text)
```
tagomoris: We can implement this code with compressed chunks with just only one `Chunk#open`, without `Chunk#bytesize(compressed: :text)` like below:

```ruby
sock.write tag.to_msgpack

chunk.open do |chunk_io|
  sz = chunk_io.size
  sock.write [0xdb, sz].pack('CN')
  IO.copy_stream(chunk_io, sock)
end
```

This code repeats the implementation of `Chunk#write_to`, but it's acceptable because `out_forward` is just the core plugin of Fluentd.

tagomoris: The code in your patch generates tempfile twice. It's too heavy for almost workloads.

ganmacs:
@tagomoris 

Updated.

Remove Compressable#bytesize.

lib/fluent/plugin/out_forward.rb
```diff
@@ -303,21 +303,23 @@ def send_data(node, tag, chunk)
...
-        #  # raw 16
-        #  sock.write [0xda, sz].pack('Cn')
-        #else
-        # raw 32
-        sock.write [0xdb, sz].pack('CN')
-        #end
-
-        # writeRawBody(packed_es)
-        chunk.write_to(sock)
+        chunk.open(compressed: :text) do |chunk_io|
```
tagomoris: This is very trivial point (because we will have another pull-request to modify here), no need to update.

This `chunk.open` SHOULD NOT have `compressed: :text` argument.
At this point of codebase, this plugin doesn't know anything about compressed chunks and buffer compression. So the plugin should behave as before by doing just `chunk.open`.
Using `chunk_io.size` instead of `chunk.bytesize` is another problem. We found that it is wrong to use `chunk.bytesize` to get the size of content in chunks, and also found that `open{|io| io.size }` is right way.

tagomoris:
LGTM. Great updates to create awesome patch!

tagomoris:
@ganmacs Could you rebase this change on master once again for merge?

ganmacs:
@tagomoris 
rebased!

tagomoris:
I checked the actual behavior of Fluentd with `compress gzip`, and found a bug that compression doesn't work when output plugin has `#format` method... investigating.

tagomoris:
I confirmed that compression doesn't work well for output plugins with custom format by configuration file below and command line: `bundle exec bin/fluentd -c example/in_dummy_with_compression.conf`

```apache
# example/in_dummy_with_compression.conf
<source>
  @type dummy
  @label @main
  tag "test.data"
  size 2
  rate 10
  dummy {"message":"yaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaay"}
  auto_increment_key number
</source>

<label @main>
  <match test.data>
    @type buffered_stdout
    # <buffer>
    #   @type memory
    #   chunk_limit_size 1m
    #   flush_interval 10s
    #   compress gzip
    # </buffer>
    <buffer>
      @type file
      path "#{Dir.pwd}/compressed_buffers"
      flush_at_shutdown false
      chunk_limit_size 1m
      flush_interval 10s
      compress gzip
    </buffer>
  </match>
</label>
```

tagomoris:
@ganmacs Could you add the example configuration above, and add fixes for it?

tagomoris:
@ganmacs One more request: could you add tests like `test/plugin/test_output_as_buffered.rb` with compression option?
We can find bugs like we found today by such tests: `test/plugin/test_output_as_buffered_compress.rb` with tests for both of standard and custom formats.

ganmacs:
@tagomoris 
I fixed a bug that occurs when output plugin has `#format` method
I add  a test `test/plugin/test_output_as_buffered.rb`

tagomoris:
Okay, LGTM!

PR #1176, opened 2016-08-21
Fix to prevent overwriting reserved placeholder keys such as tag

sonots:
when same keys exist in record. The bug appeared only with on `enable_ruby false`

Same thing with https://github.com/sonots/fluent-plugin-record-reformer/pull/42

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -204,7 +204,9 @@ module Fluent::Plugin
             end
           elsif value.kind_of?(Hash) # record, etc
             value.each do |k, v|
-              placeholders.store("${#{k}}", v) # foo
+              unless placeholder_values.has_key?(k) # prevent overwriting reserved keys such as tag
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -194,6 +194,7 @@ def preprocess_map(value, force_stringify = false)
 
       def prepare_placeholders(placeholder_values)
         placeholders = {}
+        reserved_keys = Set.new(placeholder_values.keys)
```
repeatedly: Why is this `Set` needed?
Does `placeholder_values.include?(k)` have a problem?

sonots: No problem, but for readability.

repeatedly: Creating Set needs much resources.
I think plugin should keep lower resources.

sonots: Right, fixed

PR #1178, opened 2016-08-22
Be sure to detach one-shot timer watcher

okkez:
See https://github.com/fluent/fluentd/pull/1059/files#r75269945

### DIFF
--- a/lib/fluent/plugin_helper/timer.rb
+++ b/lib/fluent/plugin_helper/timer.rb
@@ -76,8 +76,12 @@ module Fluent
         rescue => e
           @log.error "Unexpected error raised. Stopping the timer.", title: @title, error: e
           @log.error_backtrace
-          self.detach
+          detach
           @log.error "Timer detached.", title: @title
+        ensure
+          if attached?
+            detach unless @repeating
+          end
         end
       end
     end
--- a/lib/fluent/plugin_helper/timer.rb
+++ b/lib/fluent/plugin_helper/timer.rb
@@ -66,6 +66,7 @@ module Fluent
         def initialize(title, interval, repeat, log, checker, &callback)
           @title = title
           @callback = callback
+          @repeat = repeat
           @log = log
           @checker = checker
           super(interval, repeat)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin_helper/timer.rb
```diff
@@ -76,8 +76,12 @@ def on_timer
         rescue => e
           @log.error "Unexpected error raised. Stopping the timer.", title: @title, error: e
           @log.error_backtrace
-          self.detach
+          detach
           @log.error "Timer detached.", title: @title
+        ensure
+          if attached?
+            detach unless @repeating
```
tagomoris: Is `@repeating` an instance variable of `TimerWatcher`?
I think it's too internal to use it, and it looks too fragile.
We should store `repeat` into our own instance variable in `#initialize` and use it.

okkez:
I've pushed change for comment @tagomoris

test/plugin_helper/test_timer.rb
```diff
@@ -97,4 +97,30 @@ class Dummy < Fluent::Plugin::TestBase
...
+    counter = 0
+    d1.timer_execute(:test, 1, repeat: false) do
+      counter += 1
+    end
+
+    watchers = d1._event_loop.watchers.reject {|w| w.is_a?(Fluent::PluginHelper::EventLoop::DefaultWatcher) }
+    assert_equal(1, watchers.size)
+    assert(watchers.first.attached?)
+
+    sleep 3
```
tagomoris: This `sleep 3` makes tests slow and unstable.
The better way to make it well is like this (not tested code):
```ruby
waiting_assertion = true
waiting_timer = true

counter = 0
d1.timer_execute(:test, 1, repeat: false) do
  sleep 0.1 while waiting_assertion
  counter += 1
  waiting_timer = false
end

# assertions about watchers here
waiting_assertion = false

sleep 0.1 while waiting_timer

assert_equal(1, counter)
# rest of assertions
```

okkez: Thank you for sample code.
I will fix test code later.

okkez: I've pushed a change for this comment.

tagomoris:
LGTM.

okkez:
Thanks!

PR #1179, opened 2016-08-24
Support compress in forward plugin

### LINKED ISSUE #504: support the compressing stream in forward and secure forward

ganmacs:
Because this PR depends on #1172, please merge this after merging #1172.

This PR closes #504.

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -274,6 +274,53 @@ module Fluent
     end
   end
 
+  class CompressedMessagePackEventStream < MessagePackEventStream
+    def initialize(data, cached_unpacker = nil, size = 0, unpacked_times: nil, unpacked_records: nil)
+      super
+      @decompressed_data = nil
+      @compressed_data = data
+    end
+
+    def dup
+      if @unpacked_times
+        self.class.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
+      else
+        self.class.new(@data.dup, nil, @size)
+      end
+    end
+
+    def empty?
+      @data = decompressed
+      super
+    end
+
+    def ensure_unpacked!
+      @data = decompressed
+      super
+    end
+
+    def each(&block)
+      @data = decompressed unless @unpacked_times
+      super
+    end
+
+    def to_msgpack_stream(time_int: false)
+      @data = decompressed
+      super
+    end
+
+    def to_compressed_msgpack_stream(time_int: false)
+      # time_int is always ignored because @data is always packed binary in this class
+      @compressed_data
+    end
+
+    private
+
+    def decompressed
+      @decompressed_data ||= decompress(@data)
+    end
+  end
+
   module ChunkMessagePackEventStreamer
     include MessagePackFactory::Mixin
     # chunk.extend(ChunkEventStreamer)
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -213,9 +213,9 @@ module Fluent
 
     def dup
       if @unpacked_times
-        MessagePackEventStream.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
+        self.class.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
       else
-        MessagePackEventStream.new(@data.dup, nil, @size)
+        self.class.new(@data.dup, nil, @size)
       end
     end
 
@@ -281,31 +281,23 @@ module Fluent
       @compressed_data = data
     end
 
-    def dup
-      if @unpacked_times
-        self.class.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
-      else
-        self.class.new(@data.dup, nil, @size)
-      end
-    end
-
     def empty?
-      @data = decompressed
+      ensure_decompressed!
       super
     end
 
     def ensure_unpacked!
-      @data = decompressed
+      ensure_decompressed!
       super
     end
 
     def each(&block)
-      @data = decompressed unless @unpacked_times
+      ensure_decompressed!
       super
     end
 
     def to_msgpack_stream(time_int: false)
-      @data = decompressed
+      ensure_decompressed!
       super
     end
 
@@ -316,8 +308,9 @@ module Fluent
 
     private
 
-    def decompressed
-      @decompressed_data ||= decompress(@data)
+    def ensure_decompressed!
+      return if @decompressed_data
+      @data = @decompressed_data = decompress(@data)
     end
   end
 
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -340,6 +340,7 @@ module Fluent
       def initialize(sender, server, failure:)
         @sender = sender
         @log = sender.log
+        @compress = sender.compress
 
         @name = server.name
         @host = server.host
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -140,6 +140,12 @@ module Fluent
         end
       end
 
+      if @compress == :gzip && @buffer.compress == :text
+        @buffer.compress = :gzip
+      elsif @compress == :text && @buffer.compress == :gzip
+        log.info "buffer is compressed.  If you also want to save the bandwidth of a network, Add `compress` configuration in <match>"
+      end
+
       if @nodes.empty?
         raise ConfigError, "forward output plugin requires at least one <server> is required"
       end
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -80,6 +80,9 @@ module Fluent
     desc 'Enable client-side DNS round robin.'
     config_param :dns_round_robin, :bool, default: false # heartbeat_type 'udp' is not available for this
 
+    desc 'Compress buffered data.'
+    config_param :compress, :enum, list: [:text, :gzip], default: :text
+
     config_section :security, required: false, multi: false do
       desc 'The hostname'
       config_param :self_hostname, :string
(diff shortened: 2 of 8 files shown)


lib/fluent/event.rb
```diff
@@ -267,6 +274,53 @@ def to_msgpack_stream(time_int: false)
...
   end
 
+  class CompressedMessagePackEventStream < MessagePackEventStream
+    def initialize(data, cached_unpacker = nil, size = 0, unpacked_times: nil, unpacked_records: nil)
+      super
+      @decompressed_data = nil
+      @compressed_data = data
+    end
+
+    def dup
```
tagomoris: This re-implementation looks no need.

tagomoris: Ah... It's better to fix original implementation `MessagePackEventStream#dup` to use `self.class` instead of re-implementing this method in subclass.

lib/fluent/event.rb
```diff
@@ -267,6 +274,53 @@ def to_msgpack_stream(time_int: false)
...
+      if @unpacked_times
+        self.class.new(@data.dup, nil, @size, unpacked_times: @unpacked_times, unpacked_records: @unpacked_records.map(&:dup))
+      else
+        self.class.new(@data.dup, nil, @size)
+      end
+    end
+
+    def empty?
+      @data = decompressed
```
tagomoris: No need when `@data` value already exists.

tagomoris: Calling `ensure_unpacked!` is better than this code for DRY.

lib/fluent/event.rb
```diff
@@ -267,6 +274,53 @@ def to_msgpack_stream(time_int: false)
...
+      super
+    end
+
+    def ensure_unpacked!
+      @data = decompressed
```
tagomoris: Take care about the case `@data` already exists.

lib/fluent/event.rb
```diff
@@ -267,6 +274,53 @@ def to_msgpack_stream(time_int: false)
...
+    end
+
+    def to_compressed_msgpack_stream(time_int: false)
+      # time_int is always ignored because @data is always packed binary in this class
+      @compressed_data
+    end
+
+    private
+
+    def decompressed
```
tagomoris: `#decompressed` is highly confusing name, because it's name is `-ed` but it does the same thing with `#decompress`.

tagomoris: `ensure_decompressed!` looks to work in just same way with `ensure_unpacked!`

lib/fluent/plugin/out_forward.rb
```diff
@@ -144,6 +147,13 @@ def configure(conf)
         log.info "adding forwarding server '#{name}'", host: host, port: port, weight: weight, plugin_id: plugin_id
       }
 
+      if @compress == :gzip && @buffer.compress == :text
+        @buffer.compress = :gzip
+      elsif @compress == :text && @buffer.compress == :gzip
+        @compress = :gzip
```
tagomoris: IMO there are some needs to compress buffer chunks and not to compress network traffic, for who want just to reduce disk usage.
So enabling `@compress` forcedly is too much for such users. It's just needed to notice users that this configuration (buffer compress, network traffic uncompressed) consumes more CPU time.

tagomoris:
I added review comments on code.

ganmacs:
@tagomoris 
I updated.

ganmacs:
@tagomoris 

I rebased and updated code.

lib/fluent/event.rb
```diff
@@ -267,6 +274,46 @@ def to_msgpack_stream(time_int: false)
...
+      super
+    end
+
+    def ensure_unpacked!
+      ensure_decompressed!
+      super
+    end
+
+    def each(&block)
+      ensure_decompressed! unless @unpacked_times
```
tagomoris: Unnecessary `unless`

lib/fluent/plugin/out_forward.rb
```diff
@@ -303,23 +312,25 @@ def send_data(node, tag, chunk)
...
+          #  sock.write [0xda, sz].pack('Cn')
+          #else
+          # raw 32
+          sock.write [0xdb, sz].pack('CN')
+
+          # writeRawBody(packed_es)
+          IO.copy_stream(chunk_io, sock)
+        end
+
+        option = { 'size' => chunk.size_of_events, 'compress' => @compress }
```
tagomoris: The key name should be `compressed`, because this key is to show that the payload is compressed, or not.

tagomoris: And, the value should be a string, not symbol. Do `@compress.to_s` to do it.
The default representation of Symbol in msgpack-ruby is `str`, but it can be overridden. So make it sure to feed String as value.

tagomoris:
@ganmacs could you rebase/update your change?

ganmacs:
@tagomoris 
Rebased 👌

test/plugin/test_in_forward.rb
```diff
@@ -465,6 +469,68 @@ def test_skip_invalid_event(data)
...
+      events = [
+        ["tag1", time, {"a"=>1}],
+        ["tag1", time, {"a"=>2}]
+      ]
+
+      # create compressed entries
+      entries = ''
+      events.each do |_tag, _time, record|
+        v = ''
+        [_time, record].to_msgpack(v)
```
tagomoris: It's already deprecated feature to assign String objects as an argument of `#to_msgpack`.
Please update the code as `v << [_time, record].to_msgpack`.

test/plugin/test_in_forward.rb
```diff
@@ -465,6 +469,68 @@ def test_skip_invalid_event(data)
...
```
tagomoris: ditto.

test/plugin/test_in_forward.rb
```diff
@@ -465,6 +469,68 @@ def test_skip_invalid_event(data)
     end
   end
 
+  class CompressedPackedForward < self
+    extend Fluent::Test::StartupShutdown
+
+    def test_set_compress_to_option
+      d = create_driver
+
+      time = Fluent::EventTime.parse("2011-01-02 13:14:15 UTC").to_i
```
tagomoris: `event_time` helper function is available (defined in `lib/fluent/test/helpers.rb`).
E.g., `event_time("2011-01-02 13:14:15 UTC")`.

test/plugin/test_out_forward.rb
```diff
@@ -107,6 +107,28 @@ def test_configure_no_server
     end
   end
 
+  def test_configure_text_compress
```
tagomoris: This test name doesn't claim anything. Does it show the default configuration value of `compress` option, right?

test/plugin/test_out_forward.rb
```diff
@@ -107,6 +107,28 @@ def test_configure_no_server
...
+    d = create_driver
+    assert_equal :text, d.instance.compress
+  end
+
+  def test_configure_gzip_compress
```
tagomoris: Same with the other above.

test/plugin/test_out_forward.rb
```diff
@@ -107,6 +107,28 @@ def test_configure_no_server
...
+    d = create_driver(CONFIG + %[compress gzip])
+    assert_equal :gzip, d.instance.compress
+    assert_equal :gzip, d.instance.buffer.compress
+  end
+
+  def test_configure_gzip_compress_in_buffer
```
tagomoris: More detailed test description is needed.
And, please add a assertion for log message.

test/test_event.rb
```diff
@@ -402,4 +402,98 @@ def setup
...
+    include Fluent::Plugin::Compressable
+
+    def setup
+      time = Engine.now
+      @times = [Fluent::EventTime.new(time.sec), Fluent::EventTime.new(time.sec + 1)]
+      @records = [{ 'k' => 'v1', 'n' => 1 }, { 'k' => 'v2', 'n' => 2 }]
+      @entries = ''
+      @times.zip(@records).each do |_time, record|
+        v = ''
+        [_time, record].to_msgpack(v)
```
tagomoris: ditto.

test/test_event.rb
```diff
@@ -402,4 +402,98 @@ def setup
...
+      # duplicated event stream still has unpacked objects and correct size
+      dupped = @es.dup
+      assert_equal 2, dupped.instance_eval{ @size }
+    end
+
+    test 'repeatable?' do
+      assert_true @es.repeatable?
+    end
+
+    test 'size' do
```
tagomoris: `CompressedMessagePackEventStream#size` (and some other methods) has side-effect to decompress contents, but these tests don't assert anything about it.
Please add assertions for that.

ganmacs:
@tagomoris 
Updated. Could you re-review this change?

test/test_event.rb
```diff
@@ -402,4 +402,115 @@ def setup
...
+    def setup
+      time = Engine.now
+      @times = [Fluent::EventTime.new(time.sec), Fluent::EventTime.new(time.sec + 1)]
+      @records = [{ 'k' => 'v1', 'n' => 1 }, { 'k' => 'v2', 'n' => 2 }]
+      @packed_record = ''
+      @entries = ''
+      @times.zip(@records).each do |_time, record|
+        v = ''
+        v << [_time, record].to_msgpack
+        @packed_record += v
```
tagomoris: `@packed_record << [_time, record].to_msgpack` looks to work well, and it's simple.

ganmacs: oops😨
Thanks. Fixed it.

tagomoris:
LGTM!

tagomoris:
I removed my LGTM, because I found the bug and missing tests, from warning of test results.
```
/home/travis/build/fluent/fluentd/lib/fluent/plugin/out_forward.rb:444: warning: instance variable @compress not initialized
/home/travis/build/fluent/fluentd/lib/fluent/plugin/out_forward.rb:452: warning: instance variable @compress not initialized
```

This warning shows that Node instances doesn't have instance variable named as `@compress`.
Your patch doesn't feed `@compress` of output plugin into `ForwardOutput::Node` objects.
So compression transferring is always false in this code.
And there are no tests about it.

ganmacs:
@tagomoris 
I fixed a bug that `compress` option  is always `:text` and added some tests.

tagomoris:
@ganmacs It's ok, but not enough. It's needed to check whether `in_forward` (instance of `target_input_driver`) creates `CompressedMessagePackEventStream` or not.

ganmacs:
@tagomoris I added a test that `in_forward` creates `CompressedMessagePackEventStream` or not.

tagomoris:
LGTM. Thank you for great patch!

Issue #1129, opened 2016-07-28
in_forward RuntimeError after logrotate


Yanson:
After upgrading to td-agent 2.3.2-0 (fluentd **0.12.26**) I am seeing some servers crash each morning.

It appears to happen after the logroate each day.

```
$ cat /etc/logrotate.d/td-agent
/var/log/td-agent/td-agent.log {
  daily
  rotate 30
  compress
  delaycompress
  notifempty
  create 640 td-agent td-agent
  sharedscripts
  postrotate
    pid=/var/run/td-agent/td-agent.pid
    test -s $pid && kill -USR1 "$(cat $pid)"
  endscript
}
```

```
$ sudo grep SIGUSR1 /var/log/td-agent/td-agent.log.1
2016-07-28 06:29:05 +0000 [debug]: fluent/supervisor.rb:408:block in install_supervisor_signal_handlers: fluentd supervisor process get SIGUSR1
2016-07-28 06:29:05 +0000 [debug]: fluent/supervisor.rb:552:block in install_main_process_signal_handlers: fluentd main process get SIGUSR1
```

```
$ sudo grep RuntimeError /var/log/td-agent/td-agent.log
2016-07-28 06:29:56 +0000 [error]: plugin/in_forward.rb:107:rescue in run: unexpected error error=#<RuntimeError: can't add a new key into hash during iteration> error_class=RuntimeError
```

Note, there is no stack trace associated with the error.

Nothing special about the forward config:
```
<source>
  type forward
</source>
```

I have not been able to reproduce this by sending `USR1` myself.

repeatedly:
Does this happen with only fluentd v0.12.26?

repeatedly:
Hmm... error seems to happen inside Cool.io's run method.
I will check the code.

Yanson:
I don't know if it only happens in that version. I only know it started happening since upgrading to td-agent 2.3.2-0.
It is definitely ongoing, today it brought down 3 of 8 servers. Note, we are running three regions and only the high traffic region was affected (the other 6 servers across two regions remain OK).

Is there another version you would like me to try?

Is there a way to get the stack trace? From the code it appears it should print it but it never shows.

repeatedly:
I checked the cool.io code but I can't find "can't add a new key into hash during iteration" code.

> Note, there is no stack trace associated with the error.

If you okay, could you paste entire logs here?

Yanson:
[...]

PR #1181, opened 2016-08-25
Protect buffer operations in force_flush

repeatedly:
force_flush is called from another thread, so accessing buffer should be guarded.

See: https://github.com/fluent/fluentd/issues/1129

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -291,7 +291,8 @@ module Fluent
     #end
 
     def enqueue_buffer(force = false)
-      @buffer.keys.each {|key|
+      buf_keys = @buffer.synchronize { @buffer.keys }
+      buf_keys.each { |key|
         @buffer.push(key)
       }
     end
@@ -587,7 +588,8 @@ module Fluent
 
     def enqueue_buffer(force = false)
       if force
-        @buffer.keys.each {|key|
+        buf_keys = @buffer.synchronize { @buffer.keys }
+        buf_keys.each { |key|
           @buffer.push(key)
         }
       else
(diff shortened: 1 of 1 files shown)


lib/fluent/output.rb
```diff
@@ -291,7 +291,8 @@ def format_stream(tag, es)
     #end
 
     def enqueue_buffer(force = false)
-      @buffer.keys.each {|key|
+      buf_keys = @buffer.synchronize { @buffer.keys }
```
tagomoris: Do `@buffer.keys.dup` in critical sections.
Your code iterates the object in `@buffer` itself, and doesn't protect anything in `synchronize`.

repeatedly: `@buffer.keys` returns an array from internal `@map` hash keys.
What does '`@buffer` itself' mean?

tagomoris: I tested on pry:
```
hash = {a: 1, b: 2, c: 3, d: 4}
=> {:a=>1, :b=>2, :c=>3, :d=>4}
t = Thread.new{ i = 0; hash.each{|k,v| p(k: v); sleep(5) if i == 1; i += 1 } }; sleep 1; hash[:e] = 5; p hash; sleep 3; p hash
{:k=>1}
{:k=>2}
RuntimeError: can't add a new key into hash during iteration
from (pry):12:in `__pry__'
{:k=>3}
{:k=>4}
```

tagomoris: Ah, no, it's not what i want to test...

tagomoris: I confirmed that you're right. Sorry to bother you.

PR #1184, opened 2016-08-26
Call formatter_create in #configure instead of #start in filter_stdout

cosmo0920:
See: https://github.com/fluent/fluentd/pull/1059#discussion_r76362903

### DIFF
--- a/lib/fluent/plugin/filter_stdout.rb
+++ b/lib/fluent/plugin/filter_stdout.rb
@@ -37,10 +37,6 @@ module Fluent::Plugin
       @formatter = formatter_create(conf: @config.elements('format').first, default_type: DEFAULT_FORMAT_TYPE)
     end
 
-    def start
-      super
-    end
-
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/filter_stdout.rb
```diff
@@ -34,10 +34,10 @@ class StdoutFilter < Filter
     def configure(conf)
       compat_parameters_convert(conf, :inject, :formatter)
       super
+      @formatter = formatter_create(conf: @config.elements('format').first, default_type: DEFAULT_FORMAT_TYPE)
     end
 
     def start
```
tagomoris: It's possible to remove whole `#start` because it's just calling `super`.

cosmo0920: I've removed this.

tagomoris:
LGTM

cosmo0920:
Thanks!

PR #1186, opened 2016-08-26
config_param options to specify deprecated/obsoleted parameters

tagomoris:
This change adds 2 options for `config_param`.
And I marked `host`, `port` of `out_forward` without `<server>` sections as obsoleted.

### deprecated
Unrecommended parameter, and will be removed in future.
If users specify this parameter, Fluentd warns it in logs.

### obsoleted
Configuration parameter which is already removed.
If users specify this parameter, Fluentd stops with an error (a kind of configuration errors).

### DIFF
--- a/lib/fluent/config/section.rb
+++ b/lib/fluent/config/section.rb
@@ -126,6 +126,7 @@ module Fluent
             logger.error "config error in:\n#{conf}" if logger # logger should exist, but somethimes it's nil (e.g, in tests)
             raise ConfigError, "'<#{proxy.name} ARG>' section requires argument" + section_stack
           end
+          # argument should NOT be deprecated... (argument always has a value: '')
         end
 
         proxy.params.each_pair do |name, defval|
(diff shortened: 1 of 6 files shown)


tagomoris:

lib/fluent/config/section.rb
```diff
@@ -138,6 +139,20 @@ def self.generate(proxy, conf, logger, plugin_class, stack = [])
...
             section_params[varname] = self.instance_exec(val, opts, name, &block)
+
+            # Source of definitions of deprecated/obsoleted:
+            # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features
+            #
+            # Deprecated: These deprecated features can still be used, but should be used with caution
+            #             because they are expected to be removed entirely sometime in the future.
+            # Obsoleted: These obsolete features have been entirely removed from JavaScript and can no longer be used.
+            if opts[:deprecated]
+              logger.warn "'#{name}' paramenter isn't recommended to use now."
```
repeatedly: typo: parameter

repeatedly:
One idea. How about adding "message" like parameter to show better logs?.

```rb
 config_param :host, :string, default: nil, obsoleted: true, message: "Use <server> section instead"
```

result:

```
 'host' parameter is already removed. Don't use it. Use <server> section instead"
```

tagomoris:
@repeatedly When is that parameter used? only for `deprecated` and `obsoleted`?
If so, it may be better to change value type of these options from true/false to string of message to show. Isn't it?

repeatedly:
> it may be better to change value type of these options from true/false to string of message to show. Isn't it?

LGTM 👍

repeatedly:
Good. After test passed, please merge.

tagomoris:
Merged.

PR #1191, opened 2016-08-29
Port parser filter from fluent-plugin-parser. fix #1189

### LINKED ISSUE #1189: Port parser filter of fluent-plugin-parser into core plugins.
Plugin repository is here: https://github.com/tagomoris/fluent-plugin-parser

`parser` filter is for parsing a field of record and update the record by using parsed result.
This is useful when forwarding device can't generate proper structured record, e.g. docker logging driver, embedded device, etc.
fluentd and docker combination is now widely used and lots of users use `parser` filter on production.

Porting parser filter has several merits.
- Reduce installation step in many cases
- Avoid parser filter breakage by fluentd version update
- Use parser filter to check updating parsers doesn't break existing API.

So time to port parser filter.

repeatedly:
Changes from original code:
- Use v0.14 namespace
- Use v0.14 API (Now working)
- Remove warnings like unused variable

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -1,9 +1,30 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/time'
+require 'fluent/config/error'
 require 'fluent/plugin/filter'
-require 'fluent/plugin/parser'
+require 'fluent/plugin_helper/parser'
+require 'fluent/plugin_helper/compat_parameters'
 
 module Fluent::Plugin
   class ParserFilter < Filter
     Fluent::Plugin.register_filter('parser', self)
+
+    helpers :parser, :compat_parameters
 
     config_param :key_name, :string
     config_param :reserve_data, :bool, default: false
@@ -16,20 +37,22 @@ module Fluent::Plugin
 
     attr_reader :parser
 
-    def initialize
-      super
-      require 'time'
-    end
-
     def configure(conf)
+      compat_parameters_convert(conf, :parser)
+      parser_config = conf.elements('parse').first
+      unless parser_config
+        raise Fluent::ConfigError, "<parse> section is required"
+      end
+      unless parser_config["@type"]
+        raise Fluent::ConfigError, "parse/@type is required"
+      end
+
       super
 
-      @parser = Fluent::TextParser.new
-      @parser.estimate_current_event = false
-      @parser.configure(conf)
-      if !@time_parse && @parser.parser.respond_to?("time_key=".to_sym)
+      @parser = parser_create(usage: 'filter_parser', conf: parser_config)
+      if !@time_parse && @parser.respond_to?("time_key=".to_sym)
         # disable parse time
-        @parser.parser.time_key = nil
+        @parser.time_key = nil
       end
 
       self
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -58,60 +58,66 @@ module Fluent::Plugin
       self
     end
 
-    def filter_stream(tag, es)
-      new_es = Fluent::MultiEventStream.new
-      es.each do |time, record|
-        raw_value = record[@key_name]
-        if raw_value.nil?
-          log.warn "#{@key_name} does not exist" unless @ignore_key_not_exist
-          new_es.add(time, handle_parsed(tag, record, time, {})) if @reserve_data
-          next
+    FAILED_RESULT = [nil, nil].freeze # reduce allocation cost 
+
+    def filter_with_time(tag, time, record)
+      raw_value = record[@key_name]
+      if raw_value.nil?
+        log.warn "#{@key_name} does not exist" unless @ignore_key_not_exist
+        if @reserve_data
+          return time, handle_parsed(tag, record, time, {})
+        else
+          return FAILED_RESULT
         end
-        begin
-          @parser.parse(raw_value) do |t, values|
+      end
+      begin
+        @parser.parse(raw_value) do |t, values|
+          if values
+            t ||= time
+            r = handle_parsed(tag, record, t, values)
+            return t, r
+          else
+            log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
+            if @reserve_data
+              t = time
+              r = handle_parsed(tag, record, time, {})
+              return t, r
+            else
+              return FAILED_RESULT
+            end
+          end
+        end
+      rescue Fluent::TextParser::ParserError => e
+        log.warn e.message unless @suppress_parse_error_log
+      rescue ArgumentError => e
+        if @replace_invalid_sequence
+          unless e.message.index("invalid byte sequence in") == 0
+            raise
+          end
+          replaced_string = replace_invalid_byte(raw_value)
+          @parser.parse(replaced_string) do |t, values|
             if values
               t ||= time
               r = handle_parsed(tag, record, t, values)
-              new_es.add(t, r)
+              return t, r
             else
               log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
               if @reserve_data
                 t = time
                 r = handle_parsed(tag, record, time, {})
-                new_es.add(t, r)
-              end
-            end
-          end
-        rescue Fluent::TextParser::ParserError => e
-          log.warn e.message unless @suppress_parse_error_log
-        rescue ArgumentError => e
-          if @replace_invalid_sequence
-            unless e.message.index("invalid byte sequence in") == 0
-              raise
-            end
-            replaced_string = replace_invalid_byte(raw_value)
-            @parser.parse(replaced_string) do |t, values|
-              if values
-                t ||= time
-                r = handle_parsed(tag, record, t, values)
-                new_es.add(t, r)
+                return t, r
               else
-                log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
-                if @reserve_data
-                  t = time
-                  r = handle_parsed(tag, record, time, {})
-                  new_es.add(t, r)
-                end
+                return FAILED_RESULT
               end
             end
-          else
-            raise
           end
-        rescue => e
-          log.warn "parse failed #{e.message}" unless @suppress_parse_error_log
+        else
+          raise
         end
+      rescue => e
+        log.warn "parse failed #{e.message}" unless @suppress_parse_error_log
+        return FAILED_RESULT
       end
-      new_es
     end
 
     private
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -1,0 +1,114 @@
+require 'fluent/plugin/filter'
+require 'fluent/plugin/parser'
+
+module Fluent::Plugin
+  class ParserFilter < Filter
+    Fluent::Plugin.register_filter('parser', self)
+
+    config_param :key_name, :string
+    config_param :reserve_data, :bool, default: false
+    config_param :inject_key_prefix, :string, default: nil
+    config_param :replace_invalid_sequence, :bool, default: false
+    config_param :hash_value_field, :string, default: nil
+    config_param :suppress_parse_error_log, :bool, default: false
+    config_param :time_parse, :bool, default: true
+    config_param :ignore_key_not_exist, :bool, default: false
+
+    attr_reader :parser
+
+    def initialize
+      super
+      require 'time'
+    end
+
+    def configure(conf)
+      super
+
+      @parser = Fluent::TextParser.new
+      @parser.estimate_current_event = false
+      @parser.configure(conf)
+      if !@time_parse && @parser.parser.respond_to?("time_key=".to_sym)
+        # disable parse time
+        @parser.parser.time_key = nil
+      end
+
+      self
+    end
+
+    def filter_stream(tag, es)
+      new_es = Fluent::MultiEventStream.new
+      es.each do |time, record|
+        raw_value = record[@key_name]
+        if raw_value.nil?
+          log.warn "#{@key_name} does not exist" unless @ignore_key_not_exist
+          new_es.add(time, handle_parsed(tag, record, time, {})) if @reserve_data
+          next
+        end
+        begin
+          @parser.parse(raw_value) do |t, values|
+            if values
+              t ||= time
+              r = handle_parsed(tag, record, t, values)
+              new_es.add(t, r)
+            else
+              log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
+              if @reserve_data
+                t = time
+                r = handle_parsed(tag, record, time, {})
+                new_es.add(t, r)
+              end
+            end
+          end
+        rescue Fluent::TextParser::ParserError => e
+          log.warn e.message unless @suppress_parse_error_log
+        rescue ArgumentError => e
+          if @replace_invalid_sequence
+            unless e.message.index("invalid byte sequence in") == 0
+              raise
+            end
+            replaced_string = replace_invalid_byte(raw_value)
+            @parser.parse(replaced_string) do |t, values|
+              if values
+                t ||= time
+                r = handle_parsed(tag, record, t, values)
+                new_es.add(t, r)
+              else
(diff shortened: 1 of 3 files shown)


repeatedly:
@tagomoris We can't set `estimate_current_event` via `parser` helper API before `configure`. Do you know the workaround for this?

repeatedly:
Reaming task is replacing `filter_stream` with `filter_with_time`.

repeatedly:
Don't set `estimate_current_event = false` but test passed. Hmm...

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,688 @@
...
+require 'timecop'
+require 'fluent/test/driver/filter'
+require 'fluent/plugin/filter_parser'
+
+class ParserFilterTest < Test::Unit::TestCase
+  def setup
+    Fluent::Test.setup
+    @tag = 'test'
+    @default_time = Time.parse('2010-05-04 03:02:01 UTC')
+    Timecop.freeze(@default_time)
```
cosmo0920: Should we restore Time with `Timecop.return` in `#teardown`?

repeatedly: right

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,137 @@
...
+      @parser = parser_create(usage: 'filter_parser', conf: parser_config)
+      if !@time_parse && @parser.respond_to?("time_key=".to_sym)
+        # disable parse time
+        @parser.time_key = nil
+      end
+
+      self
+    end
+
+    def filter_stream(tag, es)
```
cosmo0920: It is just a comment. Feel free to ignore this.
I think that we can use `#filter_with_time` instead of `#filter_stream` here.
Adopting to filter pipeline should be next task?

lib/fluent/plugin_helper/compat_parameters.rb
```diff
@@ -48,14 +48,21 @@ module CompatParameters
...
         "delimiter"   => "delimiter",
+        "keep_time_key" => "keep_time_key",
+        "null_empty_string" => "null_empty_string", # ValuesParser
+        "null_value_pattern" => "null_value_pattern", # ValuesParser
         "json_parser"      => "json_parser", # JSONParser
         "label_delimiter"  => "label_delimiter", # LabeledTSVParser
         "format_firstline" => "format_firstline", # MultilineParser
         "message_key"      => "message_key", # NoneParser
         "with_priority"    => "with_priority", # SyslogParser
+        "types" => "types", # TypeConverter
```
tagomoris: We should use `hash` type for such parameter, instead of `types`, `types_delimiter` and `types_label_delimiter`.
It's same for `keys` with `array`... (I'll do it in another pull-request).

repeatedly: So I should remove this diff from this PR?
Removing it causes test failure.

tagomoris: No, it's ok to include it for now.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,143 @@
...
+    Fluent::Plugin.register_filter('parser', self)
+
+    helpers :parser, :compat_parameters
+
+    config_param :key_name, :string
+    config_param :reserve_data, :bool, default: false
+    config_param :inject_key_prefix, :string, default: nil
+    config_param :replace_invalid_sequence, :bool, default: false
+    config_param :hash_value_field, :string, default: nil
+    config_param :suppress_parse_error_log, :bool, default: false
```
tagomoris: It's no need to use, but to use error stream instead.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,143 @@
...
+    config_param :time_parse, :bool, default: true
+    config_param :ignore_key_not_exist, :bool, default: false
+
+    attr_reader :parser
+
+    def configure(conf)
+      compat_parameters_convert(conf, :parser)
+      parser_config = conf.elements('parse').first
```
tagomoris: Don't check `<parse>` sections and `@type` in these. Use `config_section :parse` instead.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,143 @@
...
+      unless parser_config
+        raise Fluent::ConfigError, "<parse> section is required"
+      end
+      unless parser_config["@type"]
+        raise Fluent::ConfigError, "parse/@type is required"
+      end
+
+      super
+
+      @parser = parser_create(usage: 'filter_parser', conf: parser_config)
```
tagomoris: Is it better to remove `time_parse` option and set `time_key` in default?

lib/fluent/plugin/filter_parser.rb
repeatedly: It seems good. I will rewrite it.

lib/fluent/plugin/filter_parser.rb
repeatedly: I follow this way. Should we also change this lines?https://github.com/fluent/fluentd/blob/388eefdd4dcd70b20254c0c9bc874b585a2ab3e8/lib/fluent/plugin/in_tail.rb#L81

tagomoris: It's very special case, because that plugin scans `format#{n}` from parse section by itself.

lib/fluent/plugin/filter_parser.rb
tagomoris: I meant that:

* remove `time_parse` option
* add `config_set_default :time_key, nil` in `parse` section

With this change, we perfectly delegate "parse time or not" to parser plugins.

tagomoris: But now, I'm thinking it's better NOT to set `@parser.time_key` anytime, and:

* rename `time_parse` configuration option to `reserve_time` (consistent with `reserve_data`)
* ignore parsed time (from `@parser.parse`) when `@reserve_time` is true

It makes the plugin code highly cleaner and simpler.
Parser plugins have these own default values for `time_key`, and I think it's better not to touch them from parser plugin.

repeatedly:
Rebased and apply reviews.

repeatedly:
@tagomoris Could you check this patch again?

lib/fluent/plugin_helper/compat_parameters.rb
```diff
@@ -52,12 +52,13 @@ module CompatParameters
...
         "types_delimiter" => nil,
         "types_label_delimiter" => nil,
-        "null_value_pattern" => "null_value_pattern",
-        "null_empty_string" => "null_empty_string",
         "keys" => "keys", # CSVParser, TSVParser (old ValuesParser)
         "time_key"    => "time_key",
         "time_format" => "time_format",
         "delimiter"   => "delimiter",
+        "keep_time_key" => "keep_time_key",
+        "null_empty_string" => "null_empty_string", # ValuesParser
```
tagomoris: `ValuesParser` doesn't exist now.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+  #   add_prefix    parsed
+  #   key_name      message
+  #   format        /^(?<x>.)(?<y>.) (?<time>.+)$/
+  #   time_format   %Y%m%d%H%M%S
+  #   reserve_data  true
+  # ]
+  def test_filter
+    d1 = create_driver(CONFIG)
+    time = Time.parse("2012-01-02 13:14:15").to_i
+    d1.run do
```
tagomoris: `default_tag` keyword argument is available here: `d1.run(default_tag: @tag)`

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+  #   remove_prefix test
+  #   add_prefix    parsed
+  #   key_name      message
+  #   format        /^(?<x>.)(?<y>.) (?<time>.+)$/
+  #   time_format   %Y%m%d%H%M%S
+  #   reserve_data  true
+  # ]
+  def test_filter
+    d1 = create_driver(CONFIG)
+    time = Time.parse("2012-01-02 13:14:15").to_i
```
tagomoris: use EventTime instance as time for v0.14 plugins: `event_time("2012-01-02 13:14:15")`

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+      d1.feed(@tag, time, {'message' => '34 20120402182100'})
+      d1.feed(@tag, time, {'message' => '56 20120402182100'})
+      d1.feed(@tag, time, {'message' => '78 20120402182101'})
+      d1.feed(@tag, time, {'message' => '90 20120402182100'})
+    end
+    filtered = d1.filtered
+    assert_equal 5, filtered.length
+
+    first = filtered[0]
+    assert_equal Time.parse("2012-04-02 18:20:59").to_i, first[0]
```
tagomoris: Is it integer not EventTime?

test/plugin/test_filter_parser.rb
tagomoris: If this is a test code especially for integer time, test case name should show it.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+    time = Fluent::EventTime.from_time(@default_time) # EventTime emit test
+    d2.run do
+      d2.feed(@tag, time, {'data' => '12 20120402182059'})
+      d2.feed(@tag, time, {'data' => '34 20120402182100'})
+    end
+    filtered = d2.filtered
+    assert_equal 2, filtered.length
+
+    first = filtered[0]
+    assert_equal time, first[0]
```
tagomoris: Use `assert_equal_event_time`, because `EventTime#==` compares just seconds (for compatibility reason, i guess).

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+    time = Time.parse("2012-04-02 18:20:59").to_i
+    d3.run do
+      d3.feed(@tag, time, {'data' => '12 20120402182059'})
+      d3.feed(@tag, time, {'data' => '34 20120402182100'})
+      d3.feed(@tag, time, {'data' => 'xy 20120402182101'})
+    end
+    filtered = d3.filtered
+    assert_equal 2, filtered.length
+
+    d3x = create_driver(%[
```
tagomoris: Split test cases with descriptive names.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+    </parse>
+  ]
+
+  class DefaultSuppressParseErrorLogTest < self
+    setup do
+      # default(disabled) 'suppress_parse_error_log' is not specify
+      @d = create_driver(CONFIG_DEFAULT_SUPPRESS_PARSE_ERROR_LOG)
+    end
+
+    def test_raise_exception
```
tagomoris: This test tests `emit_error_event`, not error raised.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+      flexmock(@d.instance.router).should_receive(:emit_error_event).
+        with(String, Integer, Hash, ParserError.new("pattern not match with data '#{INVALID_MESSAGE}'")).once
+      @d.run do
+        @d.feed(@tag, Fluent::EventTime.now.to_i, {'message' => INVALID_MESSAGE})
+      end
+    end
+
+    def test_nothing_raised
```
tagomoris: ditto.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+    assert_equal 1, filtered.length
+    assert_nil filtered[0][1]['data']
+    assert_equal 'bar', filtered[0][1]['foo']
+  end
+
+  # suppress_parse_error_log test
+  INVALID_MESSAGE = 'foo bar'
+  VALID_MESSAGE   = 'col1=foo col2=bar'
+
+  def test_parser_error_warning
```
tagomoris: There are no assertions about warning.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+    </parse>
+  ]
+
+  class DisabledSuppressParseErrorLogTest < self
+    setup do
+      # disabled 'suppress_parse_error_log'
+      @d = create_driver(CONFIG_DISABELED_SUPPRESS_PARSE_ERROR_LOG)
+    end
+
+    def test_raise_exception
```
tagomoris: ditto.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+      flexmock(@d.instance.router).should_receive(:emit_error_event).
+        with(String, Integer, Hash, ParserError.new("pattern not match with data '#{INVALID_MESSAGE}'")).once
+      @d.run do
+        @d.feed(@tag, Fluent::EventTime.now.to_i, {'message' => INVALID_MESSAGE})
+      end
+    end
+
+    def test_nothing_raised
```
tagomoris: ditto.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,730 @@
...
+  CONFIG_ENABELED_SUPPRESS_PARSE_ERROR_LOG = %[
+    key_name message
+    suppress_parse_error_log true
+    <parse>
+      @type regexp
+      expression /^col1=(?<col1>.+) col2=(?<col2>.+)$/
+    </parse>
+  ]
+
+  class EnabledSuppressParseErrorLogTest < self
```
tagomoris: Migrated plugin code doesn't have `suppress_parse_error_log` option.
This test code is outdated, right?

tagomoris:
Parser plugin helper can instantiate two or more parsers. So we can add "secondary" parser for the values which don't match the first parser. It's nice to have to replace `fluent-plugin-multi-format-parser`.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,136 @@
...
+          if values
+            t = if @reserve_time
+                  time
+                else
+                  t.nil? ? time : t
+                end
+            r = handle_parsed(tag, record, t, values)
+            return t, r
+          else
+            router.emit_error_event(tag, time, record, Fluent::Plugin::Parser::ParserError.new("pattern not match with data '#{raw_value}'"))
```
repeatedly: I'm not sure but we can't say all existing parsers don't return nil for time and value.
So keeping this behaviour is better for users.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,136 @@
...
+        values = Hash[values.map { |k, v| [@inject_key_prefix + k, v] }]
+      end
+      r = @hash_value_field ? {@hash_value_field => values} : values
+      if @reserve_data
+        r = r ? record.merge(r) : record
+      end
+      r
+    end
+
+    def replace_invalid_byte(string)
```
repeatedly: right.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,136 @@
...
+            else
+              return FAILED_RESULT
+            end
+          end
+        end
+      rescue Fluent::Plugin::Parser::ParserError => e
+        router.emit_error_event(tag, time, record, e)
+        return FAILED_RESULT
+      rescue ArgumentError => e
+        if @replace_invalid_sequence
```
repeatedly: good point. I will rewrite these lines.

repeatedly:
Applied reviews.

repeatedly:
@tagomoris It is a good idea > secondary parsers.
It's a TODO after merged this PR.

repeatedly:
This patch is not the cause of test failure.

repeatedly:
@tagomoris check again.

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,667 @@
...
+    </parse>
+  ]
+  CONFIG_INVALID_BYTE = CONFIG_NOT_REPLACE + %[
+    replace_invalid_sequence true
+  ]
+  def test_filter_invalid_byte
+    invalid_utf8 = "\xff".force_encoding('UTF-8')
+
+    d = create_driver(CONFIG_NOT_REPLACE)
+    d.run(shutdown: false) do
```
tagomoris: Are there any reason to specify `shutdown: false`?

test/plugin/test_filter_parser.rb
```diff
@@ -0,0 +1,667 @@
...
+  ]
+  CONFIG_PASS_SAME_RECORD = CONFIG_NOT_IGNORE + %[
+    reserve_data true
+  ]
+  def test_filter_key_not_exist
+    d = create_driver(CONFIG_NOT_IGNORE)
+    flexmock(d.instance.router).should_receive(:emit_error_event).
+      with(String, Integer, Hash, ArgumentError.new("data does not exist")).once
+    assert_nothing_raised {
+      d.run(shutdown: false) do
```
tagomoris: ditto

tagomoris:
I commented about just one point (twice). Others look good to me.

repeatedly:
Applied all reviews. After test passed, I will merge this PR.

PR #1192, opened 2016-08-30
fix code depending ServerEngine 2.0

tagomoris:


tagomoris:
Let me merge this, because failing test isn't related with this change, and it's a known pattern.

PR #1193, opened 2016-08-30
Updates dependencies and reduce warnings

tagomoris:
This change is to reduce warnings when we run `rake test`.
Almost all warnings are about uninitialized variables, and some others are about unused variables.

I did 2 changes at the same time:

### Updating dependencies

There seems no reason to specify minor version dependency for some gems. I updated these to depends on (recent) major versions. I think this operation doesn't make any problems, because all of these are development dependencies.

I also remove `json` runtime dependency, to use bundled json of Ruby runtime.
Specifying `json` version is bad practice, because it makes problem between ruby version limitations and other module's dependencies

### Fix implementation of `in_monitor_agent`

The previous implementation depends on the behavior of `eval` for string expression, NoMethodError for nil objects and rescuing all errors (and do nothing).
It hides any types of unexpected errors... very bad practice.

I rewrote that code with `instance_exec` with proc objects, and `throw-catch` for explicit skip for objects in unexpected states. It shows warnings for unexpected errors in logs.

### DIFF
--- a/test/plugin/test_in_monitor_agent.rb
+++ b/test/plugin/test_in_monitor_agent.rb
@@ -156,10 +156,11 @@ EOC
     end
 
     test "emit" do
+      port = unused_port
       d = create_driver("
   @type monitor_agent
   bind '127.0.0.1'
-  port 24200
+  port #{port}
   tag monitor
   emit_interval 1
 ")
--- a/test/plugin/test_in_monitor_agent.rb
+++ b/test/plugin/test_in_monitor_agent.rb
@@ -159,7 +159,7 @@ EOC
       d = create_driver("
   @type monitor_agent
   bind '127.0.0.1'
-  port #{@port}
+  port 24200
   tag monitor
   emit_interval 1
 ")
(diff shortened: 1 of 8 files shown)


tagomoris:
@repeatedly I will include this patch for `v0.14.3`. Could you take a glance on it?

test/plugin/test_in_monitor_agent.rb
```diff
@@ -159,7 +159,7 @@ def test_configure
       d = create_driver("
   @type monitor_agent
   bind '127.0.0.1'
-  port #{@port}
+  port 24200
```
repeatedly: Why use fixed port number?
`@port = unused_port` is better?

tagomoris: It's just smallest diff to reduce warnings.
But you're right.

tagomoris: Updated with `unused_port`.

tagomoris:
I rebased on current master branch, and add a commit to edit upper limits of dependencies.

repeatedly:
LGTM

Issue #1115, opened 2016-07-20
Test fails occasionally


tagomoris:
This failure occurs occasionally on Travis CI environment.
It shows that `Time.now` is called, but `Time` is a rr mock object without any expectations about methods.
Output doesn't contain any information about which test case includes this test.

```
===============================================================================
Failure:
  On subject Time,
  unexpected method invocation:
    now()
  expected invocations:
test_level(PluginLoggerTest)
/Users/travis/build/fluent/fluentd/lib/fluent/log.rb:306:in `dump_stacktrace'
/Users/travis/build/fluent/fluentd/lib/fluent/log.rb:259:in `error_backtrace'
/Users/travis/build/fluent/fluentd/lib/fluent/plugin/output.rb:990:in `rescue in flush_thread_run'
/Users/travis/build/fluent/fluentd/lib/fluent/plugin/output.rb:963:in `flush_thread_run'
/Users/travis/build/fluent/fluentd/lib/fluent/plugin/output.rb:369:in `block (2 levels) in start'
/Users/travis/build/fluent/fluentd/lib/fluent/plugin_helper/thread.rb:66:in `block in thread_create'
===============================================================================
/Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.2.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnermediator.rb:49:in `ensure in run': On subject Time, (RR::Errors::DoubleNotFoundError)
unexpected method invocation:
  now()
expected invocations:
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.2.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnermediator.rb:50:in `run'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.2.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunner.rb:40:in `start_mediator'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.2.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunner.rb:25:in `start'
	from /Users/travis/build/fluent/fluentd/vendor/bundle/ruby/2.2.0/gems/test-unit-3.1.9/lib/test/unit/ui/testrunnerutilities.rb:24:in `run'
[...]

PR #1194, opened 2016-08-30
Use timecop instead of stub in test_log

cosmo0920:
Related to #1115.

Using stub to freeze time is not good manner.
We should use `Timecop` to do this. :cop:


tagomoris:
LGTM.

cosmo0920:
Thanks!

PR #1200, opened 2016-09-01
Merge buffered and non buffered outputs

tagomoris:
There are two buffer plugins to output events into `null` and `stdout`.
But Fluentd v0.14 API allows plugins to provide both in a plugin implementation.

This change is to merge these plugins into an implementation.


repeatedly:
`example/out_buffered_null.conf` should be renamed to `example/out_null.conf`?

lib/fluent/plugin/out_null.rb
```diff
@@ -18,10 +18,48 @@
 
 module Fluent::Plugin
   class NullOutput < Output
+    # This plugin is for tests of non-buffered/buffered plugins
     Fluent::Plugin.register_output('null', self)
+    Fluent::Plugin.register_output('buffered_null', self)
```
repeatedly: Why `buffered_null` is needed?
This is introduced since v0.14 and there is no production use.
So we don't need to keep the compatibility.

lib/fluent/plugin/out_stdout.rb
```diff
@@ -19,30 +19,67 @@
 module Fluent::Plugin
   class StdoutOutput < Output
     Fluent::Plugin.register_output('stdout', self)
+    Fluent::Plugin.register_output('buffered_stdout', self)
```
repeatedly: ditto

lib/fluent/plugin/out_null.rb
tagomoris: Okay.

tagomoris:
Update for file/plugin names

PR #1203, opened 2016-09-05
Port parser filter to v0.12

repeatedly:
v0.12 version of https://github.com/fluent/fluentd/pull/1191.
Original parser filter is for v0.12 so no need additional changes for `filter_parser.rb`.

Fix test to use `Timecop`.

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -1,0 +1,111 @@
+require 'fluent/parser'
+
+class Fluent::ParserFilter < Fluent::Filter
+  Fluent::Plugin.register_filter('parser', self)
+
+  config_param :key_name, :string
+  config_param :reserve_data, :bool, default: false
+  config_param :inject_key_prefix, :string, default: nil
+  config_param :replace_invalid_sequence, :bool, default: false
+  config_param :hash_value_field, :string, default: nil
+  config_param :suppress_parse_error_log, :bool, default: false
+  config_param :time_parse, :bool, default: true
+  config_param :ignore_key_not_exist, :bool, default: false
+
+  attr_reader :parser
+
+  def initialize
+    super
+    require 'time'
+  end
+
+  def configure(conf)
+    super
+
+    @parser = Fluent::TextParser.new
+    #@parser.estimate_current_event = false
+    @parser.configure(conf)
+    if !@time_parse && @parser.parser.respond_to?("time_key=".to_sym)
+      # disable parse time
+      @parser.parser.time_key = nil
+    end
+
+    self
+  end
+
+  def filter_stream(tag, es)
+    new_es = Fluent::MultiEventStream.new
+    es.each do |time,record|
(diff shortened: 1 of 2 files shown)


repeatedly:
Test passed. I will merge this patch later.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,111 @@
...
+  config_param :hash_value_field, :string, default: nil
+  config_param :suppress_parse_error_log, :bool, default: false
+  config_param :time_parse, :bool, default: true
+  config_param :ignore_key_not_exist, :bool, default: false
+
+  attr_reader :parser
+
+  def initialize
+    super
+    require 'time'
```
tagomoris: Move this `require` to the top of file, and remove `#initialize`.
This lazy requirement has no positive things.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def configure(conf)
```
tagomoris: Use `compat_parameter` plugin helpers to make it possible to use both of `<parser>` section and flat parameters.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,111 @@
...
+    super
+
+    @parser = Fluent::TextParser.new
```
tagomoris: Use `parser` plugin helpers to set plugin logger into parser instance, and values of `system` section.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -0,0 +1,111 @@
...
+    @parser.configure(conf)
+    if !@time_parse && @parser.parser.respond_to?("time_key=".to_sym)
+      # disable parse time
+      @parser.parser.time_key = nil
+    end
+
+    self
+  end
+
+  def filter_stream(tag, es)
```
tagomoris: Is it for multi output records from `parser.parse` to use `#filter_stream` instead of `#filter` or `#filter_with_time`?

repeatedly: This is for v0.12. v0.12 filter doesn't have v0.14 APIs.

tagomoris: Oh, I missed it... but wait, are you going to merge this into `v0.12` at first, and then forward port into `master`?

repeatedly: No. v0.14 and v0.12 are significantly different, so cherry-pick is useless.
I don't have plan to port this change into master.
#1191 is for v0.14.

tagomoris: I got confused from similar pull-requests... Use label for `v0.12`.
Ok, got it.

repeatedly: Oops. I forgot to add label...

repeatedly:
Applied review.

PR #1207, opened 2016-09-06
Symmetric time parse and format

tagomoris:
This patch is to provide consistent API about time and tag, handling these on records.

Provided things:
* symmetric class structure: Fluent::TimeFormatter and Fluent::TimeParser with `#call` method (and `#parse`/`#format`)
* consistent API to parse / format time with configurable format and timezone
* common configuration parameters for parser/formatter plugins and inject/extract plugin helpers

With this patch, formatter plugins can get TimeFormatter instances anytime via `time_formatter_create` method with configured format/timezone by users in `<format>` section.
Parser plugins also does it via `time_parser_create`.
Input/Output plugins can use `inject`/`extract` plugin helpers with just same configuration parameter names. It provides consistent how-to-use for users.


tagomoris:

lib/fluent/plugin_helper/extract.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      module ExtractParams
+        include Fluent::Configurable
+        config_section :extract, required: false, multi: false, param_name: :extract_config do
+          config_param :tag_key, :string, default: nil
+          config_param :time_key, :string, default: nil
+          config_param :time_type, :enum, list: [:float, :unixtime, :string], default: :float
+
+          Fluent::TimeMixin::TIME_PARAMETERS.each do |name, type, opts|
```
repeatedly: Why don't you use `include Fluent::TimeMixin::TimeParameters`?

tagomoris: Because that mixin adds parameters on the top namespace, not in `<extract>` sections.

repeatedly: I see. implementation limitation...

repeatedly:
Off topic. Cache machanizm of parser / formatter becomes overhead for msec / nsec times.
We should improve this point, e,g separate TimeParser into TimeParser and CachedTimeParser, in the future.

test/test_time_parser.rb
```diff
@@ -0,0 +1,195 @@
+require_relative 'helper'
+require 'fluent/test'
+require 'fluent/time'
+
+class TimeParserTest < ::Test::Unit::TestCase
+  def setup
+    Fluent::Test.setup
+  end
+
+  def test_call_with_parse
```
repeatedly: What does 'call' mean?

test/test_time_parser.rb
```diff
@@ -0,0 +1,195 @@
...
+    assert(parser.parse('28/Feb/2013:12:00:00:123456789 +0900').is_a?(Fluent::EventTime))
+
+    time = event_time('28/Feb/2013:12:00:00:123456789 +0900', format: '%d/%b/%Y:%H:%M:%S:%N %z')
+    assert_equal_event_time(time, parser.parse('28/Feb/2013:12:00:00:123456789 +0900'))
+  end
+
+  def test_parse_with_invalid_argument
+    parser = Fluent::TimeParser.new
+
+    [[], {}, nil, true, 10000, //, ->{}, '', :symbol].each { |v|
```
repeatedly: Using `data` is good for better test result.

test/test_time_parser.rb
tagomoris: I have no idea: it's just moved from `test/plugin/test_parser_time.rb` to there.

test/test_time_parser.rb
tagomoris: ditto. IMO, CI green without any modification on tests is most important for such refactoring-ish tasks.

repeatedly:
Other LGTM.

PR #1209, opened 2016-09-06
Mark secret parameters of forward plugins as secret

tagomoris:


tagomoris:
@repeatedly Could you check please?
As far as I seen, secret parameters of `out_forward` are already marked as secret: https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward.rb#L87

repeatedly:
LGTM

PR #1223, opened 2016-09-09
Add Fluent namespace to ConfigError in v0.14 API based plugins

repeatedly:
One user reports following error:

```
2016-09-09 09:01:49 +0000 [error]: fluent/supervisor.rb:567:rescue in main_process: unexpected error error="uninitialized constant Fluent::Plugin::GrepFilter::ConfigError
```


tagomoris:
👍  - but, it's better to add tests to `assert_raised` for these configuration errors.

repeatedly:
>  but, it's better to add tests to assert_raised for these configuration errors.

I agree but this changes several plugins across plugin types.
I think separating PR is good for me.
We have a room for test improvement, use `data` instead of `each` and more.

tagomoris:
Ah, you're right.

PR #1227, opened 2016-09-11
Fix log messages and its conditions

repeatedly:


tagomoris:
Tests are not fixed with modified log messages.

repeatedly:
Yes because I want feedback for more better message.

lib/fluent/plugin/storage_local.rb
```diff
@@ -56,7 +56,7 @@ def configure(conf)
         elsif @path
           # ok
         else # @_plugin_id_configured is true
-          raise NotImplementedError, "implement this feature later with system_config"
+          raise NotImplementedError, "implement this feature later with system_config. Add <storage> section for file store or remove @id for memory store temporary"
```
tagomoris: This block is for "Make root path for buffers and storages" (root path is based on `@id`) described in wiki. So it's wrong to tell users to remove `@id`.
https://github.com/fluent/fluentd/wiki/Changes-proposed-for-v0.14.x#multi-processes--cpu-cores

repeatedly: Yeah, I know. The problem is users now hard to use <storage> without file because storage plugin doesn't have store type option.
Currently, existing users hit this error because they use following configuration for v0.12.

```
<source>
  @type dummy
  @id input_dummy
  # ...
</source>
```

I think this configuration should work with v0.14 without error. My change is temporary. Better way is fallback to on memory store, not raising `NotImplementedError`.

repeatedly: There are 2 ways for `in_dummy`.

- Add store_type like option to specify file or on-memory and log message in this line
- Fallback to on-memory storage, not raising `NotImplementedError`

tagomoris: IMO, falling back to on-memory storage is better than adding configuration parameters for temporal issue.

repeatedly:
Update test for EventRouter.

repeatedly:
Test passed excluding unstable test. Re-run test.

repeatedly:
Test are passed.

repeatedly:
If no concern, I will merge it.

PR #1232, opened 2016-09-13
Fix monitor agent compatibility

### LINKED ISSUE #1226: monitor_agent does strange output in v0.14
monitor_agent's ouput in v0.14 is weird and different from in v0.12. With folloing configuration:

```
<source>
  @type monitor_agent
</source>

<match nginx>
  @type copy
  <store>
    @type stdout
  </store>
</match>

<match nginx2>
  @type stdout
</match>
```

in v0.12.29

```
$ curl http://localhost:24220/api/plugins
plugin_id:object:3fe8b81d9c10	plugin_category:input	type:monitor_agent	output_plugin:false	retry_count:
plugin_id:object:3fe8b857d768	plugin_category:output	type:copy	output_plugin:true	retry_count:
plugin_id:object:3fe8b857c46c	plugin_category:output	type:stdout	output_plugin:true	retry_count:
plugin_id:object:3fe8b81f4038	plugin_category:output	type:stdout	output_plugin:true	retry_count:
```

in v0.14.6

```
$ curl http://localhost:24220/api/plugins
plugin_id:object:3fa09f9fad74	plugin_category:input	type:monitor_agent	output_plugin:false	retry_count:
plugin_id:object:3fa09f53ae4c	plugin_category:unknown	type:copy	output_plugin:false	retry_count:0
plugin_id:object:3fa09f4ed32c	plugin_category:output	type:stdout	output_plugin:true	buffer_total_queued_size:0	retry_count:0
plugin_id:object:3fa09f4ed32c	plugin_category:output	type:stdout	output_plugin:true	buffer_total_queued_size:0	retry_count:0
plugin_id:object:3fa09fa5e5e0	plugin_category:output	type:stdout	output_plugin:true	buffer_total_queued_size:0	retry_count:0
```

There seems to be 3 issues.

1. `copy` plugin is categorized as `unknown`
2. `stdout` plugin inside `copy` plugin is duplicate
3. `stdout` plugin is not buffered output plugin but displays metrics for buffered plugin such as `buffered_total_queue_size`

tagomoris:
fix #1226.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -326,6 +326,7 @@ module Fluent
           @buffering = prefer_buffered_processing
           if !@buffering && @buffer
             @buffer.terminate # it's not started, so terminate will be enough
+            @buffer = nil
           end
         end
 
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -326,6 +326,8 @@ module Fluent
           @buffering = prefer_buffered_processing
           if !@buffering && @buffer
             @buffer.terminate # it's not started, so terminate will be enough
+            # At here, this plugin works as non-buffered plugin.
+            # Un-assign @buffer not to show buffering metrics (e.g., in_monitor_agent)
             @buffer = nil
           end
         end
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -262,31 +262,17 @@ module Fluent::Plugin
       array.concat Fluent::Engine.root_agent.inputs
 
       # get all output plugins
-      Fluent::Engine.root_agent.outputs.each { |o|
-        MonitorAgentInput.collect_children(o, array)
-      }
+      array.concat Fluent::Engine.root_agent.outputs
+
       # get all filter plugins
-      Fluent::Engine.root_agent.filters.each { |f|
-        MonitorAgentInput.collect_children(f, array)
-      }
+      array.concat Fluent::Engine.root_agent.filters
+
       Fluent::Engine.root_agent.labels.each { |name, l|
         # TODO: Add label name to outputs / filters for identifing plugins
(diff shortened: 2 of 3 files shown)


tagomoris:
@repeatedly Please review this too.

lib/fluent/plugin/output.rb
```diff
@@ -326,6 +326,7 @@ def start
           @buffering = prefer_buffered_processing
           if !@buffering && @buffer
             @buffer.terminate # it's not started, so terminate will be enough
+            @buffer = nil
```
repeatedly: It seems error prune.
Using configured value is better for detecting output plugin is un-buffered or buffered.

tagomoris: Non-buffered output plugin (without `#write` method) doesn't have values on `@buffer`.
And, output plugin instances which runs this line is determined as non-buffered plugin. So the `@buffer` value is unassigned.
No bad/wrong things there are.

repeatedly: I see. If so, adding comment is better like "This line is needed for detecting buffer or non-buffer, e.g. used by in_monitor_agent".
Only this line, it looks like for GC.

And we should move this `:buffer` reader from `for test` to above `attr_reader` lines.
https://github.com/fluent/fluentd/blob/24e15f3cfcb46029b1dee080bf50b576f573553e/lib/fluent/plugin/output.rb#L145

tagomoris: No. `@buffering` shows this plugin is buffered or non-buffered. Unassigning `@buffer` is just to avoid side-effect in `in_monitor_agent`.

tagomoris: And `@buffer` is still only for tests.

repeatedly: I understand the situation. Adding comment is better with "Avoid side-effect that in_monitor_agent shows non-buffered plugins as buffered."

tagomoris: Done.

repeatedly:
LGTM

PR #1233, opened 2016-09-13
raise ConfigError when same label appears twice, or no match appears in label

tagomoris:
This change fixes the 2 `<label>` related bugs:

### no `<match>` sections in `<label>` isn't reported

```apache
<label @test>
  @type stdout
</label>
```

This is clearly wrong configuration, and warning "no patterns matched" appears always for `@test` label.
This change raises configuration error for such configuration.

### same label configured twice

```apache
<label @test>
  # first
</label>
<label @test>
  # last
</label>
```

Currently, Fluentd uses the last configuration for same labels without any errors nor warnings.
This configuration is clearly a misconfiguration. So this change is to make to report such configurations as errors.

### DIFF
--- a/lib/fluent/label.rb
+++ b/lib/fluent/label.rb
@@ -27,6 +27,14 @@ module Fluent
 
     attr_accessor :root_agent
 
+    def configure(conf)
+      super
+
+      if conf.elements('match').size == 0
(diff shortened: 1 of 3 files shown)


tagomoris:
@repeatedly review this please.

repeatedly:
Good catch!

PR #1235, opened 2016-09-15
Enable log rotate for log

### LINKED ISSUE #1231: Enable log rotation for `Fluent::Log`
As the log file increases, it becomes difficult to handle the log file.
So, I want to enable the log rotation for `Fluent::Log` to keep log files down to a manageable size.

I'll add these command line arguments to `fluentd` command.

* `--log-rotate-age`

The number of old log files to keep, or frequency of rotation (`daily`, `weekly` or `monthly`). The default value is 5. it is used when you set a value to `--log-rotate-size` and don't set a value to `--log-rotate-age`.If you set 0 as a value of `--log-rotate-age`,  the logger will do no log rotation.

* `--log-rotate-size`

Maximum logfile size (only applies when `log-rotate-age` is a number).  The default value is 1M.
it is used when you set a value to `--log-rotate-age` and don't set a value to `--log-rotate-size`.

ganmacs:
This PR closes #1231.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -81,6 +81,23 @@ op.on('-o', '--log PATH', "log file path") {|s|
   opts[:log_path] = s
 }
 
+op.on('--log-rotate-age AGE', 'generations to keep rotated log files') {|age|
+  rotate_ages = %w(daily weekly monthly)
+  if rotate_ages.include?(age)
+    opts[:log_rotate_age] = age
+  else
+    begin
+      opts[:log_rotate_age] = Integer(age)
+    rescue
+      usage "log-rotate-age should be #{rotate_ages.join(', ')} or a number"
+    end
+  end
+}
+
+op.on('--log-rotate-size SIZE', 'sets the size to rotate log files') {|s|
+  opts[:log_rotate_size] = s.to_i
+}
+
 op.on('-i', '--inline-config CONFIG_STRING', "inline config which is appended to the config file on-the-fly") {|s|
   opts[:inline_config] = s
 }
@@ -270,4 +287,3 @@ if opts[:supervise]
 else
   Fluent::Supervisor.new(opts).run_worker
 end
-
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -94,7 +94,7 @@ op.on('--log-rotate-age AGE', 'generations to keep rotated log files') {|age|
   end
 }
 
-op.on('--log-rotate-size SIZE', 'sets the size to rotate log files') {|s|
+op.on('--log-rotate-size BYTES', 'sets the byte size to rotate log files') {|s|
   opts[:log_rotate_size] = s.to_i
 }
 
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -81,14 +81,14 @@ op.on('-o', '--log PATH', "log file path") {|s|
   opts[:log_path] = s
 }
 
+ROTATE_AGE = %w(daily weekly monthly)
 op.on('--log-rotate-age AGE', 'generations to keep rotated log files') {|age|
-  rotate_ages = %w(daily weekly monthly)
-  if rotate_ages.include?(age)
+  if ROTATE_AGE.include?(age)
     opts[:log_rotate_age] = age
   else
     begin
       opts[:log_rotate_age] = Integer(age)
-    rescue
+    rescue TypeError
       usage "log-rotate-age should be #{rotate_ages.join(', ')} or a number"
     end
   end
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -237,7 +237,9 @@ module Fluent
 
       log_opts = {suppress_repeated_stacktrace: suppress_repeated_stacktrace}
       logger_initializer = Supervisor::LoggerInitializer.new(
-        log_path, log_level, chuser, chgroup, log_rotate_age, log_rotate_size, log_opts
+        log_path, log_level, chuser, chgroup, log_opts,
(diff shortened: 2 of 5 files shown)


lib/fluent/supervisor.rb
```diff
@@ -295,42 +299,48 @@ def self.load_config(path, params = {})
...
-        logger = ServerEngine::DaemonLogger.new(@io, dl_opts)
+        logger = ServerEngine::DaemonLogger.new(logdev, dl_opts)
         $log = Fluent::Log.new(logger, @opts)
         $log.enable_color(false) if @path
         $log.enable_debug if @level <= Fluent::Log::LEVEL_DEBUG
       end
 
       def stdout?
-        @io == STDOUT
+        $log.out == STDOUT
```
tagomoris: `$log` might be replaced with other values by code outside.

lib/fluent/supervisor.rb
```diff
@@ -295,42 +299,48 @@ def self.load_config(path, params = {})
...
       end
 
       def reopen!
         if @path && @path != "-"
-          @io.reopen(@path, "a")
+          $log.out.reopen(@path, "a")
```
tagomoris: ditto.

lib/fluent/supervisor.rb
tagomoris: Using `@io` should be kept to use here and `#reopen!`.

lib/fluent/supervisor.rb
```diff
@@ -295,42 +299,48 @@ def self.load_config(path, params = {})
     end
 
     class LoggerInitializer
-      def initialize(path, level, chuser, chgroup, opts)
+      def initialize(path, level, chuser, chgroup, log_rotate_age, log_rotate_size, opts)
```
tagomoris: It's better to use keyword arguments to add options..

lib/fluent/command/fluentd.rb
```diff
@@ -81,6 +81,23 @@
   opts[:log_path] = s
 }
 
+op.on('--log-rotate-age AGE', 'generations to keep rotated log files') {|age|
+  rotate_ages = %w(daily weekly monthly)
+  if rotate_ages.include?(age)
+    opts[:log_rotate_age] = age
+  else
+    begin
+      opts[:log_rotate_age] = Integer(age)
```
tagomoris: Are there any reason to use `Integer()` instead of `to_i`?
IMO `to_i` is widely used and normal for most ruby programmer.

lib/fluent/command/fluentd.rb
```diff
@@ -81,6 +81,23 @@
   opts[:log_path] = s
 }
 
+op.on('--log-rotate-age AGE', 'generations to keep rotated log files') {|age|
+  rotate_ages = %w(daily weekly monthly)
```
tagomoris: Use constant.

lib/fluent/command/fluentd.rb
```diff
@@ -81,6 +81,23 @@
...
+  if rotate_ages.include?(age)
+    opts[:log_rotate_age] = age
+  else
+    begin
+      opts[:log_rotate_age] = Integer(age)
+    rescue
```
tagomoris: Use `rescue` with error class name (TypeError?).

test/test_log.rb
```diff
@@ -404,6 +409,43 @@ def test_level_reload
...
+
+  DAY_SEC = 60 * 60 * 24
+  data(
+    rotate_daily_age: ['daily', 100000, DAY_SEC + 1],
+    rotate_weekly_age: ['weekly', 100000, DAY_SEC * 7 + 1],
+    rotate_monthly_age: ['monthly', 100000, DAY_SEC * 31 + 1],
+    rotate_size: [1, 100, 0, '0'],
+  )
+  def test_log_with_logdevio(expected)
+    old = ENV['TZ']
```
tagomoris: Use `with_timezone` test helper method defined in `lib/fluent/test/helpers.rb` instead of doing by itself.

lib/fluent/command/fluentd.rb
```diff
@@ -81,6 +81,23 @@
...
+  else
+    begin
+      opts[:log_rotate_age] = Integer(age)
+    rescue TypeError
+      usage "log-rotate-age should be #{rotate_ages.join(', ')} or a number"
+    end
+  end
+}
+
+op.on('--log-rotate-size SIZE', 'sets the size to rotate log files') {|s|
```
tagomoris: `BYTES` is better to show the value is an integer, not "size" (in fluentd's configuration file, e.g., 256k, 8m).

lib/fluent/supervisor.rb
tagomoris: Aren't there fixes for this?

test/test_log.rb
```diff
@@ -404,6 +409,73 @@ def test_level_reload
...
+      msg2 = 'b' * 101
+      log.info msg2
+      c = File.read(path)
+
+      assert_match msg2, c
+      assert_not_equal msg, c
+    end
+  end
+
+  def test_log_rotates_specifed_size_with_logdevio
```
tagomoris: This test should check '.1' file is NOT created, as a test for `rotate_age`.

lib/fluent/supervisor.rb
ganmacs: Sorry. I forgot it. 
I updated `@io` to `@logdev`.  Because the name of `@io` instance variable  is changed here (https://github.com/fluent/fluentd/pull/1235/files#diff-dbe0e1ec4079138e48ca6a4d7c7248f9R316).

test/test_log.rb
ganmacs: I added. https://github.com/fluent/fluentd/pull/1235/commits/7bf4ab42be90a976d5f08806707b080a91e9f700

tagomoris:
Okay, LGTM. Thank you!

PR #1236, opened 2016-09-16
add skip_accessor to supported options list for config_param

tagomoris:


tagomoris:

repeatedly:
When is this option useful?

tagomoris:
This option is already available - it's helpful to avoid redefine/overwrite Ruby's method of plugin objects.
It's already used for `format` configuration parameter of some plugins, not to overwrite `Kernel#write`.
And I re-found this option not to overwrite `Object#method` in fluent-plugin-bigquery, by using `method` configuration parameter (which configures RPC method insert/load to be used).

lib/fluent/config/configure_proxy.rb
```diff
@@ -208,9 +208,21 @@ def overwrite_defaults(other) # other is owner plugin's corresponding proxy
...
-      def option_value_type!(name, opts, key, klass)
-        if opts.has_key?(key) && !opts[key].is_a?(klass)
-          raise ArgumentError, "#{name}: #{key} must be a #{klass}, but #{opts[key].class}"
+      def option_value_type!(name, opts, key, klass=nil, type: nil)
+        if opts.has_key?(key)
+          if klass && !opts[key].is_a?(klass)
+            raise ArgumentError, "#{name}: #{key} must be a #{klass}, but #{opts[key].class}"
+          end
+          case type
+          when :boolean
```
repeatedly: `bool` is better for consistency?
`config_param` uses `:bool` for boolean type parameter.

tagomoris: It's something corresponding to Ruby classes, not Fluentd's configuration value types, for `TrueClass` and `FalseClass`.
The document of Ruby expresses such values as "boolean expressions".
https://docs.ruby-lang.org/en/2.3.0/FalseClass.html

PR #1239, opened 2016-09-23
Use compat instead of strict to follow json/yajl way. fix #1230

### LINKED ISSUE #1230: Oj in strict mode cannot format Symbols into json
As I noted on the comments below, setting Oj's default options in `parser_json` affects to the behavior of Oj in `format_json`.
https://github.com/fluent/fluentd/pull/1147#issuecomment-246307029

It's too strict, and a kind of bug.
We should use `compat` mode of Oj globally (need to consider about Oj's default options in both of formatter and parser).

repeatedly:


repeatedly:
I can write messy test like below but it seems not unittest...

```
diff --git a/test/plugin/test_formatter_json.rb b/test/plugin/test_formatter_json.rb
index 5f80dd1..f073ea2 100644
--- a/test/plugin/test_formatter_json.rb
+++ b/test/plugin/test_formatter_json.rb
@@ -17,7 +17,7 @@ class JsonFormatterTest < ::Test::Unit::TestCase
   end
 
   def record
-    {'message' => 'awesome'}
+    {'message' => 'awesome', :message2 => :awesome2}
   end
 
   data('oj' => 'oj', 'yajl' => 'yajl')
@@ -27,4 +27,14 @@ class JsonFormatterTest < ::Test::Unit::TestCase
 
     assert_equal("#{Yajl.dump(record)}\n", formatted)
   end
+
+  def test_format_with_symbol_value
+    d = create_driver('json_parser' => 'oj')
+    require 'fluent/plugin/parser_json'
+    parser = Fluent::Plugin::JSONParser.new
+    parser.configure(config_element())
+    formatted = d.instance.format(tag, @time, record)
+
+    assert_equal("#{Yajl.dump(record)}\n", formatted)
+  end
 end
```

lib/fluent/plugin/parser_json.rb
```diff
@@ -38,7 +38,7 @@ def configure(conf)
         begin
           raise LoadError unless @json_parser == 'oj'
           require 'oj'
-          Oj.default_options = {bigdecimal_load: :float, mode: :strict}
+          Oj.default_options = {bigdecimal_load: :float, mode: :compat}
```
tagomoris: This default options are different from one in `formatter_json`.
You must make a common Oj default options as a constant, and share it in everywhere.

repeatedly: > You must make a common Oj default options as a constant,

This is the trick but it seems acceptable.

tagomoris:
LGTM.

PR #1243, opened 2016-09-27
Use timecop instead of flexmock

okkez:
This can avoid occasional stack level too deep error when CI on
AppVeyor.

https://ci.appveyor.com/project/fluent/fluentd/build/1564/job/ed9y9dky4xtcucd8


tagomoris:
LGTM

PR #1244, opened 2016-09-27
Ignore duplicated inotify events

okkez:
This change can avoid occasional failure against `test_rotate_xxx`.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -248,6 +248,10 @@ module Fluent::Plugin
 
     # refresh_watchers calls @tails.keys so we don't use stop_watcher -> start_watcher sequence for safety.
     def update_watcher(path, pe)
+      unless pe.read_inode == @pf[path].read_inode
+        log.trace "Skip update_watcher because watcher has been already updated by other inotify event"
+        return
+      end
       rotated_tw = @tails[path]
       @tails[path] = setup_watcher(path, pe)
       close_watcher_after_rotate_wait(rotated_tw) if rotated_tw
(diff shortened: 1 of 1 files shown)


repeatedly:
"test: should output plain text when compress option is default(text)(FileOutputSecondaryTest::write)" test fails but in_tail test seems stable.
We re-run test more times.

tagomoris:
I requested rebuild on TravisCI, and found the failure.
https://travis-ci.org/fluent/fluentd/jobs/163013161

okkez:
Hmm, this approach may be wrong...
I will try another approach.

tagomoris:
LGTM. I'm requesting rebuild (2nd time).
@repeatedly ?

repeatedly:
I will check this change manually.
After that, I wll merge it.

repeatedly:
@okkez Could you rebase this patch?
master branch includes several fixes for other unstable tests.

okkez:
Rebased and pushed changes.

okkez:
Failed test on AppVeyor is not related to my changes.

lib/fluent/plugin/in_tail.rb
```diff
@@ -248,6 +248,10 @@ def stop_watchers(paths, immediate = false, unwatched = false)
 
     # refresh_watchers calls @tails.keys so we don't use stop_watcher -> start_watcher sequence for safety.
     def update_watcher(path, pe)
+      unless pe.read_inode == @pf[path].read_inode
```
repeatedly: Maybe, `@pf` check is needed?
@pf is `nil` when `pos_file` option is not specified.

okkez: Yes. Thank you!
I will add a change for this.

repeatedly:
Thanks!

PR #1245, opened 2016-09-27
Remove duplicate `require 'fluent/log'`

ganmacs:
https://github.com/fluent/fluentd/blob/master/test/helper.rb#L43
https://github.com/fluent/fluentd/blob/master/test/helper.rb#L47


tagomoris:
LGTM!
Failing tests are not related on this change.

PR #1249, opened 2016-09-27
Add nanoseconds to stdout plugin

luhn:
Now that Fluentd supports nanosecond-resolution timestamps as of v0.14,
it follows that the stdout output plugin should print the full
timestamp.

This is a very simple patch, instead of using the default `to_s`
formatting, we call `strftime` with a custom formatting.

Right now it always prints the nanoseconds, even on second-resolution
timestamps.  This means second-resolution timestamps will display like
`2013-06-27 12:34:56.000000000 -0700`, which may be undesirable.  If so,
we can adjust the implement to print a condensed formatting for
second-resolution timestamps.

### DIFF
--- a/lib/fluent/plugin/out_stdout.rb
+++ b/lib/fluent/plugin/out_stdout.rb
@@ -23,6 +23,7 @@ module Fluent::Plugin
     helpers :inject, :formatter, :compat_parameters
 
     DEFAULT_FORMAT_TYPE = 'json'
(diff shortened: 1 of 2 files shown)


tagomoris:
This change looks reasonable, and good to me.
Changing default format is breaking change and undesirable for many cases... but it's about `out_stdout`. I believe it's ok.

@repeatedly ?

repeatedly:
LGTM.

Changing out_stdout's result should not affect any production.

PR #1254, opened 2016-10-04
Support numeric times in time parser formatter

tagomoris:
There's many use cases to parse/format times in unix time or floating point values.
It should be supported by built-in TimeParser/TimeFormatter and common configurations.

This change also implements timezone control in extracting timestamp placeholders in compat_parameters.

### DIFF
--- a/lib/fluent/plugin_helper/compat_parameters.rb
+++ b/lib/fluent/plugin_helper/compat_parameters.rb
@@ -213,8 +213,6 @@ module Fluent
         # TODO: warn obsolete parameters if these are deprecated
         attr = compat_parameters_copy_to_subsection_attributes(conf, FORMATTER_PARAMS)
 
-        # TODO: make a utility method in TimeFormatter to handle these conversion
-        #       copies of this code: plugin_helper/compat_parameters, compat/formatter_utils and here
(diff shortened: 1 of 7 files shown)


tagomoris:
Many changes based on this follows.

lib/fluent/plugin_helper/compat_parameters.rb
```diff
@@ -199,6 +213,12 @@ def compat_parameters_formatter(conf)
         # TODO: warn obsolete parameters if these are deprecated
         attr = compat_parameters_copy_to_subsection_attributes(conf, FORMATTER_PARAMS)
 
+        # TODO: make a utility method in TimeFormatter to handle these conversion
```
repeatedly: This comment is still needed?

tagomoris: `time_as_epoch` shouldn't be implemented in TimeFormatter, because it's now have `time_type` for unixtime, float (and string). It's just for compatibility for existing plugins.

tagomoris:
Updated the change.

tagomoris:
@repeatedly No more comments? I'll merge later if so.

repeatedly:
LGTM

PR #1255, opened 2016-10-04
Validators for buffering key extraction

tagomoris:
This change is mainly to make plugin authors possible to check configuration values are valid or not in the viewpoint of placeholders, like `${tag}`, `%Y-%m-%d` or `${any_key}`s.

Users can write any placeholders in their configuration parameters if plugin author provides a feature to extract it with buffer chunk keys. But buffer chunk keys and placeholders used in these values MUST be consistent. This consistency is very easy to be broken.

Newly introduced methods provides features to validate these parameters and buffering configurations. 

This change is based on #1254.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -486,12 +486,146 @@ module Fluent
         end
       end
 
+      def placeholder_validate!(name, str)
+        placeholder_validators(name, str).each do |v|
+          v.validate!
+        end
+      end
+
+      def placeholder_validators(name, str, time_key = (@chunk_key_time && @buffer_config.timekey), tag_key = @chunk_key_tag, chunk_keys = @chunk_keys)
+        validators = []
+
+        sec, title, example = get_placeholders_time(str)
+        if sec || time_key
+          validators << PlaceholderValidator.new(name, str, :time, {sec: sec, title: title, example: example, timekey: time_key})
+        end
+
+        parts = get_placeholders_tag(str)
+        if tag_key || !parts.empty?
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -486,12 +486,146 @@ def implement?(feature)
...
+
+        keys = get_placeholders_keys(str)
+        if chunk_keys && !chunk_keys.empty? || !keys.empty?
+          validators << PlaceholderValidator.new(name, str, :keys, {keys: keys, chunkkeys: chunk_keys})
+        end
+
+        validators
+      end
+
+      class PlaceholderValidator
```
repeatedly: I think separating PlaceholderValidator into TagPlaceholderValidator, TimePlaceholderValidator, KeysPlaceholderValidator is better.
It removes `time?`, `tag?` and `keys?` helpers and reduces method call overhead in `validate!`

tagomoris: No, it doesn't work. We should provide the possibility to plugin authors, for selecting which validator should be called or not.
For example, `table` might not have timestamp placeholders even when `time` chunk key is configured for any databases which can handle automatically generated time partitions. In such cases, the validator for time can be skipped by plugin authors (by using `validator.time?`)

tagomoris:
I added a commit for tag placeholder.

lib/fluent/plugin/output.rb
repeatedly: I see.

tagomoris:
I rebased on current master and added a commit for regexp.
@repeatedly ping?

repeatedly:
It seems good.

PR #1256, opened 2016-10-04
Plugin development enhancement

tagomoris:
This change includes some small changes to enhance plugin development:

* Fix to raise errors (including errors from wrong scripting, e.g., NameError) instead of rescue&log it
* Add method to create a metadata for testing/validation using `extract_placeholders`, without adding metadata into `@buffer.metadata_list`
* Fix to wait actual flushing after force_flush in test drivers

This change is based on #1255.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -976,6 +976,10 @@ module Fluent
           log.debug "taking back chunk for errors.", plugin_id: plugin_id, chunk: dump_unique_id_hex(chunk.unique_id)
           @buffer.takeback_chunk(chunk.unique_id)
 
+          if @under_plugin_development
+            raise
+          end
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -598,6 +738,13 @@ def metadata(tag, time, record)
         end
       end
 
+      def metadata_for_test(tag, time, record)
+        raise "BUG: #test_metadata is available only when no actual metadata exists" unless @buffer.metadata_list.empty?
```
repeatedly: `#metadata_for_test`?

tagomoris:
@repeatedly Rebased on master, and pushed a commit to update comment.

repeatedly:
LGTM

PR #1257, opened 2016-10-04
migrate out_file to v0.14 API

tagomoris:
This replaces #1159, depending on #1254, #1255 and #1256.
There are no compatibility problem or regressions, as far as I tested.


tagomoris:

repeatedly:
LGTM!!!

Issue #1229, opened 2016-09-12
Errors with in_tail plugin in 0.14.6


jchauncey:
Seeing the following error in my kubernetes fluentd pod 

Version:0.14.6 running as a gem
OS: Debian

```
2016-09-12 14:56:43 +0000 [error]: can't modify string; temporarily locked
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin/in_tail.rb:563:in `readpartial'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin/in_tail.rb:563:in `on_notify'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin/in_tail.rb:444:in `on_notify'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin_helper/timer.rb:76:in `on_timer'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/cool.io-1.4.5/lib/cool.io/loop.rb:88:in `run_once'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/cool.io-1.4.5/lib/cool.io/loop.rb:88:in `run'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin_helper/event_loop.rb:71:in `block in start'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin_helper/thread.rb:66:in `block in thread_create'
2016-09-12 14:56:43 +0000 [error]: closed stream
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin/in_tail.rb:563:in `readpartial'
  2016-09-12 14:56:43 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.14.6/lib/fluent/plugin/in_tail.rb:563:in `on_notify'
[...]

PR #1261, opened 2016-10-05
Run TailWatcher#on_notify before attaching triggers

akihiro17:
relate to #1229

This commit runs `TailWatcher#on_notify` before attaching `stat_trigger` and `timer_trigger` and prevents `TailWatcher#on_notify` from running in two different threads(main and event loop thread) at the same time.

This change fixes the `can't modify string; temporarily locked` error.

I think fluentd v0.12 does not raise the `can't modify string; temporarily locked` error because it runs `stat_trigger` and `timer_trigger` after `TailInput#start` finishes and   thus `TailWatcher#on_notify` is not executed in two threads simultaneously.

```ruby
# fluentd v0.12
# in_tail.rb
def start
  @loop = Coolio::Loop.new
  refresh_watchers

  @thread = Thread.new(&method(:run))
end

def run
  @loop.run
end
```

### reproduction steps
same as described in #1229

1. Create large log files by using append.rb
2. Remove a pos file (rm tmp/logs/tail.pos)
3. Run fluentd
4. Run append.rb on another terminal

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -428,8 +428,8 @@ module Fluent::Plugin
       end
 
       def attach
-        yield self
         on_notify
+        yield self
       end
 
       def detach


repeatedly:
Thanks 👍 
This seems the root cause.

PR #1263, opened 2016-10-06
Enable chunk.each only for limited plugins using msgpack

tagomoris:
`chunk.each` (`chunk.msgpack_each`) should be available only for chunks contains msgpack binary data.

It's original design, but not implemented when v0.14 output plugins were introduced (missed to remove `include ChunkMessagePackEventStreamer` from `chunk.rb`).
Currently there are not so many output plugins using v0.14 buffered output API, so it's time to change.

`chunk.each` (`chunk.msgpack_each`) will be available only for:
* output plugins using standard format (doesn't implement #format method)
* output plugins marked to have #format method returning msgpack binary data

### DIFF
--- a/test/plugin/test_output_as_buffered.rb
+++ b/test/plugin/test_output_as_buffered.rb
@@ -56,6 +56,47 @@ module FluentPluginOutputAsBufferedTest
       super
     end
   end
+  class DummyStandardBufferedOutput < DummyBareOutput
+    def initialize
+      super
+      @prefer_delayed_commit = nil
+      @write = nil
+      @try_write = nil
+    end
+    def prefer_delayed_commit
+      @prefer_delayed_commit ? @prefer_delayed_commit.call : false
+    end
+    def write(chunk)
+      @write ? @write.call(chunk) : nil
+    end
+    def try_write(chunk)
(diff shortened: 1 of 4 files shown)


tagomoris:
@repeatedly Could you review this change? I'll merge this before releasing v0.14.7.

test/plugin/test_output_as_buffered.rb
```diff
@@ -56,6 +56,47 @@ def shutdown
...
+      @write ? @write.call(chunk) : nil
+    end
+    def try_write(chunk)
+      @try_write ? @try_write.call(chunk) : nil
+    end
+  end
+  class DummyCustomFormatBufferedOutput < DummyBareOutput
+    def initialize
+      super
+      @format_type_is_msgpack = nil
```
repeatedly: Who set this callback?

repeatedly:
Code itself is no problem.

test/plugin/test_output_as_buffered.rb
tagomoris: oops, I missed to add tests using custom formats.

tagomoris:
I've just pushed a commit to add tests about custom format plugin.

PR #1264, opened 2016-10-06
Fix to wakeup threads to quit loop in threads

tagomoris:
It's done before returning from stop/after_shutdown, but after setting flag to break while-loop.
This change suppresses log messages like 'thread doesn't exit correctly (killed or other reason)' in shutdown sequence.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -181,6 +181,7 @@ module Fluent
         @secondary = nil
         @retry = nil
         @dequeued_chunks = nil
+        @output_enqueue_thread = nil
         @output_flush_threads = nil
 
         @simple_chunking = nil
@@ -359,6 +360,9 @@ module Fluent
 
           @buffer.start
 
+          @output_enqueue_thread = nil
+          @output_enqueue_thread_running = true
+
           @output_flush_threads = []
           @output_flush_threads_mutex = Mutex.new
           @output_flush_threads_running = true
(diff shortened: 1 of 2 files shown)


tagomoris:

lib/fluent/plugin/output.rb
```diff
@@ -416,6 +420,12 @@ def before_shutdown
             force_flush
           end
           @buffer.before_shutdown
+          # Need to ensure to stop enqueueing ... after #shutdown, we cannot write any data
+          @output_enqueue_thread_running = false
+          if @output_enqueue_thread && @output_enqueue_thread.alive?
```
repeatedly: No need to check `status`?

tagomoris: No need to do so. Living threads are running or sleeping.

repeatedly: I see

lib/fluent/plugin_helper/thread.rb
```diff
@@ -110,11 +110,31 @@ def initialize
 
       def stop
         super
+        wakeup_threads = []
         @_threads_mutex.synchronize do
-          @_threads.each_pair do |obj_id, thread|
+          @_threads.values.each do |thread|
```
repeatedly: `each_value` is better for avoding temporary object allocation.

lib/fluent/plugin_helper/thread.rb
```diff
@@ -110,11 +110,31 @@ def initialize
...
+            thread.wakeup
+          end
+        end
+      end
+
+      def after_shutdown
+        super
+        wakeup_threads = []
+        @_threads_mutex.synchronize do
+          @_threads.values.each do |thread|
```
repeatedly: ditto

repeatedly:
Others are okay for me.

lib/fluent/plugin_helper/thread.rb
tagomoris: Fixed to do so.

tagomoris:
Okay. I pushed to use `each_value` instead of `values.each`.

PR #1265, opened 2016-10-06
fix the order of expected/actual, and check filesize for failures

tagomoris:


tagomoris:
This change improves stability of tests. I'll merge this after CI green.

PR #1267, opened 2016-10-07
parser/formatter plugin helper with default @type in plugin side

tagomoris:
Adding `multi: true` for `config_section :parse` and `config_section :format` helps plugin authors not to use `conf` or `default_type` argument in `parser_create` and `formatter_create`.

With this change, plugin authors can write code to create parser(formatter) instances with default parser(formatter).

```ruby
helpers :parser

config_section :parse do
  config_set_default :@type, 'myparser'
end

def configure(conf)
  super
  @parser = parser_create
end
```

After this change, parser(formatter) plugin helpers **always** requires default `@type` definition.
I believe that it's reasonable change for all plugins which uses parsers/formatters.


tagomoris:
@repeatedly could you review this change please?

repeatedly:
LGTM

Issue #1276, opened 2016-10-15
in_forward should generate "time" with sub-second precision


njam:
I noticed that `in_forward` generates event times with only seconds-precision when debugging with `fluent-cat`.

How about generating times with higher precision in [in_forward.rb](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_forward.rb)?
Currently the code looks like this:
```rb
time = Engine.now if time.to_i == 0
```

repeatedly:
in_forward is no problem.
The problem is fluent-cat sends second-precision time.
I will fix it.

njam:
Aha, thanks!

PR #1277, opened 2016-10-15
Fix fluent-cat command to send sub-second precision time

repeatedly:
Fix https://github.com/fluent/fluentd/issues/1276

### DIFF
--- a/lib/fluent/command/cat.rb
+++ b/lib/fluent/command/cat.rb
@@ -16,6 +16,8 @@
 
 require 'optparse'
 require 'fluent/env'
+require 'fluent/time'
+require 'fluent/engine'
 
 op = OptionParser.new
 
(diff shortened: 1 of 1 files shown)


repeatedly:
@tagomoris If no problem, I will merge this.

lib/fluent/command/cat.rb
```diff
@@ -200,7 +202,8 @@ def write_impl(array)
     end
 
     begin
-      socket.write [@tag, array].to_msgpack
+      packer = Fluent::Engine.msgpack_factory.packer
```
tagomoris: It's better to use `Fluent::MessagePackFactory.packer` than using `Fluent::Engine`, because it makes internal dependency problem minimized.

tagomoris:
I pointed just one problem. Outline and others look good to me.

repeatedly:
Applied reviews and test passed.

PR #1282, opened 2016-10-19
test driver: Return the return value of yield

### LINKED ISSUE #1268: Test driver's `#run` method doesn't return the return value of yield
It's usually expected by ruby programmers.

okkez:
Fix #1268


tagomoris:
Fix for `Fluent::Test::Driver::MultiOutput` is missing... and duplicated with #1284.
Let me close this change...

PR #1283, opened 2016-10-19
Event router for secondary output

### LINKED ISSUE #1271: Current Output plugin base class has a probrem that it needs `router` method if secondary uses router.
- fluentd (v0.14.7)
- All envs

Current `Fluent::Plugin::Output` has following statement in `configure` method.

```
          @secondary.router = router if @secondary.has_router?

```

In current version, If secondary output uses compat layer (compat layer always include `:event_emitter`) or secondary output uses `:event_emitter` helper, primary output always needs `:event_emitter`.
In short, If user wants to use secondary output, almost output plugin needs `:event_emitter` at present.

tagomoris:
This change introduces `context_router` for all plugin types to provide event_router to secondary output plugins.
`context_router` is the default event router for filters/outputs in `<label>` sections.

This fixes #1271.

### DIFF
--- a/lib/fluent/plugin_helper/event_emitter.rb
+++ b/lib/fluent/plugin_helper/event_emitter.rb
@@ -26,10 +26,14 @@ module Fluent
 
       def router
         @_event_emitter_used_actually = true
+        if @_event_emitter_lazy_init
+          @router = @primary_instance.router
+        end
         @router
       end
 
       def router=(r)
+        # not recommended now...
         @router = r
       end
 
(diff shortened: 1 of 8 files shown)


tagomoris:

repeatedly:
Others are good.

lib/fluent/plugin_helper/event_emitter.rb
```diff
@@ -26,10 +26,14 @@ module EventEmitter
...
       def router
         @_event_emitter_used_actually = true
+        if @_event_emitter_lazy_init
+          @router = @primary_instance.router
+        end
         @router
       end
 
       def router=(r)
+        # not recommended now...
```
tagomoris: I thought about it at first, but didn't it finally. Because we can't remove `#router=` method between v0.x and v1.x for compatibility reason. So displaying warning doesn't help anything.
IMO we should show warning here at the same time with deprecating v0.12 plugin APIs (and remove this method at v2).

repeatedly: I see. I understood the situation.

PR #1284, opened 2016-10-19
fix to return the block value as expected by many rubyists

### LINKED ISSUE #1268: Test driver's `#run` method doesn't return the return value of yield

tagomoris:
`test_driver.run` doesn't return the value of block specified. In standard ruby way, the behavior below is expected (but not yet).
```ruby
v = d.run do
  1
  2
end
v == 2 # expected
```

This change fixes #1268.

### DIFF
--- a/lib/fluent/test/driver/base.rb
+++ b/lib/fluent/test/driver/base.rb
@@ -139,8 +139,9 @@ module Fluent
             raise ArgumentError, "no stop conditions nor block specified"
           end
 
+          return_value = nil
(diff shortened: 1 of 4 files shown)


tagomoris:
This change is only for test driver, so I'll merge this after CI green.

cosmo0920:
Hm, it looks that https://github.com/fluent/fluentd/pull/1282 is same purpose PR.
Which is better?

PR #1286, opened 2016-10-21
Merge ValuesParser and TypeConverter into Parser

tagomoris:
This change does:
* merge features of ValuesParser and TypeConverter into Parser
* move ValuesParser implementation under Compat (TypeConverter is already there)
* update built-in parsers with this change

Many parser plugins (including 3rd party ones) needs features of TypeConverter and time parser.
These features are provided only in ValuesParser, but ValuesParser is originally for parsers which supports fixed number fields, like tsv and csv.
(ltsv parser inherits ValuesParser for TypeConverter, and it has unbelievable implementation...)

This change makes parser plugins to:
* use type converters, null field handling and time parsing very easily in the user-configurable way
* update configuration parameters to use `hash` and `array` to provide safe structured values without additional parameters
* make `estimate_current_event` as a configuration parameter (can be overloaded by `config_set_default` of 3rd party parser plugins)

See `parser_ltsv.rb` and `parser_regexp.rb` to show how this change makes writing plugins understandable.


tagomoris:

tagomoris:
@repeatedly ping?

lib/fluent/compat/parser.rb
```diff
@@ -116,48 +120,170 @@ def self.lookup(format)
...
-      class TimeParser < Fluent::Plugin::Parser::TimeParser
+      module TypeConverterCompatParameters
+        def convert_type_converter_parameters!(conf)
+          if conf["types"]
+            delimiter = conf["types_delimiter"] || ','
+            label_delimiter = conf["types_label_delimiter"] || ':'
+            types = {}
+            conf['types'].split(delimiter).each do |pair|
+              key, value = pair.split(label_delimiter, 2)
+              if value.start_with?("time#{label_delimiter}")
```
repeatedly: `array` should be also care?

repeatedly:
Idea is good for me and add 2 comments.

One concern is regexp parser performance because this parser is used in high-volume environment.
I will check it with micro-benchmark.

lib/fluent/compat/parser.rb
```diff
@@ -116,48 +120,170 @@ def self.lookup(format)
...
+          end
+          if value and @null_value_pattern
+            value = ::Fluent::StringUtil.match_regexp(@null_value_pattern, value) ? nil : value
+          end
+          value
+        end
       end
 
       class JSONParser < Fluent::Plugin::JSONParser
+        include TypeConverterCompatParameters
```
tagomoris: Yes. It is a subclass of `Fluent::Plugin::JSONParser`, and it doesn't include it.

repeatedly: JSONParser doesn't use TypeConverter feature.
Do you have a plan to update JSONParser?

tagomoris: Yes, I missed to update code of JSONParser. I pushed some commits including one to do it.

repeatedly:
I re-think the `convert_values` API.
Current API creates temporary array for each record.
In high-volume environment, it increase CPU usage / memory usage.
I changes Parser API to avoid this problem in v0.12, so v0.14 should keep same performance.

tagomoris:
@repeatedly I made a very simple benchmark, and found that multi value return is faster than passing blocks: https://gist.github.com/tagomoris/7bca9fe4fdc3201f691094e6aad3b177

tagomoris:
@repeatedly do you have anything more?

repeatedly:
LGTM. If we have a performance issue, we should rewrite parser code in the future.

PR #1294, opened 2016-10-25
Fix to set valid configured in section with nil of base class

tagomoris:
There's a bug that the sub-section names configured by `configured_in` are not used correctly.

It is because that the value of `configured_in_section` in superclass is always used in getting `merged_configure_proxy`. But, all plugin classes have the root superclass, `Fluent::Plugin::Base`. It has a `nil` value as `configured_in_section`, and it was used always.
This breaks behaviors of buffer, storage, parser and formatter plugins. Owner plugins cannot overwrite default values in buffer, storage, parse and format sections.

I also fixed a bug of `in_syslog` (it shows the bug in test cases with a overwritten default configuration correctly works), and also turned timeout always enabled to show never ending test cases.


tagomoris:
@repeatedly Could you check this change?

test/config/test_configurable.rb
```diff
@@ -346,6 +346,19 @@ class BufferChild
...
+      include Fluent::Configurable
+    end
+
+    class BufferSubclass < BufferBase
+      attr_accessor :owner
+      configured_in :buffer
+      config_param :size_of_something, :size, default: 512
+    end
+
+    class BufferSubclass2 < BufferSubclass
```
repeatedly: `BufferSubclass2` looks like same hierarchy with `BufferSubclass`.
`BufferSubSubclass` or other name is better.

repeatedly:
LGTM and one comment.

test/config/test_configurable.rb
tagomoris: done.

PR #1296, opened 2016-10-26
Fix a bug not to overwrite default value with nil

tagomoris:


tagomoris:
@repeatedly could you check this?

repeatedly:
Test failed.
Is this change the cause?

tagomoris:
Originally, it's a wrong manner code... I added a commit to fix that.

tagomoris:
Can I merge this?

repeatedly:
LGTM

PR #1297, opened 2016-10-26
Migrate exec plugins to v0.14 api

tagomoris:
This change is to migrate `out_exec_filter` to v0.14 API, and make `in/out_exec` plugins free from `Fluent::ExecUtil::*` classes.

This change includes the changes to:
- make all (built-in or 3rd party) formatter/parser plugins available in exec plugins
- make inject/extract plugin helpers available in exec plugins
- add a method to return parser/formatter type for data handling (binary, text or text-per-line)
- add methods to pass io or partial data to parsers to parse binary or multiline text data in consistent way
- add options of child_process plugin helper to wait process exits and callback for exit status

This change is successor of #1048.

### DIFF
--- a/lib/fluent/plugin/formatter_tsv.rb
+++ b/lib/fluent/plugin/formatter_tsv.rb
@@ -14,24 +14,20 @@
 #    limitations under the License.
 #
 
-require 'fluent/plugin/parser'
-
-require 'csv'
+require 'fluent/plugin/formatter'
 
 module Fluent
   module Plugin
-    class CSVParser < Parser
-      Plugin.register_parser('csv', self)
+    class TSVFormatter < Formatter
+      Plugin.register_formatter('tsv', self)
 
       desc 'Names of fields included in each lines'
       config_param :keys, :array, value_type: :string
-      desc 'The delimiter character (or string) of CSV values'
-      config_param :delimiter, :string, default: ','
+      desc 'The delimiter character (or string) of TSV values'
+      config_param :delimiter, :string, default: "\t"
 
-      def parse(text, &block)
-        values = CSV.parse_line(text, col_sep: @delimiter)
-        r = Hash[@keys.zip(values)]
-        convert_values(parse_time(r), r, &block)
+      def format(tag, time, record)
+        @keys.map{|k| record[k].to_s }.join(@delimiter)
       end
     end
   end
--- a/lib/fluent/plugin/formatter_tsv.rb
+++ b/lib/fluent/plugin/formatter_tsv.rb
@@ -21,7 +21,7 @@ module Fluent
     class TSVFormatter < Formatter
       Plugin.register_formatter('tsv', self)
 
-      desc 'Names of fields included in each lines'
+      desc 'Field names included in each lines'
       config_param :keys, :array, value_type: :string
       desc 'The delimiter character (or string) of TSV values'
       config_param :delimiter, :string, default: "\t"
--- a/lib/fluent/plugin/in_exec.rb
+++ b/lib/fluent/plugin/in_exec.rb
@@ -14,136 +14,104 @@
 #    limitations under the License.
 #
 
-require 'strptime'
-require 'yajl'
-
 require 'fluent/plugin/input'
-require 'fluent/time'
-require 'fluent/timezone'
-require 'fluent/config/error'
+require 'yajl'
 
 module Fluent::Plugin
   class ExecInput < Fluent::Plugin::Input
     Fluent::Plugin.register_input('exec', self)
 
-    helpers :child_process
-
-    def initialize
-      super
-      require 'fluent/plugin/exec_util'
-    end
+    helpers :compat_parameters, :extract, :parser, :child_process
 
     desc 'The command (program) to execute.'
     config_param :command, :string
-    desc 'The format used to map the program output to the incoming event.(tsv,json,msgpack)'
-    config_param :format, :string, default: 'tsv'
-    desc 'Specify the comma-separated keys when using the tsv format.'
-    config_param :keys, default: [] do |val|
-      val.split(',')
+
+    config_section :parse do
+      config_set_default :@type, 'tsv'
+      config_set_default :time_type, :float
+      config_set_default :time_key, nil
+      config_set_default :estimate_current_event, false
     end
+
+    config_section :extract do
+      config_set_default :time_type, :float
+    end
+
     desc 'Tag of the output events.'
     config_param :tag, :string, default: nil
-    desc 'The key to use as the event tag instead of the value in the event record. '
-    config_param :tag_key, :string, default: nil
-    desc 'The key to use as the event time instead of the value in the event record.'
-    config_param :time_key, :string, default: nil
-    desc 'The format of the event time used for the time_key parameter.'
-    config_param :time_format, :string, default: nil
     desc 'The interval time between periodic program runs.'
     config_param :run_interval, :time, default: nil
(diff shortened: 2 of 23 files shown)


tagomoris:

tagomoris:
Rebased on #1305.
@repeatedly ping?

repeatedly:
@tagomoris I will review it. You pushed some changes after previous comment, so I waited to finish your additional commit ;p

tagomoris:
:)
And... currently, CI on Windows is failing. It's problem about just test code, and I'll push some commits for that. Please review the code anyway.

tagomoris:
Hmm, `test_out_exec.rb` is also unstable a little.

repeatedly:
@sonots Could you also check this PR? This patch is very large so need more reviewers.

lib/fluent/plugin/formatter_tsv.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+
+require 'fluent/plugin/formatter'
+
+module Fluent
+  module Plugin
+    class TSVFormatter < Formatter
+      Plugin.register_formatter('tsv', self)
+
+      desc 'Names of fields included in each lines'
```
repeatedly: "Field names" is more simpler than "Names of fields"

lib/fluent/plugin/in_exec.rb
```diff
@@ -14,136 +14,92 @@
...
+    def on_record(time, record)
+      tag = nil
+      tag = extract_tag_from_record(record)
+      tag ||= @tag
+      time ||= extract_time_from_record(record) || Fluent::EventTime.now
       router.emit(tag, time, record)
     rescue => e
-      log.error "exec failed to emit", error: e, tag: tag, record: Yajl.dump(record)
+      log.error "exec failed to emit", tag: tag, record: Yajl.dump(record), error: e
+      router.emit_error_event(tag, time, record, "exec failed to emit")
```
repeatedly: 4th error argument should be an exception object.

lib/fluent/plugin/out_exec.rb
```diff
@@ -18,97 +18,82 @@
...
-      @formatter.call(record, out)
-      out
-    end
-
-    def write(chunk)
-      if chunk.respond_to?(:path)
-        prog = "#{@command} #{chunk.path}"
+      record = inject_values_to_record(tag, time, record)
+      if @formatter.formatter_type == :text_per_line
+        @formatter.format(tag, time, record).chomp + "\n"
```
repeatedly: Using constant value for "\n"  is better for reducing temporary object allocation.

lib/fluent/plugin/out_exec.rb
```diff
@@ -18,97 +18,82 @@
...
-
-      if ecode != 0
-        raise "command returns #{ecode}: #{prog}"
-      end
+    def try_write(chunk)
+      tmpfile = nil
+      prog = if chunk.respond_to?(:path)
+               "#{@command} #{chunk.path}"
+             else
+               tmpfile = Tempfile.new("fluent-plugin-exec-")
```
repeatedly: "fluent-plugin-out-exec-" is more better.

lib/fluent/plugin/out_exec_filter.rb
```diff
@@ -202,192 +172,141 @@ def configure(conf)
...
+          @parser.parse(line.chomp, &method(:on_record))
+        end
       else
-        tag = @tag
+        @parser.parse(io.read, &method(:on_record))
       end
+    end
 
+    def on_record(time, record)
+      tag = nil
```
repeatedly: ditto

lib/fluent/plugin/out_exec_filter.rb
```diff
@@ -202,192 +172,141 @@ def configure(conf)
...
+      tag = extract_tag_from_record(record)
+      tag = @added_prefix_string + tag if tag && @add_prefix
+      tag ||= @tag
```
repeatedly: Off topic. We should care extrated result and `@tag` is `nil` case.

lib/fluent/plugin/out_exec_filter.rb
```diff
@@ -202,192 +172,141 @@ def configure(conf)
...
-    rescue
+    rescue => e
       if @suppress_error_log_interval == 0 || Time.now.to_i > @next_log_time
-        log.error "exec_filter failed to emit", error: $!, record: Yajl.dump(record)
-        log.warn_backtrace $!.backtrace
+        log.error "exec_filter failed to emit", record: Yajl.dump(record), error: e
+        log.error_backtrace e.backtrace
         @next_log_time = Time.now.to_i + @suppress_error_log_interval
       end
+      router.emit_error_event(tag, time, record, "exec_filter failed to emit") if tag && time && record
```
repeatedly: ditto. 4th argument should be an error object.

lib/fluent/plugin/parser_msgpack.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def parse(data)
+        @unpacker.feed_each(data) do |obj|
+          yield convert_values(parse_time(obj), obj)
+        end
+      end
+      alias parse_partial_data parse
+
+      def parse_io(io)
```
repeatedly: should be use `parse_io(io, &block)` for API signature?

lib/fluent/plugin_helper/compat_parameters.rb
```diff
@@ -77,10 +79,20 @@ module CompatParameters
         "utc" => nil,
       }
 
+      EXTRACT_PARAMS = {
+        "time_key"      => "time_key",
+        "time_format"   => "time_format",
```
repeatedly: More spaces.

lib/fluent/plugin/out_exec.rb
```diff
@@ -18,97 +18,82 @@
...
+               tmpfile.binmode
+               chunk.write_to(tmpfile)
+               tmpfile.close
+               "#{@command} #{tmpfile.path}"
+             end
+      chunk_id = chunk.unique_id
+      callback = ->(status){
+        begin
+          if tmpfile
```
tagomoris: No. `ensure` of `#try_write` doesn't take care about child process status.
This callback is called after child process plugin helper find the child process's exit status. So here is the only place to delete tmpfile.

repeatedly: Hmm... what happen if exception happens before calling callback?

tagomoris:
I've pushed commits for review comments.

tagomoris:
@repeatedly ping

lib/fluent/plugin/out_exec_filter.rb
```diff
@@ -202,192 +172,140 @@ def configure(conf)
...
-          @log.error "exec_filter process unexpectedly exited.", command: @command, ecode: stat.to_i
-          unless @respawns == 0
-            @log.warn "exec_filter child process will respawn for next input data (respawns #{@respawns})."
-          end
-        end
+    def format(tag, time, record)
+      tag = tag_remove_prefix(tag)
+      record = inject_values_to_record(tag, time, record)
+      if @formatter.formatter_type == :text_per_line
+        @formatter.format(tag, time, record).chomp + "\n"
```
repeatedly: Use `NEWLINE = "\n"` like above.

repeatedly:
LGTM.

But my code review doesn't guarantee new process handling code has no problem on actual workloads.
So merging it and test by users is better.

PR #1302, opened 2016-11-01
Show errors on console under plugin development

tagomoris:
This change fixes code to show error/fatal logs in testing console.

Raising errors doesn't help situation because:
* errors raised in threads will be raised in main thread by abort_on_exception=true
* main thread of test-unit will rescues and ignore errors

So we can't find any errors in unit tests without showing these on console.

I found that some `#try_write` method call `#commit_write` in it, and it misses to remove the chunk from `@dequeued_chunks`. It's a kind of bug.

Adding a chunk into `@dequeued_chunks` after calling `#try_write` has thread-unsafe bug. (because plugin may commit the chunk in another threads just after `#try_write` called, and plugin's main thread may have NOT added that chunk into `@dequeued_chunks`.)
So, I fixed it at the same time.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -150,10 +150,9 @@ module Fluent
 
       # for tests
       attr_reader :buffer, :retry, :secondary, :chunk_keys, :chunk_key_time, :chunk_key_tag
-      attr_accessor :output_enqueue_thread_waiting, :in_tests
+      attr_accessor :output_enqueue_thread_waiting, :dequeued_chunks, :dequeued_chunks_mutex
 
       # output_enqueue_thread_waiting: for test of output.rb itself
-      # in_tests: for tests of plugins with test drivers
 
       def initialize
         super
@@ -161,7 +160,6 @@ module Fluent
         @buffering = false
         @delayed_commit = false
         @as_secondary = false
-        @in_tests = false
         @primary_instance = nil
 
         # TODO: well organized counters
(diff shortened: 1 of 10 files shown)


tagomoris:

lib/fluent/plugin/output.rb
```diff
@@ -1059,7 +1061,11 @@ def submit_flush_once
         @output_flush_thread_current_position = (@output_flush_thread_current_position + 1) % @buffer_config.flush_thread_count
         state = @output_flush_threads[@output_flush_thread_current_position]
         state.next_time = 0
-        state.thread.run
+        if state.thread && state.thread.status # "run"/"sleep"/"aborting" or false(successfully stop) or nil(killed by exception)
+          state.thread.run
+        else
+          log.warn "thread is already dead"
```
repeatedly: If one thread is dead, we should replace dead thread with new active one?

repeatedly:
Commented. Changes are LGTM.

lib/fluent/plugin_helper/thread.rb
```diff
@@ -66,6 +66,12 @@ def thread_create(title)
             yield
             thread_exit = true
           rescue Exception => e
+            if @under_plugin_development
+              STDERR.puts "\nError raised in thread from #thread_create, #{e.class}:#{e.message}"
```
tagomoris: Yes, this code expects that leading text is '.' (means progress of test cases) from test runner, without tailing newline.

lib/fluent/plugin/output.rb
tagomoris: I don't think so. Dead flush thread means any kind of bug of `output.rb`.

tagomoris:
OK, let me merge this later.

Issue #608, opened 2015-06-04
Nginx log format configuration


reachgopal:
Hello:

I am using default Nginx log format provided as below: (mentioned in fluentd site http://docs.fluentd.org/articles/in_tail). 

```
format /^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)")?$/
time_format %d/%b/%Y:%H:%M:%S %z
```

I use standard nginx docker container and try to access the site and found that the format mentioned but it does not cover the gzip ratio as mentioned below.

```
^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)" "(?<gzip_ratio>[^\"]*)")?$
```

Is the format not updated in the portal as default format? Please advice.

Thanks.

repeatedly:
Sorry I don't have a log including gzip ratio part.
Could you paste actual log example here?

repeatedly:
ping?

repeatedly:
ping?
We need actual log to check and fix this issue.

repeatedly:
There is no response.
Please reopen it if you have actual logs.

PR #1303, opened 2016-11-01
Default nginx access log format parse fix

ljagiello:
Default nginx access log format (http://nginx.org/en/docs/http/ngx_http_log_module.html) looks like:
```
log_format compression '$remote_addr - $remote_user [$time_local] '
                       '"$request" $status $bytes_sent '
                       '"$http_referer" "$http_user_agent" "$gzip_ratio"';
```

Fluentd parser was missing `"$gzip_ratio"`.

How to reproduce the bug:
```
➜  ~ docker run --name test-nginx -P -d nginx
[…]
➜  ~ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                           NAMES
3b445987ac4d        nginx               "nginx -g 'daemon off"   16 hours ago        Up 2 minutes        0.0.0.0:32769->80/tcp, 0.0.0.0:32768->443/tcp   test-nginx
➜  ~ curl -so /dev/null http://localhost:32769
➜  ~ docker logs 3b445987ac4d
172.17.0.1 - - [01/Nov/2016:01:25:20 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.49.1" "-"
```

Fluentd was unable to parse default nginx log format without this patch.

My patch adds missing field and keep backward compatibility.

The issue was reported also here - https://github.com/fluent/fluentd/issues/608

### DIFF
--- a/lib/fluent/plugin/parser_nginx.rb
+++ b/lib/fluent/plugin/parser_nginx.rb
@@ -21,7 +21,7 @@ module Fluent
     class NginxParser < RegexpParser
       Plugin.register_parser("nginx", self)
 
-      config_set_default :expression, %q{/^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*?)(?: +\S*)?)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)")?$/}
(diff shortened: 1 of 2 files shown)


tagomoris:
@ljagiello It's just example, not default. See "Directives":
```
Syntax: access_log path [format [buffer=size] [gzip[=level]] [flush=time] [if=condition]];
Default: access_log logs/access.log combined;
```

repeatedly:
Default `combined` is also described in same page:

```
The configuration always includes the predefined “combined” format:
log_format combined '$remote_addr - $remote_user [$time_local] '
                    '"$request" $status $body_bytes_sent '
                    '"$http_referer" "$http_user_agent"';
```

ljagiello:
@tagomoris @repeatedly 

Inside official nginx package - http://nginx.org/packages/mainline/debian/pool/nginx/n/nginx/nginx_1.11.5-1~jessie_amd64.deb

configuration looks now:
```
    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log  /var/log/nginx/access.log  main;
```

I wrongly assume it's a gzip ratio, in fact it's `$http_x_forwarded_for`. IMHO if it's part of official package it should be supported. As a result of that docker/kubernetes containers will be supported by default.

PR #1305, opened 2016-11-04
Test driver fixes for error handling

tagomoris:
This change follows #1302.
After #1302, I found that any errors in `#write` of output plugins are not raised in test case code like below:
```ruby
assert_raise AnyErrors do
  d.run
    d.feed(events)
  end # events flushed and #write called - "AnyErrors" raised
end
```

Errors are shown on console, but we can't test it.
This change is to check all threads via thread plugin helper, and re-raise errors from these threads at the end of `#run` method. It shows all errors in all threads related with plugins explicitly as test results - so there's no need to show these errors on console.

Some changes are also included:
* add an option to retry for flush errors even in plugin development: to emulate write operations 2 or more times over failures
* use `Process.clock_gettime` instead of `Time` to work well even when `timecop` is used
* raise TestTimedOut error for hard timeout to show problematic test cases clearly
* change to wait post conditions after `#run_actual`, because output plugin test driver flushes and wait for completion of flushing in `#run_actual`

### DIFF
--- a/lib/fluent/test/driver/base.rb
+++ b/lib/fluent/test/driver/base.rb
@@ -91,8 +91,17 @@ module Fluent
             raise ArgumentError, "no stop conditions nor block specified"
           end
 
+          sleep_with_watching_threads = ->(){
+            if @instance.respond_to?(:_threads)
+              @instance._threads.values.each{|t| t.join(0) }
+            end
+            sleep 0.1
+          }
+
           begin
-            run_actual(timeout: timeout, &block)
+            retval = run_actual(timeout: timeout, &block)
+            sleep_with_watching_threads.call until stop?
+            retval
           ensure
             instance_shutdown if shutdown
           end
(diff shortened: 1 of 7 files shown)


tagomoris:
This is required for both of migrations of exec plugins and forward plugin.

tagomoris:
@repeatedly ping?

lib/fluent/test/driver/base.rb
```diff
@@ -79,8 +83,25 @@ def run(timeout: nil, start: true, shutdown: true, &block)
...
+          stop_at = Process.clock_gettime(@test_clock_id) + timeout
+          @run_breaking_conditions << ->(){ Process.clock_gettime(@test_clock_id) >= stop_at }
+
+          if !block_given? && @run_post_conditions.empty? && @run_breaking_conditions.empty?
+            raise ArgumentError, "no stop conditions nor block specified"
+          end
+
+          sleep_with_watching_threads = ->(){
+            if @instance.respond_to?(:_threads)
+              @instance._threads.values.each{|t| t.join(0) }
```
repeatedly: What happen with `0` limit?

repeatedly:
LGTM

PR #1306, opened 2016-11-07
Migrate forward plugin to v0.14 api

tagomoris:
This change is to migrate in/out_forward to v0.14 APIs, except for socket/server plugin helpers, to keep commits few and diff small.

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -149,7 +149,7 @@ module Fluent::Plugin
       event_loop_attach(@hbr)
     end
 
-    def stop
+    def close
       @lsock.close if @lsock
       @usock.close if @usock
       super
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -14,23 +14,26 @@
 #    limitations under the License.
 #
 
-require 'fcntl'
 
-require 'cool.io'
+require 'fluent/plugin/input'
+require 'fluent/msgpack_factory'
 require 'yajl'
+require 'digest'
 
-require 'fluent/input'
+require 'fluent/plugin/socket_util'
+require 'fcntl'
+require 'cool.io'
 
-module Fluent
+module Fluent::Plugin
   class ForwardInput < Input
-    Plugin.register_input('forward', self)
+    Fluent::Plugin.register_input('forward', self)
+
+    # See the wiki page below for protocol specification
+    # https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1
+
+    helpers :event_loop
 
     LISTEN_PORT = 24224
-
-    def initialize
-      super
-      require 'fluent/plugin/socket_util'
-    end
 
     desc 'The port to listen to.'
     config_param :port, :integer, default: LISTEN_PORT
@@ -130,8 +133,6 @@ module Fluent
     def start
       super
 
-      @loop = Coolio::Loop.new
-
       socket_manager_path = ENV['SERVERENGINE_SOCKETMANAGER_PATH']
       if Fluent.windows?
         socket_manager_path = socket_manager_path.to_i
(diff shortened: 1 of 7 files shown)


tagomoris:
Pull requests to add socket/server plugin helpers will follow this change.

repeatedly:
Okay. I will review it later 👍

tagomoris:
@repeatedly ping?

lib/fluent/plugin/in_forward.rb
```diff
@@ -125,46 +128,30 @@ def configure(conf)
...
-      # This problem occurs possibly because those watchers are thought to be related to each other
-      # and when detaching one of them the other is also detached for some reasons.
-      # As a workaround, check if watchers are attached before detaching them.
-      @loop.watchers.each {|w| w.detach if w.attached? }
-      @loop.stop
-      @usock.close
-      @thread.join
-      @lsock.close
-
+    def stop
```
repeatedly: `stop` is called before `shutdown`, right?
What happens when close socket before shutdown/close event loop?
Is it safe?

lib/fluent/plugin/in_forward.rb
```diff
@@ -14,23 +14,26 @@
...
-require 'cool.io'
+require 'fluent/plugin/input'
+require 'fluent/msgpack_factory'
 require 'yajl'
+require 'digest'
 
-require 'fluent/input'
+require 'fluent/plugin/socket_util'
+require 'fcntl'
+require 'cool.io'
```
repeatedly: `cool.io` is required by event_loop helper.

repeatedly: Ah wait. This commit doesn't include socket helpers. Ignore this comment.

lib/fluent/plugin/out_forward.rb
```diff
@@ -14,42 +14,23 @@
...
+require 'fluent/compat/socket_util'
 
-  class ForwardOutputACKTimeoutError < ForwardOutputResponseError
-  end
+module Fluent::Plugin
+  class ForwardOutput < Output
+    class Error < StandardError; end
+    class ResponseError < Error; end
+    class ConnectionClosedError < Error; end
+    class ACKTimeoutError < Error; end
```
repeatedly: In v0.12, ForwardOutputACKTimeoutError is subclass of ForwardOutputResponseError.
Change from ResponseError to Error is intented?

test/plugin/test_in_forward.rb
```diff
@@ -74,148 +88,153 @@ def test_auth
...
+      begin
+        Timecop.freeze(Time.at(time))
+        @d = d = create_driver
+
+        records = [
+          ["tag1", time, {"a"=>1}],
+          ["tag2", time, {"a"=>2}],
+        ]
+
+        d.run(expect_records: records.length, timeout: 5) do
```
repeatedly: Why change timeout to 5?

lib/fluent/plugin/out_forward.rb
tagomoris: It's intended. There are no actual benefits to make these errors with sub-super classes.
And, in fact, there are no responses under ACK timeout situation.d

test/plugin/test_in_forward.rb
```diff
@@ -74,148 +88,153 @@ def test_auth
...
         ["tag1", time, {"a" => 1}],
         ["tag2", time, {"a" => 2}],
       ]
 
-      d.run do
-        entries = records.map { |tag, _time, record| [tag, _time, record] }
+      d.run(shutdown: false, expect_records: 2, timeout: 10) do
+        entries = []
         # These entries are skipped
         entries << ['tag1', true, {'a' => 3}] << ['tag2', time, 'invalid record']
```
repeatedly: Why use 2 lines intead of `entries = [[...], [...]]`?

test/plugin/test_in_forward.rb
tagomoris: Just for stability of CI tests.

test/plugin/test_in_forward.rb
tagomoris: To reduce the number of lines of diff.

test/plugin/test_out_forward.rb
```diff
@@ -29,9 +32,13 @@ def setup
   ]
 
   def create_driver(conf=CONFIG)
-    d = Fluent::Test::BufferedOutputTestDriver.new(Fluent::ForwardOutput) {
+    Fluent::Test::Driver::Output.new(Fluent::Plugin::ForwardOutput) {
       attr_reader :responses, :exceptions
 
+      def write(chunk)
```
repeatedly: Why is this definition needed?

lib/fluent/plugin/out_forward.rb
repeatedly: okay

repeatedly:
Commented. Others looks good.

test/plugin/test_out_forward.rb
tagomoris: `Output#implement?` checks the features of that plugin by checking instance methods implemented in just that class, without methods of superclasses.
This test code defines a one-time class inherits ForwardOutput. If this method is not defined, `Output#implement?` returns false for check of buffered output feature.

It's better to be fixed in future, but currently, this code works fine.

repeatedly: I see. I understood the situation.

tagomoris:
I pushed a commit to close sockets in `#close`, and will merge this after CI green.

PR #1308, opened 2016-11-09
migrating in_http to v0.14 API

tagomoris:
This change is to migrate `in_http` to v0.14 APIs, without socket/server plugin helpers (like #1306).

Currently, this change disables DetachMultiProcessMixin feature, because `event_loop` plugin helper can't work with it (error below occurs when DetachMultiProcessMixin is mixed in):
```
Assertion failed: (("libev: ev_loop recursion during release detected", loop_done != EVBREAK_RECURSE)), function ev_run, file ./../libev/ev.c, line 3437.
rake aborted!
SignalException: SIGABRT
```

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -14,28 +14,31 @@
 #    limitations under the License.
 #
 
+require 'fluent/plugin/input'
+require 'fluent/plugin/parser'
+require 'fluent/event'
+require 'fluent/process'
+
+require 'http/parser'
+require 'webrick/httputils'
 require 'uri'
 require 'socket'
 require 'json'
 
-require 'cool.io'
-
-require 'fluent/input'
-require 'fluent/event'
-require 'fluent/process'
-
-module Fluent
-  class HttpInput < Input
-    Plugin.register_input('http', self)
-
-    include DetachMultiProcessMixin
-
-    require 'http/parser'
-
-    def initialize
-      require 'webrick/httputils'
-      super
+module Fluent::Plugin
+  class InHttpParser < Parser
+    Fluent::Plugin.register_parser('in_http', self)
+    def parse(text)
+      # this plugin is dummy implementation not to raise error
+      yield nil, nil
     end
+  end
+
+  class HttpInput < Input
+    Fluent::Plugin.register_input('http', self)
+
+    # include DetachMultiProcessMixin
+    helpers :parser, :compat_parameters, :event_loop
 
     EMPTY_GIF_IMAGE = "GIF89a\u0001\u0000\u0001\u0000\x80\xFF\u0000\xFF\xFF\xFF\u0000\u0000\u0000,\u0000\u0000\u0000\u0000\u0001\u0000\u0001\u0000\u0000\u0002\u0002D\u0001\u0000;".force_encoding("UTF-8")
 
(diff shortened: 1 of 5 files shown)


tagomoris:
How do you think about dropping support of detach_process?

repeatedly:
I will check it later.

> How do you think about dropping support of detach_process?

I think this is no problem. `detach_process` itself doesn't boost the performance and current detach_process has several problems, e.g. internal pipe sometine stuck.
And detach_process in `in_http` is un-documented feature: http://docs.fluentd.org/articles/in_http
So we can remove it. I will update v0.12 in_http to show warning message in configure.

lib/fluent/plugin/in_http.rb
```diff
@@ -52,25 +53,36 @@ def initialize
...
     def configure(conf)
+      compat_parameters_convert(conf, :parser)
+
       super
 
-      m = if @format == 'default'
+      m = if @parser_configs.first['@type'] == 'in_http'
+            @parser_msgpack = parser_create(type: 'msgpack')
+            @parser_msgpack.estimate_current_event = false
+            @parser_json = parser_create(type: 'json')
```
repeatedly: parser_create uses `usage` for `@_parsers` key.
This second `parser_create` seems to overwrite `msgpack` parser.

lib/fluent/plugin/in_http.rb
```diff
@@ -52,25 +53,36 @@ def initialize
...
-    desc 'The format of the HTTP body.'
-    config_param :format, :string, default: 'default'
     config_param :blocking_timeout, :time, default: 0.5
     desc 'Set a white list of domains that can do CORS (Cross-Origin Resource Sharing)'
     config_param :cors_allow_origins, :array, default: nil
     desc 'Respond with empty gif image of 1x1 pixel.'
     config_param :respond_with_empty_img, :bool, default: false
 
+    config_section :parse do
+      config_set_default :@type, 'in_http'
```
repeatedly: Is using `nil` bad idea?
It can avoid `InHttpParser` definition.

test/plugin/test_in_http.rb
```diff
@@ -44,228 +49,309 @@ def test_configure
...
-    d.run do
-      d.expected_emits.each {|tag,_time,record|
-        res = post("/#{tag}", {"json"=>record.to_json, "time"=>_time.to_s}, {"X-Forwarded-For"=>"129.78.138.66, 127.0.0.1"})
-        assert_equal "200", res.code
-      }
+    time = event_time("2011-01-02 13:14:15 UTC")
+    time_i = time.to_i
+
+    events = [
+      ["tag1", time, {"REMOTE_ADDR"=>"129.78.138.66", "a"=>1}],
```
repeatedly: `"REMOTE_ADDR"=>"129.78.138.66"` is assigined by in_http, not emitted record.

test/plugin/test_in_http.rb
```diff
@@ -44,228 +49,309 @@ def test_configure
...
   def test_multi_json_with_add_remote_addr_given_x_forwarded_for
     d = create_driver(CONFIG + "add_remote_addr true")
 
-    time = Time.parse("2011-01-02 13:14:15 UTC").to_i
-    events = [{"a"=>1},{"a"=>2}]
+    time = event_time("2011-01-02 13:14:15 UTC")
+    time_i = time.to_i
+    records = [{"a"=>1},{"a"=>2}]
+    events = [
+      ["tag1", time, {"REMOTE_ADDR"=>"129.78.138.66", "a"=>1}],
```
repeatedly: ditto

repeatedly:
Commented. Others are looks good.

lib/fluent/plugin/in_http.rb
tagomoris: IMO, using `nil` is bad idea for here, because we can't specify `nil` explicitly in configuration file.

lib/fluent/plugin/in_http.rb
tagomoris: Oops, that's right.

test/plugin/test_in_http.rb
tagomoris: Right. I'll fix it.

test/plugin/test_in_http.rb
tagomoris: No. Data to be committed is `records` in this test case, and events is used just for expectation.

tagomoris:
I pushed two commits for review comments.

repeatedly:
LGTM

PR #1311, opened 2016-11-14
filter-record-transformer: remove old behaviours

repeatedly:
We got many questions "Why transformed result is string?"
Time to change default behaviour before stable v0.14.

@sonots Could you check this patch?


sonots:
I want to drop https://github.com/fluent/fluentd/blob/6367d0cc8b1c09eb3e051e14cc40ae342047c788/lib/fluent/plugin/filter_record_transformer.rb#L323 too. 
Dropping this enforces users to specify record keys as "record[key]", it enables to remove method_missing and we can improve performance as record_modifier does.

sonots:
Codes look good, okay to merge this individually.

repeatedly:
@sonots Thanks for the review.

> Dropping this enforces users to specify record keys as "record[key]"

This is good idea. Could you write a patch?
You are best person to handle it.

PR #1312, opened 2016-11-15
Introduce server plugin helpers

tagomoris:
Server plugin helpers are to make plugin authors to write network servers very easily.
These make authors free from writing code to create sockets, handle event loops or threads.

The key features are:
* callback-based network daemon execution (plugin helper style)
* test driver support to wait startup/shutdown servers before actual test code running
* creating sockets via socket manager in default for coming multi-process configuration support

The full features of socket/server plugin helpers and TBDs are:
* TCP, UDP and TLS support (TLS will be implemented in future change)
* TCP/TLS keepalive (in future)
* IPv6 support

I committed all changes for this pull-request
* server plugin helper: TCP/UDP implementation complete
* tcp/udp input plugin: migrated to v0.14 APIs with server plugin helper
* forward input plugin: updated with server plugin helper
* tests of server plugin helper

To reduce the size of diff, keepalive support and socket plugin helper support will be implemented in following pull requests.
* socket plugin helper and its tests
* update of output plugins using sockets (including out_forward)
* http input plugin update (in future with keepalive support)

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -20,10 +20,6 @@ require 'fluent/msgpack_factory'
 require 'yajl'
 require 'digest'
 
-require 'fluent/plugin/socket_util'
-require 'fcntl'
-require 'cool.io'
-
 module Fluent::Plugin
   class ForwardInput < Input
     Fluent::Plugin.register_input('forward', self)
@@ -31,7 +27,7 @@ module Fluent::Plugin
     # See the wiki page below for protocol specification
     # https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1
 
-    helpers :event_loop
+    helpers :server
 
     LISTEN_PORT = 24224
 
@@ -39,12 +35,15 @@ module Fluent::Plugin
     config_param :port, :integer, default: LISTEN_PORT
     desc 'The bind address to listen to.'
     config_param :bind, :string, default: '0.0.0.0'
+
     config_param :backlog, :integer, default: nil
     # SO_LINGER 0 to send RST rather than FIN to avoid lots of connections sitting in TIME_WAIT at src
     desc 'The timeout time used to set linger option.'
     config_param :linger_timeout, :integer, default: 0
     # This option is for Cool.io's loop wait timeout to avoid loop stuck at shutdown. Almost users don't need to change this value.
     config_param :blocking_timeout, :time, default: 0.5
+    desc 'Try to resolve hostname from IP addresses or not.'
+    config_param :resolve_hostname, :bool, default: nil
     desc 'Connections will be disconnected right after receiving first message if this value is true.'
     config_param :deny_keepalive, :bool, default: false
 
@@ -91,6 +90,15 @@ module Fluent::Plugin
     def configure(conf)
       super
 
+      if @source_hostname_key
+        # TODO: add test
+        if @resolve_hostname.nil?
+          @resolve_hostname = true
+        elsif !@resolve_hostname # user specifies "false" in config
+          raise Fluent::ConfigError, "resolve_hostname must be true with source_hostname_key"
+        end
+      end
+
       if @security
         if @security.user_auth && @security.users.empty?
           raise Fluent::ConfigError, "<user> sections required if user_auth enabled"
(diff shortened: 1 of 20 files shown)


tagomoris:
@repeatedly Could you take a look for API design and basic features? If you are ok, I'll write tests and move forward to implement socket plugin helper.

lib/fluent/plugin/in_forward.rb
```diff
@@ -315,29 +310,31 @@ def on_message(msg, chunk_size, peeraddr)
...
-        record[@source_hostname_key] = peeraddr[2] if @source_hostname_key
+        record[@source_hostname_key] = remote_host if @source_hostname_key
         router.emit(tag, time, record)
         option = msg[3]
       end
 
       # return option for response
       option
+    ensure
+      p(here: "ensure of on_message", error: $!) if $!
```
repeatedly: Why use `p`?

tagomoris: It's just about debugging. I missed to remote it.

lib/fluent/plugin_helper/server.rb
```diff
@@ -0,0 +1,375 @@
...
+          def remote_host; peeraddr[2]; end
+          def remote_port; peeraddr[1]; end
+
+          def write(data)
+            send(data, 0) # flag - nothing specified
+          end
+        end
+
+        def initialize(sock, resolve_name, max_bytes, flags, log, under_plugin_development, &callback)
+          raise ArgumentError, "socket is a UDPSocket" unless sock.is_a?(UDPSocket)
```
repeatedly: `socket must be a UDPSocket: socket = #{sock}` or `socket is not a UDPSocket: socket = #{sock}`  is better.

repeatedly:
Basic implementation looks good.

Off topic: How about moving `under_plugin_development` into Engine like `Engine.under_plugin_development?`
Passing `under_plugin_development` to each method seems bit messy...

lib/fluent/plugin_helper/server.rb
```diff
@@ -0,0 +1,375 @@
...
+            on_read_with_connection(data)
+          else
+            on_read_without_connection(data)
+          end
+        end
+
+        def on_read_without_connection(data)
+          @data_callback.call(data)
+        rescue => e
+          p(here: "error without connection", error: e)
```
repeatedly: `p` debug.

lib/fluent/plugin_helper/server.rb
```diff
@@ -0,0 +1,375 @@
...
+          @log.error "unexpected error on reading data", host: remote_host, port: remote_port, error: e
+          @log.error_backtrace
+          close(true) rescue nil
+          raise if @under_plugin_development
+        end
+
+        def on_read_with_connection(data)
+          @data_callback.call(data, self)
+        rescue => e
+          p(here: "error with connection", error: e)
```
repeatedly: ditto

tagomoris:
@repeatedly No way. It's very bad practice to use Engine for such purpose. Engine is a singleton object, and using Engine for such purpose makes many messy code to overriding `Engine.under_plugin_development?` and to revert overriding in tests.

tagomoris:
I pushed updated commits to follow review comments, and add some changes:
* add `Fluent::PluginHelper::Server::EventHandler` module and added TCPServer/UDPServer under it
* add `Fluent::PluginHelper::Server::CallbackSocket` and its subclasses of it for tcp/udp

I changed to pass an instance object of CallbackSocket to callbacks, to avoid to pass instances of cool.io classes. If plugin helper passes instances of cool.io classes, users may use methods of these classes (even if we want to keep these methods as semi-private).
For instances of original classes, we can control the variations of these methods (for compatibility).

tagomoris:
There's no change about server plugin helper API for plugins in updated commits.

tagomoris:
I pushed all commits to add/fix server plugin helper, and tests which pass on my laptop.
I'll check CI failure and fix it, but anyway, @repeatedly could you review this change?

lib/fluent/plugin/in_tcp.rb
```diff
@@ -14,28 +14,66 @@
...
+                next
+              end
 
-      socket_manager_path = ENV['SERVERENGINE_SOCKETMANAGER_PATH']
-      if Fluent.windows?
-        socket_manager_path = socket_manager_path.to_i
+              tag = extract_tag_from_record(record)
+              tag ||= @tag
+              time ||= extract_time_from_record(record) || Fluent::EventTime.now
+              record[@source_host_key] = conn.remote_host if @source_host_key
```
repeatedly: use `@source_hostname_key` instead.

lib/fluent/plugin/in_udp.rb
```diff
@@ -14,24 +14,58 @@
...
+          @parser.parse(data) do |time, record|
+            unless time && record
+              log.warn "pattern not match", data: data
+              next
+            end
+
+            tag = extract_tag_from_record(record)
+            tag ||= @tag
+            time ||= extract_time_from_record(record) || Fluent::EventTime.now
+            record[@source_host_key] = sock.remote_host if @source_host_key
```
repeatedly: ditto

lib/fluent/plugin_helper/event_loop.rb
```diff
@@ -30,12 +30,16 @@ module EventLoop
...
       EVENT_LOOP_RUN_DEFAULT_TIMEOUT = 0.5
+      EVENT_LOOP_SHUTDOWN_TIMEOUT = 5
+      EVENT_LOOP_CLOCK_ID = Process::CLOCK_MONOTONIC_RAW rescue Process::CLOCK_MONOTONIC
 
       attr_reader :_event_loop # for tests
 
       def event_loop_attach(watcher)
         @_event_loop_mutex.synchronize do
           @_event_loop.attach(watcher)
+          @_event_loop_attached_watchers << watcher
```
repeatedly: This means `Loop#watchers` sometimes returns incorrect watchers?

lib/fluent/plugin_helper/server.rb
```diff
@@ -0,0 +1,494 @@
...
+          on(:data, &callback)
+        end
+
+        def on(event, &callback)
+          raise "BUG: this event is disabled for #{@server_type}: #{event}" unless @enabled_events.include?(event)
+          case event
+          when :data
+            @sock.data(&callback)
+          when :write_complete
+            cb = ->(){
```
repeatedly: Use one line for `:close` code.

lib/fluent/plugin_helper/server.rb
```diff
@@ -0,0 +1,494 @@
...
+            @log.error_backtrace
+            raise if @under_plugin_development
+          end
+        end
+
+        class TCPServer < Coolio::TCPSocket
+          SOCK_OPT_FORMAT = 'I!I!' # { int l_onoff; int l_linger; }
+
+          def initialize(sock, close_callback, resolve_name, linger_timeout, log, under_plugin_development, connect_callback)
+            raise ArgumentError, "socket must be a TCPSocket" unless sock.is_a?(TCPSocket)
```
repeatedly: add `: sock = #{sock}` to show actual object.

repeatedly:
New code/classes looks good.

nurse:
treasure-data/serverengine#82 fixes the test failure.

tagomoris:
@nurse @unak Great work, thank you!

lib/fluent/plugin/in_tcp.rb
```diff
@@ -14,28 +14,66 @@
...
+
+    config_param :blocking_timeout, :time, default: 0.5
 
-    config_set_default :port, 5170
     desc 'The payload is read up to this character.'
     config_param :delimiter, :string, default: "\n" # syslog family add "\n" to each message and this seems only way to split messages in tcp stream
 
-    def listen(callback)
-      log.info "listening tcp socket on #{@bind}:#{@port}"
+    def configure(conf)
```
tagomoris: Right.

lib/fluent/plugin_helper/event_loop.rb
tagomoris: `Loop#watchers` contains watchers which are NOT registered by event loop helpers, like TCP connection I/O watchers registered by TCP server watcher.
These watchers should be detached/closed by these watchers itself. Detaching such watchers by event loop helpers forcedly makes troubles.

tagomoris:
I pushed some commits for review comments, and will merge after CI green.

PR #1317, opened 2016-11-18
in_monitor_agent: Add include_config parameter to remove config field from response

repeatedly:
In our experience, almost users don't need config param because this parameter is not needed for buffer monitoring.
Removing it avoids unnecessary json processing and reduces response size.

`config` param itself is useful when user wants to check actual buffer configuration.
So add `with_config` query parameter to keep previous behaviour.

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -34,6 +34,7 @@ module Fluent::Plugin
     config_param :port, :integer, default: 24220
     config_param :tag, :string, default: nil
     config_param :emit_interval, :time, default: 60
+    config_param :include_config, :bool, default: true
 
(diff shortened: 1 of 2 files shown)


repeatedly:
@tagomoris How about this change?

tagomoris:
This fix changes default behavior, and there's no way to change plugin's default behavior (the only way to change output is to change fetcher - any other scripts or agents which users wrote).

I think it's better to have a new `config_param` to control whether output includes configuration or not without any query parameters.

repeatedly:
Applied review and change PR title.

tagomoris:
`test_configure` should have an assertion about `include_config`. Others look good to me.

repeatedly:
Good point. I just added assertion. After test passed, I will merge this patch.

PR #1319, opened 2016-11-21
Remove 'utc' field from TimeSlicedOutput conf to prevent ConfigError in plugin helpers

repeatedly:
Helpers/Compat utils  can't judge user specifies both localtime/utc or converted result.
This changes conf dumpped result but it seems acceptable.

Without this change, v0.12 TimeSlicedOutput plugins with `utc` parameter don't work.

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -81,7 +81,7 @@ module Fluent
       end
 
       def assume_timekey!
-        @_formatter = Fluent::Timezone.formatter(@_timezone, @_time_slice_format)
+        @_formatter = Fluent::TimeFormatter.new(@_time_slice_format, nil, @_timezone)
 
         return if self.metadata.timekey
         if self.respond_to?(:path) && self.path =~ /\.(\d+)\.(?:b|q)(?:[a-z0-9]+)/
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -621,6 +621,7 @@ module Fluent
           elsif conf['utc']
             conf['timezone'] = "+0000"
             conf['localtime'] = "false"
+            conf.delete('utc')
           elsif conf['localtime']
             conf['timezone'] = Time.now.strftime('%z')
             conf['localtime'] = "true"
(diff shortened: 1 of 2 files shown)


repeatedly:
This is incomplete. Incompatibility still exists by `conf['timezone'] = "+0000"`.
v0.12 formatter doesn't generate "+0000" suffix time.
So existing plugins assume formatted result is "Z" suffix time when `utc` is specified.

repeatedly:
@tagomoris Could you handle this problem? Currently, v0.14 has incompatibility issue with v0.12 TimeSlicedOutput.
S3 plugin has tests for thsese cases, so passed S3 test may cover other TimeSlicedOutput plugin cases.

repeatedly:
@tagomoris I added test for this case. Now, test failed.

```
Failure: test: check formatted time compatibility with utc. Should Z, not +00:00(FluentOutputTest::TimeSlicedOutputTest::test emit)
/Users/repeatedly/dev/fluentd/fluentd/lib/fluent/test/output_test.rb:139:in `block in run'
/Users/repeatedly/dev/fluentd/fluentd/lib/fluent/test/input_test.rb:124:in `block in run'
/Users/repeatedly/dev/fluentd/fluentd/lib/fluent/test/base.rb:71:in `run'
/Users/repeatedly/dev/fluentd/fluentd/lib/fluent/test/input_test.rb:123:in `run'
/Users/repeatedly/dev/fluentd/fluentd/lib/fluent/test/output_test.rb:124:in `run'
/Users/repeatedly/dev/fluentd/fluentd/test/test_output.rb:261:in `block (2 levels) in <class:TimeSlicedOutputTest>'
     258:         time = Time.parse("2016-11-08 12:00:00 UTC").to_i
     259:         d.emit({"a" => 1}, time)
     260:         d.expect_format %[2016-11-08T12:00:00Z\ttest\t{"a":1,"time":"2016-11-08T12:00:00Z"}\n]
  => 261:         d.run
     262:       end
     263:     end
     264:   end
<"2016-11-08T12:00:00Z\ttest\t{\"a\":1,\"time\":\"2016-11-08T12:00:00Z\"}\n"> expected but was
<"2016-11-08T12:00:00+00:00\ttest\t{\"a\":1,\"time\":\"2016-11-08T12:00:00+00:00\"}\n">
```

tagomoris:
I pushed two commits to fix this problem.
@repeatedly Could you check these and merge if these are ok?

repeatedly:
I tested the patch with s3 plugin test and it fixes broken tests.
Thanks for quick fix.

repeatedly:
And LGTM

PR #1325, opened 2016-11-25
optimize to split lines

### LINKED ISSUE #1324: in_tail uses regexp to split lines
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L573

It's much slower than using `index` and `slice!`.
Here's a micro bench code and its result:
https://gist.github.com/tagomoris/d9cf1e961378ab553b65fbaa844ba6df

tagomoris:
fixes #1324.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -570,8 +570,8 @@ module Fluent::Plugin
                   else
                     @buffer << @io.readpartial(2048, @iobuf)
                   end
-                  while line = @buffer.slice!(/.*?\n/m)
-                    @lines << line
+                  while idx = @buffer.index("\n".freeze)
+                    @lines << @buffer.slice!(0, idx + 1)
                   end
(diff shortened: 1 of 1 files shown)


tagomoris:
Benchmark is much nice to have.

lib/fluent/plugin/in_tail.rb
```diff
@@ -570,8 +570,8 @@ def on_notify
                   else
                     @buffer << @io.readpartial(2048, @iobuf)
                   end
-                  while line = @buffer.slice!(/.*?\n/m)
-                    @lines << line
+                  while idx = @buffer.index("\n")
```
repeatedly: For more small optimization, `"\n".freeze` is better,

tagomoris: Right.

tagomoris:
I pushed an updated commit.

repeatedly:
Looks good to me

PR #1326, opened 2016-11-25
Store regexp options properly

okkez:


okkez:
With Fluentd v0.14.6, fluent-plugin-grok-parser's test passed.
But with Fluentd v0.14.9 , fluent-plugin-grok-parser's test failed.

I noticed that `Fluent::Compat::TextParser::RegexpParser#configure` does not copy `Regexp#options`.

tagomoris:
LGTM

PR #1327, opened 2016-11-25
Close files explicitly in in_tail

tagomoris:
File handles are not closed explicitly now, and it makes troubles on Windows (especially in tests).
After detaching watchers (and stops all threads including event loop), no threads touches to IO objects and file handlers.
File handles should be closed after that.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -164,6 +164,12 @@ module Fluent::Plugin
       super
     end
 
+    def close
+      super
+      # close file handles after all threads stopped (in #close of thread plugin helper)
+      close_watcher_handles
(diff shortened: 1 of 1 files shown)


tagomoris:
I'm seriously confused about why #1312 fails on Windows CI only with ruby 2.1 and 2.2, but root cause of that problem is not to close files in IOHandler.
This change fixes that problem.

tagomoris:
@repeatedly could you check this?

repeatedly:
Will check :)

tagomoris:
@repeatedly ping

repeatedly:
LGTM

tagomoris:
Failing CI task is fixed at #1331.

PR #1328, opened 2016-11-25
Goodbye legacy compatibility

repeatedly:
- Remove `${message}` like short cut
- Use `emit_error_event` for using `@ERROR` label instead of dumping log

@sonots Could you check?


sonots:
Thanks!! LGTM

PR #1331, opened 2016-11-28
Make sure to specify use_to_json option

tagomoris:
For the behavior change of Oj 2.18.0.
CI tasks newly kicked are now failing with Oj 2.18.0. This change fixes it.


tagomoris:
@repeatedly could you check please?
Failing CI task is here: https://travis-ci.org/fluent/fluentd/jobs/178796884

repeatedly:
Good catch. LGTM

PR #1332, opened 2016-11-28
Add null check for `logger`

okkez:
If plugin uses `config_param :name, :string, deprecated: "This is
deprecated"` then we cannot use the plugin with plugins using v0.12
compatible layer.

### DIFF
--- a/test/config/test_configurable.rb
+++ b/test/config/test_configurable.rb
@@ -1307,7 +1307,7 @@ module Fluent::Config
           assert_nil(obj.log)
         end
 
-        test 'nothing raised if obsoleted parameter is configured' do
+        test 'NoMethodError is not raised if obsoleted parameter is configured' do
           obj = ConfigurableSpec::UnRecommended.new
           obj.log = nil
           assert_raise Fluent::ObsoletedParameterError.new("'key2' parameter is already removed: key2 has been removed.") do
--- a/test/config/test_configurable.rb
+++ b/test/config/test_configurable.rb
@@ -1298,6 +1298,24 @@ module Fluent::Config
         first_log = obj.log.logs.first
         assert{ first_log && first_log.include?("[error]") && first_log.include?("'key2' parameter is already removed: key2 has been removed.") }
       end
+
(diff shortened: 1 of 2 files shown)


okkez:
I've added `deprecated` option to fluent-plugin-grok-parser, but I could not complete it.
Because `in_tcp` in Fluentd v0.14.9 does not use v0.14 API.

[logger is nil](https://github.com/fluent/fluentd/blob/master/lib/fluent/config/section.rb#L105) and [parent is nil](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin.rb#L115)

See also
* https://github.com/fluent/fluent-plugin-grok-parser/commit/86067fac42d4871e8c8efbeeb001ec6a158db23f
* https://github.com/fluent/fluent-plugin-grok-parser/commit/e69a47b63cd43c379c15bc4bbe5b038b1230ee99

tagomoris:
LGTM.
Can you add any tests about it?

okkez:
I've added test and pushed changes.
@tagomoris could you review again?

test/config/test_configurable.rb
```diff
@@ -1295,6 +1295,26 @@ class Buggy1
...
+
+      sub_test_case 'logger is nil' do
+        test 'nothing raised if deprecated parameter is configured' do
+          obj = ConfigurableSpec::UnRecommended.new
+          obj.log = nil
+          obj.configure(config_element('ROOT', '', {'key1' => 'yay'}, []))
+          assert_nil(obj.log)
+        end
+
+        test 'nothing raised if obsoleted parameter is configured' do
```
tagomoris: This test case shows that "nothing raised", but test code does "assert_raise". Which one is correct?

okkez: Wow, I've fixed :bow:

tagomoris:
@okkez could you rebase this branch on master and push again? CI jobs are failing with many reasons which are already solved in other pull-requests.

okkez:
Rebased and pushed.

tagomoris:
LGTM. Thank you!

PR #1349, opened 2016-12-05
fluent-cat: Add --time-as-integer option for v0.12 or earlier versions

repeatedly:
For testing, re-install v0.12 is hard for users.


repeatedly:
@tagomoris How about this option?

lib/fluent/command/cat.rb
```diff
@@ -123,7 +128,7 @@ def run
     end
   end
 
-  def initialize(tag, connector)
+  def initialize(tag, connector, time_as_integer)
```
tagomoris: It's better to use keyword arguments for possible further options.

repeatedly:
applied review.

tagomoris:
LGTM

PR #1351, opened 2016-12-06
Add priority_key/facility_key options to in_syslog

knqyf263:
I want that `priority` and `facility` are in event record.
Currently, I use `fluent-plugin-record-reformer` to extract `priority` and `facility` values from tags.

However, these values have already been parsed in `in_syslog` plugin.
https://github.com/fluent/fluentd/blob/406c2a1dbfe0c2f14618b9251b67bc1e4f637813/lib/fluent/plugin/in_syslog.rb#L172-L173

I think it is useless to use another plugin for extracting `priority` and `facility`.
Then, I added `include_priority/facility` options to in_syslog.

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -85,6 +85,10 @@ module Fluent::Plugin
     config_param :include_source_host, :bool, default: false
     desc 'Specify key of source host when include_source_host is true.'
     config_param :source_host_key, :string, default: 'source_host'.freeze
+    desc 'The field name of the priority.'
+    config_param :priority_key, :string, default: nil
+    desc 'The field name of the facility.'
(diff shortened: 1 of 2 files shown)


repeatedly:
in_tcp's `source_hostname_key` approach is better than `include_priority` because someone wants to specify key.

knqyf263:
@repeatedly Thanks for your feedback!
I fixed. How about this?

tagomoris:
LGTM.

knqyf263:
Thanks!
How about a document?
Shoud I send a PR to `fluent/fluentd-docs` ? 0.12 or 0.14 ?

repeatedly:
I will backport this feature to v0.12 so v0.12 is enough.

knqyf263:
OK!
I sent a PR.
https://github.com/fluent/fluentd-docs/pull/249

PR #1356, opened 2016-12-07
Introduce socket plugin helper

tagomoris:
This change introduces socket plugin helper, and migrate `out_forward` to v0.14 API fully with plugin helpers.

This change does:
* add `fluent/plugin_helper/socket` and `#socket_create` (`_tcp`, `_udp`) methods
* fix server plugin helper to call it with existing socket (only for UDP, to create a server without any assigned port)
* fix output plugin to move it into recovery (retry) state when chunk is rollbacked in async mode
* rewrite `out_forward` with socket/timer/server/thread plugin helpers using `#try_write` for at-least-once mode

This change doesn't:
* add tests for socket plugin helper itself ... not yet
* tcp connection pools and/or keepalive implementation


tagomoris:

lib/fluent/plugin/output.rb
```diff
@@ -898,9 +900,9 @@ def commit_write(chunk_id, delayed: @delayed_commit, secondary: false)
         @retry_mutex.synchronize do
           if @retry # success to flush chunks in retries
             if secondary
-              log.warn "retry succeeded by secondary.", plugin_id: plugin_id, chunk_id: dump_unique_id_hex(chunk_id)
+              log.warn "retry succeeded by secondary.", chunk_id: dump_unique_id_hex(chunk_id)
```
repeatedly: Why remove plugin_id from the logs? Redundancy?

lib/fluent/plugin_helper/socket.rb
```diff
@@ -0,0 +1,101 @@
...
+      # close    : [-]
+      # terminate: [-]
+
+      include Fluent::PluginHelper::SocketOption
+
+      attr_reader :_sockets # for tests
+
+      # TODO: implement connection pool for specified host
+
+      def socket_create(host, port, proto: :tcp, **kwargs, &block)
```
repeatedly: How about `proto` argument is changed to required?

lib/fluent/plugin_helper/socket.rb
```diff
@@ -0,0 +1,101 @@
...
+        when :tcp
+          socket_create_tcp(host, port, **kwargs, &block)
+        when :udp
+          socket_create_udp(host, port, **kwargs, &block)
+        when :tls
+          socket_create_tls(host, port, **kwargs, &block)
+        when :unix
+          raise "not implemented yet"
+        else
+          raise ArgumentError, "invalid protocol"
```
repeatedly: "invalid protocol: #{proto}" is better.

lib/fluent/plugin/output.rb
tagomoris: Redundancy. Plugin logger fills it automatically now.

lib/fluent/plugin/output.rb
```diff
@@ -997,43 +1005,60 @@ def try_flush
...
+          update_retry_state(chunk.unique_id, using_secondary, e)
+
+          raise if @under_plugin_development && !@retry_for_error_chunk
+        end
+      end
+
+      def update_retry_state(chunk_id, using_secondary, error = nil)
+        @retry_mutex.synchronize do
+          @counters_monitor.synchronize{ @num_errors += 1 }
+          chunk_id_hex = dump_unique_id_hex(chunk_id)
```
tagomoris: There are almost no pass not to output a log with `chunk_id_hex`.
It's just:
* `@retry` exists, and
  * without error (called from `#rollback_commit`)
  * or
  * `@retry.limit?` is true (all chunks will be cleared)

That variable is referred in most cases, so I think it's worth to assign into a variable to reduce code complexity.

lib/fluent/plugin_helper/socket.rb
tagomoris: Good idea. I'll update to `socket_create(proto, host, port, **kwargs, &block)`.

lib/fluent/plugin_helper/socket.rb
tagomoris: gotcha.

PR #1357, opened 2016-12-08
fix to raise error if injected hostname_key is also specified in buffer chunk key

### LINKED ISSUE #1353: extract_placeholders cannot extract ${hostname} properly
- fluentd v0.14.9
- Debian GNU/Linux

config:
```
<source>
  @type dummy
  tag raw.dummy
  dummy [
  {"host":"66.102.9.80", "message":"test1"},
  {"host":"66.102.9.80", "message":"test2"},
  {"host":"66.102.9.80", "message":"test3"},
  {"host":"66.102.9.80", "message":"test4"},
  {"host":"66.102.9.80", "message":"test5"}
  ]
</source>

<match raw.**>
  @type geoip
  tag ${tag[0]}.geoip.${hostname}
  geoip_lookup_key  host
  #geoip_database    "/path/to/your/GeoIPCity.dat"
  <record>
    city            ${city["host"]}
    latitude        ${latitude["host"]}
    longitude       ${longitude["host"]}
    country_code3   ${country_code3["host"]}
    country         ${country_code["host"]}
    country_name    ${country_name["host"]}
    dma             ${dma_code["host"]}
    area            ${area_code["host"]}
    region          ${region["host"]}
  </record>
  <buffer tag, hostname>
    flush_interval    1s
  </buffer>
  <inject>
    hostname_key hostname
  </inject>
</match>

<match raw.**>
  @type stdout
</match>
```

https://github.com/okkez/fluent-plugin-geoip/blob/migrate-to-v0.14-api/lib/fluent/plugin/out_geoip.rb

I'm trying to migrate fluent-plugin-geoip's out_geoip to Fluentd v0.14 API.
I want to extract placeholders `${hostname}` using hostname field added by inject plugin helper.

Expected behavior:

tag will include hostname added by inject plugin helper.

Actual behavior:

tag does not include hostname added by inject plugin helper.

tagomoris:
Inject plugin helper does record.dup, so that Output plugin code cannot find any
side effect of inject plugin helper. It means output cannnot do chunking with injected values.

On the other hand, there is just few good use cases about chunking with injected hostname.

* chunking hostname (just after injected) is to route events with hostname... and hostname is unique on a host
* just one good use-case, to extract hostname into ${hostname} in configured values
  * but it can be done by ruby code eval

This fixes #1353.

### DIFF
--- a/lib/fluent/plugin_helper/inject.rb
+++ b/lib/fluent/plugin_helper/inject.rb
@@ -97,6 +97,17 @@ module Fluent
         if @inject_config
           @_inject_hostname_key = @inject_config.hostname_key
           if @_inject_hostname_key
+            if self.respond_to?(:buffer_config)
+              # Output plugin cannot use "hostname"(specified by @hostname_key),
+              # injected by this plugin helper, in chunk keys.
+              # This plugin helper works in `#format` (in many cases), but modified record
+              # don't have any side effect in chunking of output plugin.
+              if self.buffer_config.chunk_keys.include?(@_inject_hostname_key)
+                log.error "Use filters to inject hostname to use it in buffer chunking."
+                raise Fluent::ConfigError, "the key specified by 'hostname_key' in <inject> cannot be used in buffering chunk key."
+              end
+            end
+
             @_inject_hostname =  @inject_config.hostname
             unless @_inject_hostname
               @_inject_hostname = Socket.gethostname
--- a/test/plugin_helper/test_inject.rb
+++ b/test/plugin_helper/test_inject.rb
@@ -1,5 +1,6 @@
 require_relative '../helper'
 require 'fluent/plugin_helper/inject'
+require 'fluent/plugin/output'
 require 'fluent/event'
 require 'time'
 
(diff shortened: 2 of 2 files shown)


tagomoris:

lib/fluent/plugin_helper/inject.rb
```diff
@@ -97,6 +97,17 @@ def configure(conf)
...
           @_inject_hostname_key = @inject_config.hostname_key
           if @_inject_hostname_key
+            if self.respond_to?(:buffer_config)
+              # Output plugin cannot use "hostname"(specified by @hostname_key),
+              # injected by this plugin helper, in chunk keys.
+              # This plugin helper works in `#format` (in many cases), but modified record
+              # don't have any side effect in chunking of output plugin.
+              if self.buffer_config.chunk_keys.include?(@_inject_hostname_key)
+                log.error "Use filters to inject hostname to use it in buffer chunking."
+                raise Fluent::ConfigError, "the key specified in 'hostname_key' cannot be used in buffering chunk key."
```
repeatedly: `the key specified in 'hostname_key' of <inject>` is better.

repeatedly:
Others LGTM

tagomoris:
Pushed the fixed commit.

PR #1359, opened 2016-12-08
introduce unrecoverable error type, and logs to show Fluentd process/worker started

tagomoris:
Currently, unrecoverable errors (which serverengine doesn't try to restart workers about) are only configuration errors.
But some other situation cannot be retried. For example:
* per-worker root directory (will be introduced next) exists but it is a file
* failed to bind an address/port in server plugin helper
* failed to read a file from in_tail by permission error
* or many others

We are using ConfigError for such cases, but it's nonsense in fact.

This change also adds unrecoverable cases for scripting errors while loading plugin files.
Fluentd is now restarting for such cases, but it's just wrong behavior.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -636,31 +636,31 @@ module Fluent
 
       begin
         block.call
-      rescue Fluent::ConfigError
+      rescue Fluent::ConfigError => e
         logging_with_console_output do |log|
-          log.error "config error", file: @config_path, error: $!
+          log.error "config error", file: @config_path, error: e
           log.debug_backtrace
         end
         unrecoverable_error = true
-      rescue Fluent::UnrecoverableError
+      rescue Fluent::UnrecoverableError => e
         logging_with_console_output do |log|
-          log.error $!.message, error: $!
+          log.error e.message, error: e
           log.error_backtrace
         end
         unrecoverable_error = true
-      rescue ScriptError # LoadError, NotImplementedError, SyntaxError
+      rescue ScriptError => e # LoadError, NotImplementedError, SyntaxError
         logging_with_console_output do |log|
-          if $!.respond_to?(:path)
-            log.error $!.message, path: $!.path, error: $!
+          if e.respond_to?(:path)
+            log.error e.message, path: e.path, error: e
           else
-            log.error $!.message, error: $!
+            log.error e.message, error: e
           end
           log.error_backtrace
         end
         unrecoverable_error = true
-      rescue
+      rescue => e
         logging_with_console_output do |log|
-          log.error "unexpected error", error: $!
+          log.error "unexpected error", error: e
           log.error_backtrace
         end
       end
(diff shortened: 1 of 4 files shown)


tagomoris:
Actually, the biggest improvement in this change is testing about:
* Fluentd process started correctly
* Fluentd process failed to start correctly

tagomoris:

lib/fluent/supervisor.rb
```diff
@@ -617,39 +618,54 @@ def flush_buffer
...
+
     def main_process(&block)
       Process.setproctitle("worker:#{@process_name}") if @process_name
 
-      configuration_error = false
+      unrecoverable_error = false
 
       begin
         block.call
       rescue Fluent::ConfigError
```
repeatedly: Use `Fluent::ConfigError => e` and remove `$!`.

lib/fluent/supervisor.rb
```diff
@@ -617,39 +618,54 @@ def flush_buffer
...
-          log.level = @log_level
-          console = log.enable_debug
-          console.error "config error", file: @config_path, error: $!.to_s
-          console.debug_backtrace
+        logging_with_console_output do |log|
+          log.error "config error", file: @config_path, error: $!
+          log.debug_backtrace
+        end
+        unrecoverable_error = true
+      rescue Fluent::UnrecoverableError
```
repeatedly: ditto

lib/fluent/supervisor.rb
```diff
@@ -617,39 +618,54 @@ def flush_buffer
...
+        logging_with_console_output do |log|
+          log.error $!.message, error: $!
+          log.error_backtrace
+        end
+        unrecoverable_error = true
+      rescue ScriptError # LoadError, NotImplementedError, SyntaxError
```
repeatedly: ditto

test/command/test_fluentd.rb
```diff
@@ -0,0 +1,276 @@
...
+        "error_class=SyntaxError",
+        "in_buggy.rb:5: syntax error, unexpected end-of-input, expecting keyword_end",
+      )
+      assert !process_exist?(@pid)
+    end
+  end
+
+  sub_test_case 'configuration to load plugin which raises unrecoverable error in #start' do
+    test 'failed to start' do
+      script =  "require 'fluent/plugin/input'\n"
```
repeatedly: Use heredocument is better.

repeatedly:
Others are LGTM and adding start test sis awesome.
BTW, test on Windows is unstable...

test/command/test_fluentd.rb
tagomoris: Writing ruby code in here document breaks indentation on my environment... :P

tagomoris:
I added some fixes about executing worker processes and showing process ids in log.
@repeatedly could you review these changes again?

repeatedly:
LGTM and good fix for windows environment 👍

PR #1361, opened 2016-12-09
formatter_csv: Change fields parameter to required

repeatedly:
Empty fields is error prone because it generates empty line.


tagomoris:
IMO it should raise error for empty field configured.

repeatedly:
It seems good > raise when fields is empty.

repeatedly:
Review applied. After test passed, will merge.

tagomoris:
Please send another pull-req to master branch right now to raise error for empty fields.
It may make regressions if missed now.

PR #1362, opened 2016-12-10
allow to raise Errno::EACCESS to bind sockets already bound

tagomoris:
Windows may raise Errno::EACCESS for such case, when
bind operation fails to access port/file descriptor.

This change is to improve test stability.


tagomoris:
Let me merge this now... it's just about testing.

Issue #1364, opened 2016-12-12
NoMethodError when using in_monitor_agent with a unbuffered plugin


shuji-koike:
When using `in_monitor_agent` with ~`out_null` plugin~ a unbuffered plugin (which `@buffer` is `nil`), NoMethodError will occur each time `http://0.0.0.0:24220/api/plugins` is accessed.

Minimal config to reproduce the problem.

```
<source>
  @type monitor_agent
  bind 0.0.0.0
  port 24220
</source>

<match **>
  @type null
</match>
```

start fluentd and, kick monitor_agent.

```
curl http://0.0.0.0:24220/api/plugins
```

NoMethodError occurs.

```
2016-12-12 15:16:34 +0900 [warn]: plugin/in_monitor_agent.rb:341:rescue in block in get_monitor_info: unexpected error in monitoring plugins key="buffer_queue_length" plugin=Fluent::Plugin::NullOutput error_class=NoMethodError error="undefined method `queue' for nil:NilClass"
2016-12-12 15:16:34 +0900 [warn]: plugin/in_monitor_agent.rb:341:rescue in block in get_monitor_info: unexpected error in monitoring plugins key="buffer_total_queued_size" plugin=Fluent::Plugin::NullOutput error_class=NoMethodError error="undefined method `stage_size' for nil:NilClass"
```

shuji-koike:
https://github.com/fluent/fluentd/pull/1193/files#diff-0ad62dd2012fecaa6ed2232f8918c528R351

Seems like This fix around `catch(:skip)` need some more work to do, right?

shuji-koike:
Turns out that, not only `out_null` but some plugins witch `@buffer` is `nil` will cause the problem.

repeatedly:
Patch is merged.

Issue #1335, opened 2016-11-29
Fluent::Plugin::MonitorAgentInput and RewriteTagFilterOutput


DanKans:
Hi,
I'm using fluentd 0.14 and during querying endpoint for monitor I got these warnings from fluentd agent:

```
2016-11-29 16:03:27 +0000 [warn]: [Fluent::Plugin::MonitorAgentInput] unexpected error in monitoring plugins key="buffer_queue_length" plugin=Fluent::RewriteTagFilterOutput error_class=NoMethodError error="undefined method `queue' for nil:NilClass"
2016-11-29 16:03:27 +0000 [warn]: [Fluent::Plugin::MonitorAgentInput] unexpected error in monitoring plugins key="buffer_total_queued_size" plugin=Fluent::RewriteTagFilterOutput error_class=NoMethodError error="undefined method `stage_size' for nil:NilClass"
```

P.S
How it is possible that buffer_queue_length > buffer_queue_limit? is it calculated according to buffer_queue_limit and buffer_chunk_limit? 

```json
    {
      "plugin_id": "object:1e89580",
      "plugin_category": "output",
      "type": "elasticsearch",
      "config": {
        "@type": "elasticsearch",
        "@log_level": "info",
        "logstash_format": "true",
        "logstash_prefix": "application",
        "type_name": "k8s",
        "host": "ELK",
        "port": "9200",
        "request_timeout": "2s",
        "buffer_type": "memory",
        "buffer_chunk_limit": "1m",
        "buffer_queue_limit": "1024",
        "flush_interval": "2s",
        "retry_wait": "2s",
        "max_retry_wait": "10s"
      },
      "output_plugin": true,
      "buffer_queue_length": 3630,
      "buffer_total_queued_size": 128630,
      "retry_count": 24
    },
```
Thanks

tagomoris:
@DanKans Could you paste the configuration file and the version of Fluentd you are using?

Fluentd v0.14 controls buffer queue with its bytesize, not length. So your buffer contains smaller chunks than buffer chunk limit (1MB, chunk_limit_size in v0.14 style), your buffer queue can contain more chunks than 1024.

tagomoris:
This issue looks duplicated one of #1365.

PR #1365, opened 2016-12-12
in_monitor_agent: get_monitor_info: fix NoMethodError

shuji-koike:
> NoMethodError when using in_monitor_agent with out_null plugin #1364

This is my fix for #1364.

----

I thought it may be better to add a `resuce e: NoMethodError` around here -> https://github.com/shuji-koike/fluentd/blob/be7ffb2/lib/fluent/plugin/in_monitor_agent.rb#L344-L346
If so, I will fix my commit :trollface:

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -216,6 +216,12 @@ module Fluent::Plugin
       end
     end
 
+    def initialize
+      super
+
+      @first_warn = false
+    end
+
     def start
       super
 
@@ -259,8 +265,8 @@ module Fluent::Plugin
 
     MONITOR_INFO = {
       'output_plugin' => ->(){ is_a?(::Fluent::Plugin::Output) },
-      'buffer_queue_length' => ->(){ throw(:skip) unless instance_variable_defined?(:@buffer); @buffer.queue.size },
-      'buffer_total_queued_size' => ->(){ throw(:skip) unless instance_variable_defined?(:@buffer); @buffer.stage_size },
+      'buffer_queue_length' => ->(){ throw(:skip) unless instance_variable_defined?(:@buffer) && !@buffer.nil?; @buffer.queue.size },
+      'buffer_total_queued_size' => ->(){ throw(:skip) unless instance_variable_defined?(:@buffer) && !@buffer.nil?; @buffer.stage_size },
       'retry_count' => ->(){ instance_variable_defined?(:@num_errors) ? @num_errors : nil },
     }
 
@@ -345,6 +351,12 @@ module Fluent::Plugin
         begin
           catch(:skip) do
             obj[key] = pe.instance_exec(&code)
+          end
+        rescue NoMethodError => e
+          unless @first_warn
+            log.error "NoMethodError in monitoring plugins", key: key, plugin: pe.class, error: e
+            log.error_backtrace
+            @first_warn = true
           end
         rescue => e
           log.warn "unexpected error in monitoring plugins", key: key, plugin: pe.class, error: e


tagomoris:
Code looks good to me.
Adding `rescue NoMethodError` is good idea. That should have `log.error_backtrace` to show where the bug is clearly.

repeatedly:
Good catch :)
And additional commits are welcome.

shuji-koike:
@tagomoris san

Adding `rescue NoMethodError` with `log.error_backtrace` would rather increase logs.
I want to reduce (suppress) logs for this occasion, because `@buffer` may be nil in some plugins and it is not a bug or an error.

So, may I leave my fix as is?

tagomoris:
@shuji-koike it makes sense. But it's worth to provide information to make it easy to debug such problems.
So, how about to show warnings at the only first time of rescue such exception? It's a kind of a bit messy code, but it satisfies requirements both of users and developers.

```ruby
# initialize @first_warn = false # in #initialize
 rescue => e
  log.error # brabrabra
  if !@first_warn
    @first_warn = true
    log.error_stacktrace
  end
```

shuji-koike:
@tagomoris san

I've updated my fix -> b38a791

I didn't notice there was a similar issue -> #1335, sorry!

tagomoris:
Hmm, This fix still have some problems:
* Including the whole logging in `unless @first_warn` makes that it warns NoMethodError only once (maybe just after Fluentd startup)
 * Users may miss just only one warning... we should warn it everytime (without backtrace)
* this patch uses `warn` about NoMethodError itself, but uses `error_backtrace` (it shows backtrace in `ERROR` level)

We should show `error` level log everytime (out of `unless @first_warn` clause), and use `error_backtrace` in `unless @first_warn`.

tagomoris:
@shuji-koike ping?

shuji-koike:
@tagomoris san
Sorry for my late reply, and a Happy New Year!
I updated my fix -> 625752c
Since fluentd will suppress duplicate stacktraces, I removed the `@first_warn` flag.

shuji-koike:
I would have to say that I am not really convinced with this fix.

I understand that logging is important.
But users have nothing to do with this error.

I think the real problem is...

In `0.12.x`, `@buffer` is defined only in `BufferedOutput` plugins.
https://github.com/shuji-koike/fluentd/blob/v0.12.30/lib/fluent/output.rb#L213

`in_monitor_agent` checks whether `@buffer` is defined to avoid (suppress) errors.
https://github.com/shuji-koike/fluentd/blob/v0.14.9/lib/fluent/plugin/in_monitor_agent.rb#L253-L254

In `0.14.x`, `@buffer` will be defined in all `Output` plugins.
https://github.com/shuji-koike/fluentd/blob/v0.14.9/lib/fluent/plugin/output.rb#L185

Therefore, checking `instance_variable_defined?(:@buffer)` may have become pointless.

I removed the `@buffer.nil?` check from my fix, but may I put them back?

tagomoris:
"suppress repeated stacktrace" option will suppress the stacktraces which is just equal to the previous one. If any error occurs in other place of code, the same stacktrace will be shown in log.
Otherwise, "suppressed same stacktrace" log messages are shown always. It looks a bit messy for me.
So, IMO it's better not to call `log.error_backtrace` twice and more.

tagomoris:
Basically, checking `@buffer.nil?` is good idea, in addition to `instance_variable_defined?(:@buffer)`, because there are still some plugins without `@buffer` instance variables (in `match` sections), like `Fluent::Plugin::BareOutput`.
But it's not the main issue about this pull-request.

This pull-request is to show where the NoMethodError occurs and where the bug hits. This requires stacktrace for further investigation, so we need to handle NoMethodError in special way.

shuji-koike:
@tagomoris san

Thanks for your advice 😸 
I added `nil?` check "in addition to instance_variable_defined", and brought back the `@first_warn` flag. -> 919ebb2

tagomoris:
LGTM

tagomoris:
@shuji-koike Merged. Thank you for contribution!

PR #1366, opened 2016-12-13
Add slow_flush_log_threshold parameter

repeatedly:
This is useful for debugging of buffer flush issue.

I set default value is `5.0`. This is good for local destination but `5.0` seems small for cloud based services.
Should we set large value for default setting?

This patch is for v0.12 but support this feature in v0.14 is also good.

### DIFF
--- a/lib/fluent/output.rb
+++ b/lib/fluent/output.rb
@@ -204,7 +204,7 @@ module Fluent
     desc 'The number of threads to flush the buffer.'
     config_param :num_threads, :integer, default: 1
     desc 'The threshold to show slow flush logs'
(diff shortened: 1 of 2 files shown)


tagomoris:
`slow_flush_log_threshold`?

repeatedly:
It is also good. I refer `long_query_time` in mysql first.
Adding log is no problem for me.

repeatedly:
Rename and change default values.

repeatedly:
After test passed, I will merge this patch.

tagomoris:
Could you create a pull-request for master branch before merging?
I want to make it sure to port this feature to v0.14.

PR #1369, opened 2016-12-14
raise ConfigError when fields parameter is empty array

repeatedly:


tagomoris:
This code doesn't raise errors for configurations including both of empty string and non-empty string: `fields a,,b`.
Is it intentional?

repeatedly:
Yes

tagomoris:
Okay. LGTM.

PR #1370, opened 2016-12-14
Add slow_flush_log_threshold parameter

repeatedly:
master version of https://github.com/fluent/fluentd/pull/1366

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1032,7 +1032,7 @@ module Fluent
         elapsed_time = Process.clock_gettime(PROCESS_CLOCK_ID) - start
         if elapsed_time > @slow_flush_log_threshold
           log.warn "buffer flush took longer time than slow_flush_log_threshold:",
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -985,21 +987,28 @@ def try_flush
         end
 
         begin
+          chunk_write_start = Time.now
```
tagomoris: Use `Process.clock_gettime(PROCESS_CLOCK_ID)` instead of `Time.now`.
`Time.now` is affected by system clock modification.

lib/fluent/plugin/output.rb
```diff
@@ -1019,6 +1028,14 @@ def try_flush
         end
       end
 
+      def check_slow_flush(start)
+        elapsed_time = Time.now - start
```
tagomoris: ditto.

repeatedly:
Applied review.

tagomoris:
LGTM.

PR #1374, opened 2016-12-15
Root directory per worker process

tagomoris:
This change introduces system-wide "root directory" and root directory per plugins.
It's enabled when:
* `root_dir` specified in `<system>` section
* `@id` configured in each plugin configuration sections

The root directories per plugin instances must be unique in a filesystem (between many workers of fluentd). So it uses `worker_id` provided by supervisor process via environment variable.
The path of directories are: `/system-wide-root/workerN/plugin-id` (generated and provided via `Fluent::Plugin::Base#plugin_root_dir`). Plugins will creates files or directories under that.

This change also adds 2 implementations to use this new feature:
* `buf_file`: configure buffer file `path` automatically
* `storage_local`: configure storage file `path` automatically

I have an additional plan to fix `in_tail` to use storage plugin as replacement of `pos_file`, but it's will be added by another pull-request.

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -32,9 +32,8 @@ module Fluent
 
       DIR_PERMISSION = 0755
 
-      # TODO: buffer_path based on system config
       desc 'The path where buffer chunks are stored.'
-      config_param :path, :string
+      config_param :path, :string, default: nil
 
       config_set_default :chunk_limit_size, DEFAULT_CHUNK_LIMIT_SIZE
       config_set_default :total_limit_size, DEFAULT_TOTAL_LIMIT_SIZE
@@ -52,6 +51,14 @@ module Fluent
       def configure(conf)
         super
 
+        unless @path
+          if root_dir = owner.plugin_root_dir
+            @path = File.join(root_dir, 'buffer')
+          else
+            raise Fluent::ConfigError, "buffer path is not configured. specify 'path' in <buffer>"
+          end
+        end
+
         type_of_owner = Plugin.lookup_type_from_class(@_owner.class)
         if @@buffer_paths.has_key?(@path) && !buffer_path_for_test?
           type_using_this_path = @@buffer_paths[@path]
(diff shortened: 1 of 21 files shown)


tagomoris:

lib/fluent/plugin/buf_file.rb
```diff
@@ -56,6 +51,14 @@ def initialize
       def configure(conf)
         super
 
+        unless @path
```
repeatedly: In the future, should we validate `@path` value for avoding conflict trouble?
Fixed path should not allowed with multiprocess configuration or
support `%{worker_id}`/`%{plugin_id}` placeholder?

tagomoris: Buffer chunks have `unique_id`, which are generated with time and random value.
So it SHOULD NOT conflict with other chunks in other processes. No need to do such thing.

tagomoris: Ahhh, there are some problem about resuming at startup...

tagomoris: Anyway, `plugin_id` can't help that situation. `plugin_id` is (should be) unique in a worker process, not server global.

lib/fluent/plugin_id.rb
```diff
@@ -59,5 +64,17 @@ def plugin_id
         "object:#{object_id.to_s(16)}"
       end
     end
+
+    def plugin_root_dir
+      return @_plugin_root_dir if @_plugin_root_dir
+      return nil unless system_config.root_dir
```
repeatedly: Put `log.warn` is better?
Maybe, users hard to find the cause why `buffer_path`/`storage` is not configured automatically.

tagomoris: No, it's not about users.
This code is called by plugins, and plugin should check whether `plugin_root_dir` is available or not by return value, and raise errors with messages which shows how to configure that plugin.

tagomoris:
Can I merge this?

repeatedly:
LGTM

PR #1378, opened 2016-12-16
Fix warnings

repeatedly:
Since ruby 2.4, `forwardable` checks the method visibility and it shows the warnings.
This patch simply removes private for targets methods.


tagomoris:
LGTM

PR #1382, opened 2016-12-20
Migrate plugins with server plugin helper

tagomoris:
This change updates `in_syslog` and `in_forward`.

Updating `in_syslog` is to use server plugin helper to enable multi-worker processing.
I also added/updated some parameters to inject hostname/address of data source, for consistent parameter names with `in_forward`.

I also found that `in_forward` don't have a parameter to inject source host address (not hostname).
So I added it too.


tagomoris:

tagomoris:
@repeatedly ping

lib/fluent/plugin/in_forward.rb
```diff
@@ -340,12 +350,31 @@ def check_and_skip_invalid_event(tag, es, remote_host)
       new_es
     end
 
-    def add_source_host(es, host)
+    def add_source_host(es, conn)
```
repeatedly: `add_source_info` or something is better.

test/plugin/test_in_forward.rb
```diff
@@ -34,7 +34,9 @@ def teardown
     port #{PORT}
     bind 127.0.0.1
   ]
-  PEERADDR = ['?', '0000', '127.0.0.1', '127.0.0.1']
+  LOCALHOST_HOSTNAME_GETTER = ->(){sock = UDPSocket.new(::Socket::AF_INET); sock.do_not_reverse_lookup = false; sock.connect("127.0.0.1", 2048); sock.peeraddr[2] }
+  LOCALHOST_HOSTNAME = LOCALHOST_HOSTNAME_GETTER.call
+  DUMMY_SOCK = Struct.new(:remote_host, :remote_addr, :remote_port).new(LOCALHOST_HOSTNAME_GETTER.call, "127.0.0.1", 0)
```
repeatedly: Use `LOCALHOST_HOSTNAME` instead of `LOCALHOST_HOSTNAME_GETTER.call` in new

test/plugin/test_in_syslog.rb
```diff
@@ -3,21 +3,21 @@
...
-      socket_manager_path = ServerEngine::SocketManager::Server.generate_path
-      @server = ServerEngine::SocketManager::Server.open(socket_manager_path)
-      ENV['SERVERENGINE_SOCKETMANAGER_PATH'] = socket_manager_path.to_s
-    end
-
-    def shutdown
-      @server.close
-    end
-  end
+  # class << self
```
repeatedly: Why keep these commented out code?

test/plugin/test_in_forward.rb
tagomoris: ok.

test/plugin/test_in_syslog.rb
tagomoris: I missed to delete these lines... will remove.

tagomoris:
I pushed commits for review comments. I'll merge this change after CI green.

PR #1389, opened 2016-12-23
fix bug about chunk response handling

### LINKED ISSUE #1388: at-least-once acknowledgements are not processed
My setup is like this:

forwarder -> aggregator
td-agent 0.14.10 on both

When I set require_ack_response to true in my forwarder, 

I can see acks being logged on the aggregator:

2016-12-23 11:20:06 +1100 [trace]: [Fluent::Plugin::ForwardInput] sent response to fluent socket address="10.18.68.219" response={"ack"=>"VESFkVa4eEpn+/hwFcOpLw==\n"}

But the forwarder still thinks ACKs have not been received
2016-12-23 11:20:36 +1100 [warn]: [Fluent::Plugin::ForwardOutput] failed to flush the buffer chunk, timeout to commit. chunk_id="5444859156b8784a67fbf87015c3a92f" flushed_at=2016-12-23 11:20:06 +1100
[more of these..]
2016-12-23 11:29:28 +1100 [warn]: [Fluent::Plugin::ForwardOutput] failed to flush the buffer chunk, timeout to commit. chunk_id="5444859156b8784a67fbf87015c3a92f" flushed_at=2016-12-23 11:28:57 +1100

Chunk id matches:
bin2hex(base64_decode('VESFkVa4eEpn+/hwFcOpLw==')); => "5444859156b8784a67fbf87015c3a92f"

And I can see that the buffer is not cleared until the retry_limit in the file system.

so far I have tried:
* increasing/decreasing flush_interval on the forwarder
* increasing/decreasing buffer_chunk_limit in the forwarder
* increasing/decreasing retry_wait in the forwarder
* increasing/decreasing ack_response_timeout in the forwarder

tcpdump is confirming traffic from the aggregator to forwarder.

Is there another setting that needs to be looked at?

tagomoris:
fixes #1388.

v0.14.11 should be released after this change is merged for quick regression fix.
The changes between v0.14.10 release and this change will be introduced at v0.14.12.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -178,7 +178,7 @@ module Fluent::Plugin
       # But it should be overwritten by ack_response_timeout to rollback chunks after timeout
       if @ack_response_timeout && @delayed_commit_timeout != @ack_response_timeout
         log.info "delayed_commit_timeout is overwritten by ack_response_timeout"
-        @delayed_commit_timeout = @ack_response_timeout
+        @delayed_commit_timeout = @ack_response_timeout + 2 # minimum ack_reader IO.select interval is 1s
       end
 
       @rand_seed = Random.new.seed
@@ -214,22 +214,23 @@ module Fluent::Plugin
       select_a_healthy_node{|node| node.send_data(tag, chunk) }
     end
 
-    ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :node, :time, :timeout) do
+    ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :chunk_id_base64, :node, :time, :timeout) do
       def expired?(now)
         time + timeout < now
       end
     end
 
     def try_write(chunk)
+      log.trace "writing a chunk to destination", chunk_id: dump_unique_id_hex(chunk.unique_id)
       if chunk.empty?
         commit_write(chunk.unique_id)
         return
       end
       tag = chunk.metadata.tag
       sock, node = select_a_healthy_node{|n| n.send_data(tag, chunk) }
-      chunk_id = Base64.encode64(chunk.unique_id)
+      chunk_id_base64 = Base64.encode64(chunk.unique_id)
       current_time = Process.clock_gettime(PROCESS_CLOCK_ID)
-      info = ACKWaitingSockInfo.new(sock, chunk_id, node, current_time, @ack_response_timeout)
+      info = ACKWaitingSockInfo.new(sock, chunk.unique_id, chunk_id_base64, node, current_time, @ack_response_timeout)
       @sock_ack_waiting_mutex.synchronize do
         @sock_ack_waiting << info
       end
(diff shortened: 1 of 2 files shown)


tagomoris:

lib/fluent/plugin/out_forward.rb
```diff
@@ -214,22 +214,23 @@ def write(chunk)
       select_a_healthy_node{|node| node.send_data(tag, chunk) }
     end
 
-    ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :node, :time, :timeout) do
+    ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :chunk_id_base64, :node, :time, :timeout) do
```
repeatedly: Just one idea.
How about using `def chunk_id_base64` instead of struct member?
It reduces additional member cost.

lib/fluent/plugin/out_forward.rb
```diff
@@ -178,7 +178,7 @@ def start
       # But it should be overwritten by ack_response_timeout to rollback chunks after timeout
       if @ack_response_timeout && @delayed_commit_timeout != @ack_response_timeout
         log.info "delayed_commit_timeout is overwritten by ack_response_timeout"
```
repeatedly: OT. Adding `why` message is better for users to understand why `delayed_commit_timeout` is overwritten in out_forward.

lib/fluent/plugin/out_forward.rb
```diff
@@ -178,7 +178,7 @@ def start
       # But it should be overwritten by ack_response_timeout to rollback chunks after timeout
       if @ack_response_timeout && @delayed_commit_timeout != @ack_response_timeout
```
repeatedly: If `@delayed_commit_timeout` and `@ack_response_timeout` are the same, `@delayed_commit_timeout = @ack_response_timeout + 2` line is skipped. Is this okay?

tagomoris: ok, because these are configured intentionally by users, and don't cause problems for longer configured values.

lib/fluent/plugin/out_forward.rb
tagomoris: It's not bad idea, but it looks better for me to add these methods to `chunk`:
* `chunk_id_hex`
* `chunk_id_base64`

These values are called many times (especially _hex for logging), and should be cached not to calculate 2 or more times.
How do you think about it?

repeatedly: Hm... it also seems good.

tagomoris: Another one: `chunk.unique_id(:hex)` (or :base64)

lib/fluent/plugin/out_forward.rb
repeatedly: I see.

lib/fluent/plugin/out_forward.rb
repeatedly: I like method approach, `unique_id_hex`, rathar than `unique_id(:hex)`.

tagomoris: Fixing this point makes many bad things (one is unreasonable SEGV in Travis) and doesn't bring so big happy.
I'll leave this as is.

PR #1393, opened 2016-12-28
in_monitor_agent: enable to select instance variables of plugins

autopp:
I often use `in_monitor_agent` with `/api/plugins.json&debug=1` for more complexity plugin monitoring and inspecting.
In this case, only some instance variables of a specific plugin is needed.
Unfortunary, `&debug=1` makes responce size too large. (Also, pretty printing is not needed...)

To reduce size of response, this PR add query parameter `with_ivars` for `/api/plugins.json`.
`with_ivars` should be a names sepalated by comma.
When `with_ivars` is given, the response has `instance_variables` that contains only specified instance variables for each plugin(This is the same as '&debug=1').

E.g.

```
$ curl 'http//localhost:24220/api/plugins.json?with_ivars=var1,var2'
```

If `with_ivars` is not good name, please give me other idea. (E.g. `with_instance_variables`)


test/plugin/test_in_monitor_agent.rb
```diff
@@ -313,10 +313,10 @@ def get(uri, header = {})
       assert_equal(expected_null_response, null_response)
     end
 
-    data(:with_config_and_retry_yes => [true, true, "?with_config=yes&with_retry"],
-         :with_config_and_retry_no => [false, false, "?with_config=no&with_retry=no"])
-    test "/api/plugins.json with query parameter. query parameter is preferred than include_config" do |(with_config, with_retry, query_param)|
-
+    data(:with_config_and_retry_yes => [true, true, false, "?with_config=yes&with_retry"],
+         :with_config_and_retry_no => [false, false, false, "?with_config=no&with_retry=no"],
+         :with_ivars_given => [false, false, true, "?with_config=no&with_retry=no&with_ivars=id,num_errors"])
```
repeatedly: Please seprate tests.

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -364,6 +368,13 @@ def get_monitor_info(pe, opts={})
           }
         end
         obj['instance_variables'] = iv
+      elsif ivars = opts[:ivars]
+        iv = {}
+        ivars.each {|name|
+          iname = :"@#{name}"
```
repeatedly: "@#{name}" should work.

test/plugin/test_in_monitor_agent.rb
autopp: @repeatedly OK.

lib/fluent/plugin/in_monitor_agent.rb
autopp: @repeatedly I chose symbol literal instead of string because reduce garbage. Is this too careful?

repeatedly: What does 'garbage' mean?
`:"@#{name}"` is dynamic symbol genaration via string, not static literal.

autopp: @repeatedly Sorry, I had misunderstood. I will update patch, so please wait.

autopp:
@repeatedly Thanks for review and merge! But, I need this feature for v0.12.x also.
Can I create a pull request for the `v0.12` branch with the same patch?

repeatedly:
@autopp Yeah. This is good > for v0.12

autopp:
@repeatedly OK, please wait...

autopp:
I created pull request #1402 for port this feature.

Issue #1260, opened 2016-10-05
Fluentd process never exits when blocking for full buffer queue


tagomoris:
I confirmed this problem on Fluentd v0.12.29, but I know that v0.14.x has this problem too.

When we enable `<buffer> overflow_action block </buffer>` or `buffer_queue_full_action block` and the buffer queue is actually full, input threads will block, and never returns if output plugins doesn't go back to normal.

If users will try to stop Fluentd process under this situation, Fluentd don't stop these threads, and the process never stops.

tagomoris:
@repeatedly will you fix this problem in v0.12 versions?

repeatedly:
@tagomoris Okay, I will check it.

tagomoris:
#1396 fixes this problem for v0.14 or later.

zg:
Why is this issue still open if #1396 got merged?

kenhys:
Released as v0.14.12
https://github.com/fluent/fluentd/blob/master/CHANGELOG.md#release-v01412---20170130 v0.14.12

cosmo0920:
Fixed in v0.14.12. And v0.12 is now EOL? (No activities until end of 2019 as described in https://www.fluentd.org/blog/drop-schedule-announcement-in-2019)

kenhys:
v0.12 is not maintained anymore, I'll close it.

PR #1396, opened 2017-01-04
fix to break infinite-loop when shutdown goes on

tagomoris:
Current Fluentd master:HEAD doesn't run into infinite loop with the configuration file added by this change because thread plugin helper kills all threads (including input plugin thread emitting events).

But this change fixes the problem even with plugins creating threads by plugins themselves.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -782,13 +782,19 @@ module Fluent
         begin
           block.call
         rescue Fluent::Plugin::Buffer::BufferOverflowError
-          log.warn "failed to write data into buffer by buffer overflow"
+          log.warn "failed to write data into buffer by buffer overflow", action: @buffer_config.overflow_action
           case @buffer_config.overflow_action
           when :throw_exception
             raise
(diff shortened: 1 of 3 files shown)


tagomoris:
This change relates #1260 only for v0.14.

tagomoris:

lib/fluent/plugin/output.rb
```diff
@@ -782,13 +782,19 @@ def write_guard(&block)
...
             raise
           when :block
             log.debug "buffer.write is now blocking"
             until @buffer.storable?
+              if self.stopped?
+                log.error "breaking block behavior to shutdown Fluentd"
+                # to break infinite loop to exit Fluentd process
+                raise
+              end
+              log.trace "sleepling until buffer can store more data"
```
repeatedly: `sleeping`

repeatedly:
LGTM

lib/fluent/plugin/output.rb
tagomoris: oops

PR #1398, opened 2017-01-04
Fix bug to ignore some command line options

tagomoris:
Currently, supervisor overwrites values from command line options by system config values (even if no parameters are configured in `<system>` sections).
This makes a bug to ignore some command line options, which can be configured via both of command line options and system config section (For example, `--suppress-config-dump`).

This change makes command line options prior than system config parameters.
It's just standard way in all daemon software.


tagomoris:

repeatedly:
LGTM

PR #1401, opened 2017-01-05
close listening socket in #stop not to accept new connections anymore

tagomoris:
This change is revival of #861, to close listening connections in #stop, not to accept connections to tell clients that this node denies incoming data now.

When `shared: true`, SocketManager in ServerEngine accepts connections. So plugin side cannot control it (this change doesn't affect to SM's behavior).
But it's correct, because other workers may be able to handle connections newly coming.

### DIFF
--- a/lib/fluent/plugin_helper/server.rb
+++ b/lib/fluent/plugin_helper/server.rb
@@ -205,28 +205,23 @@ module Fluent
         @_server_mutex = Mutex.new
       end
 
-      def shutdown
-        @_server_connections.each do |conn|
-          conn.close rescue nil
-        end
+      def stop
         @_server_mutex.synchronize do
           @_servers.each do |si|
             si.server.detach if si.server.attached?
+            # to refuse more connections: (connected sockets are still alive here)
+            si.server.close rescue nil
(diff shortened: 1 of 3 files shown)


tagomoris:
@repeatedly can you review this?

tagomoris:
@repeatedly ping?

repeatedly:
With `shared: false`, how handle tcp connections in multi processing mode?

repeatedly:
Ah, wait. Multi processing patch changes its value to `true`.

lib/fluent/plugin_helper/server.rb
```diff
@@ -205,28 +205,23 @@ def initialize
...
       end
 
-      def shutdown
-        @_server_connections.each do |conn|
-          conn.close rescue nil
-        end
+      def stop
         @_server_mutex.synchronize do
           @_servers.each do |si|
             si.server.detach if si.server.attached?
```
repeatedly: `server.close` calls `detach`, so it seems redundant.

repeatedly:
One comment. Others are good for me.

Issue #1038, opened 2016-06-07
Route fluentd internal log events to specified label


tagomoris:
Currently, fluentd's internal logs can be captured by using `<match fluent.*>` sections, but it can't be routed to labels.
I think fluentd should be able to do it, to standardize routing by labels.

tagomoris:
It's better to suppress log forwarding into plugins before `#start` completion, and just before calling `#stop`.
Plugins can process events between `#start` and `#stop`. Otherwise, plugins cannot process events in general.

So we should stop emitting Fluentd's log events out of these state, and also should add an option to emit events forcedly always.

tagomoris:
#485 relates with this.

tagomoris:
#1405 fixes this issue.

PR #1405, opened 2017-01-06
Update log events and its handling

### LINKED ISSUE #485: A counter-intuitive behavior for fluent.(error|warn|info|trace) logs
Currently (at least for v0.10.57), the only way to match Fluentd's own logs is to use the `fluent.**` match pattern.

E.g.

The following configuration outputs fluent.warn events to stdout

```
<match fluent.**>
  type stdout
</match>
```

but the following configuration does NOT

```
<match fluent.warn>
  type stdout
</match>
```

This happens because
1. `fluent.*` events are emitted by `Engine.log_event_loop` which is only turned on if the tag `fluent` is matched by an output (So, `<match fluent.warn>` does NOT turn this on. `<match fluent.**>` and `<match fluent>` does).
2. But individual events have `fluent.<LOG_LEVEL>` tags.

This behavior breaks Fluentd's match semantics.

tagomoris:
There are some issues and unexpected behavior about log events (`fluent.**` events).
This change is to fix these problems and add features.

* log events can be routed only by `<match fluent.**>` sections
  * unexpectedly, `<match fluent.info>` section doesn't work #485 
* log events can be matched only by top-level match sections
  * we are recommending label routing for all events #1038 
* log events sometimes occurs unexpected errors in shutting down
  * log events may be routed into plugins already shutdown

Added features and changes:
* logger emits log events into `@FLUENT_LOG` labels if exists
  * In that sections, mis-matched logs will be ignored without warnings (missed tags are warned at startup)
  * This behavior estimates log levels of standard loggers
* Log events will NOT be emitted before completion of startup (including `#configure` and `#start` of each plugiins), and after entering shutdown sequence (`#stop`, `#shutdown`...)
  * `--log-event-verbose` (or `log_event_verbose` in system config) turns log events on even in startup and shutting down
* Fix logger to prohibit changing tag prefix from "fluent"
  * There is no use-case to change that tag prefix

Fixes #485, #1038

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -17,6 +17,7 @@
 require 'fluent/configurable'
 require 'fluent/plugin'
 require 'fluent/output'
+require 'fluent/match'
 
 module Fluent
   #
@@ -157,36 +158,6 @@ module Fluent
     end
 
     def handle_emits_error(tag, es, error)
-    end
-
-    class NoMatchMatch
-      def initialize(log)
-        @log = log
-        @count = 0
-      end
-
-      def emit_events(tag, es)
-        # TODO use time instead of num of records
-        c = (@count += 1)
-        if c < 512
-          if Math.log(c) / Math.log(2) % 1.0 == 0
-            @log.warn "no patterns matched", tag: tag
-            return
-          end
-        else
-          if c % 512 == 0
-            @log.warn "no patterns matched", tag: tag
-            return
-          end
-        end
-        @log.on_trace { @log.trace "no patterns matched", tag: tag }
-      end
-
-      def start
-      end
-
-      def shutdown
-      end
     end
   end
 end
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -29,13 +29,15 @@ module Fluent
 
     def initialize
       @root_agent = nil
-      @event_router = nil
       @default_loop = nil
       @engine_stopped = false
 
+      @log_event_router = nil
       @log_emit_thread = nil
       @log_event_loop_stop = false
+      @log_event_loop_graceful_stop = false
       @log_event_queue = []
+      @log_event_verbose = false
 
       @suppress_config_dump = false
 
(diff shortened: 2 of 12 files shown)


tagomoris:

lib/fluent/engine.rb
```diff
@@ -111,7 +115,30 @@ def configure(conf)
...
+        log_event_router = @root_agent.event_router
+      end
+
+      if Fluent::Log.event_tags.any?{|t| log_event_router.match?(t) }
+        @log_event_router = log_event_router
+
+        unmatched_tags = Fluent::Log.event_tags.select{|t| !@log_event_router.match?(t) }
+        unless unmatched_tags.empty?
+          $log.warn "match for some tags of log events are not defined (to be ignored)", tags: unmatched_tags
+        end
```
repeatedly: Check tag typo is better. Currently, there is no warning with following conf.

```
<label @FLUENT_LOG>
  <match fleunt.info>
    @type stdout
  </match>
</label>
```

Log example: `There is no 'fluent' prefix <match> in <label @FLUENT_LOG>. Check typo.`

lib/fluent/agent.rb
```diff
@@ -158,35 +159,5 @@ def emit_error_event(tag, time, record, error)
 
     def handle_emits_error(tag, es, error)
     end
-
-    class NoMatchMatch
```
tagomoris: Oops. Why do the such thing...

tagomoris: I checked the code of `out_reemit`. It doesn't refer `Fluent::NoMatchMatch`.
(It just refer `Fluent::EngineClass::NoMatchMatch` in `V10Engine`).
Isn't it?

lib/fluent/engine.rb
tagomoris: I updated the code to warn that `<label @FLUENT_LOG>` doesn't have any matches for `fluent.*` tags.

lib/fluent/log.rb
```diff
@@ -338,13 +345,13 @@ def event(level, args)
...
 
-      unless @threads_exclude_events.include?(Thread.current)
+      if @log_event_enabled && !@threads_exclude_events.include?(Thread.current)
         record = map.dup
         record.keys.each {|key|
           record[key] = record[key].inspect unless record[key].respond_to?(:to_msgpack)
         }
         record['message'] = message.dup
-        @engine.push_log_event("#{@tag}.#{level}", time.to_i, record)
+        @engine.push_log_event("#{LOG_EVENT_TAG_PREFIX}.#{level}", time.to_i, record)
```
tagomoris: Right... I added a commit to emit events with `Fluent::EventTime.from_time(time)`.

lib/fluent/agent.rb
repeatedly: Ah, I see.
I missed the `if` condition for versions.

tagomoris:

PR #1409, opened 2017-01-08
in_tail: add `open_on_every_update` setting / support for UTF-16 and UTF-32

moriyoshi:
## When do we find this useful?

* On Windows, the tail target may well be opened in "exclusive" sharing mode, where no other processes are allowed to open it.  This patch adds an option called `open_on_every_update`, which enables the plugin to tail such a target by letting the plugin open it only when it gets updated.
* In the current line splitter code, lookup for end-of-line characters is always performed byte-wise, which means it cannot deal with the file in an encoding whose minimum encoded unit consist of more than a byte (namely UTF-16(LE|BE), UCS-2(LE|BE) and UTF-32(LE|BE)).  This patch removes such a limitation.

## Known problem(s)

* Access to the file by a process that tries to exclusively open it is effectively prevented by any preceding process that opens it in any mode (https://msdn.microsoft.com/library/windows/desktop/aa363874).  For getting the things to work well, the implementation simply lays down a premise that the application writing the log to the target file would try to open and write to it later again when it fails to open it.

## Why did you end up with a single patch addressing two different issues?

* First I thought of splitting it into two, but the parts those patches would modify overlaps each other and it wouldn't make much sense to me if I had just one of them merged.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -587,8 +587,7 @@ module Fluent::Plugin
           if @from_encoding == @encoding
             s
           else
-            c = Encoding::Converter.new(@from_encoding, @encoding)
-            c.convert(s) + c.finish()
+            s.encode(@encoding, @from_encoding)
           end
         end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -586,7 +586,7 @@ module Fluent::Plugin
           convert(@buffer.slice!(0, idx + 1)) unless idx.nil?
         end
 
-        def size
+        def bytesize
           @buffer.bytesize
         end
       end
@@ -626,7 +626,7 @@ module Fluent::Plugin
 
               unless @lines.empty?
                 if @receive_lines.call(@lines)
-                  @watcher.pe.update_pos(io.pos - @fifo.size)
+                  @watcher.pe.update_pos(io.pos - @fifo.bytesize)
                   @lines.clear
                 else
                   read_more = false
@@ -649,7 +649,7 @@ module Fluent::Plugin
 
         def open
           io = Fluent::FileWrapper.open(@watcher.path)
-          io.seek(@watcher.pe.read_pos + @fifo.size)
+          io.seek(@watcher.pe.read_pos + @fifo.bytesize)
           io
         rescue Errno::ENOENT
           nil
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -566,9 +566,20 @@ module Fluent::Plugin
         attr_reader :from_encoding, :encoding, :buffer
 
         def <<(chunk)
+          # Although "chunk" is most likely transient besides String#force_encoding itself
+          # won't affect the actual content of it, it is also probable that "chunk" is
+          # a reused buffer and changing its encoding causes some problems on the caller side.
+          #
+          # Actually, the caller here is specific and "chunk" comes from IO#partial with
+          # the second argument, which the function always returns as a return value.
+          #
+          # Feeding a string that has its encoding attribute set to any double-byte or
+          # quad-byte encoding to IO#readpartial as the second arguments results in an
+          # assertion failure on Ruby < 2.4.0 for unknown reasons.
           orig_encoding = chunk.encoding
           chunk.force_encoding(from_encoding)
           @buffer << chunk
+          # Thus the encoding needs to be reverted back here
           chunk.force_encoding(orig_encoding)
         end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -73,6 +73,8 @@ module Fluent::Plugin
     config_param :from_encoding, :string, default: nil
     desc 'Add the log path being tailed to records. Specify the field name to be used.'
     config_param :path_key, :string, default: nil
+    desc 'Open and close the file on every update instead of leaving it open until it gets rotated.'
+    config_param :open_on_every_update, :bool, default: false
 
     attr_reader :paths
 
@@ -213,7 +215,7 @@ module Fluent::Plugin
 
     def setup_watcher(path, pe)
       line_buffer_timer_flusher = (@multiline_mode && @multiline_flush_interval) ? TailWatcher::LineBufferTimerFlusher.new(log, @multiline_flush_interval, &method(:flush_buffer)) : nil
-      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher, &method(:receive_lines))
+      tw = TailWatcher.new(path, @rotate_wait, pe, log, @read_from_head, @enable_watch_timer, @read_lines_limit, method(:update_watcher), line_buffer_timer_flusher, @from_encoding, @encoding, open_on_every_update, &method(:receive_lines))
       tw.attach do |watcher|
         watcher.timer_trigger = timer_execute(:in_tail_timer_trigger, 1, &watcher.method(:on_notify)) if watcher.enable_watch_timer
         event_loop_attach(watcher.stat_trigger)
@@ -297,13 +299,6 @@ module Fluent::Plugin
     def flush_buffer(tw)
       if lb = tw.line_buffer
         lb.chomp!
-        if @encoding
-          if @from_encoding
-            lb.encode!(@encoding, @from_encoding)
-          else
-            lb.force_encoding(@encoding)
-          end
-        end
         @parser.parse(lb) { |time, record|
           if time && record
             tag = if @tag_prefix || @tag_suffix
@@ -345,13 +340,6 @@ module Fluent::Plugin
     def convert_line_to_event(line, es, tail_watcher)
       begin
         line.chomp!  # remove \n
-        if @encoding
-          if @from_encoding
-            line.encode!(@encoding, @from_encoding)
-          else
-            line.force_encoding(@encoding)
-          end
-        end
         @parser.parse(line) { |time, record|
           if time && record
             record[@path_key] ||= tail_watcher.path unless @path_key.nil?
@@ -410,7 +398,7 @@ module Fluent::Plugin
     end
 
     class TailWatcher
-      def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, read_lines_limit, update_watcher, line_buffer_timer_flusher, &receive_lines)
+      def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, read_lines_limit, update_watcher, line_buffer_timer_flusher, from_encoding, encoding, open_on_every_update, &receive_lines)
         @path = path
         @rotate_wait = rotate_wait
         @pe = pe || MemoryPositionEntry.new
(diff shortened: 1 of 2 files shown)


moriyoshi:
It seems some builds on travis-ci were stuck.  Could anyone please rebuild them?

tagomoris:
I cannot understand why `open_on_every_update` works. It seems to work well to open files in shared mode via FileWrapper.

Another question: What's happen when someone opens the file tailed by `in_tail` with `open_on_every_update` in exclusive sharing mode? Does it crash?

moriyoshi:
> I cannot understand why open_on_every_update works. It seems to work well to open files in shared mode via FileWrapper.

Generally, if a file is opened by a process with neither of sharing modes (`FILE_SHARE_READ` and `FILE_SHARE_WRITE`) specified, the file cannot be opened by other processes.  Inversely, even if a file is opened by a process with some sharing modes specified, it cannot be opened by another process that tries to open it with no sharing modes specified because the process requires exclusive access to it.

FileWrapper opens the given file R/W shared (`FILE_SHARE_READ` and `FILE_SHARE_WRITE`).  If another process opens the target file for exclusive access that has already been opened by fluentd, the attempt simply fails.  If the file remains opened by fluentd,  the other process will have no chance to open it.  `open_on_every_update` instructs fluentd to open the target file only when it gets updated so that such other process can touch it.  That's how it works.

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
+            s
+          else
+            c = Encoding::Converter.new(@from_encoding, @encoding)
+            c.convert(s) + c.finish()
+          end
+        end
+
+        def next_line
+          idx = @buffer.index(@eol)
+          convert(@buffer.slice!(0, idx + 1)) unless idx.nil?
```
tagomoris: This line should be `@buffer.slice!(0, idx + @eol.bytesize)`

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
         mpe.update(pe.read_inode, pe.read_pos)
         @pe = mpe
-        @io_handler.pe = mpe # Don't re-create IOHandler because IOHandler has an internal buffer.
-
         pe # This pe will be updated in on_rotate after TailWatcher is initialized
       end
 
       class StatWatcher < Coolio::StatWatcher
-        def initialize(path, log, &callback)
+        def initialize(watcher, &callback)
```
tagomoris: This class just refers `watcher.path` and `watcher.log`, and doesn't refer any other attributes of `watcher`.
What is this change for?

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
+            c.convert(s) + c.finish()
+          end
+        end
+
+        def next_line
+          idx = @buffer.index(@eol)
+          convert(@buffer.slice!(0, idx + 1)) unless idx.nil?
+        end
+
+        def size
```
tagomoris: Use `bytesize` for bytesize.

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
-                    read_more = true
-                    break
+          with_io do |io|
+            begin
+              read_more = false
+
+              if !io.nil? && @lines.empty?
+                begin
+                  while true
+                    @fifo << io.readpartial(2048, @iobuf)
```
tagomoris: Are there any reasons to use `@iobuf`?

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
+          super(watcher.path)
         end
 
         def on_change(prev, cur)
           @callback.call
         rescue
           # TODO log?
-          @log.error $!.to_s
-          @log.error_backtrace
+          @watcher.log.error $!.to_s
```
tagomoris: Please add any error messages to make us known where the error occurs, and use `error "message", error: e` style logging instead of `$!`.

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
+        def initialize(from_encoding, encoding)
+          @from_encoding = from_encoding
+          @encoding = encoding
+          @buffer = ''.force_encoding(from_encoding)
+          @eol = "\n".encode(from_encoding).freeze
+        end
+
+        attr_reader :from_encoding, :encoding, :buffer
+
+        def <<(chunk)
```
tagomoris: Is the argument chunk object reused in anywhere? (it's related with my comment about using `@iobuf`).
If no, there seems no reason to re-do `force_encoding(orig_encoding)`.

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,144 @@ def swap_state(pe)
...
+                yield io
+              ensure
+                io.close unless io.nil?
+              end
+            else
+              @io ||= open
+              yield @io
+            end
+          rescue
+            @watcher.log.error $!.to_s
```
tagomoris: Add error message to find where the error occurs (and use `rescue => e` and `error: e` to dump error objects).

lib/fluent/plugin/in_tail.rb
```diff
@@ -632,44 +688,40 @@ def on_notify
...
+              @on_rotate.call(stat)
             end
             @inode = inode
             @fsize = fsize
           end
 
         rescue
-          @log.error $!.to_s
-          @log.error_backtrace
+          @watcher.log.error $!.to_s
```
tagomoris: ditto.

lib/fluent/plugin/in_tail.rb
```diff
@@ -410,7 +398,7 @@ def parse_multilines(lines, tail_watcher)
     end
 
     class TailWatcher
-      def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, read_lines_limit, update_watcher, line_buffer_timer_flusher, &receive_lines)
+      def initialize(path, rotate_wait, pe, log, read_from_head, enable_watch_timer, read_lines_limit, update_watcher, line_buffer_timer_flusher, from_encoding, encoding, open_on_every_update, &receive_lines)
```
repeatedly: I think time to add TailWatcherSetting like object to pass parameters.
Hard to maintain these long arguments.

lib/fluent/plugin/in_tail.rb
moriyoshi: Other helper classes needed a similar change for constructor parameters, and it made sense for me to apply the same policy to `StatWatcher` as well.

lib/fluent/plugin/in_tail.rb
moriyoshi: It's just an adoption for the changes to the instance attributes.  Totally agreed but I'd open the separate issue for this.

lib/fluent/plugin/in_tail.rb
moriyoshi: Absolutely, I hate these long arguments!

lib/fluent/plugin/in_tail.rb
moriyoshi: I'll fix this.

lib/fluent/plugin/in_tail.rb
moriyoshi: No, this is correct.  `@buffer` is searched character-wise.

lib/fluent/plugin/in_tail.rb
moriyoshi: That's exactly what I was wondering about.  I just did so as the original code does.

lib/fluent/plugin/in_tail.rb
tagomoris: Ah, right.

lib/fluent/plugin/in_tail.rb
moriyoshi: Uh, this is the weirdest part.  As `IO.readpartial(len, buf)` returns the same object as the second argument,  the encoding of the reused buffer (`@iobuf`) eventually becomes the same encoding as that specfied to `chunk.force_encoding()`.  Reverting back to the original encoding is necessary because for unknown reasons an assertion failure occurs for older versions of Ruby when a buffer with any double-byte or quad-byte encoding (UTF-16 / UCS-2 / UTF-32) is given to `readpartial()` as the second argument.

Try this:
```
a = open('/dev/urandom')
buf = ''.force_encoding(Encoding::UTF_16BE)
a.readpartial(2048, buf)
```

tagomoris: Okay - we need code comment for that here :)

lib/fluent/plugin/in_tail.rb
tagomoris: I tried this code:
```ruby
"\n".encode("utf-16").size #=> 2
```

moriyoshi: When encoded in Encoding::UTF_16, a BOM is prepended to the resulting string.

```
irb(main):005:0> "\n".encode("utf-16le").size
=> 1
irb(main):006:0> "\n".encode("utf-16").size
=> 2
irb(main):007:0> "\n".encode("utf-16le").size
=> 1
irb(main):008:0> "\n".encode("utf-16")[0]
=> "\uFEFF"
```

tagomoris: Ahh, I've forgotten about BOMs. Sorry.

tagomoris:
LGTM.

lib/fluent/plugin/in_tail.rb
```diff
@@ -498,36 +495,37 @@ def on_rotate(io)
...
+          watcher_needs_update = false
 
-          if io
-            stat = io.stat
+          if stat
             inode = stat.ino
             if inode == @pe.read_inode # truncated
-              @pe.update_pos(stat.size)
-              io_handler = IOHandler.new(io, @pe, @log, @read_lines_limit, &method(:wrap_receive_lines))
+              @pe.update_pos(0)
```
repeatedly: `0` instead of `stat.size` is correct?
What happen when file is truncated to non-zero size?

moriyoshi: The current behavior is somewhat strange.  Because the file position of `io`, which must have been newly opened by the caller, isn't updated while the position is updated to `stat.size` here, IOHandler later starts reading from the head anyway.

The new code tries to update the file handle's position to that stored in `@pe` before reading from the target file, so the position has to be set to 0 in order to preserve the behavior.

If this is a bug, it needs to be corrected to `@pe.update_pos(stat.size)`.

lib/fluent/plugin/in_tail.rb
```diff
@@ -537,86 +535,155 @@ def swap_state(pe)
...
+          @buffer << chunk
+          # Thus the encoding needs to be reverted back here
+          chunk.force_encoding(orig_encoding)
+        end
+
+        def convert(s)
+          if @from_encoding == @encoding
+            s
+          else
+            c = Encoding::Converter.new(@from_encoding, @encoding)
```
moriyoshi: There seems to be no such descriptions in the document that indicates the method resets the internal state of a converter.  It would definitely be better if we could do so.

repeatedly: Currently, `in_tail` sometimes consumes lots of CPU and it causes high pressure on application server.
So I want to avoid more CPU usage with typical cases if possible.

repeatedly: @nurse How about this point?

nurse: `Encoding::Converter#finish` just finishes the converter; it doesn't reset the state. And there's no way to recycle the converter.

For this code, it should be `s.encode(@encoding, @from_encoding)`

lib/fluent/plugin/in_tail.rb
repeatedly: I see.

repeatedly:
LGTM

moriyoshi:
Sorry, I overlooked the notice.  Thank you all for looking into this!

PR #1410, opened 2017-01-10
Fix plugin load order. Loading built-in plugins should be last

repeatedly:
`Registry` now loads built-in plugins first, but it should be last.
Fluentd should prefer 3rd party plugins to overwrite behaviour and keep to use same plugin, e.g. parser plugin.

### DIFF
--- a/lib/fluent/registry.rb
+++ b/lib/fluent/registry.rb
@@ -19,13 +19,14 @@ require 'fluent/config/error'
 module Fluent
   class Registry
     DEFAULT_PLUGIN_PATH = File.expand_path('../plugin', __FILE__)
+    FLUENT_LIB_PATH = File.dirname(File.dirname(DEFAULT_PLUGIN_PATH))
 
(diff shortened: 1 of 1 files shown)


repeatedly:
@tagomoris Check this patch.

repeatedly:
Windows environment tests are toooooo unstable.
I re-run AppVeyor test over 20+ times but tests are still failed...

repeatedly:
@tagomoris ping?

repeatedly:
Which point? fluentd gem is removed from gem search > https://github.com/fluent/fluentd/pull/1410/files#diff-da556f72b0de83b0e659240bf9d879bbR92

tagomoris:
Oh, I missed it.
Looks good to me - Of course, it's better to have some tests.

repeatedly:
yeah, adding test is better but it needs extra gem installation.
It is hard in unit test. This is TODO of fluentd's integration tests.

PR #1411, opened 2017-01-10
Encode utf8 for configuration string values

tagomoris:
The parameters defined as `config_param :name, :string` may produce ASCII-8BIT String values, but it's unexpected values for most plugin authors and users (these values will be packed into `bin` in msgpack).
This change encodes string values in utf8.


tagomoris:

repeatedly:
LGTM

PR #1412, opened 2017-01-10
Rename parameter names with the rule of other params

tagomoris:
This change is to rename two buffer parameters:
* `chunk_records_limit` -> `chunk_limit_records`
* `queue_length_limit` -> `queue_limit_length`

These new names are named under the same rule with other parameters:
* `chunk_limit_size`
* `total_limit_size`

Once (in very early era in v0.14) these parameters were renamed from `chunk_size_limit` and `total_size_limit` with consultation of some English native speakers, but some others were unchanged at that time.
But I think these should be named under the same rule for consistency.

These parameters are still undocumented, so I believe we can do it now, only right now.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -47,10 +47,10 @@ module Fluent
 
       # If user specify this value and (chunk_size * queue_length) is smaller than total_size,
       # then total_size is automatically configured to that value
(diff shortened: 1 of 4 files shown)


tagomoris:
@repeatedly How do you think about this change?

repeatedly:
okay. Go merge.

PR #1413, opened 2017-01-10
Ensure to write str(not bin) objects, for compatibility reason

### LINKED ISSUE #1406: Forwarded tag is violating the specification

- fluentd 0.14.11
- CentOS 7.2
- [configuration](https://github.com/fluent/fluentd-benchmark/blob/master/in_forward/agent.conf)

Fluentd 0.14 seems to encode tags into msgpack's bin format on forwarding.
But [The specification](https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1) says that tags should be str format and 0.12 is encoding tags into str.

I found this problem by the following process and packet capturing(but lost logs……).
https://github.com/okumin/influent-benchmark

tagomoris:
fixes #1406.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -532,16 +532,22 @@ module Fluent::Plugin
         option = { 'size' => chunk.size, 'compressed' => @compress }
         option['chunk'] = Base64.encode64(chunk.unique_id) if @sender.require_ack_response
 
-        # out_forward always uses Raw32 type for content.
-        # Raw16 can store only 64kbytes, and it should be much smaller than buffer chunk size.
+        # https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1#packedforward-mode
+        # out_forward always uses str32 type for entries.
+        # str16 can store only 64kbytes, and it should be much smaller than buffer chunk size.
 
-        sock.write @sender.forward_header        # beginArray(3)
-        sock.write tag.to_msgpack                # 1. writeRaw(tag)
+        tag = tag.dup.force_encoding(Encoding::UTF_8)
+
+        sock.write @sender.forward_header                    # array, size=3
+        sock.write tag.to_msgpack                            # 1. tag: String (str)
         chunk.open(compressed: @compress) do |chunk_io|
-          sock.write [0xdb, chunk_io.size].pack('CN') # 2. beginRaw(size) raw32
-          IO.copy_stream(chunk_io, sock)              # writeRawBody(packed_es)
+          entries = [0xdb, chunk_io.size].pack('CN')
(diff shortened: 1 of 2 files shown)


tagomoris:
@repeatedly could you check this fix?

lib/fluent/plugin/out_forward.rb
```diff
@@ -532,16 +532,22 @@ def send_data_actual(sock, tag, chunk)
...
-        sock.write tag.to_msgpack                # 1. writeRaw(tag)
+        tag = tag.dup.force_encoding(Encoding::UTF_8)
+
+        sock.write @sender.forward_header                    # array, size=3
+        sock.write tag.to_msgpack                            # 1. tag: String (str)
         chunk.open(compressed: @compress) do |chunk_io|
-          sock.write [0xdb, chunk_io.size].pack('CN') # 2. beginRaw(size) raw32
-          IO.copy_stream(chunk_io, sock)              # writeRawBody(packed_es)
+          entries = [0xdb, chunk_io.size].pack('CN')
+          sock.write entries.force_encoding(Encoding::UTF_8) # 2. entries: String (str32)
```
repeatedly: If the entries are compressed, we should not use UTF-8 for such content.

tagomoris: It's true for any `entries` objects, but it's for compatibility.
If we pack it as ASCII-8BIT, Fluentd v0.14 will pack it as `bin` data, and will break v0.12 Fluentd.

repeatedly: We can't use compressed mode with fluentd v0.12 because v0.12's in_forward doesn't support compressed mode.
So it seems no problem for me.

tagomoris: No, this code is for both of compressed and uncompressed transfer mode (by passing `compressed: false` above).

repeatedly: Yes. We should change encoding by `compressed` value.

tagomoris: This code (packing header, not entry itself) doesn't contain compressed/uncompressed content body. It depends on `chunk_io` below.
Do you mean that we should take care about the class of `chunk` and switch `0xdb` to `0xc6`? I think it's nonsense because It increases code complexity and doesn't compatibility problem because we can't throw away `str` types for uncompressed data.

repeatedly: Ah, I missed the code. I thought `entries` means content body.
And I noticed the commde. After drop older msgpack, we will change it to bin for binary data?

tagomoris: Yes. I added a comment for it.
https://github.com/fluent/fluentd/pull/1413/files#diff-40c0e8c88efdc9a297a9fc63bc2bc69fR550

repeatedly: Okay.

PR #1415, opened 2017-01-11
Use BasicObject for cleanroom

okkez:
BasicObject has following methods:

    >> BasicObject.instance_methods
    => [:!, :==, :!=, :__send__, :equal?, :instance_eval, :instance_exec, :__id__]

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -307,13 +307,9 @@ module Fluent::Plugin
         raise "failed to expand `#{str}` : error = #{e}"
       end
 
-      class CleanroomExpander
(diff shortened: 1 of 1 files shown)


okkez:
BasicObject has been added since Ruby 1.9.

repeatedly:
Thanks

okkez:
I'm not sure about that. I've created new issue on bugs.ruby-lang.org.
https://bugs.ruby-lang.org/issues/13204

okkez:
nobu said:
> Top level constants are constants defined in Object, but BasicObject does not inherit Object.

repeatedly:
Hmm... could you revert it?

repeatedly:
ah, wait. easy to revert via web. I will do it.

PR #1417, opened 2017-01-11
Fix a bug not to use plugin root directory for buffer path even if configured correctly

### LINKED ISSUE #1408: file output plugin adds a buffer path even if root_dir is enabled
Current Fluentd file buffer plugin may not have `path` in `<buffer>` section, because of `root_dir`.
But file output plugin adds `path` if missing.
It causes unexpected result when users expect to use root dir for buffer paths.

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_file.rb#L103

tagomoris:
fixes #1408.

### DIFF
--- a/test/plugin/test_out_file.rb
+++ b/test/plugin/test_out_file.rb
@@ -56,6 +56,10 @@ class FileOutputTest < Test::Unit::TestCase
     end
 
     test 'path should be writable' do
+      assert_raise(Fluent::ConfigError.new("'path' parameter is required")) do
+        create_driver ""
+      end
+
       assert_nothing_raised do
         create_driver %[path #{TMP_DIR}/test_path]
       end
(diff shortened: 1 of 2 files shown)


tagomoris:
@repeatedly could you review this bug fix?

test/plugin/test_out_file.rb
```diff
@@ -37,6 +37,24 @@ def create_driver(conf = CONFIG)
       assert_equal :gzip, d.instance.instance_eval{ @compress_method }
     end
 
+    test 'using root_dir for buffer path' do
```
repeatedly: Does test_out_file have tests for `buffer_conf['path'] = conf['path'] || '/tmp/dummy_path'` line?

tagomoris: It's just configuration error for `config_param :path` (in `out_file`), so it's clear that `out_file` raises configuration error for missing path...
But anyway, i added it.

tagomoris:
Can I merge this?

repeatedly:
yes

PR #1420, opened 2017-01-17
formatter: Add add_newline parameter to remove '\n' from the result

repeatedly:
KVS or Queue like plugins don't need `\n` in the json formatted result.
This parameter controlls `\n` for such plugins.


repeatedly:

repeatedly:
@tagomoris Good idea. I will add this parameter to other text formatters.

repeatedly:
Added `add_newline` paramter to ltsv, hash and csv formatters. `single_line` has already had this parameter.

moris says `tsv` formatter is for `out_exec` plugin and current implementation is not good for other cases, e.g. don't espace `\t` / quote field. So I exclude tsv from this patch.
CSV with `\t` delimiter is better for formatter plugin usecase.

@tagomoris Could you re-check this?

tagomoris:
LGTM.

PR #1423, opened 2017-01-18
Support TLS by socket/server plugin helper

tagomoris:
This change is to add a functionality to support TLS, mainly for in/out forward plugins.

The server plugin helper's TLS support is:
* plugins can use `tls_options` hash argument to specify various options about TLS options and certifications in programatic way
* plugins can load configurations about TLS via `<transport tls>` section of configuration transparently to control them by users directly (not to write any code in plugins)

This feature supports these certificate management:
* load a server certificate (signed by public/private CA)
* load a private CA certificate and private key, and generate server certificates dynamically
* generate self-signed server certificate dynamically

With this feature, v0.14 forward input/output plugins are compatible with 3rd party secure-forward plugin.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -43,7 +43,7 @@ module Fluent::Plugin
     # TODO: add linger_timeout, recv_timeout
 
     desc 'The protocol to use for heartbeats (default is the same with "transport").'
-    config_param :heartbeat_type, :enum, list: [:transport, :udp, :none], default: :transport
+    config_param :heartbeat_type, :enum, list: [:transport, :tcp, :udp, :none], default: :transport
     desc 'The interval of the heartbeat packer.'
     config_param :heartbeat_interval, :time, default: 1
     desc 'The wait time before accepting a server fault recovery.'
@@ -153,6 +153,11 @@ module Fluent::Plugin
 
       @read_interval = @read_interval_msec / 1000.0
       @recover_sample_size = @recover_wait / @heartbeat_interval
+
+      if @heartbeat_type == :tcp
+        log.warn "'heartbeat_type tcp' is deprecated. use 'transport' instead."
+        @heartbeat_type = :transport
+      end
 
       if @dns_round_robin
         if @heartbeat_type == :udp
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -33,12 +33,17 @@ module Fluent::Plugin
 
     LISTEN_PORT = 24224
 
+    desc 'The transport protocol.'
+    config_param :transport, :enum, list: [:tcp, :tls], default: :tcp
+    # TODO: TLS session cache/tickets
+    # TODO: Connection keepalive
+
     desc 'The timeout time when sending event logs.'
     config_param :send_timeout, :time, default: 60
     # TODO: add linger_timeout, recv_timeout
 
-    desc 'The transport protocol to use for heartbeats.(udp,tcp,none)'
-    config_param :heartbeat_type, :enum, list: [:tcp, :udp, :none], default: :tcp
+    desc 'The protocol to use for heartbeats (default is the same with "transport").'
+    config_param :heartbeat_type, :enum, list: [:transport, :udp, :none], default: :transport
     desc 'The interval of the heartbeat packer.'
     config_param :heartbeat_interval, :time, default: 1
     desc 'The wait time before accepting a server fault recovery.'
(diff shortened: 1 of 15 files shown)


tagomoris:

repeatedly:
Okay. I will review it later.

lib/fluent/plugin_helper/socket.rb
```diff
@@ -79,12 +86,145 @@ def socket_create_udp(host, port, resolve_name: false, connect: false, **kwargs,
...
+          if allow_self_signed_cert && OpenSSL::X509.const_defined?('V_FLAG_CHECK_SS_SIGNATURE')
+            cert_store.flags = OpenSSL::X509::V_FLAG_CHECK_SS_SIGNATURE
+          end
+          begin
+            if enable_system_cert_store
+              log.trace "loading system default certificate store"
+              cert_store.set_default_paths
+            end
+          rescue OpenSSL::X509::StoreError
+            log.warn "failed to load system default certificate store", error: e
```
repeatedly: Is `warn` ok?
This failure is not critical for actual connection?

lib/fluent/plugin/out_forward.rb
```diff
@@ -76,6 +81,19 @@ class ConnectionClosedError < Error; end
     desc 'Compress buffered data.'
     config_param :compress, :enum, list: [:text, :gzip], default: :text
 
+    desc 'The default version of TLS transport.'
+    config_param :tls_version, :enum, list: Fluent::PluginHelper::Socket::TLS_SUPPORTED_VERSIONS, default: Fluent::PluginHelper::Socket::TLS_DEFAULT_VERSION
```
repeatedly: Using `tls_` prefix instead of `<tls>` is for avoiding confustion with `<transport tls>`?

lib/fluent/plugin/out_forward.rb
```diff
@@ -137,9 +155,29 @@ def configure(conf)
...
-          raise Fluent::ConfigError, "forward output heartbeat type must be 'tcp' or 'none' to use dns_round_robin option"
+          raise Fluent::ConfigError, "forward output heartbeat type must be 'transport' or 'none' to use dns_round_robin option"
+        end
+      end
+
+      if @transport == :tls
+        if @tls_cert_path && !@tls_cert_path.empty?
+          @tls_cert_path.each do |path|
+            raise Fluent::ConfigError, "specified cert path does not exist:#{path}" unless File.exist?(path)
+            raise Fluent::ConfigError, "specified cert path is not readable:#{path}" unless File.exist?(path)
```
tagomoris: Oops. It's just mistake.

lib/fluent/plugin_helper/socket.rb
```diff
@@ -79,12 +86,145 @@ def socket_create_udp(host, port, resolve_name: false, connect: false, **kwargs,
...
+          end
+          if cert_paths
+            if cert_paths.respond_to?(:each)
+              cert_paths.each do |cert_path|
+                log.trace "adding CA cert", path: cert_path
+                cert_store.add_file(cert_path)
+              end
+            else
+              cert_path = cert_paths
```
tagomoris: Right.

lib/fluent/plugin_helper/socket.rb
tagomoris: It depends on many things including env variables to specify default cert store path configured by end users, and not to make connections failure directly (because many other parameters can control it - e.g., self signed certs, manually specified cert path and others).
So, I left it as warned, not critical.

lib/fluent/plugin/out_forward.rb
```diff
@@ -456,6 +514,14 @@ def initialize(sender, server, failure:)
         @available = true
         @state = nil
 
+        # @hostname is used for certificate verification & TLS SNI
+        host_is_hostname = !(IPAddr.new(@host) rescue false)
+        @hostname = case
+                    when host_is_hostname then @host
+                    when @name then @name
```
tagomoris: Yes, I changed the meaning of `name` in `<server>` section.
But in past versions, this `name` is used to indicate the role (or other human readable thing) of this server node in logging.
The "human readable role" is just FQDN server name signed by server certificate, isn't it?

lib/fluent/plugin/out_forward.rb
tagomoris: As you mentioned, these parameters are named in different way from servers' `<transport>` section.
I think that almost users just specify one or (at most) two parameters in these `tls_` prefixed parameters (`tls_cert_path`, and sometimes `tls_allow_self_signed_cert`). Many users don't change the default TLS version nor TLS ciphers.
In such cases, `tls_cert_path ...` is much more simple than `<tls> tls_cert_path ... </tls>`.

repeatedly:
Sorry for the late. LGTM

tagomoris:
Rebased. I'll merge this after CI green.

tagomoris:
Green!

PR #1425, opened 2017-01-19
Clock api for incremental second unit counter

tagomoris:
Currently there are many code to use `Time.now` to control flush timing and event invoke timing.
It's basically wrong, because time may move to future or past by time adjustment, or affected by leap seconds (and Timecop).
We can use `Process.clock_gettime` to avoid this problem, but it's a bit troublesome to specify clock type available. It's too much and too raw for almost core code and plugins. Using `Process.clock_gettime` is also not controlled by test code.

So, this change will introduce a new internal API to provide clock value, which can be controlled to be frozen for tests.

### DIFF
--- a/lib/fluent/clock.rb
+++ b/lib/fluent/clock.rb
@@ -1,0 +1,62 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Fluent
+  module Clock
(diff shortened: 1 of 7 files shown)


tagomoris:

test/test_clock.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert{ Fluent::Clock.now > c0 }
+    end
+  end
+
+  sub_test_case 'using #freeze with Time argument' do
+    test 'Clock.freeze returns the clock value which should be produced when the time is at the specified time' do
+      c0 = Fluent::Clock.now
+      t0 = Time.now
+      t1 = t0 - 30
+      assert_kind_of Time, t1
```
repeatedly: This checks ruby spec, not fluentd code.
So it seems not needed.

lib/fluent/clock.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    @@block_level = 0
+    @@frozen_clock = nil
+
+    def self.now
+      @@frozen_clock || now_raw
+    end
+
+    def self.freeze(dst = nil, &block)
+      return freeze_block(dst, &block) if block_given?
```
repeatedly: `if block` is enough but trivial

repeatedly:
LGTM and add comments.

test/test_clock.rb
tagomoris: That's true, but it assures this test code uses Time argument.
Assertions are sometimes used for such usage.

tagomoris:
Let me merge this right now... Windows CI is failing with known timing bug of test code (and it's already fixed in #1426)

Issue #1376, opened 2016-12-16
Plugin Generator


repeatedly:
For easy plugin development, generating plugin template is useful for users.

## Command

```
$ fluent-plugin-generate type name
```

Based on Embulk: http://www.embulk.org/docs/customization.html#id1

## Generated files

```
.
├── ChangeLog
├── Gemfile
├── README.md
├── Rakefile
├── fluent-plugin-xxx.gemspec
├── lib
│   └── fluent
│       └── plugin
│           └── out_xxx.rb
└── test
    └── test_out_xxx.rb
```

Generated plugin file is based on `type` parameter. If `type` is input, file is `in_xxx.rb`
If same file exist, command prefer existing file.

tagomoris:
I have an idea to add a command to:
* load plugin .rb file
* get the list of configuration parameters (and its type, options and descriptions)
* get the list of plugin helper loaded
* generate a well-formatted Markdown section (for copy-and-paste to README) to describe configuration parameters and links to the page of available plugin helper features (i know it doesn't exist yet)

It's better to add a placeholder in default README.md (generated by template) to be replaced by the generated section above.

okkez:
I will work for this issue if nobody has started development.

repeatedly:
Fixed by #1427

PR #1427, opened 2017-01-20
Add fluent-plugin-generate command

okkez:
See #1376


okkez:
This is workin progress.
I will add one more feature https://github.com/fluent/fluentd/issues/1376#issuecomment-267515326

okkez:
Done my work.
Could you check this PR @tagomoris @repeatedly ?

lib/fluent/command/plugin_config_formatter.rb
```diff
@@ -0,0 +1,228 @@
...
+
+  def call
+    parse_options!
+    init_engine
+    @plugin = Fluent::Plugin.__send__("new_#{@plugin_type}", @plugin_name)
+    dumped_config = {}
+    dumped_config[:plugin_helpers] = @plugin.class.plugin_helpers
+    @plugin.class.ancestors.reverse_each do |plugin_class|
+      next unless plugin_class.respond_to?(:dump)
+      next if plugin_class == Fluent::Plugin::Base
```
tagomoris: Why is `Fluent::Plugin::Base` removed? There seems no actual benefits to do it.

lib/fluent/command/plugin_config_formatter.rb
```diff
@@ -0,0 +1,228 @@
...
+    Fluent::Engine.init(system_config)
+
+    @libs.each do |lib|
+      require lib
+    end
+
+    @plugin_dirs.each do |dir|
+      if Dir.exist?(dir)
+        dir = File.expand_path(dir)
+        Fluent::Engine.add_plugin_dir(dir)
```
tagomoris: `Engine.add_plugin_dir` just calls `Plugin.add_plugin_dir`. So there seems no reason to require/init `Fluent::Engine` only for that.
Are there any other reason to require/initialize Engine?

lib/fluent/command/plugin_config_formatter.rb
```diff
@@ -0,0 +1,228 @@
...
+require "pathname"
+require "fluent/plugin"
+require "fluent/env"
+require "fluent/engine"
+require "fluent/system_config"
+require "fluent/config/element"
+
+class FluentPluginConfigFormatter
+
+  AVAILABLE_FORMATS = [:txt, :markdown, :json]
```
tagomoris: This constant is not used anywhere.
At least, `--format` option value should be checked with this.

lib/fluent/command/plugin_generator.rb
```diff
@@ -0,0 +1,229 @@
...
+#
+
+require "optparse"
+require "pathname"
+require "fileutils"
+require "erb"
+
+class FluentPluginGenerator
+  attr_reader :type, :name
+  attr_reader :license
```
tagomoris: It's ok to fix the license as `Apache-2.0`.
If this feature supports GPLv2 or GPLv3 or any others, it means that Fluentd source code tree have the license files/templates of these licenses, or code to download these license files. It sounds very bad situation.

lib/fluent/command/plugin_generator.rb
```diff
@@ -0,0 +1,229 @@
...
+    src = preamble_file.read
+    ERB.new(src, nil, "-").result(binding).lines.map {|line| "# #{line}".gsub(/ +$/, "") }.join
+  end
+
+  def license_full_name
+    LICENSES_MAP[license]
+  end
+
+  def copy_license
+    # in gem_name directory
```
tagomoris: This method should download the original license from https://www.apache.org/licenses/LICENSE-2.0.txt and use it.
We can't assure the correctness of license text (Please remove the license file from template tree).

lib/fluent/command/plugin_generator.rb
```diff
@@ -0,0 +1,229 @@
...
+    if label == "conflict"
+      return unless overwrite?(dest)
+    end
+    FileUtils.cp(source, dest)
+  end
+
+  def prepare_parser
+    @parser = OptionParser.new
+    @parser.banner = <<BANNER
+Usage: fluent-plugin-generate [options] <type> <name>
```
tagomoris: It's better to show the list of available types.

lib/fluent/command/templates/new_gem/test/plugin/test_filter.rb.erb
```diff
@@ -0,0 +1,16 @@
...
+class <%= class_name %>Test < Test::Unit::TestCase
+  def setup
+    Fluent::Test.setup
+  end
+
+  def create_driver(conf)
+    Fluent::Test::Driver::Filter.new(Fluent::Plugin::<%= class_name %>).configure(conf)
+  end
+
+  def test_failure
```
tagomoris: Use `test 'failure'` type testcases to introduce that feature to plugin authors.

test/helper.rb
```diff
@@ -146,6 +146,15 @@ def ipv6_enabled?
   end
 end
 
+def capture_stdout
```
tagomoris: Move this method to `fluent/test/helpers` because that file already have `#capture_log` method.

tagomoris:
@okkez Nice implementation! I added general review comment and comments for some points.
@repeatedly please add some other points if you have.

lib/fluent/command/plugin_config_formatter.rb
okkez: Because `Fluent::Plugin::Base` has no `config_param` and `config_section`.
`Fluent::Plugin::Base.dump_config_definition` always returns empty hash.
I think it is noise for us.

lib/fluent/command/plugin_config_formatter.rb
okkez: Oh, I forgot to check available format. I will check.

lib/fluent/plugin/base.rb
```diff
@@ -28,6 +28,17 @@ class Base
 
       attr_accessor :under_plugin_development
 
+      def self.plugin_helpers
```
okkez: Sounds nice. I will change implementation.

lib/fluent/command/plugin_config_formatter.rb
tagomoris: How about to omit classes which returns empty hash, instead of using class name?
There are some other classes which doesn't have any `config_param` and `config_section`.

okkez: Sounds good. Thanks!

okkez:
@tagomoris @repeatedly I've pushed changes for all comments.
Could you check again?

repeatedly:
LGTM but all tests failed.
Could you check test?

okkez:
@repeatedly @tagomoris Passed all tests.

repeatedly:
Thanks for your work!

PR #1433, opened 2017-01-25
record_transformer: remove_keys processing should be last

repeatedly:
Currently, `remove_keys` doesn't work with `renew_time_key` parameter, so
user needs one more filter to remove `renew_time_key` field from the record.

```aconf
<filter pattern>
  @type record_transformer
  renew_time_key event_time
  auto_typecast true
  <record>
    event_time ${Convert record["time_str"] into unixtime or float type}
  </record>
</filter>

<filter pattern>
  @type record_transformer
  remove_keys event_time
</filter>
```

This is verbose and not good performance. `remove_keys` should remove `renew_time_key`.

```aconf
<filter pattern>
  @type record_transformer
  renew_time_key event_time
  remove_keys event_time
  auto_typecast true
  <record>
    event_time ${Convert record["time_str"] into unixtime or float type}
  </record>
</filter>
```

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -101,6 +101,8 @@ module Fluent::Plugin
           if @renew_time_key && new_record.has_key?(@renew_time_key)
             time = Fluent::EventTime.from_time(Time.at(new_record[@renew_time_key].to_f))
           end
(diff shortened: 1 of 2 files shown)


repeatedly:
@sonots Could you check this patch?

sonots:
LGTM

PR #1437, opened 2017-01-27
Add config param option validator

tagomoris:
Currently, plugin authors can add any keyword options on `config_param` and `config_argument`, and these are not checked.
This may occur that plugin author want to set any options, but it doesn't work well by typo or any other reasons.

So I added a validator for these options, not to be set unexpected parameter options.

### DIFF
--- a/lib/fluent/config/configure_proxy.rb
+++ b/lib/fluent/config/configure_proxy.rb
@@ -226,19 +226,41 @@ module Fluent
         end
       end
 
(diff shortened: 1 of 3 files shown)


tagomoris:

repeatedly:
LGTM

tagomoris:
Rebased. I'll merge this after CI green.

tagomoris:
Let me merge this, because failing CI on AppVeyor is about `in_tail`.

PR #1438, opened 2017-01-30
fix to generate storage local path using usage

tagomoris:
This feature is very useful for plugins which uses 2 or more storage instances, with `root_dir` configuration.

### DIFF
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -64,7 +64,8 @@ module Fluent
             end
           end
         elsif root_dir = owner.plugin_root_dir
-          @path = File.join(root_dir, 'storage.json')
+          basename = (conf.arg && !conf.arg.empty?) ? "storage.#{conf.arg}.json" : "storage.json"
+          @path = File.join(root_dir, basename)
           @multi_workers_available = true
         else
           if @persistent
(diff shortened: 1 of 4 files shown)


tagomoris:

lib/fluent/plugin/storage_local.rb
```diff
@@ -64,7 +64,8 @@ def configure(conf)
             end
           end
         elsif root_dir = owner.plugin_root_dir
-          @path = File.join(root_dir, 'storage.json')
+          basename = (conf.arg && !conf.arg.empty?) ? "storage.#{conf.arg}.json" : "storage.json"
```
repeatedly: `conf.arg` can contain space like `<storage foo bar>`.
So need to validate or convert unexpected charactors.

tagomoris: Okay, I'll add validation for usage.

repeatedly:
One comment. Others are good.

tagomoris:
Pushed a commit to add validation for usage.

lib/fluent/plugin/storage_local.rb
cosmo0920: This `conf.arg` will be given by :usage in storage_create?
Or users should specify `<storage args>` by themselves?

cosmo0920: Ah, I got it.
We should fill up `usage` when conf is `Fluent::Config::Element`'s instance case: https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin_helper/storage.rb#L56

cosmo0920: This behavior is a limitations for now.

PR #1442, opened 2017-01-31
Improve flush_interval handling

repeatedly:
This patch improves flush_interval parameter handing.

- Set `flush_mode interval` when v0.12 configuration has `flush_interval`

This avoids a trouble for existing out_file and out_s3 configurations

- Log warning message for flush_interval without `flush_mode interval` in `<buffer>`

Currenlty, `flush_interval` is ignored silently with following configuration. It is hard to debug for users.

```
<buffer time>
  timekey 3600
  flush_interval 30s
</buffer>
```

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -310,6 +310,10 @@ module Fluent
             log.warn "'flush_at_shutdown' is false, and buffer plugin '#{buf_type}' is not persistent buffer."
             log.warn "your configuration will lose buffered data at shutdown. please confirm your configuration again."
           end
+
+          if @flush_mode != :interval && buffer_conf.has_key?('flush_interval')
+            log.warn "'flush_interval' is ignored because 'flush_mode' is not 'interval': '#{@flush_mode}'"
+          end
         end
 
         if @secondary_config
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -229,6 +229,7 @@ module Fluent
         end
 
         has_buffer_section = (conf.elements(name: 'buffer').size > 0)
+        has_flush_interval = conf.has_key?('flush_interval')
 
         super
 
@@ -290,7 +291,12 @@ module Fluent
 
           @flush_mode = @buffer_config.flush_mode
           if @flush_mode == :default
-            @flush_mode = (@chunk_key_time ? :lazy : :interval)
+            if has_flush_interval
+              log.info "'flush_interval' is configured at out side of <buffer>. 'flush_mode' is set to 'interval' to keep existing behaviour"
+              @flush_mode = :interval
+            else
+              @flush_mode = (@chunk_key_time ? :lazy : :interval)
+            end
           end
 
           buffer_type = @buffer_config[:@type]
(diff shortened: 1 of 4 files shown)


repeatedly:

lib/fluent/plugin/output.rb
```diff
@@ -310,6 +310,10 @@ def configure(conf)
             log.warn "'flush_at_shutdown' is false, and buffer plugin '#{buf_type}' is not persistent buffer."
             log.warn "your configuration will lose buffered data at shutdown. please confirm your configuration again."
           end
+
+          if @flush_mode != :interval && buffer_conf.has_key?('flush_interval')
```
tagomoris: It looks better to set `:interval` when `flush_mode` is `:default`, even when `timekey` is specified, here:
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L292
We need to know `flush_interval` is set in `<buffer>` section or not before setting `:interval`... how about to add `has_flush_interval` before `super` as same with `has_buffer_section`?

tagomoris: And it's better to warn when both of `flush_interval` and `flush_mode lazy`(or immediate) are configured explicitly by users (these are exclusive parameters).

repeatedly: > It looks better to set :interval when flush_mode is :default

https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L293

From above line, `:default` is changed to `:lazy` or `:interval` so `:default` is not appeared here, right?

>  how about to add has_flush_interval before super as same with has_buffer_section?

It seems good.

> these are exclusive parameters

If so, should we raise an ConfigError when `flush_interval` and `flush_mode lazy` are set explictly?

tagomoris: I mean to set `@flush_mode = :interval` at L293 when `flush_interval` is explicitly specified in configuration even if `@chunk_key_time` is true.

tagomoris: If we doesn't take care about compatibility layer, it's better to raise configuration errors when both of `flush_mode lazy` and `flush_interval` are specified in configuration file.
But compatibility layer may specify these both if a plugin is TimeSlicedOutput and `flush_interval` exists in configuration. So we can't raise errors for such cases. `warn` is better.

tagomoris:
@repeatedly any updates?

lib/fluent/plugin/output.rb
repeatedly: > flush_interval is explicitly specified in configuration even if @chunk_key_time is true.

This means we should use `:interval` with following configuration?

```
<buffer>
  flush_interval 10s # use flush_mode interval
</buffer>
```

And warning for follwing configuration?

```
<buffer>
  flush_interval 10s
  flush_mode lazy
</buffer>
```

tagomoris: 1st example: No. It depends on default of plugins, and `output.rb` should warn about ignored `flush_interval` when plugins' default is not `interval`.

My first point is that we cannot know the `:interval` set in `@flush_mode` is from plugin default or configuraitons at L314 (because it can be set at L293). So we should check `conf` before `super` to show appropriate warning.

2nd example: No. IMO it's better to raise error for inconsistent configuration.

The last my comment was wrong, because `Fluent::Compat::TimeSlicedOutput#configure` will set `flush_mode` as `interval` if `conf` has `flush_interval`. So we can raise exception if both of `flush_mode lazy` and `flush_interval ...` are specified in configurations explicitly.

repeatedly:
Applied reviews.

repeatedly:
@tagomoris could you check again?

tagomoris:
LGTM

PR #1443, opened 2017-01-31
Fix regression of require process

repeatedly:


repeatedly:

lib/fluent/compat/detach_multi_process_mixin.rb
```diff
@@ -0,0 +1,22 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Fluent
+  module Compat
+    module DetachMultiProcessMixin
```
tagomoris: There were `DetachProcessMixin` too. So it's also needed.

repeatedly: You are right. I forgot to add it.

repeatedly:
Add `DetachProcessMixin`

tagomoris:
LGTM.

repeatedly:
Mac tests are canceled by tracis-ci.
Maybe, we should move Mac environment  tests to allow_failures...

PR #1450, opened 2017-02-06
Keep Fluent::BufferQueueLimitError for exsting plugins

repeatedly:
This is for v0.12 based plugins.
Plugin should be moved to BufferOverflowError after v0.14 migration.
But fluentd should keep API compatibility for now.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -614,4 +614,7 @@ module Fluent
       end
     end
   end
+
+  # Some input plugins refer BufferQueueLimitError for throttling
+  BufferQueueLimitError = ::Fluent::Plugin::Buffer::BufferOverflowError
 end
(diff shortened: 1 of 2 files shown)


repeatedly:

tagomoris:
What is this class for? As far as I know, there are no cases that plugin code raise this error.
Are there any actual issue for this?

repeatedly:
Input plugin like in_tail uses this class to stop generate new events.
Such plugins doesn't work with v0.14.

tagomoris:
OK, it does make sense.

lib/fluent/plugin/buffer.rb
```diff
@@ -614,4 +614,7 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
       end
     end
   end
+
+  # Some input plugins refer BufferQueueLimitError for throttling
+  BufferQueueLimitError = ::Fluent::Plugin::Buffer::BufferOverflowError
```
tagomoris: Define this constant should be under `lib/fluent/compat` because it's only for compat plugins.

tagomoris:
I reverted this change at #1454.

PR #1452, opened 2017-02-08
Secondary calculation should consider retry_max_times. fix #1449

### LINKED ISSUE #1449: secondary mechanism broken
- fluentd `0.14.11`
- Ubuntu 14.04 LTS on EC2

In the course of attempting to debug this plugin issue: https://github.com/uken/fluent-plugin-elasticsearch/issues/235 I've been unable to get fluentd to fail over to the secondary *using only the built-in output plugins*.

The configuration I have had working under 0.12.x and that fails under 0.14.11 is below. This fluentd sits on the edge of a logging service and receives logs from host clients:

```
<source>
  @type tail
  path /var/log/syslog
  pos_file /var/spool/td-agent/syslog.pos
  tag locallogs
  format /^(?<time>[^\s]+) (?<facility>\w+)\.(?<priority>\w+) (?<hostname>[^\s]+) (?<process>[^:]+): *(?<message>.*)$/

  # Support ms time resolution
  time_format %Y-%m-%dT%H:%M:%S.%L%z
  keep_time_key true
</source>

# This source receives logs from remote client fluentd installations:
# the primary manner in which logs are ingested into the service
<source>
  @type secure_forward
  self_hostname xxx.xxx.com
  shared_key xxx

  secure true
  cert_path /etc/td-agent/ssl_cert.pem
  private_key_path /etc/td-agent/ssl_key.pem
  private_key_passphrase

  # Support ms time resolution
  time_as_integer false
</source>

# This source is present so that it can be used to
# ingest the files from /var/spool/td-agent/overflow
# http://docs.fluentd.org/articles/in_forward
<source>
  @type forward
  port 24224
  bind 127.0.0.1
</source>

# Filters
<filter locallogs>
  @type record_transformer
  <record>
    statsd_type increment
    statsd_key syslog.${record["facility"]}.${record["priority"]}
  </record>
  <record>
    ident_product xxx
    ident_environment xxx
    ident_role fluentd-server
    ident_hostname xxx
    host_ident i-1234567890
  </record>
</filter>

# Matched outputs
<match **>
  type copy

  <store>
    @type statsd
    host localhost
    port 8125
    flush_interval 1s
  </store>

  <store>
    @type "aws-elasticsearch-service"
    logstash_format true
    include_tag_key true
    tag_key "@log_name"

    # Implement this fix:
    # https://discuss.elastic.co/t/elasitcsearch-ruby-raises-cannot-get-new-connection-from-pool-error/36252/9
    reload_connections false

    # We'd like to gain the .raw fields mapping, like logstash:
    # http://stackoverflow.com/questions/33979226/where-do-raw-fields-come-from-when-using-logstash-with-elasticsearch-output
    #
    # This is so that we can do visualizations on non-analyzed fields.
    #
    # Further reading:
    #
    # https://github.com/uken/fluent-plugin-elasticsearch/pull/194
    # https://github.com/uken/fluent-plugin-elasticsearch/issues/33
    template_name "logstash"
    template_file "/etc/td-agent/elasticsearch-template-es2x.json"

    # Implement this fix:
    # https://github.com/fluent/fluentd/issues/607
    request_timeout 10

    # Support ms time resolution
    time_key 'time'
    time_as_integer false

    # AWS Elasticsearch has a 10M maximum request size.
    # Fluentd serializes messages sent to the Elasticsearch
    # output plugin with msgpack.
    #
    # A small subset of benchmarked messages were shown to
    # compress to between 85 and 66% of their original size.
    #
    # Therefore the buffer_chunk_limit below assumes that
    # msgpack will never be more than 50% efficient.
    buffer_type "file"
    buffer_path "/var/spool/td-agent/"
    buffer_chunk_limit 5m    # AWS ES Service max request size is 10m
    buffer_queue_limit 17800 # 5m chunk with a 90G spool volume, plus headroom
    flush_at_shutdown true
    flush_interval 5s
    retry_limit 2            # Write to secondary quickly

    <endpoint>
      url https://xxx.eu-west-1.es.amazonaws.com
      region eu-west-1
    </endpoint>

    <secondary>
      @type secondary_file
      directory /var/spool/td-agent/overflow/
      basename failed.log
    </secondary>

  </store>
</match>
```

Under 0.12.x all `time_*` options are not present (the primary reason for the upgrade is to gain ms time resolution). Furthermore the `<secondary>` under 0.12.x was plainly `@type file` (with associated configuration), however fluentd refuses to start in this configuration for us unless we amend this to `secondary_file` as shown.

In this configuration a failure of the elasticsearch endpoint, which previously caused logs to spool to the secondary, drops chunks on the floor once the various retry thresholds have been breached.

In order to eliminate 0.12.x plugins as being the problem I amended the `<match **>` configuration to be:

```
<match **>
  @type "forward"
  send_timeout 60s
  recover_wait 10s
  heartbeat_interval 1s
  phi_threshold 16
  hard_timeout 60s

    <server>
      name localhost
      host localhost
      port 1234
      weight 60
    </server>

    <secondary>
      @type secondary_file
      directory /var/spool/td-agent/overflow/
      basename failed.log
    </secondary>

</match>
```

I then ran netcat, correctly bound (`nc -lk 1234`) and observed msgpack output on netcat's stdout. Killing netcat started the retry procedure (as expected). Once that is exhausted I would have expected a failover to the secondary (as happens under 0.12.x). *This does not happen*

My only conclusion is that under 0.14.11 the secondary mechanism does not work, given that all plugins involved in this test are 0.14.x native and shipped with the gem.

Furthermore, I see that the documentation on the secondary mechanism has been removed from the 0.14.x documentation: http://docs.fluentd.org/v0.14/articles/buffer-plugin-overview

Without support for ms timestamp resolution fluentd is basically useless to us, which is a great shame given the investment we've already given it.

repeatedly:
Current secondary transition calculation checks only `retry_timeout` parameter, but
it should consider `retry_max_times`.

### DIFF
--- a/lib/fluent/plugin_helper/retry_state.rb
+++ b/lib/fluent/plugin_helper/retry_state.rb
@@ -64,7 +64,12 @@ module Fluent
           @secondary_threshold = secondary_threshold
           if @secondary
             raise "BUG: secondary_transition_threshold MUST be between 0 and 1" if @secondary_threshold <= 0 || @secondary_threshold >= 1
-            @secondary_transition_at = @start + timeout * @secondary_threshold
+            max_retry_timeout = timeout
+            if max_steps
+              timeout_by_max_steps = calc_max_retry_timeout(max_steps)
+              max_retry_timeout = timeout_by_max_steps if timeout_by_max_steps < max_retry_timeout
+            end
+            @secondary_transition_at = @start + max_retry_timeout * @secondary_threshold
             @secondary_transition_steps = nil
           end
         end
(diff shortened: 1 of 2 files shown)


repeatedly:

repeatedly:
@tagomoris ping?

PR #1453, opened 2017-02-09
parser_multiline.rb must require parser_regexp

richm:
Because parser_multiline uses Fluent::Plugin::RegexpParser it must
require parser_regexp, otherwise, you get error messages like this
when using in_tail with format multiline:
```
2017-02-09 02:02:25 +0000 [error]: #0 fluent/supervisor.rb:674:block in main_process: config error file="/etc/fluent/fluent.conf" error_class=Fluent::ConfigError error="Invalid regexp '^(?<time>\\S+\\s+\\S+\\s+\\S+)\\s+(?<host>\\S+)\\s+(?<ident>[\\w\\/\\.\\-]*)(?:\\[(?<pid>[0-9]+)\\])?[^\\:]*\\:\\s*(?<message>.*)$': uninitialized constant Fluent::Plugin::RegexpParser\nDid you mean?  RegexpError"
```

### DIFF
--- a/lib/fluent/plugin/parser_multiline.rb
+++ b/lib/fluent/plugin/parser_multiline.rb
@@ -15,6 +15,7 @@
 #
 
 require 'fluent/plugin/parser'
+require 'fluent/plugin/parser_regexp'
 
 module Fluent
   module Plugin


repeatedly:
Oops. Good catch!

PR #1456, opened 2017-02-09
Keep Fluent::BufferQueueLimitError for exsting plugins

repeatedly:
#1450 again by using compat module


repeatedly:
@tagomoris Could you check again?

PR #1468, opened 2017-02-15
buf_file: handle "Too many open files" error to keep buffer and metadata pair

repeatedly:
This is not incomplete but keep buffer and metadata pair phase as much as possible.

Use `BufferOverflowError` in `create_new_chunk` for input plugin retry.

TODO: Writing unittest for this case is hard. It will be added to integration test.

### DIFF
--- a/lib/fluent/plugin/buffer/file_chunk.rb
+++ b/lib/fluent/plugin/buffer/file_chunk.rb
@@ -106,10 +106,36 @@ module Fluent
 
           write_metadata(update: false) # re-write metadata w/ finalized records
 
-          file_rename(@chunk, @path, new_chunk_path, ->(new_io){ @chunk = new_io })
-          @path = new_chunk_path
+          begin
+            file_rename(@chunk, @path, new_chunk_path, ->(new_io) { @chunk = new_io })
+          rescue => e
+            begin
+              file_rename(@chunk, new_chunk_path, @path, ->(new_io) { @chunk = new_io }) if File.exist?(new_chunk_path)
+            rescue => re
+            end
+            if re
+              raise "can't enqueue buffer file. This may causes inconsistent state: path = #{@path}, error = '#{e}', retry error = '#{re}'"
+            else
+              raise "can't enqueue buffer file: path = #{@path}, error = '#{e}'"
+            end
+          end
 
-          file_rename(@meta, @meta_path, new_meta_path, ->(new_io){ @meta = new_io })
+          begin
+            file_rename(@meta, @meta_path, new_meta_path, ->(new_io) { @meta = new_io })
+          rescue => e
+            begin
+              file_rename(@chunk, new_chunk_path, @path, ->(new_io) { @chunk = new_io }) if File.exist?(new_chunk_path)
+              file_rename(@meta, new_meta_path, @meta_path, ->(new_io) { @meta = new_io }) if File.exist?(new_meta_path)
+            rescue => re
+            end
+            if re
+              raise "can't enqueue buffer metadata. This may causes inconsistent state: path = #{@meta_path}, error = '#{e}', retry error = '#{re}'"
+            else
+              raise "can't enqueue buffer metadata: path = #{@meta_path}, error = '#{e}'"
+            end
+          end
+
+          @path = new_chunk_path
           @meta_path = new_meta_path
 
           super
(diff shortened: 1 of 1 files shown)


repeatedly:

tagomoris:
@repeatedly Do you have any issues that this pull-request is about? Add it in description if exists.

repeatedly:
> why rescue => re can be ignored,

Because it can't be restored. We can add retry mechanizm for it but it may cause livelock under too many open and high load environment .
This patch adds "This may causes inconsistent state" message instead.
I will add comments.

repeatedly:
Added comment.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -106,10 +106,41 @@ def enqueued!
...
+          rescue => e
+            begin
+              file_rename(@chunk, new_chunk_path, @path, ->(new_io) { @chunk = new_io }) if File.exist?(new_chunk_path)
+            rescue => re
+              # In this point, restore buffer state is hard because previous `file_rename` failed by resource problem.
+              # Retry is one possible approach but it may cause livelock under limited resources or high load environment.
+              # So we ignore such errors for now and log better message instead.
+              # "Too many open files" should be fixed by proper buffer configuration and system setting.
+            end
+            if re
```
tagomoris: It's better to use else clause to simplify code.

```ruby
begin
  file_rename
rescue => re
  raise "" # using re
else
  raise "" # using e
end
```

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -106,10 +106,41 @@ def enqueued!
...
+          begin
+            file_rename(@meta, @meta_path, new_meta_path, ->(new_io) { @meta = new_io })
+          rescue => e
+            begin
+              file_rename(@chunk, new_chunk_path, @path, ->(new_io) { @chunk = new_io }) if File.exist?(new_chunk_path)
+              file_rename(@meta, new_meta_path, @meta_path, ->(new_io) { @meta = new_io }) if File.exist?(new_meta_path)
+            rescue => re
+              # See above
+            end
+            if re
```
tagomoris: ditto

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -242,14 +273,24 @@ def file_rename(file, old_path, new_path, callback=nil)
...
-          @meta.set_encoding(Encoding::ASCII_8BIT)
-          @meta.sync = true
-          @meta.binmode
+          begin
+            @chunk = File.open(@path, 'wb+', perm)
+            @chunk.set_encoding(Encoding::ASCII_8BIT)
+            @chunk.sync = true
+            @chunk.binmode
+          rescue => e
+            raise BufferOverflowError, "can't create buffer file for #{path}. Stop creating buffer files: error = #{e}"
```
tagomoris: Does this statement expect "too many open files" or exceptions like that?
Please add a comment about expected exception types - it seems not the direct "buffer overflow" error.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -242,14 +273,24 @@ def file_rename(file, old_path, new_path, callback=nil)
...
+          end
+          begin
+            @meta = File.open(@meta_path, 'wb', perm)
+            @meta.set_encoding(Encoding::ASCII_8BIT)
+            @meta.sync = true
+            @meta.binmode
+          rescue => e
+            @chunk.close
```
tagomoris: this requires `rescue nil`.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -242,14 +273,24 @@ def file_rename(file, old_path, new_path, callback=nil)
...
+            File.unlink(@path)
```
tagomoris: ditto.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -242,14 +273,24 @@ def file_rename(file, old_path, new_path, callback=nil)
...
+            raise BufferOverflowError, "can't create buffer metadata for #{path}. Stop creating buffer files: error = #{e}"
```
tagomoris: Same comment required above.

PR #1469, opened 2017-02-16
Skip record when Output#format returns nil

repeatedly:
Current Output API can't skip invalid record with custom format.
Old API uses `emit` for such purpose but new API doesn't have it.
Omitting invalid record in format is good for reducing formatting cost.


repeatedly:

Issue #1466, opened 2017-02-15
in_tail: Watch only recently modified files


t-osakada:
Would you consider adding an option like `limit_recent_modified :time` ?

The purpose of this option is to solve the following:

- I want to skip logs of expiration when Fluentd stops for a long time due to failure or maintenance.

- In addition, suppress unnecessary watcher creating to save resources.

ex.
```diff
     def expand_paths
       date = Time.now
       paths = []

       excluded = @exclude_path.map { |path| path = date.strftime(path); path.include?('*') ? Dir.glob(path) : pat\
h }.flatten.uniq
       @paths.each { |path|
         path = date.strftime(path)
         if path.include?('*')
           paths += Dir.glob(path).select { |p|
             if File.readable?(p) && !File.directory?(p)
-              true
+              if @limit_recent_modified and File.mtime(p) < (date - @limit_recent_modified)
+                false
+              else
+                true
+              end
             else
```

repeatedly:
It seems good idea.

PR #1474, opened 2017-02-17
Watch only recently modified files

t-osakada:
Add option `limit_recently_modified` to in_tail. 

See: https://github.com/fluent/fluentd/issues/1466


repeatedly:
Thanks!

PR #1477, opened 2017-02-20
Use `String#force_encoding` instead of `String#encode`

okkez:
Because:

  * `String#encode` will try to encode self, it will raise
    `Encoding::UndefinedConversionError`.
  * fluent.conf will read as ASCII-8BIT

See also https://github.com/fluent/fluent-plugin-grok-parser/issues/27


repeatedly:
LGTM

repeatedly:
Merged

PR #1478, opened 2017-02-21
Add `Fluent::Log#<<`

okkez:
This is useful for AWS related plugins to set Aws::XXX::Client's logger.
AWS logger requires `#<<` method.

See https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html#initialize-instance_method

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -357,6 +357,7 @@ module Fluent
     def write(data)
       @out.write(data)
     end
+    alias << write
 
     def flush
       @out.flush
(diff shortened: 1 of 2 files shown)


test/test_log.rb
```diff
@@ -720,6 +720,10 @@ def test_write
       @log.write("log")
     end
 
+    def test_write_alias
+      assert { @log.respond_to?(:<<) }
```
repeatedly: Is `{}` needed? It seems `assert @log.respond_to?(:<<)` is enough.

lib/fluent/log.rb
```diff
@@ -357,6 +357,7 @@ def puts(msg)
     def write(data)
       @out.write(data)
     end
+    alias << write
```
repeatedly: Please add comment why this alias is needed.

PR #1479, opened 2017-02-23
Allow escape sequence in Apache access log

sergeyfedotov:
http://httpd.apache.org/docs/current/mod/mod_log_config.html#format-notes


repeatedly:
Hmm... 2.4 on Mac OS X test failed.
Maybe, your patch is not related. Please wait.

repeatedly:
Thanks!

PR #1483, opened 2017-03-01
Fix cleanup resource

repeatedly:
I noticed fluentd doesn't remove SocketManager's socket file.
It remains lots of garbage files under `/tmp`.

I added `cleanup_resources` function to handle such case.


repeatedly:
@tagomoris How about this change?

tagomoris:
LGTM.

Issue #1465, opened 2017-02-15
in_tail: Reduce startup time


t-osakada:
Currently, in_tail startup process take very long time when reading huge files.
Would you consider adding an option like `skip_refresh_on_startup :bool` to reduce startup time?

ex.
```diff
     def start
       super

       if @pos_file
         @pf_file = File.open(@pos_file, File::RDWR|File::CREAT|File::BINARY, @file_perm)
         @pf_file.sync = true
         @pf = PositionFile.parse(@pf_file)
       end

-      refresh_watchers
+      refresh_watchers unless @skip_refresh_on_startup
       timer_execute(:in_tail_refresh_watchers, @refresh_interval, &method(:refresh_watchers))
     end
```

repeatedly:
It seems no problem for me.

t-osakada:
Thanks!

PR #1487, opened 2017-03-02
Skip the refresh of watching list on startup.

t-osakada:
Add option `skip_refresh_on_startup` to in_tail.

See: https://github.com/fluent/fluentd/issues/1465


repeatedly:
Thanks!

PR #1490, opened 2017-03-03
in_forward: backport source / address key features

repeatedly:
- Fix source_hostname_key feature to get hostname, not address.
- Add source_address_key feature


repeatedly:
This patch follows v0.14's `in_forward` behaviour.

tagomoris:
LGTM.

Issue #1488, opened 2017-03-02
Add the 'newsyslog' gem to Azure OMS Agent (fluentd) crashes the agent


SamirFarhat:
Hi All,
I have installed the OMS Linux Agent in order to send syslog messages to OMS.
I'm using a version which is based on ruby 2.3 (No yet published)
The aim is to use the 'newsyslog' input type which solves some limitations encountered with the 'syslog' input type
I have installed the newsyslog gem using this command : 
./fluent-gem install fluent-plugin-newsyslog
The installation was successful

The following is my omsagent.conf file (which uses the same fluentd syntax)
<source>
  type newsyslog
 # format none
  port 20514
  bind 0.0.0.0
  protocol_type udp
  tag oms.networkdevices
</source>

When restarting the OMS Agent, i encountered the following error : ( service omsagent status -l)
![snag-0041](https://cloud.githubusercontent.com/assets/14938496/23511135/7b83f062-ff5c-11e6-9e14-6bd857308541.png).

I also noticed that the in_newsyslog.rb is not located under "/opt/microsoft/omsagent/ruby/lib/ruby/gems/2.3.0/gems/fluentd-0.12.24/lib/fluent/plugin/" where all the rb files were supposed to be located.
Fluentd version : fluentd-0.12.24
OS : Centos 7.3

What do you think that is missing ?
Thanks

SamirFarhat:
I think that the best to help me is @repeatedly (if he can)

repeatedly:
Maybe, this is the cause: https://github.com/athenahealth/fluent-plugin-newsyslog/pull/3

repeatedly:
I checked newsyslog plugin code and the main difference is supporting rfc5424 format.
in_syslog with newsyslog rfc5424 regexp format seems to realize same thing.

SamirFarhat:
Thanks repeatedly. So what m'i supposed to do please. Use the time_format with the appropriate regexp ?
Can i use multiple time formats and use the 'OR' (|) statement on the regexp ? Do you think it's supported.
If this will cover my needs, i will wait for the final OMS agent with the newsyslog format and the right documentation.

repeatedly:
> Can i use multiple time formats and use the 'OR' (|) statement on the regexp ? Do you think it's supported.

This is hard for now. BTW, if your case is in_syslog receives rfc3614 and rfc5424 formats at the same time, supporting it in `in_syslog` is good idea.

SamirFarhat:
[...]

Issue #636, opened 2015-07-24
syslog input has no support for RFC 5424


huntc:
I note that there is no direct support for [RFC 5424](https://tools.ietf.org/html/rfc5424) by fluentd, only [RFC 3164](https://tools.ietf.org/html/rfc3164), the latter being inconsistent across Unix variants. A summary of the differences between the two: [https://tools.ietf.org/html/rfc5424#page-34](https://tools.ietf.org/html/rfc5424#page-34)

Given the popularity of Ubuntu and RHEL, and the RFC 5424 compliant rsyslog, I recommend supporting it. Perhaps a new `format` type of `syslog-5424`...

A work-around for RFC 5424 version 1 style producers of syslog:

```
<source>
  type syslog
  format syslog
  time_format 1 %Y-%m-%dT%H:%M:%S.%L%z
  port 5140
  protocol_type tcp
  tag system
</source>
```

Related:
- https://github.com/fluent/fluentd/issues/363
- http://stackoverflow.com/questions/24398756/ietf-syslog-input-to-fluentd
- https://groups.google.com/d/msg/fluentd/k0HU1Dkbazs/qVRipdLsIpwJ

repeatedly:
Could you show me the summery of differences between 5424 nad 3164 or pointer to the article?

huntc:
@repeatedly Description has been updated to refer to the summary of differences. HTH.

repeatedly:
Thanks! I will check it later.

huntc:
If I were to create a PR with changes in support of RFC 5424 and RFC 5425/6587, would there be interest? Also what is the release cadence typically i.e. when would we expect to see a new version including the changes given a PR to your liking?

repeatedly:
I don't have a enough time to create RFC5424 parser.
So if you can create a PR, it would be awesome.

huntc:
OK, I'll give it a go. I don't have time either - but I think that the world needs it - I sort of do right now! ;-)

repeatedly:
Thanks!

huntc:
@repeatedly  https://github.com/fluent/fluentd/pull/640 is a start on supporting modern syslog. Please review that and let me know if it is to your liking (it is a WIP, but quite complete as a PR).

lwh:
[...]

PR #1492, opened 2017-03-08
Support rfc5424 syslog format

cosmo0920:
This PR is related to #636 and #1488.


lib/fluent/plugin/parser_syslog.rb
```diff
@@ -39,15 +43,44 @@ def initialize
...
+                    nil
+                  end
         @time_parser = time_parser_create
       end
 
       def patterns
         {'format' => @regexp, 'time_format' => @time_format}
       end
 
-      def parse(text)
```
repeatedly: Keep `parse(text)` for method check safety.

```
def parse(text)
  # This is overwritten by configure
end
```

repeatedly:
Thanks for great patch!
Could you send same patch to v0.12 branch?

cosmo0920:
Sure, I'll do it! :smile:

PR #1493, opened 2017-03-08
in_monitor_agent: Start one HTTP server per worker on sequential port numbers

mururu:
MonitorAgentInput doesn't work on multi worker environment because workers
try to start servers with the same port.

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -226,6 +226,11 @@ module Fluent::Plugin
       @first_warn = false
     end
 
+    def configure(conf)
+      super
+      @port += fluentd_worker_id
+    end
+
     def multi_workers_ready?
       true
     end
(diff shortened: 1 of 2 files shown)


repeatedly:
Good catch.
BTW, how about adding `launch_server` like option to enable/disable server launching?
It allows users can use `in_monitor_agent` to emit buffer parameters in multi-process environment without this error.

mururu:
@repeatedly Sounds good, but there is an another idea (@tagomotis sugestted) that each worker launch a server on different port(port + worker_id) on multi worker environment.

```
<system>
  worker 4
<system>
<source>
  @type monitor_agent
  port 24220
</source>
```

Using this conf, workers launche servers with port number 24220, 24221, 24222 and 24223. 

I think that it is better for us to be able to use same feature on single worker environment and multi worker environment if possible. 
How about that?

repeatedly:
It is also good.

mururu:
Thanks, I'll go with it.

mururu:
Updated.

repeatedly:
Thanks!

Issue #1480, opened 2017-02-24
fluentd 0.14.11 & 0.14.13 parser plugin 'suppress_parse_error_log' not used


fmax:
fluentd 0.14.11 & 0.14.13
parser plugin 'suppress_parse_error_log' not used.

filter parser since parser is built in  core - no parser plugin needed any more
[warn]: parameter 'suppress_parse_error_log' in <filter project.nginx*> is not used.
so this results in lot's of unwanted log messages, because there are two filters for ngnix logs:
* one for parsing access log nginx regex
* one for parsing error log nginx regex

so when access log filter matches, error log filter writes to td-agent.log:
'error_class=Fluent::Plugin::Parser::ParserError error="pattern not match'
same for error log filter matches, access log filter filling up td-agent.log
but different filters are needed to get different prefix attribute injected for elasticsearch output.

related config part:
<filter project.nginx*>
  @type parser
  key_name log
  suppress_parse_error_log true
  format /^nginx-access-log-regex$/
  inject_key_prefix nginx.acc.
</filter>
<filter project.nginx*>
  @type parser
  key_name log
  suppress_parse_error_log true
  format /^nginx-error-log-regex$/
  inject_key_prefix nginx.err.
</filter>

OS: ubuntu 16 LTS docker container
package: td-agent                     3.0.0-0
gems - manually updated gem 'fluentd' to 0.14.13 :
/opt/td-agent/embedded/bin/gem list --local | egrep "action|active|rail|fluent"
actionmailer (4.2.8)
actionpack (4.2.8)
actionview (4.2.8)
activejob (4.2.8)
activemodel (4.2.8)
activerecord (4.2.8)
activesupport (4.2.8)
fluent-logger (0.6.2)
fluent-mixin-config-placeholders (0.4.0)
fluent-mixin-plaintextformatter (0.2.6)
fluent-plugin-concat (2.0.1)
fluent-plugin-elasticsearch (1.9.2)
fluent-plugin-kafka (0.5.0)
fluent-plugin-rewrite (0.1.0)
fluent-plugin-rewrite-tag-filter (1.5.5)
fluent-plugin-s3 (1.0.0.rc1)
fluent-plugin-td (0.10.29)
fluent-plugin-td-monitoring (0.2.2)
fluent-plugin-webhdfs (0.4.2)
fluentd (0.14.13)
fluentd-ui (0.4.3)
font-awesome-rails (4.7.0.1)
haml-rails (0.5.3)
[...]

PR #1494, opened 2017-03-08
filter_parser: Add emit_invalid_record_to_error parameter

repeatedly:
Related to https://github.com/fluent/fluentd/issues/1480
This is acceptable for me because configuration becomes error prone without this parameter for ignoring unneeded log lines case.

```
<label @ERROR>
  <match {tag1,tag2,...}>
    @type null
  </match>
</label>
```

`@ERROR` label is for capturing unexpected error event so blackhoke usecase is misuse.

Parameter name comes from v0.12's `filter_parser`.


repeatedly:

repeatedly:
@tagomoris ping?

repeatedly:
Use `raise` when no emit case.
With `reserve_data true`, filter_parser passes incoming data to next filter so can't use `raise`.

tagomoris:
LGTM.

PR #1495, opened 2017-03-09
Backport #1492 to v0.12

cosmo0920:


lib/fluent/parser.rb
```diff
@@ -549,6 +553,26 @@ def configure(conf)
         super
 
         @regexp = @with_priority ? REGEXP_WITH_PRI : REGEXP
```
repeatedly: Can we remove this line?

lib/fluent/parser.rb
```diff
@@ -557,6 +580,23 @@ def patterns
       end
 
       def parse(text)
+        # This is overwritten in configure
+      end
+
+      def parse_auto(text, &block)
+        if @message_format == :auto
```
repeatedly: We can remove this condition because `parse_auto` is not called directly.

repeatedly:
Commented. I will merged this patch after applied reviews.

lib/fluent/parser.rb
cosmo0920: Oh! I've overlooked this. :confounded:

lib/fluent/parser.rb
```diff
@@ -548,7 +552,26 @@ def initialize
...
+                  when :rfc3164
+                    class << self
+                      alias_method :parse, :parse_plain
+                    end
+                    @with_priority ? REGEXP_WITH_PRI : REGEXP
+                  when :rfc5424
+                    class << self
+                      alias_method :parse, :parse_plain
+                    end
+                    REGEXP_RFC5424
```
repeatedly: Assigninig `@time_format = @rfc5424_time_format` is needed?

repeatedly:
I re-checked the code and I have one question. Commented it.

lib/fluent/parser.rb
cosmo0920: Ah! Your suggestion means that we should set `@time_format` in the following situation?

```ruby
@time_format = @rfc5424_time_format unless @time_format
```

It is reasonable for me.

repeatedly: Yes, but time_format has default value.
So there is two approaches:

- Change time_format's default to `nil` and set each default time_format in `message_format` condition
- Use conf.has_key?('time_format')

cosmo0920: I want to choose `conf.has_key?('time_format')` approach.

repeatedly:
Thanks!
Could you add test for only `message_format rfc5424` configuration?

cosmo0920:
> Could you add test for only message_format rfc5424 configuration?

I've added in https://github.com/fluent/fluentd/pull/1495/commits/223d249d212563809b5752a90c2c10b55b992075.

repeatedly:
Thanks!

Issue #1485, opened 2017-03-01
Fluentd 0.14.13 gets buffer overflow errors with the same settings as Fluentd 0.12.32


qingling128:
During some basic load testing, we found that the upgrade to Fluentd 0.14.13 from 0.12.32 caused buffer overflow errors to be triggered very frequently.

[Error]
```
2017-03-01 17:02:20 +0000 [warn]: #0 failed to write data into buffer by buffer overflow action=:block
```

[Steps to reproduce]
1. Create a GCP instance (1 vCPU, 3.75GB, Debian GNU/Linux 8)
2. Install docker:
```
$ sudo apt-get install docker.io
```
3. Create some test logs with the logs generator container:
```
$ mkdir ~/logs
$ sudo docker run -i -e "LOGS_GENERATOR_DURATION=1s" -e "LOGS_GENERATOR_LINES_TOTAL=1000000" gcr.io/google_containers/logs-generator:v0.1.0 2>&1 | awk '{print "{\"log\":\"" $0 "\"}"}' > ~/logs/log.log
```
4. Create configs
```
$ mkdir ~/config.d
$ cat ~/config.d/config.conf
<match fluent.**>
  type null
</match>

<source>
  type tail
  format json
  time_key time
  path /var/log/containers/*.log
  pos_file /var/log/gcp-containers.log.pos
  time_format %Y-%m-%dT%H:%M:%S.%NZ
  tag reform.*
  read_from_head true
</source>

<match reform.**>
  type record_reformer
  enable_ruby true
  tag kubernetes.${tag_suffix[4].split('-')[0..-2].join('-')}
</match>

# We use 2 output stanzas - one to handle the container logs and one to handle
# the node daemon logs, the latter of which explicitly sends its logs to the
# compute.googleapis.com service rather than container.googleapis.com to keep
# them separate since most users don't care about the node logs.
<match kubernetes.**>
  type google_cloud
  # Set the buffer type to file to improve the reliability and reduce the memory consumption
  buffer_type file
  buffer_path /var/log/fluentd-buffers/kubernetes.containers.buffer
  # Set queue_full action to block because we want to pause gracefully
  # in case of the off-the-limits load instead of throwing an exception
  buffer_queue_full_action block
  # Set the chunk limit conservatively to avoid exceeding the GCL limit
  # of 10MiB per write request.
  buffer_chunk_limit 2M
  # Cap the combined memory usage of this buffer and the one below to
[...]

PR #1502, opened 2017-03-15
Fix Input and Output deadlock when buffer is full during startup

repeatedly:
Output threads need calling `after_start` to run but
`after_start` is sometimes not called when Input#start is blocked.
This happens when in_tail with `read_from_head true` and
output with `overflow_action block`.
This fixes the problem by checking `after_start` immediately after start.

This may be related with #1485

### DIFF
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -136,8 +136,9 @@ module Fluent
     def start
       lifecycle(desc: true) do |i| # instance
         i.start unless i.started?
-      end
-      lifecycle(desc: true) do |i|
+        # Input#start emits lots of evetns with in_tail/`read_from_head true` case and
+        # it causes deadlock for small buffer/queue output. To avoid such problem,
+        # buffer related output threads should be run before `Input#start`.
         i.after_start unless i.after_started?
       end
     end
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -136,9 +136,12 @@ module Fluent
     def start
       lifecycle(desc: true) do |i| # instance
         i.start unless i.started?
-        # Input#start emits lots of evetns with in_tail/`read_from_head true` case and
-        # it causes deadlock for small buffer/queue output. To avoid such problem,
+        # Input#start sometimes emits lots of evetns with in_tail/`read_from_head true` case
+        # and it causes deadlock for small buffer/queue output. To avoid such problem,
         # buffer related output threads should be run before `Input#start`.
+        # This is why after_start should be called immediately after start call.
+        # This depends on `desc: true` because calling plugin order of `desc: true` is
+        # Output, Filter, Label, Output with Router, then Input.
         i.after_start unless i.after_started?
       end
     end
--- a/test/test_plugin_classes.rb
+++ b/test/test_plugin_classes.rb
@@ -21,6 +21,28 @@ module FluentTest
     end
   end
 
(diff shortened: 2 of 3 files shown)


repeatedly:
@tagomoris Do you see any concern for this patch?

tagomoris:
@repeatedly this change breaks many tests (makes tests unstable at least). read this commit comment: https://github.com/fluent/fluentd/commit/bd5841e930dfca8f73181d77d00ad8a2de98f6c6

IMO it's wrong to emit events in `TailInput#start`. It should be done in another threads executed from `#start` thread.

repeatedly:
> this change breaks many tests (makes tests unstable at least). 

Which point? input or filter `start` assumes `Output#after_start` is not called? 

> IMO it's wrong to emit events in TailInput#start. It should be done in another threads executed from #start thread.

Yes but current plugin API, it is hard, right?
How to know other threads are running inside input plugin?

tagomoris:
About the order of `#start` and `#after_start`, I made wrong comment. As you mentioned, there are no barrier to call `#after_start` immediately after calling `#start` per plugins.

But it depends the order of `RootAgent#lifecycle(desc: true)` why this patch solves #1485. So the comment in `root_agent.rb` should mention that `desc: true` means the order from `Output`, `Filter`, `Output with event_emitter` and `Input`.

About the threading, there are some options to do it - running long-running process in `#start` is bad practice anyway.
1. adding `immediate: (default false)` option to `PluginHelper::Timer#timer_create` to invoke callback immediately in timer thread (by creating TimerWatcher with `0` interval and `repeat: false`)
2. using `thread_create` to run `refresh_watchers` in `TailInput#start`

The first option is more general solution. `ChildProcess#child_process_execute` method also has `immediate:` optional argument besides `interval:` argument, so having same option in `timer_create` looks consistent solution. (in that case, `child_process_execute(immediate: true)` can be re-implemented on `timer_create(immediate: true)`).

The second option is to call `thread_create` and call `refresh_watchers` by TailInput itself. It may call `#refresh_watchers` in parallel from the thread created and timer threads without any control, so it requires mutex to prohibit it.

repeatedly:
> adding immediate: (default false) option to PluginHelper::Timer#timer_create to invoke callback immediately in timer thread (by creating TimerWatcher with 0 interval and repeat: false)

It seems good. I will implement this feature for in_tail like case in other patch.
in_tail specific fix is not good for plugin authors. So 2 approach is less merit.

repeatedly:
Update comment of `RootAgent#start`

lib/fluent/root_agent.rb
```diff
@@ -136,8 +136,12 @@ def lifecycle(desc: false, kind_callback: nil)
...
       lifecycle(desc: true) do |i| # instance
         i.start unless i.started?
-      end
-      lifecycle(desc: true) do |i|
+        # Input#start sometimes emits lots of evetns with in_tail/`read_from_head true` case
+        # and it causes deadlock for small buffer/queue output. To avoid such problem,
+        # buffer related output threads should be run before `Input#start`.
+        # This is why after_start should be called immediately after start call.
+        # This depends on `desc: true` because calling plugin order of `desc: true` is
+        # Output, Filter, Label, Output with Router and Input.
```
tagomoris: "Output with Router, then Input" looks better to emphasis the order.

tagomoris:
Ah, please open an issue to add the `immediate` option to `#timer_create` before closing this pull-request.

PR #1498, opened 2017-03-13
msgpack format chosen with content-type header

jdconley:
Adds support for msgpack record(s) if the content-type header is set to “application/msgpack”.

This is to support the format of messages sent by the fluentbit project’s out_http plugin.

PR #1505, opened 2017-03-16
Msgpack content type 0.12

jdconley:
This is the same as #1498 but for the 0.12 branch. Test has been adapted to the 0.12 style.


jdconley:
Oops, put this in here on accident.

PR #1506, opened 2017-03-16
msgpack format chosen with content-type header

jdconley:
This is the same as #1498 but for the 0.12 branch. Test has been adapted to the 0.12 style.


repeatedly:
Thanks!

PR #1507, opened 2017-03-17
Add <worker n> section to set a configuration for a specific worker

### LINKED ISSUE #1392: `<worker>` section to enable a set of configuration in a specified worker
Symmetric multi worker processing #1386 doesn't permit to execute any plugins and configurations in just one process.
But there are requirements and feature requests for such use cases. For example, executing plugins for monitoring (should be unique in a server).

This is an idea to keep symmetric semantics for multi process workers, but to realize such workloads.
* add `<worker N>` section in configuration top-level
* enable that section only in the specified worker (specified by worker id)
* allow to run plugins only in `<worker>` sections, which doesn't support multi worker feature

```apache
<system>
  workers 8
  root_dir /path/fluentd/root
</system>
<source>  # top-level sections works on all workers in parallel
  @id edge_input
  @type forward
  @label @traffic
  port 24224
</source>
<label @traffic>  # this section works on all workers too
  <match **>
    @id backend_output
    @type forward
    <buffer>
      @type file
      flush_interval 8m
    </buffer>
  </match>
</label>
<worker 0> # this section works only on first worker process
  <source>
    @id system_log_reader
    @type tail
    tag monitoring
    path /path/now/watching/...
  </source>
  <match monitoring>
    @id output_for_monitoring
    @type elasticsearch
    # ...
  </match>
</worker>
```

mururu:
Closes #1392.

This PR introduce `<worker n>` section. Configurations under this section is enabled only in specified worker and act as if they work in single worker environment. More precisely, config elements in worker section get `worker_target_id` recursively. Prior to configure phase of corresponding plugins, `system_config.workers` for the plugins is overwritten as `1` if target_worker_id equals the worker's worker_id.
It also means that plugins(including owned plugins) which doesn't support multi worker feature can run only in worker section.
A worker section can have `<label>`, `<source>`, `<filter>` and `<match>` as a child element.

Example from #1392

```
<system>
  workers 8
  root_dir /path/fluentd/root
</system>
<source>  # top-level sections works on all workers in parallel
  @id edge_input
  @type forward
  @label @traffic
  port 24224
</source>
<label @traffic>  # this section works on all workers too
  <match **>
    @id backend_output
    @type forward
    <buffer>
      @type file
      flush_interval 8m
    </buffer>
  </match>
</label>
<worker 0> # this section works only on first worker process
  <source>
    @id system_log_reader
    @type tail
    tag monitoring
    path /path/now/watching/...
  </source>
  <match monitoring>
    @id output_for_monitoring
    @type elasticsearch
    # ...
  </match>
</worker>
```

### DIFF
--- a/example/worker_section.conf
+++ b/example/worker_section.conf
@@ -1,5 +1,5 @@
 <system>
-  workers 8
+  workers 4
   root_dir /path/fluentd/root
 </system>
 
@@ -8,7 +8,7 @@
   port 24224
 </source>
 
-<match all> # this section works on all workers too
+<match all> # this sections also works on all workers in parallel
   @type stdout
   <inject>
     worker_id_key worker_id
@@ -16,7 +16,18 @@
 </match>
 
 <worker 0> # this section works only on first worker process
-  <match worker0>
+  <source>
+    @type tail
+    format none
+    path /var/log/fluentd_test.log
+    pos_file /var/log/fluentd_test.pos
+    tag tail
+    rotate_wait 5
+    read_from_head true
+    refresh_interval 60
+  </source>
+
+  <match tail>
     @type stdout
     <inject>
       worker_id_key worker_id
--- a/example/worker_section.conf
+++ b/example/worker_section.conf
@@ -1,0 +1,25 @@
+<system>
+  workers 8
+  root_dir /path/fluentd/root
+</system>
+
+<source> # top-level sections works on all workers in parallel
+  @type forward
+  port 24224
+</source>
+
+<match all> # this section works on all workers too
+  @type stdout
+  <inject>
+    worker_id_key worker_id
+  </inject>
+</match>
+
+<worker 0> # this section works only on first worker process
+  <match worker0>
+    @type stdout
+    <inject>
+      worker_id_key worker_id
+    </inject>
+  </match>
+</worker>
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -62,7 +62,7 @@ module Fluent
 
       # initialize <match> and <filter> elements
       conf.elements('filter', 'match').each { |e|
-        next if e.has_target? && e.target_worker_id != Fluent::Engine.worker_id
+        next if e.has_target_worker_id? && e.target_worker_id != Fluent::Engine.worker_id
         pattern = e.arg.empty? ? '**' : e.arg
         type = e['@type']
         raise ConfigError, "Missing '@type' parameter on <#{e.name}> directive" unless type
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -62,6 +62,7 @@ module Fluent
 
       # initialize <match> and <filter> elements
       conf.elements('filter', 'match').each { |e|
+        next if e.has_target? && e.target_worker_id != Fluent::Engine.worker_id
         pattern = e.arg.empty? ? '**' : e.arg
         type = e['@type']
         raise ConfigError, "Missing '@type' parameter on <#{e.name}> directive" unless type
@@ -121,7 +122,8 @@ module Fluent
     end
 
     def add_match(type, pattern, conf)
-      log.info :worker0, "adding match#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
+      log_type = conf.target_worker_id == Fluent::Engine.worker_id ? :default : :worker0
+      log.info log_type, "adding match#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
 
       output = Plugin.new_output(type)
       output.context_router = @event_router
@@ -142,7 +144,8 @@ module Fluent
     end
 
     def add_filter(type, pattern, conf)
-      log.info :worker0, "adding filter#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
+      log_type = conf.target_worker_id == Fluent::Engine.worker_id ? :default : :worker0
+      log.info log_type, "adding filter#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
 
       filter = Plugin.new_filter(type)
       filter.context_router = @event_router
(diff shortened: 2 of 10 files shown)


mururu:
Updated.
@repeatedly @tagomoris Could you review this?

tagomoris:
Please add an example configuration which use this feature under `example/`.

mururu:
Done.

mururu:
Updated the example to try this feature easily.

example/worker_section.conf
```diff
@@ -0,0 +1,25 @@
...
+</source>
+
+<match all> # this section works on all workers too
+  @type stdout
+  <inject>
+    worker_id_key worker_id
+  </inject>
+</match>
+
+<worker 0> # this section works only on first worker process
```
tagomoris: This configuration doesn't use the `<worker>` section feature to execute plugins which doesn't support multi process workers.

lib/fluent/config/element.rb
```diff
@@ -213,6 +216,17 @@ def self.unescape_parameter(v)
...
       end
+
+      def set_target_worker_id(worker_id)
+        @target_worker_id = worker_id
+        @elements.each { |e|
+          e.set_target_worker_id(worker_id)
+        }
+      end
+
+      def has_target?
```
tagomoris: `#has_target` doesn't show the correct meaning of this method. What's the target?
`#has_target_worker_id` looks better.

lib/fluent/engine.rb
```diff
@@ -251,8 +257,13 @@ def push_log_event(tag, time, record)
       @log_event_queue.push([tag, time, record])
     end
 
-    private
+    def worker_id
+      return @_worker_id if @_worker_id
+      @_worker_id = (ENV['SERVERENGINE_WORKER_ID'] || 0).to_i
```
repeatedly: Above deleted comment should be here.

lib/fluent/root_agent.rb
```diff
@@ -64,11 +64,34 @@ def initialize(log:, system_config: SystemConfig.new)
...
+      # initialize <worker> elements
+      conf.elements(name: 'worker').each do |e|
+        target_worker_id_str = e.arg
+        if target_worker_id_str.empty?
+          raise ConfigError, "Missing worker id on <worker> directive"
+        end
+
+        target_worker_id = target_worker_id_str.to_i
+        if target_worker_id < 0 || target_worker_id > (Fluent::Engine.system_config.workers - 1)
+          raise ConfigError, "worker#{target_worker_id} specified by <worker> directive doesn't exist. Specify id between 0 and #{(Fluent::Engine.system_config.workers - 1)}"
```
repeatedly: If specify `-1`, log message is `worker-1`. This is unclear, `worker` `-` `1` or `worker` `-1`.

"worker id #{target_worker_id} specified in <worker> directive is not allowed. Available worker id is between 0 and #{(Fluent::Engine.system_config.workers - 1)}"

lib/fluent/agent.rb
```diff
@@ -62,6 +62,7 @@ def configure(conf)
 
       # initialize <match> and <filter> elements
       conf.elements('filter', 'match').each { |e|
+        next if e.has_target_worker_id? && e.target_worker_id != Fluent::Engine.worker_id
```
repeatedly: Extract to method is better?

lib/fluent/agent.rb
```diff
@@ -121,7 +122,8 @@ def lifecycle(desc: false)
     end
 
     def add_match(type, pattern, conf)
-      log.info :worker0, "adding match#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
+      log_type = conf.target_worker_id == Fluent::Engine.worker_id ? :default : :worker0
```
repeatedly: Extract to method is better too?

test/command/test_fluentd.rb
```diff
@@ -614,5 +614,205 @@ def assert_fluentd_fails_to_start(cmdline, *pattern_list, timeout: 10)
         "config error file=\"#{conf_path}\" error_class=Fluent::ConfigError error=\"Plugin 'file' does not support multi workers configuration (Fluent::Plugin::FileBuffer)\"",
       )
     end
+
+    test 'failed to start workers when configured plugins as chidren of MultiOutput do not support multi worker configuration' do
+      script =  "require 'fluent/plugin/output'\n"
```
repeatedly: Here document is better.

test/command/test_fluentd.rb
```diff
@@ -614,5 +614,205 @@ def assert_fluentd_fails_to_start(cmdline, *pattern_list, timeout: 10)
...
+        create_cmdline(conf_path),
+        "#0 fluentd worker is now running worker=0",
+        "#1 fluentd worker is now running worker=1",
+        /(?!#\d) adding source type="dummy"/,
+        '#1 adding source type="dummy"'
+      )
+    end
+
+    test 'success to start workers when configured plugins only for specific worker do not support multi worker configuration' do
+      script =  "require 'fluent/plugin/input'\n"
```
repeatedly: ditto

test/command/test_fluentd.rb
```diff
@@ -614,5 +614,205 @@ def assert_fluentd_fails_to_start(cmdline, *pattern_list, timeout: 10)
...
+      assert_log_matches(
+        create_cmdline(conf_path),
+        "#0 fluentd worker is now running worker=0",
+        "#1 fluentd worker is now running worker=1",
+        '#1 adding match pattern="dummy" type="null"'
+      )
+    end
+
+    test 'success to start workers when configured plugins as a chidren of MultiOutput only for specific worker do not support multi worker configuration' do
+      script =  "require 'fluent/plugin/output'\n"
```
repeatedly: ditto

mururu:
I added `#for_every_workers?`, `#for_this_worker?` and `#for_another_worker?` instead of `#has_target_worker_id?` and `#target_worker_id ==/!= Fluent::Engine.worker_id`. How about that?

repeatedly:
LGTM :)

repeatedly:
Merged. Thanks for hard work!

mururu:
Thanks!

PR #1515, opened 2017-03-22
Unify parameter name between udp and syslog plugins

repeatedly:
`in_syslog` is widely used so using `message_length_limit` is better for users.


repeatedly:

repeatedly:
I will merge this patch tomorrow.

PR #1522, opened 2017-03-30
Fix backward compatibility with Fluent::DetachProcess and Fluent::DetachMultiProcess

okkez:
Some 3rd party plugins use these methods in their code:

```ruby
class HTTPServer # w/o inheriting Fluent::Input, Fluent::Output
  include DetachMultiProcess

  def start
    detach_multi_process do
      # Do something to run HTTP server
    end
  end
end
```

See also:
  * https://github.com/groonga/fluent-plugin-groonga/blob/530f421080cd4ad131eacc1cfd6ade474fabc19d/lib/fluent/plugin/in_groonga.rb#L144
  * https://github.com/groonga/fluent-plugin-groonga/blob/530f421080cd4ad131eacc1cfd6ade474fabc19d/lib/fluent/plugin/in_groonga.rb#L83-L85

### DIFF
--- a/lib/fluent/compat/detach_process_mixin.rb
+++ b/lib/fluent/compat/detach_process_mixin.rb
@@ -17,9 +17,17 @@
 module Fluent
   module Compat
     module DetachProcessMixin
+      def detach_process
+        log.warn "#{__method__} is not supported in this version. ignored."
+        yield
+      end
     end
 
     module DetachMultiProcessMixin
+      def detach_multi_process
+        log.warn "#{__method__} is not supported in this version. ignored."
+        yield
+      end
     end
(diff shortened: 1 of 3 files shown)


okkez:
Failed CI but the failure seems not to be related to my changes.

kou:
@repeatedly Could you apply this change and release a new version?
Our plugin, fluent-plugin-groonga, doesn't work with the latest Fluentd...

(fluent-plugin-groonga still uses 0.12 API. It will migrate to 0.14 API later. The problem will be solved at the timing.)

repeatedly:
Sorryr for the late. Merged.

PR #1540, opened 2017-04-12
Do not warn that directories are unreadable in the in_tail plugin.

iantalarico:
Currently a warning will appear for each directory in globbed results.  They will never be readable so the message appears on each refresh.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -189,15 +189,15 @@ module Fluent::Plugin
         path = date.strftime(path)
         if path.include?('*')
           paths += Dir.glob(path).select { |p|
-            is_directory = File.directory?(p)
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -189,14 +189,17 @@ def expand_paths
         path = date.strftime(path)
         if path.include?('*')
           paths += Dir.glob(path).select { |p|
-            if File.readable?(p) && !File.directory?(p)
+            is_directory = File.directory?(p)
```
repeatedly: `is_file = !File.directory?(p)` is better for removing `!is_directory`.

iantalarico: Fixed

repeatedly:
Thx!

PR #1541, opened 2017-04-12
Handle wildcards in windows files paths in the in_tail plugin.

iantalarico:
Currently tail plugins on windows will work if a specific file is targeted:
```
<source>
  type tail
  format none
  path C:\SomeDir\Logs\the.logs
  tag custom-logs
</source>
```

However with wildcards the log files will not be picked up:
```
<source>
  type tail
  format none
  path C:\SomeDir\Logs\**\*'
  tag custom-logs
</source>
```

This is due to Dir.glob not handing '\'s properly.  The '\' must be converted to '/'.  FIle.* methods will take either a '\' or '/'.

Users could define paths with '/'s but this is not idiomatic for windows users.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -100,7 +100,7 @@ module Fluent::Plugin
 
       super
 
-      @paths = @path.split(',').map {|path| path.strip }
+      @paths = @path.split(',').map {|path| path.strip.gsub('\\', '/') }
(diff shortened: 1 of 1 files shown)


repeatedly:
Your approach is incomplete.
See previous discussion: https://github.com/fluent/fluentd/pull/754

iantalarico:
Ops I missed that issue, sorry!

PR #1544, opened 2017-04-18
Fix spelling/grammar in an error message

jwolski2:
This is a simple PR that fixes the spelling/grammar in the error message used to let users know that their plugin does not implement the correct filter methods.


repeatedly:
Thanks!

PR #1546, opened 2017-04-19
Fix out_file test with timezone

repeatedly:
@mururu Could you test this on your environment?

In v0.14, compat parameters handles `utc` parameter for parsers / formatters.
So if we want to use `utc` in buffer, need to specify `timekey_use_utc` in `<buffer>`.

### DIFF
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -105,6 +105,11 @@ module Fluent::Plugin
         buffer_conf['path'] = conf['path'] || '/tmp/dummy_path'
       end
 
+      if conf.has_key?('utc') || conf.has_key?('localtime')
+        param_name = conf.has_key?('utc') ? 'utc' : 'localtime'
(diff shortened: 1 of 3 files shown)


mururu:
Passed tests on my environment.
IIRC, v0.14 should work with v0.12's output configuration parameters? If so, we should fix the implementation of the compatibility layer but not tests? Otherwise, it looks good to me.

repeatedly:
> If so, we should fix the implementation of the compatibility layer but not tests?

Yeah, this is one idea. The problem is old out_file uses `utc` for both time slice and formatter.
In this result, hard to identify this parameter is for output or formatter.
We want to resolve this problem with v0.14.
Other idea is logging warning message like below.

"'utc' parameter is deprecated for output and this parameter is used for formatter plugin. Use timekey_use_utc in <buffer> instead"

mururu:
That makes sense. 
Another concern is that if a user explicitly set `time_slice_format` in addition to `utc`, then `<buffer> timekey_use_utc true </buffer>` is inserted automatically, right? It is confusing for me. Is it no problem?

repeatedly:
> then <buffer> timekey_use_utc true </buffer> is inserted automatically, right?

repeatedly:
Add warning message.

mururu:
Makes sense. LGTM.

PR #1547, opened 2017-04-19
Add formatted_to_msgpack_binary? to Output plugin API. fix #1538

### LINKED ISSUE #1538: More rubyish way for Fluent::Plugin::Output#formatted_to_msgpack_binary
This is suggestion about method name about Fluent::Plugin::Output#formated_to_msgpack_binary

- fluentd version:
  * fluentd 0.14.x or later.
- environment infomation:
   * OS independent
- Your problem explanation. If you have an error logs, write it together.

  * Expected method name
    Fluent::Plugin::Output#formatted_to_msgpack_binary?

  * Actual method name
    Fluent::Plugin::Output#formatted_to_msgpack_binary

Traditionally, the method which returns boolean result is named with `?` suffix. That is more rubyish way to do it and easy to understand the behavior of #formated_to_msgpack_binary. How about `alias`?

repeatedly:


repeatedly:
@mururu Could you also check this patch?

mururu:
@repeatedly sure.

test/plugin/test_output_as_buffered.rb
```diff
@@ -262,9 +288,12 @@ def waiting(seconds)
       end
     end
 
-    test 'plugin using custom format cannot iterate chunk in #write' do
+    data('formatted_to_msgpack_binary?' => :custom,
+         'formatted_to_msgpack_binary' => :old_custom)
+    test 'plugin using custom format cannot iterate chunk in #write' do |out_type|
```
mururu: This testcase doesn't verify this change because `@format_type_is_msgpack.call` always returns `false` regardless of overriding `#formatted_to_msgpack_binary?`.

mururu:
This test case is better to add tests for `:custom` and `:old_custom`?
https://github.com/fluent/fluentd/pull/1547/files#diff-9b239225006e1bf484998d3e72b28704R338

repeatedly:
You are right. I just changed test case.

mururu:
LGTM

repeatedly:
Thx!

PR #1548, opened 2017-04-19
Allow the Windows Service name Fluentd runs as to be configurable.

iantalarico:
This will allow the name, display name and description fluentd is running as when it is a Windows Service to be configured.

This also removes an unneeded option 'winsvcreg'

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -385,6 +385,9 @@ module Fluent
         standalone_worker: false,
         signame: nil,
         winsvcreg: nil,
+        winsvc_name: 'fluentdwinsvc',
+        winsvc_display_name: 'Fluentd Windows Service',
+        winsvc_desc: 'Fluentd is an event collector system.',
       }
     end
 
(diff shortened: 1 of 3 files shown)


lib/fluent/supervisor.rb
```diff
@@ -385,6 +385,9 @@ def self.default_options
         standalone_worker: false,
         signame: nil,
         winsvcreg: nil,
+        winsvc_name: 'fluentdwinsvc',
```
repeatedly: default_options is bad place for these options because
default_options is for Supervisor, not command line.
This is why test failed.

BTW, `winsvcreg` can be removed from `default_options`.
There is no use for now.

iantalarico: Is there another similar place I could add these? I couldn't find one. I can make one in fluentd?

iantalarico: I just moved these under the fluentd.windows check.  PTAL

repeatedly:
LGTM

@nurse Do you see any concern?

nurse:
@repeatedly Looks good!

repeatedly:
Thanks!

PR #1562, opened 2017-05-09
Support %iso8601 special case for time_format. fix #1528

### LINKED ISSUE #1528: Unable to parse ISO8601
Currently fluentd branch 0.12 does not contain code to allow parsing ISO8601 time format.
For some reason unknown to me, when `time_format` is omitted in JSON parser, time is interpreted as integer value (UNIX-time). However, there is no way to parse ISO 8601: it cannot be described as a `Time#strptime` format, since it allows ambiguities. On the onther hand, it is trivial to implement UNIX-time parising with `Time#strptime`: `Time.strptime(value.to_s, '%s')`. This may be further optimized to parse integers directly (like it's done currently in the code), but it's only applicable if `time_format == '%s'`!
It just seems like a poor design decision to make default format different from ISO8601 like in the rest of the parsers and inputs.

Can this be fixed now, or should this maintain backward-compatibility with the previous design? In case we have to maintain backwards compatibility, maybe we could add a special case to allow proper ISO8601 parsing? Maybe add logic for `%iso8601` format string to force parsing with `Time.iso8601`?

repeatedly:
Ruby's iso8601 parser is a restrict version and it doesn't support `20170101T120000+0900` like format.
But this parser can handle `2017-01-01T12:00:00+09:00` and `2017-01-01T12:00:00.123+09:00` cases, so supporting it makes configuration simpler for iso8601 logs.

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -208,6 +208,7 @@ module Fluent
       @parse = case
                when format_with_timezone && strptime then ->(v){ Fluent::EventTime.from_time(strptime.exec(v)) }
                when format_with_timezone             then ->(v){ Fluent::EventTime.from_time(Time.strptime(v, format)) }
+               when format == '%iso8601'             then ->(v){ Fluent::EventTime.from_time(Time.iso8601(v)) }
                when strptime then ->(v){ t = strptime.exec(v);         Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
                when format   then ->(v){ t = Time.strptime(v, format); Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
                else ->(v){ Fluent::EventTime.parse(v) }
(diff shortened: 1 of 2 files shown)


repeatedly:

repeatedly:
Yes. I assume timestamp has a offset because iso8601 defines offset in the spec.

repeatedly:
If we need to handle timezone seprately, we can add timezone parameter handling later.

mururu:
Sorry, I misunderstood Time.iso8601. When the function gets a timestamp without an offset or "Z", it considers the timezone as (fluentd's) localtime. If we want to handle the timestamp as UTC, we have to add `Time.now.localtime.utc_offset` to the result of Time.iso8601 like the strptime case.

> Yes. I assume timestamp has a offset because iso8601 defines offset in the spec.

You mean the spec says the timestamp must have the offset or "Z"? It sounds good.
I think it is better to write the fact the iso8601 parser accept the timestamp without offset and the timezone is recognized as the same with server's localtime in the document.

mururu:
anyway, LGTM.

repeatedly:
Yeah, we should mention the timezone in the document.

PR #1564, opened 2017-05-10
Enable in_monitor_agent to find plugin type from its class

mururu:
Configurations for plugins generated dynamically sometimes don't have @type and type. (e.g. fluent-plugin-forest)


repeatedly:
Good catch.

PR #1568, opened 2017-05-12
Error: uninitialized constant Fluent::Engine

sampointer:
```
# /opt/td-agent/embedded/bin/fluent-binlog-reader cat -f json failed.log.0 
Usage: fluent-binlog-reader cat [options] file
Options:
    -p, --plugin DIR                 add library directory path
    -f, --format TYPE                configure output format
    -e KEY=VALUE                     configure formatter config params
    -n COUNT                         Set the number of lines to display
Error: uninitialized constant Fluent::Engine
```


repeatedly:
Good catch!

PR #1574, opened 2017-05-16
Add ignore_repeated_permission_error

repeatedly:
This option is useful when non-readable file exist in tailing directory and hard to exclude via `exclude_path`, e.g. file is generated dynamically.


repeatedly:
@mururu Could you chec kthis?

PR #1575, opened 2017-05-18
Accept private key for TLS server without passphrase

mururu:
It is convenient for users to leave the way to manage private keys to them like other many middlewares.

### DIFF
--- a/lib/fluent/plugin_helper/cert_option.rb
+++ b/lib/fluent/plugin_helper/cert_option.rb
@@ -46,12 +46,12 @@ module Fluent
         case
         when conf.cert_path
           raise Fluent::ConfigError, "private_key_path is required when cert_path is specified" unless conf.private_key_path
-          raise Fluent::ConfigError, "private_key_passphrase is required when cert_path is specified" unless conf.private_key_passphrase
+          log.warn "For security reason, setting private_key_passphrase is recommended when cert_path is specified" unless conf.private_key_passphrase
           cert_option_load(conf.cert_path, conf.private_key_path, conf.private_key_passphrase)
 
         when conf.ca_cert_path
           raise Fluent::ConfigError, "ca_private_key_path is required when ca_cert_path is specified" unless conf.ca_private_key_path
-          raise Fluent::ConfigError, "ca_private_key_passphrase is required when ca_cert_path is specified" unless conf.ca_private_key_passphrase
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin_helper/cert_option.rb
```diff
@@ -46,12 +46,10 @@ def cert_option_server_validate!(conf)
         case
         when conf.cert_path
           raise Fluent::ConfigError, "private_key_path is required when cert_path is specified" unless conf.private_key_path
-          raise Fluent::ConfigError, "private_key_passphrase is required when cert_path is specified" unless conf.private_key_passphrase
```
repeatedly: How about logging warn message like "For security reason, setting private_key_passphrase is recommended"?

lib/fluent/plugin_helper/cert_option.rb
```diff
@@ -46,12 +46,10 @@ def cert_option_server_validate!(conf)
...
           cert_option_load(conf.cert_path, conf.private_key_path, conf.private_key_passphrase)
 
         when conf.ca_cert_path
           raise Fluent::ConfigError, "ca_private_key_path is required when ca_cert_path is specified" unless conf.ca_private_key_path
-          raise Fluent::ConfigError, "ca_private_key_passphrase is required when ca_cert_path is specified" unless conf.ca_private_key_passphrase
```
repeatedly: ditto

test/plugin_helper/test_server.rb
```diff
@@ -982,7 +985,7 @@ def open_tls_session(addr, port, verify: true, cert_path: nil, selfsigned: true,
         assert_equal "yay\nfoo\n", received
       end
 
-      test 'load chained server cert by private CA cert file, verified from clients using CA cert file as root' do
+      test 'load chained server cert by private CA cert file, verified from clients using CA cert file as root' do |ca_key_passphrase, private_key_passphrase|
```
repeatedly: Nedd `data`?

test/plugin_helper/test_server.rb
```diff
@@ -1137,20 +1142,18 @@ def open_tls_session(addr, port, verify: true, cert_path: nil, selfsigned: true,
...
```
repeatedly: ditto

lib/fluent/plugin_helper/cert_option.rb
mururu: Fair enough.

mururu:
Updated.

Issue #1484, opened 2017-03-01
sometimes out_forward heartbeat(tcp) failed with fluentd>0.14.9


bungoume:
sometimes out_forward heartbeat(tcp) failed

when sender fluentd: 0.14.11 and 0.14.13
aggregator fluentd: 0.14.11 and 0.14.13

This problem does not have occurred in sender v0.14.8.

fluentd.log
```
2017-03-01 06:49:37 +0000 [warn]: recovered forwarding server '' host="log-aggregator.local" port=10224
2017-03-01 06:49:37 +0000 [warn]: recovered forwarding server '' host="log-aggregator.local" port=10224
2017-03-01 06:49:45 +0000 [warn]: retry succeeded. chunk_id="549a573de08a4928f31d10f236155293"
2017-03-01 06:49:51 +0000 [warn]: retry succeeded. chunk_id="549a579f25b7d32e5db46d5b6ab57ce5"
2017-03-01 06:54:18 +0000 [warn]: detached forwarding server '' host="log-aggregator.local" port=10224 hard_timeout=true
2017-03-01 06:54:18 +0000 [warn]: buffer flush took longer time than slow_flush_log_threshold: elapsed_time=60.00606061401777 slow_flush_log_threshold=20.0 plugin_id="object:2b093717af00"
2017-03-01 06:55:18 +0000 [warn]: failed to flush the buffer. retry_time=0 next_retry_seconds=2017-03-01 06:55:19 +0000 chunk="549a5c80fadbc463279c0c9408e0c210" error_class=Fluent::Plugin::ForwardOutput::NoNodesAvailable error="no nodes are available"
  2017-03-01 06:56:18 +0000 [warn]: suppressed same stacktrace
2017-03-01 06:56:18 +0000 [warn]: failed to flush the buffer. retry_time=0 next_retry_seconds=2017-03-01 06:55:19 +0000 chunk="549a5c80fadbc463279c0c9408e0c210" error_class=Fluent::Plugin::ForwardOutput::NoNodesAvailable error="no nodes are available"
  2017-03-01 06:56:18 +0000 [warn]: suppressed same stacktrace
2017-03-01 06:57:18 +0000 [warn]: detached forwarding server '' host="log-aggregator.local" port=10224 hard_timeout=true
[...]

PR #1581, opened 2017-05-24
Fix ack handling

repeatedly:


repeatedly:

mururu:
#1484 is about failure of heartbeat, but ecbdf8d027960af3fd73bb3ceba41c43edcc0181 is a fix for socket leak. So the commit message seems wrong. Except for that, LGTM.

repeatedly:
Yeah, because tcp heartbeat is hung by socket leak. commit message shows it.

mururu:
> because tcp heartbeat is hung by socket leak

I see. It looks perfect.

PR #1597, opened 2017-06-14
in_tail: There are cases where the excluded_path doesn't apply by timing.

t-osakada:
Newly created files after creating the `excluded` list may not be excluded.

Fix the order of creating the `paths`/`excluded` list. Creating the `excluded` list should be last for certainly excluding.


repeatedly:
Good catch!

PR #1600, opened 2017-06-20
Add rfc5424 regex without priority 

cosmo0920:


mururu:
Should we also fix `REGEXP_DETECT_RFC5424`?

cosmo0920:
I have no idea for now. 
In without priority case, should we detect with `time_format` and @rfc5424_time_format?

repeatedly:
I don't think so. `REGEXP_DETECT_RFC5424` is mainly used for in_syslog and there is no use case for syslog based file log.

mururu:
I see.

repeatedly:
Thanks!
@cosmo0920 Could you send same patch to v0.12 branch?

cosmo0920:
Sure. I'll do it.  :)

PR #1609, opened 2017-06-30
Add more available characters for tag part of syslog format

mururu:
About the TAG part of syslog format (RFC3164).
RFC3164 says "Any non-alphanumeric character will terminate the TAG field". But some syslog senders send non-alphanumeric characters as a part of TAG, so currently fluentd allowed some non-alphanumeric characters: `[a-zA-Z0-9_\/\.\-]`. In the real world, there are cases that require more characters. For example, Cisco ASA Series Syslog Messages includes `%` like `%ASA-1-101001`. It is useful if fluentd can handle such messages.

rsyslog syslog-ng and accept all characters except ` `, `:` and `[`:

https://github.com/rsyslog/rsyslog/blob/master/tools/pmrfc3164.c#L314-L347
https://github.com/balabit/syslog-ng/blob/master/modules/syslogformat/syslog-format.c#L558-L609

This PR follows them.

Once this PR is merged, I will send a PR for the master branch.

### DIFF
--- a/lib/fluent/parser.rb
+++ b/lib/fluent/parser.rb
@@ -537,9 +537,9 @@ module Fluent
 
     class SyslogParser < Parser
       # From existence TextParser pattern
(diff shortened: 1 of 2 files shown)


repeatedly:
Looks good 👍

mururu:
Thanks!

PR #1611, opened 2017-07-03
Use config_section for filter_grep

okkez:
Users can use both `regexpN` and `<regexp>` section.
But `regexpN` displays deprecation warnings on boot.


lib/fluent/plugin/filter_grep.rb
```diff
@@ -24,43 +24,67 @@ class GrepFilter < Filter
...
+        @_excludes[key] = Regexp.compile(exclude)
       end
     end
 
     def filter(tag, time, record)
       result = nil
       begin
         catch(:break_loop) do
-          @regexps.each do |key, regexp|
+          @regexps.each do |e|
```
repeatedly: This implementation is slow.
Transform `@regexps` into `@_regexps` is better.

repeatedly:
Thanks!

@okkez Could you backport this change to v0.12?

okkez:
Sure. I will try it later.

PR #1615, opened 2017-07-05
Add resolve_hostname config param in in_syslog, fix #1614

### LINKED ISSUE #1614: Please add resolve_hostname parameter for in_syslog plugin
Please add resolve_hostname parameter for in_syslog plugin , like what in_forward does.
When I enable include_source_host, fluentd auto lookup the hostname. But sometimes we do not need it, we just need IP address.
Thanks

cosmo0920:


cosmo0920:
I'll send this patch for master after merging this PR.

repeatedly:
How about adding `source_address_key` together?

cosmo0920:
I've added source_address_key config parameter.

repeatedly:
Thanks!

PR #1618, opened 2017-07-06
Make sure all log events are emitted before shutdown

mururu:
Engine emits all log events before shutdown RootAgent.

I'll send a same patch to the master branch after this PR is merged.

### DIFF
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -165,7 +165,6 @@ module Fluent
       $log.disable_events(Thread.current)
 
       while sleep(LOG_EMIT_INTERVAL)
-        break if @log_event_loop_stop
         next if @log_event_queue.empty?
 
         # NOTE: thead-safe of slice! depends on GVL
@@ -179,6 +178,7 @@ module Fluent
             $log.error "failed to emit fluentd's log event", tag: tag, event: record, error_class: e.class, error: e
           end
         }
+        break if @log_event_loop_stop
       end
     end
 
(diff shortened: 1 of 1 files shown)


lib/fluent/engine.rb
```diff
@@ -209,11 +209,11 @@ def run
         $log.error_backtrace
       ensure
         $log.info "shutting down fluentd"
-        shutdown
         if @log_emit_thread
           @log_event_loop_stop = true
           @log_emit_thread.join
         end
+        shutdown
```
repeatedly: Does calling `shutdown` after log thread stop cause log inside `shutdown` is dropped?

mururu: Yes. I wanted to make sure that the output plugin capturing events isn't terminated when log_emit_thread emit events. It seems to cause non-deterministic behaviour. But I'll try to capture more events event in `shutdown`.

repeatedly: I see. We should dump buffered log event `log_event_queue` after shutdowned.

repeatedly:
Merged. Wait next patch :)

repeatedly:
Can we release new v0.12 or should wait new patch?

mururu:
I think we can release the new version.

mururu:
^ I'll send a new patch probably tomorrow.

mururu:
@repeatedly Here is the next patch: https://github.com/fluent/fluentd/pull/1625. If you have time before releasing the new version, please review it.

Issue #1555, opened 2017-04-28
error_class=NoMethodError error="undefined method `next_time' for nil:NilClass"


cosmok:
- fluentd 0.14.14

getting the following error;
```
2017-04-28 07:06:07 +1000 [error]: #0 error on output thread error_class=NoMethodError error="undefined method `next_time' for nil:NilClass"
  2017-04-28 07:06:07 +1000 [error]: #0 /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/fluentd-0.14.14/lib/fluent/plugin/output.rb:1262:in `block in flush_thread_run'
  2017-04-28 07:06:07 +1000 [error]: #0 /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/fluentd-0.14.14/lib/fluent/plugin/output.rb:1262:in `synchronize'
  2017-04-28 07:06:07 +1000 [error]: #0 /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/fluentd-0.14.14/lib/fluent/plugin/output.rb:1262:in `flush_thread_run'
  2017-04-28 07:06:07 +1000 [error]: #0 /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/fluentd-0.14.14/lib/fluent/plugin/output.rb:415:in `block (2 levels) in start'
```

flow is; tail a few files, process then store results.

Not sure where exactly in the flow the error is being triggered.

Result being that fewer than expected lines are being processed.

igorpeshansky:
Hi. Can you please explain why you closed this? Is this no longer an issue for you?

cosmok:
Hi,

I closed it primarily because it got no response ;( . But now that it's attracted some attention, I would like to add some additional details.

My setup is like I mentioned here:

https://github.com/fluent/fluentd/issues/1516#issuecomment-290606804

I noticed that when I changed my script to ignore jumbled records (and not exit(1) inside the script -  as I had done before) this error went away.  But, at the cost of not processing some rows.

gcs-github:
I think I just ran into a similar problem on fluentd 0.14.13.

My forward instance became unavailable for a moment -> receiving an ack failed -> started to see errors about next_time lacking from the chunks to retry

```
2017-05-30 17:15:11 +0000 [warn]: #0 destination node closed the connection. regard it as unavailable. host="127.0.0.1" port=24224
[...]

PR #1623, opened 2017-07-11
Fix race condition of retry state in flush thread

repeatedly:
This is simple solution but performance is bit decreased.
The better solution is changing retry mechanizm, e.g. don't share state, but it makes
the code complicated. This is feature task.

This fixes https://github.com/fluent/fluentd/issues/1555

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1264,7 +1264,7 @@ module Fluent
             current_clock = Fluent::Clock.now
             interval = state.next_clock - current_clock
 
-            if state.next_clock <= current_clock && (!@retry || @retry_mutex.synchronize{ @retry.next_time } <= Time.now)
+            if state.next_clock <= current_clock && @retry_mutex.synchronize { @retry ? @retry.next_time <= Time.now : true }
               try_flush
 
               # next_flush_time uses flush_thread_interval or flush_thread_burst_interval (or retrying)


repeatedly:
Reproduce step:

- conf

```
# dummy plugin is also ok. Need to generate lots of events.
<source>
  @type tail
  path /pat/to/log
  tag test.tail
  read_from_head true
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type buf_test
  @id buf_output

  <buffer>
    chunk_limit_size 500000
    flush_interval 5s
    flush_thread_count 16
    #retry_max_times 15
    retry_max_interval 2s
  </buffer>
</match>
```

- out_buf_test.rb

```rb
require 'fluent/output'

module Fluent::Plugin
  class BufTestOutput < Output
    Fluent::Plugin.register_output('buf_test', self)

    config_section :buffer do
      config_set_default :chunk_keys, ["tag"]
    end

    def initialize
      super
      @num = 0
      @m = Mutex.new
    end

    def configure(conf)
      super
    end

    def multi_workers_ready?
      true
    end

    def start
      super
    end

    def close
      log.error @num
      super
    end

    def write(chunk)
      if rand(10) > 7
        raise
      end

      @m.synchronize {
        @num += chunk.size
      }
    end
  end
end

```

repeatedly:
I forgot to send a patch...

PR #1627, opened 2017-07-13
Use config_section for filter_grep

okkez:
Users can use both `regexpN` and `<regexp>` section.
But `regexpN` displays deprecation warnings on boot.

Backport #1611


repeatedly:
Thanks!

Issue #1589, opened 2017-06-01
undefined method '<' for nil:NilClass


chrisrecalis:
### Problem

Consistently getting the following error `undefined method '<' for nil:NilClass` inside lib/fluent/plugin/output.rb. From first glance it looks as though the meta data timekey is nil somehow. Fluent then backs up and is unable to move on and flush anymore buffers. I'm unable to tell which chunk is causing this

### Configuration

fluentd version: `0.14.17`

```
<source>
    @type forward
    bind 0.0.0.0
    port 24284
</source>

<match account_*.event>
    @type bigquery

    json_key "#{ENV['BIGQUERY_JSON_KEY']}"
    project_id "#{ENV['BIGQUERY_PROJECT_ID']}"
    dataset ${tag_parts[0]}

    time_partitioning_type DAY

    num_threads 4

    <tabletracker>
        key "#{ENV['TABLE_TRACKER_KEY'] || 'name' }"
        include_key "#{ENV['TABLE_TRACKER_INCLUDE_KEY'] || 'false' }"
    </tabletracker>

    <buffer tag,time>
        @type file
        path /var/log/cruncher/
        chunk_limit_size 300000000 
        timekey "#{ENV['TIME_KEY'] || '15m' }"
        flush_mode "#{ENV['FLUSH_MODE'] || 'default' }"
    </buffer>
</match>
```

type `bigquery` is a custom plugin which inherits from Fluent::Plugin::Output

### Error Logs
```
#0 fluent/engine.rb:218:run: starting fluentd worker pid=10 ppid=1 worker=0
2017-06-01 21:42:07 +0000 [debug]: #0 plugin/buffer.rb:114:start: buffer started instance=26110880 stage_size=2104194374 queue_size=0
2017-06-01 21:42:07 +0000 [info]: #0 plugin/in_forward.rb:157:start: listening port port=24284 bind="0.0.0.0"
2017-06-01 21:42:07 +0000 [info]: #0 fluent/engine.rb:227:run: fluentd worker is now running worker=0
2017-06-01 21:42:07 +0000 [debug]: #0 plugin/output.rb:1207:enqueue_thread_run: enqueue_thread actually running
2017-06-01 21:42:07 +0000 [error]: #0 plugin/output.rb:1235:rescue in enqueue_thread_run: unexpected error while checking flushed chunks. ignored. error_class=NoMethodError error="undefined method `<' for nil:NilClass"
[...]

PR #1628, opened 2017-07-13
Write buffer metadata into file ASAP

tagomoris:
As shown in #1589, file buffer chunks possibly make broken/empty .meta files for its metadata.
I found that some points to be improved to prevent such cases:

* write metadata into files just after initializing buffer chunk object
  * before: .meta content is written at the time of `#commit` or `enqueue!` - it's too late
* write metadata msgpack binary, then truncate .meta file into the size of metadata
  * before: .meta was truncated into 0 at first, then msgpack binary was written

### DIFF
--- a/lib/fluent/plugin/buffer/file_chunk.rb
+++ b/lib/fluent/plugin/buffer/file_chunk.rb
@@ -285,6 +285,7 @@ module Fluent
             @meta.set_encoding(Encoding::ASCII_8BIT)
             @meta.sync = true
             @meta.binmode
+            write_metadata(update: false)
           rescue => e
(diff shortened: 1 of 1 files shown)


tagomoris:
The behavior of msgpack unpacker for non-truncated files is proven here: https://gist.github.com/tagomoris/4b22651072aadc74b0ab9ecdfcf658a2
```
[a loooooooooooooooooooooong msgpack binary]
[overwriting short binary]oooong msgpack binary] # => unpacker unpacks shorter one only
```

repeatedly:
I tested this patch with several days and no problem on my mac.
So merged and released rc version for testing at other environment.

PR #1633, opened 2017-07-14
match doc string with its config_param

mindw:


repeatedly:
It seems `resolve_hostname`'s `desc` is incorrect.
Could you update the desc with `https://github.com/fluent/fluentd/blob/c8c4ff7ffa0de175c422f2bc588e0b15179c87cc/lib/fluent/plugin/in_forward.rb#L46` ?

mindw:
updated, squashed and rebased.

repeatedly:
Thanks!

PR #1635, opened 2017-07-18
Add Ruby 2.4 on appveyor

nurse:


appveyor.yml
```diff
@@ -24,6 +24,10 @@ branches:
 # https://www.appveyor.com/docs/installed-software/#ruby
 environment:
   matrix:
+    - ruby_version: "24-x64"
+      devkit: C:\Ruby23-x64\DevKit
```
cosmo0920: `devkit:` section is meaningless for RubyInstaller2 version of  Ruby2.4.
We can enable C extension building via `ridk.cmd` when using RubyInstaller2's Ruby.
ref: https://github.com/oneclick/rubyinstaller2#the-ridk-command

nurse: Sure, I fixed to call `ridk.cmd`.

repeatedly:
Hmm... ruby 2.4 test always failed...
Something is changed between ruby 2.3 and ruby 2.4

cosmo0920:
I think that Ruby 2.4 x86-mingw32 hits this issue: https://github.com/cosmo0920/windows-pr/issues/21 
But I think that it is caused by win32-api gem in [api_call](https://github.com/cosmo0920/win32-api/blob/master/ext/win32/api.c#L762)
And I reported it in [bugs.ruby-lang.org](https://bugs.ruby-lang.org/issues/13746)

I have no idea root cause of the above issue for Ruby 2.4 x86-mingw32, though....

PR #1637, opened 2017-07-19
Record accessor helper

repeatedly:
Implement record_accessor helper for nested record support.
Support two jsonpath like syntaxes.

- dot notation: `$.key1.key2[0]`
- bracket notation: `$['key1'][0]['this.is.key2']`
- single key: `key`

In this patch, I also update `filter_grep` to use this helper.

NOTE: I implement Accessor with normal way. If need the performance, I will change its implementaion to eval way: https://gist.github.com/repeatedly/ab553ed260cd080bd01ec71da9427312

### DIFF
--- a/test/plugin_helper/test_record_accessor.rb
+++ b/test/plugin_helper/test_record_accessor.rb
@@ -30,8 +30,7 @@ class RecordAccessorHelperTest < Test::Unit::TestCase
       assert_equal ['key1', 'key2', 0], result
     end
 
-    data('dot' => '$.key1[0].ke y2',
-         'bracket' => "$['key1'][0]['ke y2']")
+    data('bracket' => "$['key1'][0]['ke y2']")
     test "nested keys ['key1', 0, 'ke y2']" do |param|
       result = Fluent::PluginHelper::RecordAccessor::Accessor.parse_parameter(param)
       assert_equal ['key1', 0, 'ke y2'], result
@@ -48,6 +47,7 @@ class RecordAccessorHelperTest < Test::Unit::TestCase
          "missing array index with dot" => "$.hello[]",
          "missing array index with braket" => "$[]",
          "more chars" => "$.key1[0]foo",
+         "whitespace char included key in dot notation" => "$.key[0].ke y",
          "empty keys with dot" => "$.",
          "empty keys with bracket" => "$[")
     test 'invalid syntax' do |param|
@@ -85,8 +85,7 @@ class RecordAccessorHelperTest < Test::Unit::TestCase
       assert_equal 1, accessor.call(r)
     end
 
-    data('dot' => '$.key1[0].ke y2',
-         'bracket' => "$['key1'][0]['ke y2']")
+    data('bracket' => "$['key1'][0]['ke y2']")
     test "nested keys ['key1', 0, 'ke y2']" do |param|
       r = {'key1' => [{'ke y2' => "value"}]}
       accessor = @d.record_accessor_create(param)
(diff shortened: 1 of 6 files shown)


repeatedly:

repeatedly:
Bench result:

```
% ruby dig_bench.rb
baseline     0.570000   0.000000   0.570000 (  0.573822)
expand args  1.100000   0.000000   1.100000 (  1.093701)
pre-eval     0.860000   0.000000   0.860000 (  0.867493)
```

sonots:
I thought jsonpath supports `$.key1[0]['this.is.key2']`, but this patch does not?

repeatedly:
@sonots yes. In fluentd case, mix case seems no merits.

sonots:
Well, I feel allowing users to write `$.key1[0]['this.is.key2']` is a merit.
I agree that it makes parsing a bit complex and consumes time, but I think it is acceptable tradeoff because parser usually runs only once at "configure".

repeatedly:
> I think it is acceptable tradeoff because parser usually runs only once at "configure".

Yes, it doesn't have a impact at runtime.

This is just jsonpath like, not jsonpath, so I don't think following similar syntax is not must. For documentation and maintainance, separating two way is less cost at the baseline.

sonots:
We need to get users' opinions. As one user, I like to get allowed `$.key1[0]['this.is.key2']`.

test/plugin_helper/test_record_accessor.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    data('dot' => '$.key1.key2[0]',
+         'bracket' => "$['key1']['key2'][0]")
+    test "nested keys ['key1', 'key2', 0]" do |param|
+      result = Fluent::PluginHelper::RecordAccessor::Accessor.parse_parameter(param)
+      assert_equal ['key1', 'key2', 0], result
+    end
+
+    data('dot' => '$.key1[0].ke y2',
```
sonots: Wow, this is acceptable? It looks wierd for me. I think we should not allow it.

repeatedly: Why?

sonots: I thinks typical jsonpath parser does not accept this. At least https://github.com/json-path/JsonPath did not accept this.

repeatedly: It seems javascript object limitation. javascript object field can't contain space.
But fluentd target is record, so there is no such limitation.

sonots: I am against to extend specifications because it makes difficult to refactor implementation later (for example, think of changing jsonpath parser to 3rd party implementation later) without breaking backward compatibility.

repeatedly: We don't have a plan to use 3rd party jsonpath parser because this is just jsonpath like, not jsonpath. jsonpath has lots of useless feature in fluentd, e.g.g filter, wildcard and etc.
So we want limited well-known syntax, not entire jsonpath.

repeatedly: But limiting several characters in dot notation is not bad idea because bracket notation is straight way for such cases.

mururu:
`$.hello.[0]` is parsed as `["hello", "[0]", 0]`. It should be parsed as `["hello", 0]`, `["hello", "[0]"]` or raise error.
https://github.com/joshbuddy/jsonpath parses it as `["hello", 0]`. I haven't check other implementations.

mururu:
Minor comment: When parsing `$[]` or `$.hello[]`, it says `invalid value for Integer(): "]"`, but it is better say `invalid value for Integer(): ""` .

test/plugin_helper/test_record_accessor.rb
mururu: I don't strongly oppose this, but it just looks weird for me too.

repeatedly:
@mururu Good catch. I will add more validation.

repeatedly:
Applied reviews and improve the performance for single key.

PR #1642, opened 2017-07-25
Send signal to all workes

sovioletta:
The supervisor only send a signal to last worker process before..
  (problem: after logrotate, worker processes write the log into deleted file)
I changed the supervisor send signal to send all of worker processes.


repeatedly:
Thanks for the patch.
We will check it later :)

repeatedly:
Thanks!

PR #1644, opened 2017-07-26
Add <log> directive to <system>

repeatedly:
This directive controls fluentd's log behaviour
Now, support two parameters:

```
<system>
  <log>
    format json
    time_format %F
  </log>
</system>
```

- format: Change fluentd log format. text and json

text:

```
2017-07-27 07:37:02 +0900 [info]: #0 fluentd worker is now running worker=0
```

json:

```
{"time":"2017-07-27 06:44:54 +0900","level":"info","message":"fluentd worker is now running worker=0","worker_id":0}
```

- time_format: Change time format of log

`time_format %Y`

```
2017 [info]: #0 fluentd worker is now running worker=0
```

I have a plan to move some logging related parameters to under `<log>` like `log_level`.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -96,8 +96,12 @@ module Fluent
       @level = logger.level + 1
       @debug_mode = false
       @log_event_enabled = false
-      @time_format = '%Y-%m-%d %H:%M:%S %z '
       @depth_offset = 1
+      @format = nil
+      @time_format = nil
+      @formatter = nil
+
+      self.format = :text
       enable_color out.tty?
       # TODO: This variable name is unclear so we should change to better name.
       @threads_exclude_events = []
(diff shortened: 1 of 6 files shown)


lib/fluent/log.rb
```diff
@@ -96,8 +96,12 @@ def initialize(logger, opts={})
       @level = logger.level + 1
       @debug_mode = false
       @log_event_enabled = false
-      @time_format = '%Y-%m-%d %H:%M:%S %z '
       @depth_offset = 1
+      @format = nil
+      @time_format = nil
+      @formatter = nil
+
+      self.format = :text
```
mururu: Why not `@format = :text`?

mururu:
I added a very small comment. Other than that LGTM.

lib/fluent/log.rb
repeatedly: Need to set other default values: https://github.com/fluent/fluentd/pull/1644/files#diff-d845866477e75d27b77c4b8acc63c7d9R163

mururu: I see.

lib/fluent/log.rb
```diff
@@ -426,7 +491,7 @@ def caller_line(type, time, depth, level)
                        else
                          "".freeze
                        end
-      log_msg = "#{time.strftime(@time_format)}[#{LEVEL_TEXT[level]}]: #{worker_id_part}"
+      log_msg = "#{time.strftime(@time_format)} [#{LEVEL_TEXT[level]}]: #{worker_id_part}"
```
vvk: Can't this log pattern be customizable as well?
For example, it may be necessary to have patten like `[time][log_level] message`.

It could be a customizable part of `text` format or a new format, let's say `custom`.

vvk:
Will there be any documentation updates? Unfortunately, I'm not deeply familiar with the repository's structure, but I didn't where the docs are stored.

lib/fluent/log.rb
repeatedly: > For example, it may be necessary to have patten like [time][log_level] message to parse it easily.

For easy to parse, `json` is enough.

repeatedly:
Visit fluentd-docs repository.

PR #1648, opened 2017-07-31
condition bug in compat/output.rb

drshin81:
hi 
I've just found little condition check bug in compat/output.rb
Would you check it?


repeatedly:
Could you add test?

PR #1662, opened 2017-08-10
output: Support record_accessor in buffer's chunk keys

repeatedly:
This patch add record_accessor's dot-notation support in chunk keys.
To support bracket-notation, we need to consider relax validation in many places.
So it is the future work.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -15,6 +15,7 @@
 #
 
 require 'fluent/plugin/base'
+require 'fluent/plugin_helper/record_accessor'
 require 'fluent/log'
 require 'fluent/plugin_id'
 require 'fluent/plugin_helper'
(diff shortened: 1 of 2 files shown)


repeatedly:
@mururu Could you check this?

lib/fluent/plugin/output.rb
```diff
@@ -36,7 +37,7 @@ class Output < Base
       helpers_internal :thread, :retry_state
 
       CHUNK_KEY_PATTERN = /^[-_.@a-zA-Z0-9]+$/
-      CHUNK_KEY_PLACEHOLDER_PATTERN = /\$\{[-_.@a-zA-Z0-9]+\}/
+      CHUNK_KEY_PLACEHOLDER_PATTERN = /\$\{[-_.@$a-zA-Z0-9]+\}/
```
mururu: Why do we have to add `$`?

repeatedly: Without `$`, we can't put `$` inside placeholders like `${$.nested.key}`.

mururu: I got it.

PR #1663, opened 2017-08-15
plugin: Improve flter_record_transformer

### LINKED ISSUE #1240: record_transformer keep_keys option creates keys with 'nil' values for unset keys.

I saw this on 0.12 but it looks like it still works the same way.

If you configure i.e.

```
<filter XXX>
  type record_transformer
  renew_record true
  keep_keys a,b,c
</filter>
```

Then the resulting record will always have 'a', 'b', and 'c' set, even if those keys were not present in the original record (in which case they get nil values).

instead of
`@keep_keys.each {|k| new_record[k] = record[k]} if @keep_keys and @renew_record`

should we do?
`@keep_keys.each {|k| new_record[k] = record[k] if record.has_key?(k)} if @keep_keys and @renew_record`

BTW, perhaps this should be a different issue but we also found it surprising/confusing that if you write
`keep_keys a, b, c`
those spaces get into the actual keys which are then named "a", " b", and " c".

yurano:
- Change types of `keep_keys` and `remove_keys` string into array
- Don't create new keys if the original record doesn't have keys specified `keep_keys` option

fix #1240

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -28,9 +28,9 @@ module Fluent::Plugin
     Fluent::Plugin.register_filter('record_transformer', self)
 
     desc 'A comma-delimited list of keys to delete.'
-    config_param :remove_keys, :string, default: nil
+    config_param :remove_keys, :array, default: nil
     desc 'A comma-delimited list of keys to keep.'
-    config_param :keep_keys, :string, default: nil
+    config_param :keep_keys, :array, default: nil
     desc 'Create new Hash to transform incoming data'
     config_param :renew_record, :bool, default: false
     desc 'Specify field name of the record to overwrite the time of events. Its value must be unix time.'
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -129,7 +124,9 @@ def reform(record, placeholder_values)
       placeholders = @placeholder_expander.prepare_placeholders(placeholder_values)
 
       new_record = @renew_record ? {} : record.dup
-      @keep_keys.each {|k| new_record[k] = record[k]} if @keep_keys and @renew_record
+      @keep_keys.each do |k|
+        new_record[k] = record[k] if record.has_key?(k)
+      end if @keep_keys and @renew_record
```
repeatedly: Could you use `&&` instead of `and`?
This is not related with this patch but we want to unify condition style.

yurano: I will take care of it soon.

yurano:
Please note the follows:

I'm going to change types of `keep_keys` and `remove_keys` option.
Therefore, I think fixing [this](https://docs.fluentd.org/v0.12/articles/filter_record_transformer#keepkeys-optional-string-type) document is also needed if this PR is accepted.

I couldn't find docs in this repo, so that I didn't take care of it.

repeatedly:
docs repository is here: https://github.com/fluent/fluentd-docs

yurano:
@repeatedly 
> docs repository is here: https://github.com/fluent/fluentd-docs

Thanks.
And I've just found in the contribution guide, sorry.

I'll open an issue and send PR for this problem when this PR will be merged.

repeatedly:
Thanks!

PR #1666, opened 2017-08-19
formatter_ltsv: suppress delimiters in output

MikePlayle:
Replace tabs with spaces before writing them to the output, to prevent
creating malformed LTSV files.

Also add test cases to verify this behaviour.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -166,6 +166,7 @@ module Fluent::Plugin
 
       @usock = nil
       @keep_alive_watcher_interval = 5 # TODO
+      @suspend_flush = false
     end
 
     def configure(conf)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/formatter_ltsv.rb
```diff
@@ -32,7 +32,7 @@ def format(tag, time, record)
         formatted = ""
         record.each do |label, value|
           formatted << @delimiter if formatted.length.nonzero?
-          formatted << "#{label}#{@label_delimiter}#{value}"
+          formatted << "#{label}#{@label_delimiter}#{value.to_s().gsub(@delimiter, ' ')}"
```
repeatedly: Replaced char should be configurable, not only `' '`.

repeatedly: And `to_s` is better than `to_s()`

kenhys:
@MikePlayle do you have still motivation to improve this PR?

MikePlayle:
I forgot about this completely. I thought it had been merged! I'll get to it this weekend.

ashie:
Although I felt a doubt a bit about the specification, probably it's a reasonable solution when I consider it based on philosophy of LTSV (parser should be as simple as possible). I'll merge it after checking the tests results on my environment.

ashie (approved):

ashie:
Sorry I wrongly pushed master branch.
I'll open new PR with revised patch.

PR #1671, opened 2017-08-25
Support pretty print of Fluent::Config::Element

k0kubun:
## Problem
With fluentd v0.14.20, when debugging fluentd config parser with pry, I got following output for non-empty `Fluent::Config::Element`.

screen shot 2017-08-25 at 18 20 23

I think the pretty print result (`{}`) is not so pretty and it makes it hard to debug.

## Changes
I changed `Fluent::Config::Element` to pretty print the same string as `#inspect`.

screen shot 2017-08-25 at 18 22 55


repeatedly:
LGTM. Thanks!

Issue #1674, opened 2017-08-29
fluentd gives unexcepted error when sending SIGHUP signal


mikkohoo:
Following scenario was tested with official 0.14.19 and 0.14.20 fluentd docker images.

**Description:**
We have used SIGHUP signal to reload configuration each time our configuration changes. This has been working perfectly in our use case until 0.14.20. Apparently something has been changed in 0.14.20 since now we get following error each time we send SIGHUP signal to fluentd process.
```
2017-08-29 06:16:25 +0000 [error]: #0 unexpected error error_class=SignalException error="SIGHUP"
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/engine.rb:228:in `sleep'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/engine.rb:228:in `run'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/supervisor.rb:772:in `run_engine'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/supervisor.rb:521:in `block in run_worker'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/supervisor.rb:697:in `main_process'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/supervisor.rb:516:in `run_worker'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/fluentd-0.14.20/lib/fluent/command/fluentd.rb:316:in `<top (required)>'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
  2017-08-29 06:16:25 +0000 [error]: #0 /usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #1683, opened 2017-09-05
supervisor: Clear previous worker pids when receives kill signals. ref #1674

repeatedly:
Without this patch, sending multiple signals, e.g. call `kill -HUP pid` twice, causes "No such process"


repeatedly:
@mururu Could you check?

Issue #1665, opened 2017-08-18
fluentd-sender(0.14.20) become cpu 100% after fluentd-server down


bungoume:
After the connection to the server is lost, the fluentd-sender CPU usage become 100%.
Even if the server is restored, CPU still sticks to max.

fluentd(sender) version: 0.14.20

![image](https://user-images.githubusercontent.com/507733/29460393-cb4b1da6-8462-11e7-9776-a378ece76437.png)
21:36 - server down
21:38 - server reovered

## fluentd sender config
https://github.com/bungoume/log-sender/blob/master/fluent.tmpl#L147-L168

repeatedly:
Do you have reproducible step?
I tested destination shutdown on my Mac but forwarder's CPU is lower...

GLStephen:
I've seen this behavior as well. The target server for a forwarder has a blip and then the sending server spikes and stays high CPU.

repeatedly:
@GLStephen Could you tell me your setting and traffic?
I can't reproduce on my environment so I want to reproduce the problem first.

bungoume:
It was reproduced when AWS CLB(ELB) was placed in front of the aggregator(server).

log sender -> ELB -> aggregator

when ELB->aggs connection was closed, 
sender seems to misunderstand that aggregator server has downed and recovered.
And, sender CPU will be 100%.

```
2017-08-28 10:03:15 +0000 [warn]: #0 [forward_any] destination node closed the connection. regard it as unavailable. host="log-aggregator.local" port=24224
2017-08-28 10:03:16 +0000 [warn]: #0 [forward_any] recovered forwarding server '' host="log-aggregator.local" port=24224
2017-08-28 10:03:16 +0000 [warn]: #0 [forward_any] destination node closed the connection. regard it as unavailable. host="log-aggregator.local" port=24224
2017-08-28 10:03:16 +0000 [warn]: #0 [forward_any] failed to flush the buffer. retry_time=1 next_retry_seconds=2017-08-28 10:03:17 +0000 chunk="557cd66b41ec2fe08ea752ca6ee2d50e" error_class=Fluent::Plugin::ForwardOutput::ConnectionClosedError error="failed to establish connection with node "
  2017-08-28 10:03:16 +0000 [warn]: #0 /usr/lib/ruby/gems/2.4.0/gems/fluentd-0.14.20/lib/fluent/plugin/out_forward.rb:606:in `send_data_actual'
[...]

PR #1684, opened 2017-09-06
Sleep if readable sockets are empty

okkez:
This can reduce CPU usage following case:

    [Sender] -(1)-> [ELB] -(2)-> [Aggregator]

Sender enables `require_ack_response`.

If (2) is disconnected, but sender send logs to ELB and wait for ack responses. Soeckets that are waiting for ack response cannot be readable for long time.

This will cause busy loop [here](https://github.com/fluent/fluentd/blob/v0.14.20/lib/fluent/plugin/out_forward.rb#L472-L506).

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -493,7 +493,10 @@ module Fluent::Plugin
           end
 
           readable_sockets, _, _ = IO.select(sockets, nil, nil, select_interval)
-          next unless readable_sockets
+          unless readable_sockets
+            sleep 0.01
+            next
+          end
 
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -493,7 +493,10 @@ def ack_reader
           end
 
           readable_sockets, _, _ = IO.select(sockets, nil, nil, select_interval)
-          next unless readable_sockets
+          unless readable_sockets
```
repeatedly: Could you add comment why this sleep is needed?

okkez:
I noticed that this PR cannot fix #1665. 🙇

PR #1686, opened 2017-09-06
Don't update retry state when failed to get ack response. fix #1665

### LINKED ISSUE #1665: fluentd-sender(0.14.20) become cpu 100% after fluentd-server down

repeatedly:
Ack failure is not buffer flush error so it should not update retry state.
It create invalid retry state and this causes busy loop.
The problematic node is disabled and its node is excluded at next flush after rollback.
So non-acked chunk should be retried next flush.

To resolve this problem, we add update_retry parameter to Output#rollback_write.
This parameter controls rollback considers retry or not.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -433,7 +433,7 @@ module Fluent::Plugin
       if raw_data.empty?
         log.warn "destination node closed the connection. regard it as unavailable.", host: info.node.host, port: info.node.port
         info.node.disable!
-        rollback_write(info.chunk_id, false)
+        rollback_write(info.chunk_id, update_retry: false)
         return nil
       else
         unpacker.feed(raw_data)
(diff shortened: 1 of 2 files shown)


mururu:
I can't fully get what this commit means yet, but I feel that retry strategies should be applied to the set of "commit_write + rollback_write" in the case of async buffered output.

repeatedly:
> I feel that retry strategies should be applied to the set of "commit_write + rollback_write" in the case of async buffered output.

Yeah, generally yes. But out_forward removes problematic node from the active node list, so immediate retry is no problem. If we retry flush with same destination, retry state should be updated.

GLStephen:
@repeatedly I'm currently testing this behavior and I will patch my 14.22.rc2 with this and report back

repeatedly:
Use keyword argument. This is more clear than before.

repeatedly:
@mururu How about this? I will consider changing retry status management but it needs more time to re-desing internal mechanzim. So this temporary fix seems acceptable for out_forward.

mururu:
@repeatedly Fair enough. Then it is good to add comments for that.

PR #1689, opened 2017-09-12
filter_grep: strip input regexp before compiling

icy:
The leading/trailing spaces are not easy to see in editor,
and it may make debugging process harder.

Allowing such spaces disallows us to format configuration file, e.g,

   exclude foo   test1
   exclude foo2  test2

Without stripping, "foo" only matches "  test1".

Now user should use \s to specify leading/trailing spaces explicitly:

   exclude bar   \s+test


icy:
@repeatedly Would you mind taking a a look why Travis fails? Thanks a lot.

repeatedly:
Hmm... regexpN/excludeN are now deprecated and this change breaks the compatibility.
I'm not sure users depend on current behaviour but change is not good for exsiting users.

If this problem happens in new `<regexp>`/`<exclude>`, patches are welcome.

icy:
Thanks @repeatedly . That makes sense. I will send another patch for new `<regexp>/<exclude>`. Stay tuned.

PR #1690, opened 2017-09-12
Delay time parser initialization

okkez:
In previous version,
`Fluent::Plugin::Apache2Parser.new` creates `@time_parser`
but `Fluent::Plugin::Parser#configure` overwrites it with `time_parser_create`.

In this version, `Fluent::Plugin::Apache2Parser#configure` creates
`@time_parser` proper
sequence. `Fluent::Plugin::Apache2Parser#configure` invoked after `Fluent::Plugin::Parser#configure`.


repeatedly:
Thanks!

PR #1691, opened 2017-09-12
Add support to enable writing to new lines for tsv formatter

ryysud:
I found that writing to new line is impossible with tsv formatter.
So, I add `add_newline` parameter to control writing to new line by tsv formatter.

### DIFF
--- a/lib/fluent/plugin/formatter_tsv.rb
+++ b/lib/fluent/plugin/formatter_tsv.rb
@@ -27,7 +27,7 @@ module Fluent
       config_param :delimiter, :string, default: "\t"
 
       def format(tag, time, record)
-        @keys.map{|k| record[k].to_s }.join(@delimiter)
+        @keys.map{|k| record[k].to_s }.join(@delimiter) << "\n".freeze
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/formatter_tsv.rb
```diff
@@ -25,9 +25,14 @@ class TSVFormatter < Formatter
       config_param :keys, :array, value_type: :string
       desc 'The delimiter character (or string) of TSV values'
       config_param :delimiter, :string, default: "\t"
+      desc 'The parameter to enable writing to new lines'
+      config_param :add_newline, :bool, default: true
 
       def format(tag, time, record)
-        @keys.map{|k| record[k].to_s }.join(@delimiter)
+        formatted = ""
+        formatted << @keys.map{|k| record[k].to_s }.join(@delimiter)
```
repeatedly: `formatted = @keys.map{|k| record[k].to_s }.join(@delimiter)` seems enough

ryysud: @repeatedly Thanks. I fix it.

PR #1692, opened 2017-09-12
Fix test_config_params of ltsv formatter

ryysud:
Source code
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/formatter_ltsv.rb#L26-L28

Document
https://docs.fluentd.org/v0.14/articles/formatter_ltsv#addnewline-boolean-optional-defaults-to-true


repeatedly:
Thx!

PR #1700, opened 2017-09-26
Update url from http to https

ryysud:
I updated url of the following domain from http to https.
- www.fluentd.org
- docs.fluentd.org
- kafka.apache.org
- www.splunk.com
- cncf.io
- fluentd.org
- github.com
- groups.google.com
- slack.fluentd.org
- docs.ruby-lang.org
- test-unit.github.io


repeatedly:
Thanks!

PR #1708, opened 2017-10-04
out_file/out_secondary_file: Support ${chunk_id} placeholder. fix #1705

### LINKED ISSUE #1705: No Way to tie Failed Buffer Chunks to Chunk ID
- V14
- CentOS
- Forwarder web servers, to aggregators

FluentD log files refer to chunk_id when it is emitting errors and warning. Sometimes of these lead to failed chunks. Those failed chunks can be put in a secondary location. However, the ${chunk_id} is not available in the failed log and the meta file is not in the failed folder so this chunk_id can't be used to figure out which specific failed buffer file an error relates to. 

Remedy: file & secondary_file should support something like ${chunk_id} placeholder so that the chunk_id can be retained.

repeatedly:
`${chunk_id}` placeholder is useful for out_file like plugins.
So I added `CHUNK_ID_PLACEHOLDER_PATTERN` in Output to share.
For example, out_webhdfs uses own `${chunk_id}` value but it should use this predefined constant.

### DIFF
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -142,7 +142,7 @@ module Fluent::Plugin
         dummy_record = Hash[dummy_record_keys.zip(['data'] * dummy_record_keys.size)]
 
         test_meta1 = metadata_for_test(dummy_tag, Fluent::Engine.now, dummy_record)
-        test_path = extract_placeholders(@path_template, test_meta1)
+        test_path = extract_placeholders(@path_template.gsub(CHUNK_ID_PLACEHOLDER_PATTERN, 'test'), test_meta1)
         unless ::Fluent::FileUtil.writable_p?(test_path)
           raise Fluent::ConfigError, "out_file: `#{test_path}` is not writable"
         end
@@ -178,7 +178,7 @@ module Fluent::Plugin
     end
 
     def write(chunk)
-      path = extract_placeholders(@path_template, chunk.metadata)
+      path = extract_placeholders(@path_template.gsub(CHUNK_ID_PLACEHOLDER_PATTERN, dump_unique_id_hex(chunk.unique_id)), chunk.metadata)
       FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
       writer = case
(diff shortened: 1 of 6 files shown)


lib/fluent/plugin/out_file.rb
```diff
@@ -178,7 +178,7 @@ def format(tag, time, record)
     end
 
     def write(chunk)
-      path = extract_placeholders(@path_template, chunk.metadata)
+      path = extract_placeholders(@path_template.gsub(CHUNK_ID_PLACEHOLDER_PATTERN, dump_unique_id_hex(chunk.unique_id)), chunk.metadata)
```
mururu: It seems better that the replacements of `CHUNK_ID_PLACEHOLDER_PATTERN` should be done by `extract_placeholders` instead of each plugins. For now, only these two plugins use this method, but it is useful if all plugins can use this feature automatically in the future.

repeatedly: Yeah, it is one idea.
For it, we need to change `extract_placeholders` API.

1. 2nd argument changed to chunk, not `chunk.metadata`

For compatibility, we need to check 2nd argument type.

```
metadata = chunk.metadata if metadata.is_a?(Chunk)
```

2. Add 3rd argument for chunk

`def extract_placeholders(str, metadata, chunk = nil)`

`metadata` is from `chunk` so this API is a little strange.

3. Add 3rd argument for additional params

This avoids further API changes.

```
def extract_placeholders(str, metadata, extras = nil)
  if extras && extras.has_key?(:chunk)
  end
end
```

I think 1 is better because this doesn't break exsiting API and doesn't add more argument.
We can change Plugin API before stable announcement.

mururu: 👍 > 1

repeatedly:
@okkez @cosmo0920 How about above change? Do you have other opinion?

lib/fluent/plugin/out_file.rb
cosmo0920: I agree to change `extract_placeholders` API.
:+1: > 1 and 3. The both looks good to me.

okkez: :+1:  > 1

repeatedly:
Change `extract_placeholders` API. PTAL!

PR #1718, opened 2017-10-16
log when a new file chunk has been created

tpetr:
This is useful for debugging because it makes it easier to associate chunks with their corresponding metadata in the fluentd logs.


repeatedly:
LGTM

PR #1720, opened 2017-10-19
Check pos_file duplication between in_tail sources

repeatedly:
pos_file duplication causes broken state.

In addition, move `buffer_path_for_test?` to Base class to share the routine.


repeatedly:
@mururu could you check this?

mururu:
looks good

PR #1721, opened 2017-10-20
Always lock chunks first to avoid deadlock

abicky:
This PR resolves https://github.com/fluent/fluentd/issues/1549.

For example, deadlock occurs by the following steps before this commit:

1. input plugin thread receives multiple events
  (`metadata_and_data.size > 1`)
2. input plugin thread processes the first `metadata_and_data` element
  and aquires the lock of chunk1 (`chunk.mon_exit` in `Buffer#write`)
3. enqueue thread aquires the lock of buffer
  (`synchronize` in `Buffer#enqueue_chunk`)
4. enqueue thread tries to aquire the lock of chunk1
  (`chunk.synchronize` in `Buffer#enqueue_chunk`)
5. input plugin thread processes the second `metadata_and_data` element
  and tries to aquire the lock of buffer (`synchronize` in `Buffer#write_once`)

cf. https://github.com/fluent/fluentd/issues/1549#issuecomment-335736547

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -323,7 +323,7 @@ module Fluent
             else
               @queue << chunk
               @queued_num[metadata] = @queued_num.fetch(metadata, 0) + 1
-              chunk.enqueued! if chunk.respond_to?(:enqueued!)
+              chunk.enqueued!
             end
           end
           bytesize = chunk.bytesize
@@ -340,7 +340,7 @@ module Fluent
             metadata = chunk.metadata
             @queue << chunk
             @queued_num[metadata] = @queued_num.fetch(metadata, 0) + 1
-            chunk.enqueued! if chunk.respond_to?(:enqueued!)
+            chunk.enqueued!
           end
           @queue_size += chunk.bytesize
         end
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -319,20 +319,18 @@ module Fluent
         return nil unless chunk
 
         chunk.synchronize do
-          if chunk.empty?
-            chunk.close
-          else
-            synchronize do
+          synchronize do
+            if chunk.empty?
+              chunk.close
+            else
               @queue << chunk
               @queued_num[metadata] = @queued_num.fetch(metadata, 0) + 1
+              chunk.enqueued!
             end
-            chunk.enqueued!
+            bytesize = chunk.bytesize
+            @stage_size -= bytesize
+            @queue_size += bytesize
           end
-        end
-        bytesize = chunk.bytesize
-        synchronize do
-          @stage_size -= bytesize
-          @queue_size += bytesize
         end
         nil
       end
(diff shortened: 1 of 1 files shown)


abicky:
@repeatedly Could you review the change? I want to try the latest version (v0.14.22.rc2?) which includes this change in our production environment if it looks good to you.

repeatedly:
Thanks! I will review it!

abicky:
I found there were errors like `unexpected error while checking flushed chunks. ignored. error_class=RuntimeError error="can't enqueue buffer file: path = /fluentd/log/..., error = 'closed stream'"` and I'll investigate the cause 🙇

abicky:
I fixed "closed stream" error in  https://github.com/fluent/fluentd/pull/1721/commits/1a6f16b4c2c06060c65837c1167d055ec93bded5.

@repeatedly Please review. I've checked the behavior in our development environment and it works fine. I want your review before I try it in our production environment.
Some tests failed in Travis CI, but these tests seem to sometimes fails also in master branch, aren't they?

lib/fluent/plugin/buffer.rb
```diff
@@ -340,25 +342,24 @@ def enqueue_unstaged_chunk(chunk)
...
-              v = yield metadata, chunk
-              enqueue_chunk(metadata) if v
-            end
-          else
-            @stage.keys.each do |metadata|
-              enqueue_chunk(metadata)
-            end
+        if block_given?
+          synchronize{ @stage.keys }.each do |metadata|
+            chunk = @stage[metadata]
```
repeatedly: No need `synchronized` guard here?

repeatedly: Hmm... deleting metadata from `@stage` happens in only `enqueue_chunk` so it seems safe.

abicky: I think data race doesn't occur in  CRuby thanks to GIL and `synchronized` is not necessary, but I'll change the code like below for other Ruby implementations:

```diff
diff --git a/lib/fluent/plugin/buffer.rb b/lib/fluent/plugin/buffer.rb
index 337b3b79..61a904ed 100644
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -351,11 +351,15 @@ module Fluent
       def enqueue_all
         log.trace "enqueueing all chunks in buffer", instance: self.object_id
         if block_given?
-          synchronize{ @stage.keys }.each do |metadata|
-            chunk = @stage[metadata]
-            next unless chunk
-            v = yield metadata, chunk
-            enqueue_chunk(metadata) if v
+          metadata_array = []
+          synchronize do
+            @stage.each do |metadata, chunk|
+              v = yield metadata, chunk
+              metadata_array << metadata if v
+            end
+          end
+          metadata_array.each do |metadata|
+            enqueue_chunk(metadata)
           end
         else
           synchronize{ @stage.keys }.each do |metadata|
```

repeatedly: Yes but currently fluentd focus on CRuby so I think code change is not needed for now.
But added comment is better for the future.

abicky: Okey, I added a comment https://github.com/fluent/fluentd/pull/1721/commits/2261428eab806a061e6e39824f6da125100386f1

PR #1722, opened 2017-10-21
Lock buffers in order of metadata

tagomoris:
When 2 or more input plugin threads try to emit events into an output plugin (and its buffer plugin instance), it may occur dead-lock, because the order of chunks to be locked are random for now.

This change introduces to sort metadata before locking chunks, to lock chunks in same order in different threads.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -72,14 +72,16 @@ module Fluent
             return 1
           end
           # both of v1 and v2 are non-nil
-          if v1.keys.sort != v2.keys.sort
-            if v1.keys.size == v2.keys.size
-              v1.keys.sort <=> v2.keys.sort
+          v1_sorted_keys = v1.keys.sort
+          v2_sorted_keys = v2.keys.sort
+          if v1_sorted_keys != v2_sorted_keys
+            if v1_sorted_keys.size == v2_sorted_keys.size
+              v1_sorted_keys <=> v2_sorted_keys
             else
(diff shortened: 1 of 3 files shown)


tagomoris:

lib/fluent/plugin/buffer.rb
```diff
@@ -62,6 +62,72 @@ class BufferChunkOverflowError < BufferError; end # A record size is larger than
...
+        def cmp_variables(v1, v2)
+          if v1.nil? && v2.nil?
+            return 0
+          elsif v1.nil? # v2 is non-nil
+            return -1
+          elsif v2.nil? # v1 is non-nil
+            return 1
+          end
+          # both of v1 and v2 are non-nil
+          if v1.keys.sort != v2.keys.sort
```
repeatedly: `v1.keys` and `v2.keys` should be assigned to local variables because each `keys` call creates new object.
metadata comparison is called heavily in buffers so reducing object allocation is better.

repeatedly:
BTW... with this patch, https://github.com/fluent/fluentd/blob/master/test/plugin/test_buffer.rb#L861 this assertion always fails.

lib/fluent/plugin/buffer.rb
tagomoris: I see. Will fix it.

tagomoris:
I've pushed commits to fix performance issue and existing tests.

repeatedly:
Thx!

PR #1729, opened 2017-10-27
Issue 1726: Fix for in_tcp log corruption under load.

AM-iain:
The TCP input plugin shares a single buffer across all connections.  Under load
the buffer sometimes gets truncated.  The suspicion is that concurrent
connections race to modify it.

Long message is received.
Short message is received while the long message is still being parsed.
The short message is parsed so the buffer will be truncated.
The buffer, which now contains "longmessage\nshort\n", is truncated by the
length of the short message.  It is now "longm".
Another message arrives.  Now the buffer is "longmanother\n", which does not
parse.
The whole buffer is thrown away and subsequent messages are received and
handled as usual.
Eventually the pattern repeats.

The fix is to use a per-connection buffer.

### DIFF
--- a/lib/fluent/plugin/in_tcp.rb
+++ b/lib/fluent/plugin/in_tcp.rb
@@ -55,13 +55,12 @@ module Fluent::Plugin
     def start
       super
 
-      @buffer = ''
       server_create(:in_tcp_server, @port, bind: @bind) do |data, conn|
-        @buffer << data
+        conn.buffer << data
         begin
           pos = 0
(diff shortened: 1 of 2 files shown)


repeatedly:
Reporter confirmed this patch fixed the issue. Thanks!

sampointer:
It would be a great help if this could be taged as `0.14.23`

sampointer:
I am a colleague of @AM-iain. We've tested this on our infrastructure successfully. `0.14.23` would be a great help.

PR #1733, opened 2017-11-01
filter_parser: Fix dumpped result for avoiding escape sequence injection

repeatedly:
Escape sequence is affected when user launches fluentd without log option in the console.

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -53,7 +53,7 @@ class Fluent::ParserFilter < Fluent::Filter
             if @emit_invalid_record_to_error
               router.emit_error_event(tag, time, record, ::Fluent::ParserError.new("pattern not match with data '#{raw_value}'"))
             else
-              log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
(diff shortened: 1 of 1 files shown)


MarwanBayadsi:
hi,
In which version of td-agent i can get this fix ?

repeatedly:
Yes, latest td-agent 2 doesn't include this fix yet.
We don't have new td-agent 2 plan for this patch because td-agent use-case doesn't have this problem for now.

chaosoldier:
hi,

Please tell me when will this fixed version (0.12.41) be available on the yum rpm repository?
https://github.com/fluent/fluentd/blob/v0.12/CHANGELOG.md#bug-fixes
https://td-agent-package-browser.herokuapp.com/2/redhat/6/x86_64

> use-case doesn't have this problem for now.
Is this comment means the injenction will be safe for all use-case?? 
https://docs.treasuredata.com/articles/td-agent-changelog

repeatedly:
> Is this comment means the injenction will be safe for all use-case??

If you use td-agent with default way, yes. If you use fluentd directly in td-agent and no log / daemonize option, this problem happens.
If you want to use latest version in existing td-agent, update fluentd to latest is one way.

chaosoldier:
thank you for answering my question.

PR #1738, opened 2017-11-06
Move tick check to after heartbeat for avoiding the impact of other node heartbeat. fix #1719

### LINKED ISSUE #1719: Unexpected Multiple Server Behavior
FluentD v14 - Technically google-fluentd, but observed in td-agent as well.
CentOS

```
<match tag.**>
  @type forward
<server>
    name fluentd-groupb-1
    host x.y.z.w
    port 24228
    weight 60
  </server>

  <server>
    host x.y.z.w
    port 24228
    weight 60
  </server>

  <server>
    host x.y.z.w
    port 24228
    weight 60
  </server>
.... other stuff ....
</match>
```

Problem: In this multiple server forwarding config if I take down any of the machines they will all eventually be marked as "no nodes available" even though two of them are available. Am I misunderstanding this multi-server configuration option?

repeatedly:
In the past, fluentd supports only UDP heartbeat and it uses async check between fluentd nodes.
So if previous node takes longer time for heartbeat, it doesn't affect elapsed time.
On the other hand, recent fluentd supports TCP heartbeat and it uses sync check via tcp connection.
So if previous node takes longer time, it affects elapsed time for current node and it causes misjudgment.
To avoid the problem, elapsed time calculation should be done after heartbeat.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -395,9 +395,6 @@ module Fluent::Plugin
 
     def on_timer
       @nodes.each {|n|
-        if n.tick
-          rebuild_weight_array
-        end
         begin
           log.trace "sending heartbeat", host: n.host, port: n.port, heartbeat_type: @heartbeat_type
           n.usock = @usock if @usock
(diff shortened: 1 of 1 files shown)


repeatedly:
@mururu Could you check this change?

mururu:
Looks good.

Issue #1713, opened 2017-10-09
 error="bignum too big to convert into `unsigned long long'"


kkniffin:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

fluentd 0.14.21
Linux

I am having an issue where data is being sent to FluentD that throws an error and the system is unrecoverable and logs no longer flow until I stop and restart the service again. I am unsure what specific data message is causing the issue and also why a message would cause the system to no longer respond to future messages and simply fail. I would think that it would error on just that one message but continue to process future ones?

The error I eventually receive is: 

```
2017-10-09 14:13:57 +0000 [warn]: #0 fluent/event_router.rb:87:emit: suppressed same stacktrace
fluentd-nxlog_1   | 2017-10-09 14:13:57 +0000 [warn]: #0 fluent/log.rb:336:warn: emit transaction failed: error_class=RangeError error="bignum too big to convert into `unsigned long long'" tag="nxlog"
fluentd-nxlog_1   |   2017-10-09 14:13:57 +0000 [warn]: #0 fluent/event_router.rb:87:emit: suppressed same stacktrace
fluentd-nxlog_1   | 2017-10-09 14:13:57 +0000 [error]: #0 fluent/log.rb:356:error: unexpected error on reading data host="x.x.x.x" port=60051 error_class=RangeError error="bignum too big to convert into `unsigned long long'"
```

This is the Fluentd Config I have:
```
<system>
  log_level error
</system>


####################
###### SOURCES #####
####################

<source>
        @type tcp
        format json
        port 5140
        tag nxlog
</source>


#####################
##### MANIPULATE ####
#####################

#### Add Server Received Time to Records
#<filter nxlog>
#       @type record_modifier
#       <record>
#               FluentDReceived ${Time.at(time).to_s}
#       </record>
#</filter>

######################################
##### Beat Processing ################
######################################

<match nxlog>
        @type rewrite_tag_filter
        rewriterule1 NXLogFileType ^dhcp$ nxlog.tagged.dhcp
        rewriterule2 NXLogFileType ^nps$ nxlog.tagged.nps
        rewriterule3 NXLogFileType ^dns$ nxlog.tagged.dns
        rewriterule4 NXLogFileType ^wineventlog$ nxlog.tagged.wineventlog
        rewriterule5 NXLogFileType ^iis$ nxlog.tagged.iis
[...]

Issue #1696, opened 2017-09-18
fluentd crashes with "Too many open files"


reimarstier:
- fluentd-0.14.21
- raspbian, log files reside on a cifs (samba) mount
- configuration reads lines with in_tail plugin and writes to kafka
- fluentd is reading fast rolling logs and sending them to a kafka cluster

After some time fluentd crashes with an error "Too many open files". This problem only arises in connection with fast rolling log files (sometimes a file only exists for about 10 seconds - hence the reason why fluentd was necessary to trace those logs). Other applications without the fast rolling log behavior do not have this problem. Increasing the hard/soft limits of the number of open files in /etc/security/limits.conf simply prolongs the first incident of this problem.

It seems like fluentd is not able to close those files in time to keep reading at that pace. Maybe this is not an issue as long there are not that many files to read. 

Fluentd process quits with following stacktrace:
```
2017-09-14 16:00:08 +0200 [info]: #0 following tail of /srv/cifs/Application.txt
2017-09-14 16:00:08 +0200 [error]: #0 bignum too big to convert into `long long'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:691:in `seek'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:691:in `open'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:707:in `with_io'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:645:in `on_notify'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:477:in `detach'
  2017-09-14 16:00:08 +0200 [error]: #0 /var/lib/gems/2.3.0/gems/fluentd-0.14.21/lib/fluent/plugin/in_tail.rb:306:in `detach_watcher'
[...]

PR #1742, opened 2017-11-09
in_tail: Skip setup failed watcher to avoid resource leak and log bloat. ref #1713

repeatedly:
This is temporal solution and to log the important information.

In addition, introduce `WatcherSetupError` to unify setup error.
If we have other setup failed cause, we should use this error instead of adhoc logging/error.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -34,6 +34,16 @@ module Fluent::Plugin
 
     helpers :timer, :event_loop, :parser, :compat_parameters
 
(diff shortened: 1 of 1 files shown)


repeatedly:

mururu:
@repeatedly I can't see why in_tail is related to #1713. It doesn't use in_tail. Could you explain it?

repeatedly:
Ah... sorry. The issue number is incorrect.
https://github.com/fluent/fluentd/issues/1696 is correct.
Will change the commit message.

mururu:
I see.

PR #1751, opened 2017-11-15
Consider hyphen and underscore in fluent-plugin-generate arguments

okkez:
For example:

Following commands will generate fluent-plugin-rewrite-tag-filter
and register "rewrite_tag_filter".

    $ fluent-plugin-generate output rewrite_tag_filter
    $ fluent-plugin-generate output rewrite-tag-filter

### DIFF
--- a/lib/fluent/command/plugin_generator.rb
+++ b/lib/fluent/command/plugin_generator.rb
@@ -137,32 +137,36 @@ BANNER
   end
 
   def gem_name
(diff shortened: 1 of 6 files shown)


okkez:
Addtional information, in previous version,

```
$ fluent-plugin-generate output rewrite-tag-filter
$ fluent-plugin-generate output rewrite_tag_filter
```

These command will generate broken class name.

PR #1754, opened 2017-11-16
Use singleton_class method instead of instead of legacy trick

okkez:
This makes easy to understand code.


repeatedly:
okay. Thanks!

PR #1759, opened 2017-11-22
Cleanup server helper

repeatedly:
- Add error handling
- Reduce memory allocation
- Reuse built-in buffer

### DIFF
--- a/lib/fluent/plugin_helper/server.rb
+++ b/lib/fluent/plugin_helper/server.rb
@@ -370,6 +370,7 @@ module Fluent
         sock
       end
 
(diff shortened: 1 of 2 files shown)


repeatedly:
@mururu Could you check this?

lib/fluent/plugin_helper/server.rb
```diff
@@ -370,6 +370,8 @@ def server_create_udp_socket(shared, bind, port)
         sock
       end
 
+      PEERADDR_FAILED = ["?", "?", "name resolusion failed", "?"]
```
mururu: Very minor comment: remote_port usually returns Integer, but it returns String when failed. So it seems better to have comment about that it may return string at remote_port definition.

PR #1763, opened 2017-11-27
Combine before_shutdown and shutdown call in one sequence.

repeatedly:
Currently, Output's flush_at_shutdown is executed in before_shutdown but
in_tail's multiline buffered logs are flushed in shutdown.
This mismatch causes log lost at shutdown/restart.
To avoid such problem, Input#shutdown should be called before Output#before_shutdown.

If this approach doesn't resolve all cases,
calling order will be changed to v0.12 like sequence.

In addition, we will consider moving flush_at_shutdown routine to `shutdown`.

### DIFF
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -228,11 +228,18 @@ module Fluent
             Thread.current.abort_on_exception = true
             begin
               if method == :shutdown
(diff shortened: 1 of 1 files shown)


mururu:
We can remove this line: https://github.com/fluent/fluentd/pull/1763/files#diff-1cf8bebe1ebcb9cb1a30045a003f491dR215

repeatedly:
Right, removed.

PR #1770, opened 2017-12-04
out_copy: support ignore_error in store section

repeatedly:
out_copy_ex supports `ignore_error` to ignore emit errors in non-important plugin.
First, we provide this feature via 3rd party plugin because we can't judge this feature is useful or not before.
Currenlty, many users still use copy_ex on the production so merging it into core is good.


mururu:
Looks good.

PR #1771, opened 2017-12-04
in_forward: Use exception: false if available to improve resourse usage

### LINKED ISSUE #1767: TLS listener source causes memory leak
## Description

Forward source listener with TLS causes memory leak when TCP connection closed by FIN.

Some healthcheck system (for example: AWS NLB) use TCP check, so
memory leaks occur frequently.

This problem does not occur with normal(not TLS) listener.

## Config
Fluentd 0.14.24 (Docker fluent/fluentd:edge)
```
<source>
  @type  forward
  port  24228
  bind 0.0.0.0
  <transport tls>
    insecure true
  </transport>
</source>

<match **>
  @type null
</match>
```

## Leak
```
Client - Fluentd Server
-> Syn
<- Syn+Ack
-> Ack
-> Fin+Ack
<- Rst+Ack
```

#### reproduction command:
`nping -rate 30 -c 1000 --tcp-connect localhost -p 24228`
(With this command, Fin is sent at the last)

## Not leak
```
-> Syn
<- Syn+Ack
-> R
```
`nping -rate 30 -c 1000 --tcp localhost -p 24228`

## Similar issue
https://github.com/fluent/fluentd/issues/1678

(comment in Japanese)
https://twitter.com/bungoume/status/935780821515956225

repeatedly:
Thid doesn't resolve #1767 but this is better for avoiding non-used exception allocation.
And `exception: false` is bit faster when the number of requests is large.

### DIFF
--- a/lib/fluent/plugin_helper/server.rb
+++ b/lib/fluent/plugin_helper/server.rb
@@ -661,22 +661,39 @@ module Fluent
             end
           end
 
+          if RUBY_VERSION.to_f >= 2.3
+            NONBLOCK_ARG = {exception: false}
+            def try_handshake
+              @_handler_socket.accept_nonblock(NONBLOCK_ARG)
+            end
+          else
+            def try_handshake
+              @_handler_socket.accept_nonblock
+            rescue IO::WaitReadable
+              :wait_readable
+            rescue IO::WaitWritable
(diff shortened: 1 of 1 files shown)


mururu:
Looks good.

PR #1775, opened 2017-12-06
in_tail: Add enable_stat_watcher option to disable inotify events

repeatedly:
inotify sometimes lost the events.
We want to investigate inotify in libev causes unexpected behaviour or not.
disable parameter is useful for it.


mururu:
@repeatedly Is it okay if both `enable_stat_watcher` and `enable_watch_timer` are `false`?

repeatedly:
Good catch. I added the validation.

PR #1787, opened 2017-12-12
Add recv_buffer parameter for in_udp

mururu:
It is useful for tuning of udp input.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Release v1.0.0 - 2017/12/6
 
+See [CNCF announcment](https://www.cncf.io/blog/2017/12/06/fluentd-v1-0/) :)
+
 ### New features / Enhancements
 
 * out_copy: Support ignore_error argument in `<store>`


lib/fluent/plugin/in_udp.rb
```diff
@@ -40,6 +40,8 @@ class UdpInput < Input
     config_param :message_length_limit, :size, default: 4096
     desc "Remove newline from the end of incoming payload"
     config_param :remove_newline, :bool, default: true
+    desc "The max size of socket receive buffer. SO_RCVBUF"
+    config_param :recv_buffer, :size, default: nil
```
repeatedly: `receive_buffer_size` is better.

mururu: Okay. I just followed `recv_timeout`, but it isn't a parameter name yet.

mururu:
The test fails on CI. I'll check it.

lib/fluent/plugin/in_udp.rb
repeatedly: Yeah, public name should be clear.

repeatedly:
From man page, returned value is different from get/set: http://man7.org/linux/man-pages/man7/socket.7.html

PR #1788, opened 2017-12-12
Add receive_buffer_size parameter for in_udp for v0.12

mururu:
backport of #1787 for v0.12

PR #1789, opened 2017-12-13
Add recv_buffer parameter for in_udp

mururu:
I accidentally failed to update #1787. So I create a new one.


mururu:
@repeatedly Could you check this again? If it is okay, I'll apply the change to #1788.

PR #1791, opened 2017-12-13
in_tail: Explicitly update position to 0 when the file is truncated

mururu:
Current in_tail already reads the file from head when truncated,
but the position is updated to stat.size. Basically it is misleading and
may cause unexpected behaviour.


repeatedly:
Good catch!

PR #1792, opened 2017-12-13
 in_tail: Properly handle moved back and truncated case

mururu:
When same inode file reappear and the size is smaller than before, it should be handled as truncated. It triggers reading the file from head.


repeatedly:
LGTM.

repeatedly:
I found added tests failed on Windows.
So these tests are skipped on Windows temporarily

PR #1794, opened 2017-12-18
Use dig_rb instead of ruby_dig when backport_dig is not installed

okkez:
Because ruby_dig 0.0.2 does not support a object that responds to dig
method properly. On the other hand, dig_rb 1.0.1 supports it properly.


okkez:
For example, fluent-plugin-geoip uses geoip2_c's `GeoIP2::LookupResult#dig` with hash (pseudo code):

```ruby
db = GeoIP2::Database.new(db_path)
result = db.lookup(ip)
keys = [:hostname, :city, :names, :en]
hash = { hostname: result }
hash.dig(*keys") # => "Mountain View"
```
If we use ruby_dig 0.0.2, the last line will return `nil`.

PR #1796, opened 2017-12-19
Use Strftime to improve fixed timestamp performance

repeatedly:
Use `Strftime` instead of `Time#strftime` improves the performance of time formatting.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -163,10 +163,12 @@ module Fluent
     def format=(fmt)
       return if @format == fmt
 
+      @time_format = '%Y-%m-%d %H:%M:%S %z'
+      @time_formatter = Strftime.new(@time_format)
+
       case fmt
       when :text
         @format = :text
-        @time_format = '%Y-%m-%d %H:%M:%S %z'
         @formatter = Proc.new { |type, time, level, msg|
           r = caller_line(type, time, @depth_offset, level)
           r << msg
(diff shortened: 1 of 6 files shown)


repeatedly:
@mururu Could you also check this?

lib/fluent/plugin/formatter_stdout.rb
```diff
@@ -21,13 +21,14 @@ module Plugin
...
 
-      TIME_FORMAT = '%Y-%m-%d %H:%M:%S.%9N %z'
+      TIME_FORMAT = '%Y-%m-%d %H:%M:%S.%N %z'
 
       config_param :output_type, :string, default: 'json'
 
       def configure(conf)
         super
 
+        @time_formatter = Strftime.new(TIME_FORMAT)
```
mururu: Don't we need `rescue nil` here?

lib/fluent/log.rb
```diff
@@ -163,21 +163,22 @@ def logdev=(logdev)
...
           r << msg
           r
         }
       when :json
         @format = :json
-        @time_format = '%Y-%m-%d %H:%M:%S %z'
         @formatter = Proc.new { |type, time, level, msg|
           r = {
-            'time' => time.strftime(@time_format),
+            'time' => @time_formatter ? @time_formatter.exec(time) : time.strftime(@time_format),
```
mururu: How about defining method for `@time_formatter ? @time_formatter.exec(time) : time.strftime(@time_format)`?

repeatedly: Done.

lib/fluent/plugin/formatter_stdout.rb
repeatedly: Yes because TIME_FORMAT is fixed.

mururu: I see.

mururu:
Looks good.

Issue #1578, opened 2017-05-19
Standard way for nested record support


repeatedly:
This is for fixing #649, #902 and related issues.

Currently, many plugins can't handle nested record because there is no standard way.
Many users want to access nested record. For example, grep, rewrite-tag-filter, parser and more plugins.
Fluentd core should provide the way to handle these cases.

Here is a starting point. We need the feedback and suggestion!

## Features

- Standard syntax for configuration
- Provide plugin helper for nested record

### Configuration

What is the best syntax for configuration? There are several approaches.
Use grep filter plugin as an example:

#### [jsonpath](http://goessner.net/articles/JsonPath/) like synax

```
regexp1 key1 cool  #record['key1']`
regexp2 $.key1.key2[0] hot # record['key1']['key2'][0]
```

#### `.` separated key

```
regexp1 key1 cool  #record['key1']`
regexp2 key1.key2.0 hot # record['key1']['key2'][0]
```

How to handle `record["key1"]["key2"]` and `record["key1.key2"]` pattern?

#### support array

```
regexp1 key1 cool  #record['key1']`
regexp2 ['key1', 'key2', 0] hot # record['key1']['key2'][0]
```

#### record_transformer with enable_ruby syntax

```
regexp1 record["key1"] cool  #record['key1']`
regexp2 record["key1"]["key2"][0] hot # record['key1']['key2'][0]
```

`key1` is same as `record["key"]` for backward compatibility

### Plugin helper

For internal implementation, We can use `dig` method to access nested values since ruby 2.4. For ruby 2.2 or earlier, we can use [ruby_dig](https://github.com/Invoca/ruby_dig) or [backport_dig](https://github.com/koic/backport_dig).

For plugin API, there are several approaches:

#### Return accessor object

```
accessor = create_record_accessor(conf_param) # conf_param is $.key1.key2[0]
value = accessor.call(record) # access record['key1']['key2'][0]
```

#### Provide access method

This is similar to inject helper.

```
create_record_accessor(:name, conf_param) # conf_param is $.key1.key2[0]
value = access_record(:name, record) # access record['key1']['key2'][0]
```

Plugin helper needs to maintain accessor list internally.

#### or good API?

sonots:
+1 for jsonpath like syntax, but we need to add *bracket–notation* to support a key containing `.` as

```
regexp1 key1 cool  # record['key1']`
regexp2 $.key1.key2[0] hot # record['key1']['key2'][0]
[...]

PR #1800, opened 2017-12-21
Delete nested field

repeatedly:
Additional feature for `record_accessor`.
This is useful for several mutation plugins.

Ref #1578

### DIFF
--- a/test/plugin/test_filter_record_transformer.rb
+++ b/test/plugin/test_filter_record_transformer.rb
@@ -88,7 +88,7 @@ class RecordTransformerFilterTest < Test::Unit::TestCase
       config = CONFIG + %[remove_keys $.nest.k1]
       filtered = filter(config)
       filtered.each_with_index do |(_t, r), i|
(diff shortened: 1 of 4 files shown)


repeatedly:
@mururu Could you check this?

test/plugin/test_filter_record_transformer.rb
```diff
@@ -83,6 +84,14 @@ def filter(config, msgs = [''])
       end
     end
 
+    test 'remove_keys with nested key' do
+      config = CONFIG + %[remove_keys $.nest.k1]
+      filtered = filter(config)
+      filtered.each_with_index do |(_t, r), i|
+        assert_not_include({'k2' => 'v2'}, r['nest'])
```
mururu: This line should be `assert_not_include(r['nest'], 'k1')`.

repeatedly: right. just fixed.

mururu:
Looks good.

PR #1802, opened 2017-12-26
Add parameter delimiter_pattern to parser_ltsv

okkez:
The feature is taken from
https://github.com/fluent-plugins-nursery/fluent-plugin-kv-parser

fluent-plugin-kv-parser is very simple and similar to parser_ltsv.
So we can integrate fluent-plugin-kv-parser with parser_ltsv.

### DIFF
--- a/lib/fluent/plugin/parser_ltsv.rb
+++ b/lib/fluent/plugin/parser_ltsv.rb
@@ -32,10 +32,14 @@ module Fluent
 
       config_set_default :time_key, 'time'
 
+      def configure(conf)
+        super
+        @delimiter = @delimiter_pattern || @delimiter
+      end
+
       def parse(text)
         r = {}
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/parser_ltsv.rb
```diff
@@ -23,14 +23,19 @@ class LabeledTSVParser < Parser
...
+      end
       desc 'The delimiter character between field name and value'
       config_param :label_delimiter, :string, default: ":"
 
       config_set_default :time_key, 'time'
 
       def parse(text)
         r = {}
-        text.split(@delimiter).each do |pair|
+        delimiter = @delimiter_pattern || @delimiter
```
repeatedly: How about moving this to `configure`?

okkez: Thank you for reviewing. Good catch.

PR #1805, opened 2017-12-28
Fix tcp + tls degradation

mururu:
https://github.com/fluent/fluentd/pull/1729 introduced @buffer to
TCPCallbackSocket. TLSCallbackSocket also has to have it to be used for
in_tcp.


repeatedly:
good catch!

PR #1808, opened 2017-12-29
Expose record_accessor internal instance `@keys` variable. Fix #1806

### LINKED ISSUE #1806: Expose record_accessor internal API

- fluentd or td-agent version. 

Fluentd v1

- Environment information,

All supported Operating Systems

- Your configuration

Nope. It is feature request for `record_accessor`.

- Your problem explanation. If you have an error logs, write it together.

Currently, `record_accessor` does not expose internal `@keys` instance.
We want to use this variable in [fluent-plugin-anonymizer](https://github.com/y-ken/fluent-plugin-anonymizer) like this:

https://github.com/cosmo0920/fluent-plugin-anonymizer/blob/0bb45b93cebc996b6ccb13b5b7a2a71b4682cf14/lib/fluent/plugin/filter_anonymizer.rb#L242-L249

##### Additional request

~~Now, `@keys` does not handle keys correctly when using `$['your']['some']['key']` notation.
We also want to use this variable (or API?) above use case.~~

Already supported above. :bowing_man:

cosmo0920:
Currently, `record_accessor` does not expose internal @keys instance.
We want to use this variable in [fluent-plugin-anonymizer](https://github.com/y-ken/fluent-plugin-anonymizer).

### DIFF
--- a/lib/fluent/plugin_helper/record_accessor.rb
+++ b/lib/fluent/plugin_helper/record_accessor.rb
@@ -32,6 +32,8 @@ module Fluent
       end
 
       class Accessor
+        attr_reader :keys
+
         def initialize(param)
           @keys = Accessor.parse_parameter(param)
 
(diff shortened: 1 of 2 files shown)


repeatedly:
Thx!

PR #1809, opened 2017-12-30
Improve on logging

repeatedly:
1. Improve `Log#on_xxx` API performance with `yield`.
2. Old implementation is error prune. API is `log.on_info { ... }` so error should be thrown when block is not passed.
I searched the plugin code and no `log.on_xxx` without block code so it doesn't break existing plugins.
3. Use `on_trace` API in buffer. `trace` level is rare on the production so frequent object creation for trace logging is not good.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -261,9 +261,9 @@ module Fluent
       end
     end
 
-    def on_trace(&block)
+    def on_trace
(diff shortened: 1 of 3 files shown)


repeatedly:
Exactly

mururu:
Thanks. Looks good!

PR #1813, opened 2018-01-05
Fix the method for TimeFormatter#call

mururu:
`TimeFormatter#format` and `TimeFormatter#call` should call the same method.


repeatedly:
LGTM

PR #1814, opened 2018-01-10
config: Add hostname and worker_id short-cut. fix #1473

### LINKED ISSUE #1473: configuration: built-in placecholder for popular use cases
I'm now considering adding several built-in placeholders to ruby embedded code: http://docs.fluentd.org/v0.12/articles/config-file#embedded-ruby-code

- `"#{hostname}"` for `"#{Socket.gethostname}"`

`Socket.gethostname` seems verbose and it needs ruby knowledge a bit.
In `record_transformer`, we use `${hostname}` placeholder for it. So `"#{hostname}"` is much easier.

- `"#{worker_id}"` for `"#{ENV['SERVERENGINE_WORKER_ID'] || 0}"`

With built-in multiprocess feature, users sometimes need to specify worker id to avoid destination conflict, e.g. S3 path.
With multiprocess plugin, users write unique parameter in each child configuration. We should support such use cases.
`"#{ENV['SERVERENGINE_WORKER_ID'] || 0}"` is bit messy, so adding `"#{worker_id}"` is better.

For `worker_id`, I considered `${worker_id}` support in `extract_placeholders` but it conflicts with `worker_id` field of event record. In addition, `worker_id` is not dynamic in data pipeline after fluentd started, so `"#{worker_id}"` is enough for almost cases.

## NOTE: event tag support

`"#{tag}"` is out of scope because `"#{}"` is processed during configuration parsing phase in startup. So `"#{}"` can't touch event tag, time and record. Placeholder should be available value before configuration parsing.

Other popular cases?

repeatedly:


lib/fluent/config/literal_parser.rb
```diff
@@ -162,6 +163,8 @@ def eval_embedded_code(code)
         if @eval_context.nil?
           parse_error! "embedded code is not allowed in this file"
         end
+        hostname = Socket.gethostname
+        worker_id = ENV['SERVERENGINE_WORKER_ID'] || ''
```
repeatedly: This approach shows warnings with verbose mode but I'm not sure how to avoid it...

mururu:
With this conf

```
<source>
  @type dummy
  @id "dummy#{worker_id}"
  tag test
</source>
<match test>
  @type null
</match>
```

dumped config is like this:

```
2018-01-11 01:04:34 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type dummy
    @id dummy
    tag "test"
  </source>
  <match test>
    @type null
  </match>
</ROOT>
```

but I think `@id "dummy#{worker_id}"` is better.

lib/fluent/config/literal_parser.rb
mururu: I can't see any warnings on my environment. Could you share the detail?

repeatedly: See https://travis-ci.org/fluent/fluentd/jobs/327180965#L541

repeatedly:
Yeah. The problem is `"#{}"` is processed during configuration parsing and it doesn't store raw parameter.
This is the `Element` problem so we need other issue/pr for it.

mururu:
Makes sense.

lib/fluent/config/literal_parser.rb
mururu: Ah, I see. How about using `Binding#local_variable_set`?

repeatedly: With `instance_eval`, `local_variable_set` doesn't work. Try this:

```rb
context = Kernel.binding
context.local_variable_set(:worker_id, 1)
p context.instance_eval('"foo#{worker_id}"')
```

mururu: Can't we use `#eval` instead of `#instance_eval` here?

repeatedly: I'm not sure it doesn't break existing behaviour or not.

@nurse Do you see any problem for using `eval` here?

nurse: Just add to `code` seems simple.
```ruby
code = <<eom + code
hostname = Socket.gethostname
worker_id = ENV['SERVERENGINE_WORKER_ID'] || ''
eom
         @eval_context.instance_eval(code)
```

repeatedly:
Applied reviews. Will merge after test passed.

PR #1830, opened 2018-01-25
sending count of dropped messages on retry limit hit to fluentd with tag name fluent.error

jaisharma0009:
I need a way to track number of messages fluentd is dropping for monitoring purpose. After this change i can add small config in my config file to parse and count messages fluentd is dropping because of buffer retry.

This is how my fluentd config may look like
```
<match fluent.error>
##processing logic here
</match>

```

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1141,7 +1141,6 @@ module Fluent
               records = @buffer.queued_records
               msg = "failed to flush the buffer, and hit limit for retries. dropping all chunks in the buffer queue."
               $log.error msg, retry_times: @retry.steps, records: records, error: error
-
               log.error_backtrace error.backtrace
             elsif using_secondary
               msg = "failed to flush the buffer with secondary output."
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1059,7 +1059,6 @@ module Fluent
         return unless chunk
 
         log.on_trace { log.trace "trying flush for a chunk", chunk: dump_unique_id_hex(chunk.unique_id) }
-
         output = self
         using_secondary = false
         if @retry_mutex.synchronize{ @retry && @retry.secondary? }
@@ -1141,7 +1140,8 @@ module Fluent
             if @retry.limit?
               records = @buffer.queued_records
               msg = "failed to flush the buffer, and hit limit for retries. dropping all chunks in the buffer queue."
-              log.error msg, retry_times: @retry.steps, records: records, error: error
+              $log.error msg, retry_times: @retry.steps, records: records, error: error
+
               log.error_backtrace error.backtrace
             elsif using_secondary
               msg = "failed to flush the buffer with secondary output."


jaisharma0009:
retest

repeatedly:
Sorry, I'm not sure why this change works as expected.
Could you show me your detailed case?

jaisharma0009:
Here is my use-case
I want to send some data to google stack driver through file buffer. My retry limit is 1.
Before fluentd discard chunk i want to know how many messages fluentd is discarding for monitoring/alerting.

What is my change
When retry exceeds retry limit fluentd writes message in log file. I switch from local logger to global logger which also sends fluentd message with tag name "fluent.error". Now i can get count of lost messages in realtime like this
`<filter fluent.error>
  @type parser
  key_name message
  <parse>
    @type regexp
    expression /^.*?dropping all chunks in the buffer queue. retry_times=(?<retry_count>.*?) records=(?<msgs>.*?) error_class=(.*?)$/
  </parse>
</filter>

<match fluent.error>
  @type file
  path /Users/sharmajai/Documents/workspace/fluentd/tty2
</match>
`

jaisharma0009:
@repeatedly please check my last comment and let me know if you have any questions.

repeatedly:
>  I switch from local logger to global logger which also sends fluentd message with tag name "fluent.error". Now i can get count of lost messages in realtime with this config

Why need to switch to global logger? Current implementation seems to cover your case.
For example, following configuration captures buffer error log.

```
<filter fluent.**>
  @type record_transformer
  <record>
    category internal
  </record>
</filter>

<match fluent.**>
  @type stdout
</match>

<source>
  @type forward
</source>

<match test.**>
  @type null
  never_flush true
  <buffer>
    flush_interval 2s
    retry_max_times 1
  </buffer>
</match>
```

```
2018-02-03 10:14:54.769821000 +0900 fluent.error: {"retry_times":1,"records":1,"error":"#<RuntimeError: failed to flush>","message":"failed to flush the buffer, and hit limit for retries. dropping all chunks in the buffer queue. retry_times=1 records=1 error_class=RuntimeError error=\"failed to flush\"","category":"internal"}
```

jaisharma0009:
@repeatedly  In my understanding messages are sent to engine only if [log_event_enabled](https://github.com/fluent/fluentd/blob/master/lib/fluent/log.rb#L482) is true and log_event_enabled will be false for local logger. That's why i switched to global logger.

alexxed:
I had a similar issue when the errors thrown by one [plugin](https://github.com/dlackty/fluent-plugin-remote_syslog/blob/master/lib/fluent/plugin/out_remote_syslog.rb) did not make it to the captured events. I found the problem was using a store @id which caused the logging to use a different code path, shown [here](https://github.com/fluent/fluentd/blob/c3e9dbe50c737813c9086e980215be4423953249/lib/fluent/log.rb#L583-L583).

Editing the plugin like this:
```
def configure(conf)                                                                                       
        super                                                                                                   
        @log = $log
        ...
```

fixed the problem, I was able to capture errors from it using match fluent.error.

This is what I used for matching:

```
<label @FLUENT_LOG>                                                                                             
  <match fluent.{error,fatal,warn}>                                                                             
      @type statsd                                                                                              
      namespace fluentd_service                                                                                 
      <metric>                                                                                                  
        statsd_type increment                                                                                   
        statsd_key ${tag}                                                                                       
      </metric>                                                                                                 
  </match>                                                                                                      
  <match fluent.{debug,info,trace}>                                                                             
      @type null                                                                                                
  </match>                                                                                                      
</label>    
```

PR #1835, opened 2018-01-27
Do not close connection until write is complete on failed auth PONG

### LINKED ISSUE #1834: PONG on failed authentication is never received
- fluentd 1.1.0
- Fedora 27
- Configuration
```
<system>
  log_level debug
</system>
<source>
  @type forward
  port 24224
</source>
<source>
  @type forward
  port 24225
  <security>
    self_hostname fluenttest
    shared_key abcd1234
  </security>
</source>
<source>
  @type forward
  port 24226
  <security>
    self_hostname fluenttest
    shared_key 1234abcd
    user_auth true
    <user>
      username foo
      password bar
    </user>
  </security>
</source>
<match **>
  @type file
  @id output1
  path /fluentd/log/data.*.log
  symlink_path /fluentd/log/data.log
  append true
  time_slice_format %Y%m%d
  time_slice_wait 10m
  time_format %Y%m%dT%H%M%S%z
  buffer_path /fluentd/log/data.*.log
  <format>
    @type json
  </format>
</match>
```

I'm developing a new version of fluent/fluent-logger-python#124. As part of the testing suite I'm using docker with the real-life version of fluentd (1.1.0 currently) to verify proper functionality. 

What I'm finding is that it looks like FluentD, upon receiving an incorrect shared key or user creds, is sending a PONG, but is also immediately closing the connection. This causes the receiving client to receive an EOF instead of a PONG.

What should be happening is the FluentD server sending a PONG, sending a socket RDWR shutdown, but should defer closing the actual socket for N seconds (15? 30?) to ensure that the client actually has a chance to receive the PONG with failed authentication reason.

arcivanov:
fixes #1834

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -199,8 +199,8 @@ module Fluent::Plugin
         when :pingpong
           success, reason_or_salt, shared_key = check_ping(msg, conn.remote_addr, user_auth_salt, nonce)
           unless success
+            conn.on(:write_complete) { |c| c.close }
             send_data.call(serializer, generate_pong(false, reason_or_salt, nonce, shared_key))
-            conn.close
             next
           end
           send_data.call(serializer, generate_pong(true, reason_or_salt, nonce, shared_key))


repeatedly:
LGTM

PR #1839, opened 2018-01-31
command: Show fluentd version for debug purpose

repeatedly:


lib/fluent/command/binlog_reader.rb
```diff
@@ -49,7 +50,14 @@ def call
 
   def command
     command = @argv.shift
-    if !command || !SUBCOMMAND.include?(command)
+    if command
+      if command == '--version'
+        puts "#{File.basename($PROGRAM_NAME)} #{Fluent::VERSION}"
+        exit 1
```
mururu: I think it is better to return `0`.

mururu:
Other than the return value, looks good.

repeatedly:
Applied reviews. Merged.

PR #1850, opened 2018-02-07
in_http: Emit event time instead of raw time value in batch

repeatedly:
I noticed current in_http uses 'time' field value in batch request.
Should emit Fluent::EventTime instead of Integer or Float.


mururu:
Looks good.

PR #1241, opened 2016-09-26
Counter API

ganmacs:
This PR closes #858.
- WIP design  https://gist.github.com/ganmacs/cd936bba13a612eebd0594f7a965bfee  (Sorry, This article is Japanese)
- Performance and stack profiling  https://gist.github.com/ganmacs/00b225e09063e8601cb6bf281d7f27db  (only inc)

PR #1857, opened 2018-02-14
Counter API

repeatedly:
Update version of https://github.com/fluent/fluentd/pull/1241
We continue to update a patch here.

### DIFF
--- a/lib/fluent/counter/client.rb
+++ b/lib/fluent/counter/client.rb
@@ -63,6 +63,7 @@ module Fluent
       # 3. init([{ name: 'name1',reset_interval: 20 }, { name: 'name2',reset_interval: 20 }])
       # 4. init([{ name: 'name1',reset_interval: 20 }, { name: 'name2',reset_interval: 20 }], options: {})
       def init(params, options: {})
+        exist_scope!
         params = [params] unless params.is_a?(Array)
         res = send_request('init', @scope, params, options)
 
@@ -72,6 +73,7 @@ module Fluent
       end
 
       def delete(*params, options: {})
+        exist_scope!
         res = send_request('delete', @scope, params, options)
         options[:async] ? res : res.get
       end
@@ -84,22 +86,29 @@ module Fluent
       # 3. init([{ name: 'name1',value: 20 }, { name: 'name2',value: 20 }])
       # 4. init([{ name: 'name1',value: 20 }, { name: 'name2',value: 20 }], options: {})
       def inc(params, options: {})
+        exist_scope!
         params = [params] unless params.is_a?(Array)
         res = send_request('inc', @scope, params, options)
         options[:async] ? res : res.get
       end
 
       def get(*params, options: {})
+        exist_scope!
         res = send_request('get', @scope, params, options)
         options[:async] ? res : res.get
       end
 
       def reset(*params, options: {})
+        exist_scope!
         res = send_request('reset', @scope, params, options)
         options[:async] ? res : res.get
       end
 
       private
+
+      def exist_scope!
+        raise 'Call `establish` method to get a `scope` before calling this method' unless @scope
+      end
 
       def on_message(data)
         if response = @responses.delete(data['id'])
--- a/lib/fluent/counter/client.rb
+++ b/lib/fluent/counter/client.rb
@@ -215,7 +215,7 @@ module Fluent
       def join
         until @set
           @mutex.synchronize do
-            @loop.run_once
+            @loop.run_once(0.0001) # retun a lock as soon as possible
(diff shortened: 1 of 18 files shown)


repeatedly:
@okkez If you want to test counter-api, use this branch instead.

okkez:
I've tried to test counter-api but it seems that counter-api is wok in progress.
Fluentd does not call any counter-api (client). So I cannot test counter-api with Fluentd and fluent-plugin-xxx, for now.

I will try to add some ad-hoc patch and test counter-api with Fluentd.

lib/fluent/counter/client.rb
```diff
@@ -0,0 +1,230 @@
...
+      DEFAULT_PORT = 4321
+      DEFAULT_ADDR = '127.0.0.1'
+      ID_LIMIT_COUNT = 1 << 31
+
+      def initialize(loop = nil, opt = {})
+        @loop = loop || Coolio::Loop.new
+        @port = opt[:port] || DEFAULT_PORT
+        @addr = opt[:addr] || DEFAULT_ADDR
+        @log = opt[:log] || $log
+        @conn = Connection.connect(@addr, @port, method(:on_message))
```
okkez: How about following?
`Fluent::Counter::Client` is used by plugin developers and Counter API users. Is this right?
`opt ={}` does not describe details of parameters. Thus users have to read the code to understand this API.
We may also need plugin helper to create Counter API client for plugins.

```ruby
def initialize(event_loop: Coolio::Loop.new, log: $log, host: '127.0.0.1', port: 4321)
  @event_loop = event_loop
  @host = host
  @port = port
  @log = log
  @conn = Connection.connect(@host, @port, method(:on_message))
  # ...
end
```

lib/fluent/counter/client.rb
```diff
@@ -0,0 +1,237 @@
...
+      end
+
+      def data
+        get['data']
+      end
+
+      def get
+        # Block until `set` method is called and @result is set a value
+        join if @result.nil?
+        @result
```
okkez: How about changing the return value to Response object from plain Hash?

```ruby
class Response
   attr_reader :errors, :data
   def initialize(result)
      @errors = result["errors"]
      @data = result["data"]
      # ...
   end

  def success?
    @errors.nil? || @errors.empty?
  end

  def error?
    !success?
  end
end
```

It is useful to check Counter API response as following:

```ruby
@counter_client.inc({ name: "counter", value: 1}, options: { async: true }) do |response|
   if response.success?
     log.debug("Update counter successfully")
   else
     log.warn("Counter API error: #{response.errors}")
   end
end
```

Original code:

```ruby
future = @counter_client.inc({ name: "counter", value: 1}, options: { async: true })
Thread.start do
  response = future.get
  if future.errors?
     log.error("failure")
  else
     log.debug("success")
  end
end
```

repeatedly: Looks good. I will apply the changes soon.

lib/fluent/counter/client.rb
```diff
@@ -0,0 +1,237 @@
...
+      #
+      # 1. init(name: 'name')
+      # 2. init({ name: 'name',value: 20 }, options: {})
+      # 3. init([{ name: 'name1',value: 20 }, { name: 'name2',value: 20 }])
+      # 4. init([{ name: 'name1',value: 20 }, { name: 'name2',value: 20 }], options: {})
+      def inc(params, options: {})
+        exist_scope!
+        params = [params] unless params.is_a?(Array)
+        res = send_request('inc', @scope, params, options)
+        options[:async] ? res : res.get
```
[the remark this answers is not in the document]
repeatedly: Added suggested API.

repeatedly:
@okkez Do you have other suggestion? If not, I have a plan to merge this PR.

okkez:
@repeatedly I want a shorthand for `@counter_client.inc({ name: "name", value: 1 }, options: { async: true })`. It is annoying to specify `options: { async: true }` for each asynchronous request.

How about that the counter client sends an asynchronous request by default?

Or, how about that plugin helper creates an asynchronous client?

```ruby
async_client = counter_client_create(scope: "datacounter", async: true)
async_client.inc(name: "name", value: 1) # send an asynchronous request
async_client.inc(name: "name", value: 1, options: { async: false }) # send a synchronous request
```

Or else...?
Anyway, I want a shorthand to send asynchronous requests.

okkez:
Looks good to me :+1:

repeatedly:
Remove sync API

repeatedly:
Merged. We will improve Counter API with actual plugins.

PR #1861, opened 2018-02-19
Add support for TLS Mutual authentication within forward input

martinssipenko:
Hi,

This PR add support for Client auth (also knows as [Mutual authentication](https://en.wikipedia.org/wiki/Mutual_authentication)).

This mainly consist of two things:
- Configure SSL context to load `ca_file` from `ca_path` config. In case config is not set it will default to null, which is also acceptable.
- If `client_cert_auth` config is set to `true` then configure SSL context to verify peer certificates.

This is my first ever ruby code and I'm having trouble understanding how to write test, so I would appreciate any help with that.

### DIFF
--- a/lib/fluent/plugin_helper/cert_option.rb
+++ b/lib/fluent/plugin_helper/cert_option.rb
@@ -33,6 +33,11 @@ module Fluent
           ctx.ciphers = ciphers
         end
 
+        if conf.client_cert_auth
+            ctx.verify_mode = OpenSSL::SSL::VERIFY_PEER | OpenSSL::SSL::VERIFY_FAIL_IF_NO_PEER_CERT
+        end
+
+        ctx.ca_file = conf.ca_path
         ctx.cert = cert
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks. I will check it soon!

repeatedly:
Add unit test is hard for this case. So we need to consider adding other test scenario in the feature.

repeatedly:
After CI testing is passed, I will release v1.1.1.

martinssipenko:
@repeatedly should I also update the docs to reflect this festure?

repeatedly:
Yep.

https://docs.fluentd.org/v1.0/articles/in_forward#parameters 

Here is the parameters and 

https://docs.fluentd.org/v1.0/articles/in_forward#how-to-enable-tls/ssl-encryption

is better for how to use this feature.

martinssipenko:
https://github.com/fluent/fluentd-docs/pull/465

PR #1863, opened 2018-02-20
fix: encoding forces

dev-bandai:
Regexp correctly gets what is read in UTF-8


repeatedly:
So you get the encoding problem with regexp pattern?
Could you add test?

ganmacs:
No response for a while. closing.

PR #1864, opened 2018-02-20
Fix a bug to leak non-repeating timer watchers

tagomoris:
Timer plugin helpers may be used to call delayed processing with `repeat: false` option.
But if plugin authors call this method once per minute, `@_event_loop_attached_watchers` will have 60 undeleted objects per hour: that means object leakage.

This change is to fix that problem.


repeatedly:
LGTM

PR #1865, opened 2018-02-21
in_http: Do not discard the 'time' field in incoming JSON records.

fujimotos:
### What is the problem?

It was found that `parse_params_default()` unconditionally discards the `time`
value returned by the JSON parser. This trivially leads to silent data loss.

### How to reproduce the issue

We can confirm this by following these steps:

 1. Start a Fluentd server with settings like below:

    ```xml
    <source>
      @type http
    </source>
    <match debug.**>
      @type stdout
    </match>
    ```

 2. Send a HTTP request with the following command:

    ```bash
    $ curl -X POST -d 'json={"time":1}' http://localhost:8888/debug.log
    ```

 3. Now the original JSON data has been lost, leaving a empty hash with
    a timestamp of the current server time.

    ```
    2018-02-21 15:44:57.767980983 +0900 debug.log: {}
    ```

### Solution

This patch fixes the issue by passing the timestamp (extracted from
the incoming JSON data) to the caller.

With this patch applied, the `time` field in the JSON data will be properly 
reflected to the event data:

```
1970-01-01 09:00:01.000000000 +0900 debug.log: {}
```

### Note on the context of this patch

I've found this issue while updating the documentation of in_http
at https://github.com/fluent/fluentd-docs/pull/457.


fujimotos:
Also, note that the current implementation will properly update the timestamp
of the output event, *if we send the data as a JSON array*.
 
For example, the following command:

```bash
$ curl -X POST -d 'json=[{"time":1},{"time":2}]' http://localhost:9880/debug.log
```

will produce the following output:

```
1970-01-01 09:00:01.000000000 +0900 debug.log: {}
1970-01-01 09:00:02.000000000 +0900 debug.log: {}
```

(This is thanks to the special case handling implemented at [in_http.rb#L205](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_http.rb#L205))

This patch should fix this inconsistency in handling incoming records too.

repeatedly:
in_http uses time queyr parameter by default so this is expected behaviour for now.
How about option like `use_record_time` to use time field in the record?

PR #1869, opened 2018-02-26
Support multi worker environment in in_debug_agent

eagletmt:
in_debug_agent will listen sequential port number based on worker_id
just like in_monitor_agent. #1493


repeatedly:
Good catch!
Thanks.

Issue #1055, opened 2016-06-16
After a file rotation, in_tail will write log lines in new log file before the log lines in the rotated log file


tomfitzhenry:
- fluentd or td-agent version.
  td-agent-2.3.0-0.el7.x86_64
- Environment information, e.g. OS.
  Linux i-03864a911aa975c3a 3.10.0-327.18.2.el7.x86_64 #1 SMP Thu May 12 11:03:55 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
- Your configuration

```
<source>
  type tail
  format none
  path /var/log/foo/output.log
  pos_file /var/log/td-agent/application-log_foo.pos
  read_from_head true
  tag application-log
</source>

<match application-log>
    type s3
    s3_bucket foobarbucket
    s3_region eu-west-1
    path logs/annotator/application-log/
    buffer_path /var/log/td-agent/buffer/s3-application-log
    time_slice_format %Y-%m-%d/%{hostname}/%Y-%m-%dT%H_%{hostname}
    flush_interval 600s
    buffer_chunk_limit 64m
    format single_value
    store_as gzip
    num_threads 1
    use_server_side_encryption AES256

    <instance_profile_credentials>
    </instance_profile_credentials>
</match>
```
- Your problem explanation. If you have an error logs, write it together.
## Steps to reproduce

```
$ killall -STOP ruby # to more easily create this race condition
$ echo foo >> /var/log/foo/output.log
$ mv /var/log/foo/output.log /var/log/foo/output.log.0 # to simulate a log rotation
$ echo bar >> /var/log/foo/output.log
$ killall -CONT ruby
$ sleep 5 # to give td-agent time to do its thing
$ cat /var/log/td-agent/buffer/s3-application-log*
```
## Expected

```
foo
bar
```
## Actual

```
bar
foo
```
## Output of `/var/log/td-agent/td-agent.log`

```
2016-06-16 16:46:36 +0000 [info]: detected rotation of /var/log/foo/output.log; waiting 5 seconds
2016-06-16 16:46:36 +0000 [info]: following tail of /var/log/foo/output.log
```

repeatedly:
Thanks for the reporting.
I will check it later.

repeatedly:
I confirmed this behaviour.
I will check the code.

tagomoris:
@repeatedly are there any progress?

repeatedly:
[...]

PR #1880, opened 2018-03-06
in_tail: Handle records in the correct order on file rotation

fujimotos:
This is the fix for the issue reported by the bug report #1055.

### Problem

It turned out that this bug was a matter of the calling order of functions in `TailWatcher.on_rotate()`.

In particular, when TailWatcher detects a log rotation and immediately finds a new file, it performs the following steps:

  1) Create a new TailWatcher for the new file.
      * In this process, it reads the new file and emit its contents.
      * This all happens before proceeding to (2).
  2) Dispatch the `on_notify()` for the archived file.
      * In this process, it reads the content of the old file and emit its contents.

So naturally, newer log records (written to the new log file) will be emitted before older records in the archived log file.

### Solution

This patch fixes it by enforcing flush of older log records before performing the step (1).

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -580,6 +580,8 @@ module Fluent::Plugin
               @pe.update(inode, 0)
             else # file is rotated and new file found
               watcher_needs_update = true
+              # Handle the old log file before renewing TailWatcher [fluentd#1055]
+              @io_handler.on_notify
             end
           else # file is rotated and new file not found
             # Clear RotateHandler to avoid duplicated file watch in same path.


fujimotos:
Note that this bug occurs only when TailInput is slow to notice the file rotation.

In most cases, it does not occur because a log rotation is detected immediately after
the file rename. In this ordinary case, a file rotation will be handled as follows:

  1) Detect the rename of the log file and set up a new TailWatcher.
      * In this step, the new file is NOT created yet, so nothing is emitted.
  2) Dispatch the `on_notify()` for the archived file.
  3) Detect the creation of the new log file.
      * In this step, the content of the new file gets processed.

Since it handles the content of the old file in the step (2) and the content of the new file
in the step (3), the output order of records is preserved.

repeatedly:
Thanks for the investigation and patch. Just merged :)

fujimotos:
Thanks for merging it!

PR #1883, opened 2018-03-09
Fix boolean parameters without value with following comment

tagomoris:
As reported at https://github.com/tagomoris/fluent-plugin-route/issues/10, `config_param :name, :bool` returns falsely value (`nil`) when it's configured without explicit value with following comment.

```ruby
# parameter definition
config_param :b1, :bool, default: false
config_param :b2, :bool, default: false

# configuration
# b1 will be true
b1
# b2 will be nil
b2 # comment here
```

This bug cannot be reproducible via test cases of v1 parser, but I succeeded to reproduce it using plugin code and test driver. This bug was reported to be reproducible on actual Fluentd processes.

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -60,7 +60,13 @@ module Fluent
       when ''
         true
       else
-        nil
+        # Current parser passes comment without actual values, e.g. "param #foo".
+        # parser should pass empty string in this case but changing behaviour may break existing environment so keep parser behaviour. Just ignore comment value in boolean handling for now.
+        if str.respond_to?('start_with?') && str.start_with?('#')
+          true
+        else
+          nil
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


tagomoris:
@repeatedly This looks a bug related with literal parser, v1 parser and value parser, and takes a time to fix it. Could you take ownership on this?

repeatedly:
The problem is this line: https://github.com/fluent/fluentd/blob/7c0c0127bbafe7ad1847d10c52fcd85d67434907/lib/fluent/config/v1_parser.rb#L108
Maybe, this is for this test case: https://github.com/fluent/fluentd/blob/7c0c0127bbafe7ad1847d10c52fcd85d67434907/test/config/test_config_parser.rb#L97

I think this parsed result should be `{"k1" => ""}` because we can use `'# foo'` like syntax for `#` started string. But changing this behaviour may break existing environment. So we can't use this fix.
For keeping compatibility, changing bool handling is better.

```
index f41d451..beccc18 100644
@@ -60,7 +60,11 @@ module Fluent
       when ''
         true
       else
-        nil
+        if str.start_with?('#')
+          true
+        else
+          nil
+        end
       end
     end

```

repeatedly:
Hmm... return `nil` seems error prone. Should it raise an exception for unexpected value?

repeatedly:
@tagomoris How about this patch?

PR #1887, opened 2018-03-12
Support pattern start with character classes

okkez:
In prevous version, following pattern raises syntax error:

    <filter>
      @type grep
      <regexp>
        key price
        pattern [1-9]\d*
      </regexp>
    </filter>

This example is derrived from [filter-grep](https://docs.fluentd.org/v1.0/articles/filter_grep#<regexp>-directive)

### DIFF
--- a/lib/fluent/plugin/filter_grep.rb
+++ b/lib/fluent/plugin/filter_grep.rb
@@ -41,7 +41,11 @@ module Fluent::Plugin
       config_param :key, :string
       desc "The regular expression."
       config_param :pattern do |value|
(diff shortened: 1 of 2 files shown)


okkez:
On the other hand, we want to use regular expression in configuration file sometimes.
So, should we support `config_param :pattern, :regexp` ?

repeatedly:
all tests failed.

okkez:
Sorry, my fault...

repeatedly:
Thanks!

Issue #1858, opened 2018-02-14
Feature request: Filter out specific "bad actors"


JohnStarich:
I would like to filter out messages that match multiple conditions. This would allow most logs to flow normally, while filtering out specific "bad actors" producing too many logs.

While this does sound similar to what `filter_grep` provides, it does not support this use-case.

For example, I have logs like the following:

```json
{"level":"debug", "log":"hello!",                 "container_name":"foo"}
{"level":"error", "log":"something bad happened", "container_name":"foo"}
{"level":"debug", "log":"hello!",                 "container_name":"bar"}
```

My goal is to filter out `debug` logs for container_name `foo`.

So output would be:

```json
{"level":"error", "log":"something bad happened", "container_name":"foo"}
{"level":"debug", "log":"hello!",                 "container_name":"bar"}
```

It is important that only container_name `foo` is affected, so if I add other containers with `debug` messages _later on_ then they will not be filtered out with this config.

In other words, the same config should allow new logs from container_name `thing`, even if they are `debug` level.

So with the same config, these logs should make it through:

```json
{"level":"debug", "log":"hello!", "container_name":"thing"}
{"level":"debug", "log":"hello!", "container_name":"some_app"}
```

### Problem

The current standard grep filter only supports `AND` matches between `<regexp>` blocks and `OR` matches between `<exclude>` blocks.

I can create a fluentd config for the above example that **almost** works:

```
<filter **>
    @type grep
    <exclude>
        key container_name
        pattern foo
    </exclude>

    <exclude>
        key level
        pattern debug
    </exclude>
</filter>
```

The problem with this config is that it excludes all logs from `foo` containers, including `error` logs. This also excludes `debug` logs from the `bar` containers.

Ok, so maybe we can flip it a little? Unfortunately, we now hit the complement of the same problem:

```
<filter **>
    @type grep
    <regexp>
        key container_name
        pattern (?!.*foo).*    # not foo
    </regexp>

    <regexp>
        key level
        pattern (?!.*debug).*  # not debug
    </regexp>
</filter>
```

This config still throws away all of `foo`'s logs and any debug logs.

### Possible solution

One possible solution is to modify the grep filter to support a new `<exclude-all>` block that ANDs together each block before throwing out a log.

Please let me know how I can help! 😄

repeatedly:
`<exclude-all>` idea looks good.
[...]

PR #1897, opened 2018-03-15
Support and/or section

okkez:
For example:

    <filter>
      @type grep
      <and>
        <regexp>
          key level
          pattern ^WARN|ERROR$
        </regexp>
        <regexp>
          key method
          pattern ^GET$
        </regexp>
      </and>
    </filter>

For more details, see test code

See #1858

### DIFF
--- a/lib/fluent/plugin/filter_grep.rb
+++ b/lib/fluent/plugin/filter_grep.rb
@@ -33,6 +33,8 @@ module Fluent::Plugin
       @_exclude_or_conditions = {}
     end
 
+    attr_reader :_regexp_and_conditions, :_exluce_and_conditions, :_regexp_or_conditions, :_exclude_or_conditions
+
     helpers :record_accessor
 
     REGEXP_MAX_NUM = 20
--- a/lib/fluent/plugin/filter_grep.rb
+++ b/lib/fluent/plugin/filter_grep.rb
@@ -69,38 +69,29 @@ module Fluent::Plugin
     def configure(conf)
       super
 
-      rs = {}
       (1..REGEXP_MAX_NUM).each do |i|
         next unless conf["regexp#{i}"]
         key, regexp = conf["regexp#{i}"].split(/ /, 2)
         raise Fluent::ConfigError, "regexp#{i} does not contain 2 parameters" unless regexp
-        raise Fluent::ConfigError, "regexp#{i} contains a duplicated key, #{key}" if rs[key]
-        rs[key] = Regexp.compile(regexp)
+        raise Fluent::ConfigError, "regexp#{i} contains a duplicated key, #{key}" if @_regexps[key]
+        @_regexps[key] = Expression.new(record_accessor_create(key), Regexp.compile(regexp))
       end
 
-      es = {}
       (1..REGEXP_MAX_NUM).each do |i|
         next unless conf["exclude#{i}"]
         key, exclude = conf["exclude#{i}"].split(/ /, 2)
         raise Fluent::ConfigError, "exclude#{i} does not contain 2 parameters" unless exclude
-        raise Fluent::ConfigError, "exclude#{i} contains a duplicated key, #{key}" if es[key]
-        es[key] = Regexp.compile(exclude)
+        raise Fluent::ConfigError, "exclude#{i} contains a duplicated key, #{key}" if @_excludes[key]
+        @_excludes[key] = Expression.new(record_accessor_create(key), Regexp.compile(exclude))
       end
 
       @regexps.each do |e|
-        raise Fluent::ConfigError, "Duplicate key: #{e.key}" if rs.key?(e.key)
-        rs[e.key] = e.pattern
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/filter_grep.rb
```diff
@@ -25,10 +25,14 @@ class GrepFilter < Filter
...
 
-      @_regexps = {}
-      @_excludes = {}
+      @_regexp_and_conditions = {}
+      @_exclude_and_conditions = {}
+      @_regexp_or_conditions = {}
+      @_exclude_or_conditions = {}
     end
 
+    attr_reader :_regexp_and_conditions, :_exluce_and_conditions, :_regexp_or_conditions, :_exclude_or_conditions
```
JohnStarich: small spelling mistake: `_exluce_and_conditions` -> `_exclude_and_conditions`

okkez: Thanks! I've fixed.

lib/fluent/plugin/filter_grep.rb
```diff
@@ -122,5 +201,18 @@ def filter(tag, time, record)
       end
       result
     end
+
+    class Expression
```
repeatedly: How about using Struct for it?

```
Expression = Struct.new(:key, :pattern) do
  def match?(record)
    ::Fluent::StringUtil.match_regexp(pattern, key.call(record).to_s)
  end
end
```

lib/fluent/plugin/filter_grep.rb
```diff
@@ -54,65 +59,139 @@ def initialize
...
       end
     end
 
     def filter(tag, time, record)
       result = nil
       begin
         catch(:break_loop) do
-          @_regexps.each do |key, regexp|
-            throw :break_loop unless ::Fluent::StringUtil.match_regexp(regexp, key.call(record).to_s)
+          @_regexp_and_conditions.each do |key, expression|
```
repeatedly: This is very small optimization but how about using array for conditions?
In `filter`, we don't use `key` and hash related operations so array is good structure.

lib/fluent/plugin/filter_grep.rb
```diff
@@ -54,65 +59,139 @@ def initialize
...
       super
 
-      rs = {}
       (1..REGEXP_MAX_NUM).each do |i|
         next unless conf["regexp#{i}"]
         key, regexp = conf["regexp#{i}"].split(/ /, 2)
         raise Fluent::ConfigError, "regexp#{i} does not contain 2 parameters" unless regexp
-        raise Fluent::ConfigError, "regexp#{i} contains a duplicated key, #{key}" if rs[key]
-        rs[key] = Regexp.compile(regexp)
+        raise Fluent::ConfigError, "regexp#{i} contains a duplicated key, #{key}" if @_regexp_and_conditions[key]
```
repeatedly: How about logging info or warning message for top level setting?
`Top level multiple <regexp> is intepreted as 'and' condition`.

JohnStarich:
This looks great. @repeatedly can this get merged? (The failing test in Travis looks unrelated to these changes)

repeatedly:
Sorry for the delay. Will check it again!

lib/fluent/plugin/filter_grep.rb
```diff
@@ -54,65 +59,157 @@ def initialize
...
-    # for test
-    attr_reader :_regexps
-    attr_reader :_excludes
+    config_section :and, param_name: :and_conditions, multi: true do
+      config_section :regexp, param_name: :regexps, multi: true do
+        desc "The field name to which the regular expression is applied."
+        config_param :key, :string
+        desc "The regular expression."
+        config_param :pattern do |value|
+          if value.start_with?("/") and value.end_with?("/")
```
repeatedly: Hmm... I see these patterns, we need to `:regexp` type for config_param.

@okkez Could you write a patch for adding `regexp` type in other PR?

okkez: Sure, I will write a patch for this.

lib/fluent/plugin/filter_grep.rb
```diff
@@ -54,65 +59,157 @@ def initialize
...
+          end
+        end
+      end
+    end
 
     def configure(conf)
       super
 
-      rs = {}
+      _regexp_and_conditions = {}
```
repeatedly: This name is conflicts with attr_reader's definition. Yeah, I know local variable is preferred but using other name , e.g. `_regexp_and_conds`, is better. Others are same.

lib/fluent/plugin/filter_grep.rb
okkez: I've created #1927

repeatedly:
@okkez We enabled DCO for fluentd repository. Could you re-commit last commit with `-s`?

okkez:
I've re-committed and pushed the last commit with `-s`. @repeatedly

repeatedly:
Thanks!

PR #1898, opened 2018-03-15
fix typo

JiHyunSong:
... fix typo

`/opt/td-agent/embedded/bin/fluent-binlog-reader format` does not works, but
`/opt/td-agent/embedded/bin/fluent-binlog-reader formats` works


repeatedly:
Thanks!

Issue #1889, opened 2018-03-12
Fluentd stuck.


sergeyarl:
Hi! 

We are using high availability scheme with multiple fluentd forwarders, that are running in docker containers and sending logs to active-backup fluentd aggregators pair. An aggregator node then saves logs to disk in text form and also sends a copy to Elasticsearch server. 

Every once in a while something happenes to active fluent aggregator and after that all forwarders start to respond very slowly. Sometimes it seems that they got stuck. Getting stats page `http://localhost:24220/api/plugins.json` takes about 1 minute. 

It is impossible to restart fluentd on aggregator nodes with service manager. In order to stop it I need to use `kill -9` command.  

After aggregator fluentd is restarted, forwarders start to behave normally. 

We use CentOS 7 x86_64 as an OS on all nodes.

On aggregator nodes fluentd is installed as td-agent rpm: td-agent-2.3.6-0.el7.x86_64.
On forwarder nodes we use official fluend docker images.

Forwarder config file:

```

<source>
  @type monitor_agent
  bind 127.0.0.1
  port 24220
</source>

# TCP input
<source>
  @type forward
  bind 127.0.0.1
  port 24224
</source>

# Logs Forwarding
<match **>
  @type forward

  heartbeat_type tcp
  send_timeout 60s
  recover_wait 10s
  heartbeat_interval 1s
  phi_threshold 16
  hard_timeout 120s

  <server>
    name logs1
    host logs1
    port 8889
    weight 60
  </server>

  <server>
    name logs2
    host logs2
    port 8889
    weight 60
    standby
  </server>

  flush_interval 10s

  buffer_type file
  buffer_path /fluentd/log/buffer/forward
  buffer_chunk_limit 4m
  buffer_queue_limit 4096

  num_threads 2

  expire_dns_cache 600
</match>
```

Sigdumps are attached.

[sigdump-forwarder.tar.gz](https://github.com/fluent/fluentd/files/1804720/sigdump-forwarder.tar.gz)
[sigdump-aggregator.tar.gz](https://github.com/fluent/fluentd/files/1804724/sigdump-aggregator.tar.gz)

Thank you! 

Regards,
Sergey

sergeyarl:
Hi! 

Any chances to figure out why this is happening? It is rather severe problem for us. Due to this thing we are not able to use fluentd in production any more :( 

Thanks. 

Regards,
Sergey

toru-takahashi:
You can monitor fluentd usage by following. 
[...]

Issue #1636, opened 2017-07-19
elasticsearch down caused td-agent cpu 100%


hiscal2015:
td-agent version: 2.3.4

issue:
I use td-agent with fluent-elasticsearch-plugin, when I shutdown elasticsearch, found os cpu usage was 100% because of td-agent, is this the expected behavior?

log:
```
2017-07-19 12:25:59 +0800 [warn]: emit transaction failed: error_class=Fluent::BufferQueueLimitError error="queue size exceeds limit" tag="docker.var.lib.docker.containers.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55-json.log"
  2017-07-19 12:25:59 +0800 [warn]: suppressed same stacktrace
2017-07-19 12:25:59 +0800 [warn]: emit transaction failed: error_class=Fluent::BufferQueueLimitError error="queue size exceeds limit" tag="docker.var.lib.docker.containers.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55-json.log"
  2017-07-19 12:25:59 +0800 [warn]: suppressed same stacktrace
2017-07-19 12:25:59 +0800 [warn]: emit transaction failed: error_class=Fluent::BufferQueueLimitError error="queue size exceeds limit" tag="docker.var.lib.docker.containers.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55-json.log"
  2017-07-19 12:25:59 +0800 [warn]: suppressed same stacktrace
2017-07-19 12:25:59 +0800 [warn]: emit transaction failed: error_class=Fluent::BufferQueueLimitError error="queue size exceeds limit" tag="docker.var.lib.docker.containers.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55.d8062ec627f7872f17cb48bb1e7c643a4765344bf1b3a9308a69f4aa3d62ea55-json.log"
[...]

PR #1901, opened 2018-03-16
output: Prevent flushing threads from consuming too much CPU.

fujimotos:
I believe this patch fixes the CPU spikes issue reported by #1636 and #1889
(possibly also #1665).

### Problem

If one flush thread sets the retry state, this makes other threads
wait until the next retry time. This behaviour itself is fine and ok.

However, in this case the `interval` variable has a meaningless value
because `state.next_clock` is behind the system time. And since the
sleep call gets skipped in such a case:

```ruby
while @output_flush_threads_running
  current_clock = Fluent::Clock.now
  interval = state.next_clock - current_clock
  ...
  sleep interval if interval > 0
end
```

it effectively make the waiting threads to consume the resource
aggressively with busy loops.

### Solution

This patch makes a flush thread to wait in a sleep state until the
next retry time, insteads of busy looping

### Note

This bug seems to exist from v0.14 since this commit 6c5ba1a.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1313,9 +1313,17 @@ module Fluent
           # This thread don't use `thread_current_running?` because this thread should run in `before_shutdown` phase
           while @output_flush_threads_running
             current_clock = Fluent::Clock.now
-            interval = state.next_clock - current_clock
+            next_retry_time = nil
 
-            if state.next_clock <= current_clock && @retry_mutex.synchronize { @retry ? @retry.next_time <= Time.now : true }
+            @retry_mutex.synchronize do
+              next_retry_time = @retry ? @retry.next_time : nil
+            end
+
+            if state.next_clock > current_clock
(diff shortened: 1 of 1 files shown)


fujimotos:
#### Note: How to reproduce this issue

You can reproduce this issue relatively trivially. Try the following settings
(optionally with a debug `print` statement in [plugin/output.rb#L1334](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L1334)).

```xml
<source>
  @type dummy
  tag debug.log
  dummy {"a":"b"}
</source>

# Forward to a non-existent server.
<match debug.**>
  @type forward
  <server>
    host localhost
    port 9999
  </server>
  <buffer>
    flush_interval 3
    flush_thread_count 3
  </buffer>
</match>
```

Also, you can use this config to check this patch actually fixes the problem

repeatedly:
Confirmed. Thanks!

PR #1908, opened 2018-03-23
Support negative index for tag when extract placeholders

okkez:
For example,

tag is `fluentd.test.output`

`${tag[-1]}` is extracted to `output`
`${tag[-2]}` is extracted to `test`

It is useful to extract last element of variable length tag.


okkez:
Some output plugins supports negative index for tag placeholders.
I want to use this functionality with [fluent-plugin-pagerduty](https://github.com/y-ken/fluent-plugin-pagerduty).

repeatedly:
LGTM. Thanks.

PR #1910, opened 2018-03-27
Edit Fluentd license info so GitHub recognizes it

eirinikos:
Hello! 👋 

(CCing @dankohn who has requested this work so that the license will appear correctly in https://landscape.cncf.io/grouping=landscape&landscape=logging&selected=fluentd)

GitHub uses a library called Licensee to identify a project's license
type. It shows this information in the status bar and via the API if it
can unambiguously identify the license.

This commit removes the COPYING file (which duplicates info that's
already included in LICENSE), so that the license type identification
(as Apache 2.0) will succeed.


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2018 FURUHASHI Sadayuki
```
dankohn: Should this be `Copyright 2018 Fluentd Authors`?

Cc @edsiper

eirinikos: What do you think of using `Copyright Fluentd Authors` instead?
(in accordance with the discussion we've been having for OpenTracing: https://github.com/opentracing/opentracing-go/pull/181#issuecomment-376668544)

edsiper: LGTM "Copyright Fluentd Authors" :)

kzk: Should we mass-change the Copyright in the source code as well?

dankohn: @kzk Feel free, but it's not necessary.

eirinikos:
I've updated `LICENSE` and `README` to refer to `Copyright 2011-2018 Fluentd Authors`.

I've also add my DCO sign-off (which was previously missing).

If there are no more comments, this PR is ready to be merged by a maintainer! :)

repeatedly:
Thanks! Just merged :)

Issue #1907, opened 2018-03-22
`$http_x_forwarded_for` field missing from nginx parser plugin format


yaron-idan:
- fluentd or td-agent version.
fluentd v1.1.2
- Environment information, e.g. OS.
Linux / Docker
- Your configuration
```
<source>
  @type tail
  path "/logs/*.log"
  format nginx
  tag "nginx.*"
</source>
```
- Your problem explanation. If you have an error logs, write it together.
When receiving a log from nginx, I get the following message - 
`fluentd_1        | 2018-03-22 18:07:49 +0000 [warn]: #0 pattern not match: "172.18.0.1 - - [22/Mar/2018:18:07:49 +0000] \"GET / HTTP/1.1\" 200 612 \"-\" \"curl/7.54.0\" \"-\""`
If I swap the format configuration to the one listed in the plugin docs and add the last field like so the log level goes through -
```
format /^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)" "(?<agent2>[^\"]*)")?$/
  time_format %d/%b/%Y:%H:%M:%S %z
```
I find this a bit strange since the last field (`$http_x_forwarded_for`) has been part of the default nginx logging for a long time. Is there a real reason for omitting this line from the plugins format?

If the field needs adding, I'm cool with opening the PR for it. Just wanted to check up before starting to work on it.

repeatedly:
Maybe, nginx's regexp comes from https://github.com/nginx/nginx/blob/764737256521de5c34f7ce1cc1b27bf440b6cced/src/http/modules/ngx_http_log_module.c#L227
I checked the configuration and `$http_x_forwarded_for` is included in the example.
https://github.com/nginx/nginx/blob/9207cc84b21e94283478cee7a953b1859c4434cb/conf/nginx.conf#L21
So support `$http_x_forwarded_for` seems good for many users.

repeatedly:
v1.2.0.pre1 includes this change.

PR #1912, opened 2018-03-28
added http-x-forwarded-for field to nginx parser

yaron-idan:
Closing #1907, an update to the docs has been proposed in https://github.com/fluent/fluentd-docs/issues/471.

### DIFF
--- a/test/plugin/test_parser_nginx.rb
+++ b/test/plugin/test_parser_nginx.rb
@@ -14,7 +14,8 @@ class NginxParserTest < ::Test::Unit::TestCase
       'code'    => '200',
       'size'    => '777',
       'referer' => '-',
-      'agent'   => 'Opera/12.0'
+      'agent'   => 'Opera/12.0',
+      'http_x_forwarded_for' => '-'
     }
   end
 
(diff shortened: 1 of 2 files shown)


test/plugin/test_parser_nginx.rb
```diff
@@ -24,23 +25,23 @@ def create_driver
 
   def test_parse
     d = create_driver
-    d.instance.parse('127.0.0.1 192.168.0.1 - [28/Feb/2013:12:00:00 +0900] "GET / HTTP/1.1" 200 777 "-" "Opera/12.0"') { |time, record|
+    d.instance.parse('127.0.0.1 192.168.0.1 - [28/Feb/2013:12:00:00 +0900] "GET / HTTP/1.1" 200 777 "-" "Opera/12.0" -') { |time, record|
```
repeatedly: Don't change existing test because it is needed to check compatibility.

yaron-idan: I see, this test will fail though if it won't be adapted to the nginx log format, how are you suggesting to overcome this issue? Create another test with the new format?

repeatedly: > Create another test

Yes, creating new test is needed. Many users use nginx format so we can't break it. It breaks user environment and causes log lost.

> the new format?

One regular expression seems to support log with/without http_x_forwarded_for.

yaron-idan:
I've fixed the REGEXP to make the new field optional. I've also heed the advice of an SO answer and swapped `[^ ]` with `\S` which seems more readable to me - but I can change back if you don't agree.
All existing tests pass fine, and a new test has been added to check compatibility with the extended log entry. Hope that fits the requested changes.

repeatedly:
`[^ ]` and `\S` are different and `\S` is bit slower. Please back changes to previous implementation.
In addition, we enabled DCO for our repository. Please add your sign to the commit. See https://github.com/apps/dco

yaron-idan:
OK, seems I'm gonna have to open a new PR since DCO rejects commits without a sign off, I'm gonna close this one and open a new one with the changes requested + a sign off.

Issue #1904, opened 2018-03-19
v1.1 have a serious problem, too many buffer files and 100% cpu usage when forward node has error in high concurrency log system


breath-co2:
Our fluentd(v1.1) log system have billion logs per day. When the fluent out forward node fails more than a few hours, the fluentd save millions of buffer files, and 100% cpu usage. In fluentd 0.12.* is not that.

I think it is caused by some defects. In my fluentd config, chunk_limit_size is 4m, and flush_interval is 1s, I want fluentd foward plugin send data as quickly as possible. In fluentd v0.12 if forward node is error, it will save buffer files 4MB per file, but in fluentd v1.1, it will save buffer file every second, awful, it save million files. so, the problem will appear flush_mode is interval(and flush_interval is small) or immediate.

It's test config:

```
<source>
  @type dummy
  tag log.test
  dummy {"hello":"world"}
</source>

<match log.*>
  @type forward
  heartbeat_interval 10s
  require_ack_response true
  ack_response_timeout 180

  <buffer tag>
    @type file
    path /data/fluentd-buffer/test/
    flush_mode interval
    flush_interval 1s               # flush every second, each chunk max size 4MB
    chunk_limit_size 4m
    retry_forever true
    retry_max_interval 3
    queue_limit_length 5000
  </buffer>

  <server>
    host 127.0.0.1    # error node
    port 9982
  </server>
</match>
```

then run `td-agent -c test.conf` and run `ls -al /data/fluentd-buffer/test/`:
```
-rw-r--r-- 1 root root 15 3月  19 14:45 buffer.q567be4cc51e092dea4fe2ce53920b93b.log
-rw-r--r-- 1 root root 74 3月  19 14:45 buffer.q567be4cc51e092dea4fe2ce53920b93b.log.meta
-rw-r--r-- 1 root root 15 3月  19 14:45 buffer.q567be4cd87de6afc959a0c2699f29e2e.log
-rw-r--r-- 1 root root 74 3月  19 14:45 buffer.q567be4cd87de6afc959a0c2699f29e2e.log.meta
-rw-r--r-- 1 root root 15 3月  19 14:45 buffer.q567be4cec19469d897bf3af0260362ab.log
-rw-r--r-- 1 root root 74 3月  19 14:45 buffer.q567be4cec19469d897bf3af0260362ab.log.meta
[...]

PR #1916, opened 2018-03-30
buffer: Add queued_chunks_limit_size to control the number of queued chunks

repeatedly:
Releated to https://github.com/fluent/fluentd/issues/1904
Current patch resolves generating lots of queued chunks with small flush_interval.
Note that v0.12's the number of queues is 1.

I will check how control this parameter with `write` families.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -428,12 +428,13 @@ module Fluent
         end
       end
 
-      def enqueue_all
+      # At flush_at_shutdown, all staged chunks should be enqueued for buffer flush. Set true to force_enqueue for it.
+      def enqueue_all(force_enqueue = false)
         log.on_trace { log.trace "enqueueing all chunks in buffer", instance: self.object_id }
 
         if block_given?
           synchronize{ @stage.keys }.each do |metadata|
-            return if queue_full?
+            return if !force_enqueue && queue_full?
             # NOTE: The following line might cause data race depending on Ruby implementations except CRuby
             # cf. https://github.com/fluent/fluentd/pull/1721#discussion_r146170251
             chunk = @stage[metadata]
@@ -443,7 +444,7 @@ module Fluent
           end
         else
           synchronize{ @stage.keys }.each do |metadata|
-            return if queue_full?
+            return if !force_enqueue && queue_full?
             enqueue_chunk(metadata)
           end
         end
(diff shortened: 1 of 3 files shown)


repeatedly:
@mururu How about this change?

mururu:
This patch addresses only too many buffer file problem mentioned in #1904, not high CPU usage, right? Then basically it looks good. 
But I have a concern about `flush_at_shutdown`. `flush_at_shutdown` calls `#enqueue_all` at first, so some chunks seem not to be flushed at shutdown phase if we have too many chunks.

repeatedly:
> high CPU usage

I think this problem was resolved since 1.1.2: https://github.com/fluent/fluentd/pull/1901

> But I have a concern about flush_at_shutdown. flush_at_shutdown calls #enqueue_all at first

Ah I forgot it. Yeah, v0.12 enqueues all un-flushed buffer and flushes it. How about following patch?

```
index 72b0dae..9e4c89e 100644
@@ -428,12 +428,13 @@ module Fluent
         end
       end
 
-      def enqueue_all
+      # At flush_at_shutdown, all staged chunks should be enqueued for buffer flush. Set true to force_enqueue for it.
+      def enqueue_all(force_enqueue = false)
         log.on_trace { log.trace "enqueueing all chunks in buffer", instance: self.object_id }
 
         if block_given?
           synchronize{ @stage.keys }.each do |metadata|
-            return if queue_full?
+            return if !force_enqueue && queue_full?
             # NOTE: The following line might cause data race depending on Ruby implementations except CRuby
             # cf. https://github.com/fluent/fluentd/pull/1721#discussion_r146170251
             chunk = @stage[metadata]
@@ -443,7 +444,7 @@ module Fluent
           end
         else
           synchronize{ @stage.keys }.each do |metadata|
-            return if queue_full?
+            return if !force_enqueue && queue_full?
             enqueue_chunk(metadata)
           end
         end
diff --git a/lib/fluent/plugin/output.rb b/lib/fluent/plugin/output.rb
index 4a8b42d..61d9afd 100644
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1197,7 +1197,7 @@ module Fluent
 
       def force_flush
         if @buffering
-          @buffer.enqueue_all
+          @buffer.enqueue_all(true)
           submit_flush_all
         end
       end
```

mururu:
Looks good.

Issue #1918, opened 2018-04-02
Fluent::Plugin::Buffer#queued_num grows infinitely when when chunks are time-sliced


eagletmt:
- fluentd or td-agent version.
    - fluentd v1.1.2
- Environment information, e.g. OS.
    - Linux

## Problem
When the `buffer` section has `time` as chunk key like https://docs.fluentd.org/v1.0/articles/buffer-section#time , `Fluent::Plugin::Buffer#queued_num` seems to grow infinitely.

## Step to reproduce
Start fluentd with the following configuration.

```
% bundle exec fluentd --config fluentd.conf -v
2018-04-02 16:12:15 +0900 [info]: fluent/log.rb:322:info: parsing config file is succeeded path="fluentd.conf"
2018-04-02 16:12:15 +0900 [info]: fluent/log.rb:322:info: using configuration file: <ROOT>
  <source>
    @type forward
    @id input_forward
  </source>
  <source>
    @type debug_agent
    bind "127.0.0.1"
    port 26224
  </source>
  <match debug.echo>
    @type stdout
    <buffer tag,time>
      @type "memory"
      timekey 5s
      timekey_wait 5s
    </buffer>
  </match>
</ROOT>
2018-04-02 16:12:15 +0900 [info]: fluent/log.rb:322:info: starting fluentd-1.1.2 pid=18405 ruby="2.3.7"
2018-04-02 16:12:15 +0900 [info]: fluent/log.rb:322:info: spawn command to main:  cmdline=["/usr/share/rbenv/versions/2.3.7/bin/ruby", "-Eascii-8bit:ascii-8bit", "-rbundler/setup", "/home/eagletmt/gems/ruby/2.3.0/bin/fluentd", "--config", "fluentd.conf", "-v", "--under-supervisor"]
2018-04-02 16:12:16 +0900 [info]: fluent/log.rb:322:info: gem 'fluentd' version '1.1.2'
2018-04-02 16:12:16 +0900 [info]: fluent/log.rb:322:info: adding match pattern="debug.echo" type="stdout"
2018-04-02 16:12:16 +0900 [info]: fluent/log.rb:322:info: adding source type="forward"
2018-04-02 16:12:16 +0900 [info]: fluent/log.rb:322:info: adding source type="debug_agent"
2018-04-02 16:12:16 +0900 [info]: #0 fluent/log.rb:322:info: starting fluentd worker pid=18437 ppid=18405 worker=0
[...]

PR #1919, opened 2018-04-02
Delete empty queued_num field after purging chunks

eagletmt:
This should fix the memory-leak bug described in #1918.


repeatedly:
Thanks! Will check soon.

repeatedly:
Good catch. Thanks for the patch!

PR #1927, opened 2018-04-04
Support regexp type

okkez:
This PR supports following configuration:

```ruby
config_param :pattern, :regexp
```

See also https://github.com/fluent/fluentd/pull/1897#discussion_r178730491


repeatedly:
Thanks!

PR #1928, opened 2018-04-04
Fix CHANGELOG markups

eagletmt:
`<foo>` is eaten by markdown renderer.


repeatedly:
Thanks! Could you re-commit with `-s`?
We enabled DCO so need sign in the commit.

eagletmt:
OK, I did.

repeatedly:
Thanks!

PR #1943, opened 2018-04-13
prevent flushing twice

lala7573:
If flush_mode = interval and timekey = 5m, this would be flushed twice.
Though it is configuration mistake, I think it's also unexpected thing. 
Any suggestions welcome and let me know what I do not know:)

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1333,8 +1333,7 @@ module Fluent
         value_for_interval = nil
         if @flush_mode == :interval
           value_for_interval = @buffer_config.flush_interval
-        end
-        if @chunk_key_time
+        elsif @flush_mode == :lazy
           if !value_for_interval || @buffer_config.timekey < value_for_interval
             value_for_interval = [@buffer_config.timekey, @buffer_config.timekey_wait].min
           end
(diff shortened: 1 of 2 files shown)


repeatedly:
Did you test this patch?
I think your code has syntax error.

repeatedly:
BTW, could you paste your reproducible configuration here?

lala7573:
@repeatedly 
I corrected the wrong syntax. Thanks :)
As I mentioned, my configure file was like below. 
```
<buffer tag, time>
		@type               file
		flush_mode          interval
		timekey             5m
</buffer>
```
The default value of the flush_interval is 60 seconds. and there's no setting flush_interval outside the buffer tag.
Then I got the result applied both configure, timekey and flush_interval, means flush_mode=lazy and interval.
Every one minutes later, it's flushed and also timekey works.

The full configuration is like this.
```
<match sometag>
	@type						webhdfs
	namenode					hdfs_url
	append						no
	path						/some_path/%Y%m%d/%H/${hostname}.${chunk_id}.log
	<buffer tag, time>
		@type               file
		flush_mode          interval
		timekey             5m
		path                /var/log/td-agent/buffer
		chunk_limit_size    128mb
		flush_at_shutdown   true
		retry_forever       true
	</buffer>
</match>
```

lala7573:
@repeatedly
Hi, repeatedly
I think branching by flush_mode is more clear and it can prevent issue I mentioned. 
If I misunderstand about flush_mode and chunk_time_key, please let me know.

PR #1948, opened 2018-04-16
add dlq handler for buffered outputs to handle corrupt chunks

### LINKED ISSUE #1911: [Question] Partial Retries Support
Hi,

Any chance there is a way to do partial retries in Fluentd? Right now we maintain an output plugin https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud. For each buffer chunk the plugin processes, we might have some "good" entries and some "bad" entries. We'd like to enable users to flush the "good" entries right away but retry the "bad" entries later.

In the current implementation, seems that an output plugin can only throw an exception to indicate that this chunk is not processed and should be retried. The problem is that we either block the "good" entries as well, or we retry the "good" entries again and generate duplicate logs.

Is there a known way to do partial retries correctly with the built-in support?

Alternatively we are thinking about writing a new filter plugin to handle this case or dig into the core code and contribute / maintain a fluentd fork with partial retries.

Wanna get some sanity check before we went too far away unnecessarily.

Thanks,
Ling

jcantrill:
This PR is the upstream equivalent to https://github.com/uken/fluent-plugin-elasticsearch/pull/398 which:

* Provides a DLQ handler for corrupt buffer chunks.  
* Handler is 'opt-in' for buffered output implementations

This is to resolve [known issues](https://bugzilla.redhat.com/show_bug.cgi?id=1562004) where the chunks are corrupt

Possibly fixes #1911

### DIFF
--- a/lib/fluent/plugin/dead_letter_queue_drop_handler.rb
+++ b/lib/fluent/plugin/dead_letter_queue_drop_handler.rb
@@ -1,0 +1,15 @@
+module Fluent
+  module Plugin
+    module DeadLetterQueueDropHandler
+
+      def handle_chunk_error(tag, error, time, record)
+        begin
+          log.error("Dropping record from '#{tag}': error:#{error} time:#{time} record:#{record}")
+        rescue=>e
+          log.error("Error while trying to log and drop message from chunk '#{tag}' #{e.message}")
+        end
+      end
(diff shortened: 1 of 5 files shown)


jcantrill:
cc @portante @richm

Issue #1856, opened 2018-02-14
Output: Backup for bad chunk


repeatedly:
Fluentd's output plugin somtimes hit un-recoverable error during chunk flush.

- chunk contains wrong record for output configuration
- output plugin has a bug for specific record
- broken chunk is generated by hardware problem
- wrong setup for destination

Currently, we use retry limit and secondary for handling these chunks but it has several problems.

- bad chunk occupy flush threads until reach retry limit
- non retry limit environment can't rescue bad chunk

So we should care bad chunk for stability and performance.
The idea is if output plugin raises un-recoverable error during chunk flush, such chunks are routed to backup directory.

- The recoverable errors are TypeError, NoMethodError, ArgumentError, etc
- Use UnrecoverableError for plugin specific error

In addition, `<system>` directive provide `backup_dir` parameter. The default is `/tmp/fluentd`.

mururu:
Is secondary ignored when `UnrecoverableError` is thrown? The behaviour looks good for "broken chunk" case, but not good for "wrong setup for destination" case.
I think that skipping retry of the output plugin is enough when `UnrecoverableError` is thrown. It is because secondary output plugin can also raise `UnrecoverableError` if they can't handle the chunk.

artbeglaryan:
Hello @repeatedly. Maybe this is the wrong place, but I didn't find any information in existing issues or in documentation. I have my own output plugin, which do some specific job with buffered data via network. Sometimes there may be some issue with network and my plugin failed to flush data after reaching retry_wait, retry_max_interval, retry_max_times and put that buffer chunk in secondary output plugin which is file. So I didn't lose any data. I have that chunks in my backup directory. Is there any way(I really didn't find anywhere documented or answered) to process that chunks later with the same plugin? by hand, by some external command or maybe there is some existing plugin which can process that data again?

repeatedly:
@artbeglaryan Currently, writing script is better for it. Here is an example:

https://groups.google.com/d/msg/fluentd/6Pn4XDOPxoU/CiYFkJXXfAEJ

I will add this to documetation.

PR #1952, opened 2018-04-18
output: Add backup feature for bad chunks

repeatedly:
Fix https://github.com/fluent/fluentd/issues/1856

Current implementation covers most bad chunk cases. I need feedback before writing test.

-  use `root_dir` parameter for backup directory when specified, `/${root_dir}/backup/worker${worker_id}/${plugin_id}/${chunk_id}.log`.
- Call secondary directly when unrecoverable error happens

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1057,7 +1057,7 @@ module Fluent
         end
       end
 
-      UNRECOVERABLE_ERRORS = [TypeError, ArgumentError, NoMethodError, Fluent::UnrecoverableError]
+      UNRECOVERABLE_ERRORS = [Fluent::UnrecoverableError, TypeError, ArgumentError, NoMethodError]
 
       def try_flush
         chunk = @buffer.dequeue_chunk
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -1115,6 +1149,21 @@ def try_flush
         end
       end
 
+      def backup_chunk(chunk, using_secondary, delayed_commit)
+        unique_id = dump_unique_id_hex(chunk.unique_id)
+        safe_plugin_id = plugin_id.gsub(/[ \/\\:]/, '_')
```
cosmo0920: This feature assumes to work on Windows?
If so, we should add more prohibiting to use characters:

`/ > < ? : " \ * | ;` are not permitted to use on NTFS.

cosmo0920:
I've added a comment about concerns on Windows environment.

JohnMealy23:
The usage of this feature is hard to track down.  Can it be added to the docs?

ganmacs:
@repeatedly Could you add the document about it. or had you already added?

repeatedly:
Yes: https://docs.fluentd.org/buffer#handling-unrecoverable-errors

jvshahid:
I Apologize for replying to an old issue. I am curious if there is a way to monitor the number of bad chunks, e.g., using Prometheus?

PR #1963, opened 2018-04-28
change read_pos and read_ino. relate #1953 #1804

jl2005:
Signed-off-by: 焦龙 <jiaolongdy@163.com>

Issue #1804, opened 2017-12-27
can't add a new key into hash during iteration


jl2005:
- fluentd or td-agent version.

  fluentd-0.12.42

- Environment information, e.g. OS.

  Kubernetes DaemonSet

- Your configuration

```
   <source>
      @type tail
      tag kubernetes.*
      path /var/log/containers/*.log
      pos_file /var/log/fluentd-containers.log.pos
      format json
      time_key not-parse-time
      read_from_head true
    </source>
```

- Your problem explanation. If you have an error logs, write it together.

```
2017-12-26 07:41:05 +0000 [error]: can't add a new key into hash during iteration
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:247:in `block in start_watchers'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:228:in `each'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:228:in `start_watchers'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:214:in `refresh_watchers'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:577:in `on_timer'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/cool.io-1.5.3/lib/cool.io/loop.rb:88:in `run_once'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/cool.io-1.5.3/lib/cool.io/loop.rb:88:in `run'
  2017-12-26 07:41:05 +0000 [error]: /var/lib/gems/2.3.0/gems/fluentd-0.12.42/lib/fluent/plugin/in_tail.rb:316:in `run'
2017-12-26 11:14:05 +0000 [error]: can't add a new key into hash during iteration
  2017-12-26 11:14:05 +0000 [error]: suppressed same stacktrace
```

repeatedly:
Does this happen during shutdown?

[...]

PR #1966, opened 2018-04-30
make FilePositionEntry thread-safe

kazegusuri:
ref https://github.com/fluent/fluentd/issues/1804 and https://github.com/fluent/fluentd/pull/1963

fluent-plugin-protometheus calls `FilePositionEntry#read_pos` and `FilePositionEntry#read_inode` by using `instance_variable_get` to get current position and inode as metrics. I understand it is unsupported use.
FilePositionEntry's methods are not thread-safe because each methods share `@file` to read data from file. So race happens between in_tail plugin and fluent-plugin-protometheus plugin.
But now I found instance variable `@pe` of `Fluent::Plugin::TailInput::TailWatcher` is accessible by outside. So I think it is better to be thread-safe to use FilePositionEntry without considering my usecase.
How do you think?

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -912,32 +912,43 @@ module Fluent::Plugin
         @file = file
         @seek = seek
         @pos = nil
+        @mutex = Mutex.new
       end
 
       def update(ino, pos)
-        @file.pos = @seek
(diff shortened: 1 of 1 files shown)


kazegusuri:
Oh, part of tests failed. Is is flaky?

PR #1984, opened 2018-05-15
Fix DataError exception

### LINKED ISSUE #1903: buffer flush failed by unexpected Zlib::DataError
fluted v1.0,this is my configure rule:

```
<match myprofile.*.*>
  @type file
  path /var/log/td-agent/${tag[1]}/%Y%m/${tag[2]}/myprofile.${tag[2]}.%Y%m%d
  compress gzip
  append true
  <buffer tag, time>
    @type memory
    flush_mode interval
    flush_interval 30s 
  </buffer>
  <format>
    @type single_value
  </format>
</match>
```

there are so many warn log like this:
2018-03-19 00:18:51 +0900 [warn]: #0 failed to flush the buffer. retry_time=0 next_retry_seconds=2018-03-19 00:18:52 +0900 chunk="567b15a13fd0f003c72c7e0e534dbf3e" error_class=Zlib::DataError error="data error"
  2018-03-19 00:18:51 +0900 [warn]: #0 suppressed same stacktrace
2018-03-19 00:18:52 +0900 [warn]: #0 retry succeeded. chunk_id="567b15a13fd0f003c72c7e0e534dbf3e"

the result file cannot open correctly:
gzip: invalid compressed data--crc error
gzip: invalid compressed data--length error
what happend??

jl2005:
Fixed #1903  #1847

This exception is caused by `submit_flush_once` wake up the output thread and interrupting gzip compression, causing a `DataError` exception to be thrown in the `close`.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -153,7 +153,7 @@ module Fluent
       end
 
       # Internal states
-      FlushThreadState = Struct.new(:thread, :next_clock)
+      FlushThreadState = Struct.new(:thread, :next_clock, :mutex, :cond)
       DequeuedChunkInfo = Struct.new(:chunk_id, :time, :timeout) do
         def expired?
           time + timeout < Time.now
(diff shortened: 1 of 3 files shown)


repeatedly:
This patch has a bug of thread control so I wrote new patch based on this change and merged.

https://github.com/fluent/fluentd/pull/2170/commits/1c79a234ef7711bf1a77771dba458481aec3ec5d

Thanks for the contribution!

PR #1995, opened 2018-05-21
out_file: Temporal fix for broken gzipped files with gzip and append. ref #1903

repeatedly:
Direct append causes broken gzipped file when Zlib::DataError happens.
This patch uses Tempfile for gzip compression to avoid broken compression.


repeatedly:
@nurse How about this code?
Of course, we need to investigate zlib issue further.

PR #1997, opened 2018-05-21
counter: Add wait API

repeatedly:
Similar to "get" but raises an error when API call has a problem.


repeatedly:
If no objection, I will merge this patch.

PR #2001, opened 2018-05-23
Allow escaping "," in paths for tail input

herrsebi:
Currently there is no way to tail files that include a comma (,) in their file name. Switching to a neg-lookbehind regex should enable this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,8 @@ submitting an issue to Fluentd. Even better you can submit a Pull Request with a
 If you find a bug of 3rd party plugins, please submit an issue to each plugin repository.
 And use [omnibus-td-agent](https://github.com/treasure-data/omnibus-td-agent) repository for td-agent releated issues.
 
+Note: Before report the issue, check latest version first. Sometimes users report fixed bug with older version.
+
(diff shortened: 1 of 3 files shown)


repeatedly:
Could you show me the path example?

herrsebi:
Sure: `/tmp/test,dev.log` would be a vaild file name on unix system. In windows environments, a long filename (longer than 8 chars) may also contain a comma.

repeatedly:
Hmm... support special case is not good for me.
How about using dummy's message like approach?

https://github.com/fluent/fluentd/blob/ced6387cd8f00feede48de2ceb005d72f9a477e9/lib/fluent/plugin/in_dummy.rb#L42

For ',' or other special case, use `path ["p,ath1", "path2"]`.
This approach keeps to support `path path1,path2` string parameter with `split(',')`.

herrsebi:
I've added an alternative 'paths' config setting, that would allow the form of configuration that you mentioned previously.
So now it's either 
` path: "test.log, test2.log"`
or
` paths: ["test.log", "test2.log"]`

The latter form can then also include commas.

mkoerperriag:
Can I help somehow to get this further?

PR #2003, opened 2018-05-24
Dump config_argument by dump_config_definition

okkez:
In previous version, fluent-plugin-config-format command does not dump
information defined by config_argument. We cannot reconstruct a plugin
config definition from dumped config.


repeatedly:
One test fails. Could you check it?

repeatedly:
Thanks!

PR #2007, opened 2018-05-31
Fix mistook weblink in README

chigichan24:
I think the link about fluentd was mistook (may be old link). 
So did small change and send a PullRequest for README.


repeatedly:
Thx!

PR #2020, opened 2018-06-14
in_http: Honor `keep_time_key` parameter in batch mode.

fujimotos:
This patch fixes the issue reported by Eugen Cristea at fluent/fluentd-docs#507.
### Problem

`in_http` plugin supports "batch" mode for sending multiple events in a single HTTP request.
The problem is that in_http does not honor the `keep_time_key` option when processing
those bulk-insertion requests.

This means that "time" key in each record gets discarded unconditionally in batch mode.
For example, the following request:

    $ curl -d '[{"time":1514736000,"key":"foo"}]' http://localhost/

will produces a log record:

    2018-01-01 01:00:00.000000000 +0900 : {"key":"foo"}

... even if the `keep_time_key` option is enabled.

### Solution

This patch fixes this issue by teaching `in_http` to honor the option value of its parser
plugin instance.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -202,11 +202,17 @@ module Fluent::Plugin
             if @add_remote_addr
               single_record['REMOTE_ADDR'] = params['REMOTE_ADDR']
             end
-            single_time = if t = single_record.delete('time')
-                            Fluent::EventTime.from_time(Time.at(t))
-                          else
-                            time
-                          end
+
+            if single_record.has_key?('time')
+                single_time = Fluent::EventTime.from_time(Time.at(single_record['time']))
+            else
+                single_time = time
+            end
+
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_http.rb
```diff
@@ -202,11 +202,17 @@ def on_request(path_info, params)
...
-                            time
-                          end
+
+            if single_record.has_key?('time')
+                single_time = Fluent::EventTime.from_time(Time.at(single_record['time']))
+            else
+                single_time = time
+            end
+
+            unless @parser and @parser.keep_time_key
```
repeatedly: This condition can be moved to under above `if single_record._has_key?`

fujimotos: I fixed this point by 39f3a0da

repeatedly:
Could you add test?

PR #2025, opened 2018-06-21
in_http: Prefer parser's time_key setting in batch. fix #2018

### LINKED ISSUE #2018: HTTP input plugin failed to parse array of json when there is "time" property that is not a number
Fluentd Version: 1.2.1
Environment: Windows 10
Configuration: 
```
<source>
  @type http
  port 8887
  <parse>
    @type json
    time_key nil
  </parse>
</source>

<match **>
  @type stdout
</match>
```
Problem:
Running this command `curl -X POST -d '[{"foo":"bar", "time": "2018-06-12T18:35:48.1488850Z"},{"abc":"def"},{"xyz":"123"}]' http://localhost:8887/app.log
` gives me this error `can't convert String into an exact number
`
This parsing logic should respect the `time_key nil` in the setting. It's not a problem if I'm only sending one json object like `curl -X POST -d '{"foo":"bar", "time": "2018-06-12T18:35:48.1488850Z"}' http://localhost:8887/app.log`

repeatedly:


test/plugin/test_in_http.rb
```diff
@@ -162,6 +162,29 @@ def test_multi_json_with_time_field
     assert_equal_event_time time, d.events[1][1]
   end
 
+  def test_multi_json_with_time_key
```
cosmo0920: `test_multi_json_with_time_key` seems to be testing for existent time_key.
I perfer to use `test_multi_json_with_nonexistent_time_key` or `test_multi_json_with_missing_time_key` for test case name.

repeatedly:
Apply review. Merged.

PR #2026, opened 2018-06-21
Fix Parser plugin template for fluent-plugin-generate

hana-da:
In Parser plugin template for fluent-plugin-generate, parse method definition is in `Fluent::Plugin` module.
I think it should be in `Fluent::Plugin::<class_name>` class.


templates/new_gem/lib/fluent/plugin/parser.rb.erb
```diff
@@ -6,10 +6,10 @@ module Fluent
   module Plugin
     class <%= class_name %> < Fluent::Plugin::Parser
       Fluent::Plugin.register_parser("<%= plugin_name %>", self)
-    end
 
-    def parse(text)
-      yield time, record
+      def parse(text)
```
cosmo0920: :+1:

repeatedly:
Thanks!

PR #2037, opened 2018-06-26
monitor_agent: Add missing descriptions for configurable options.

fujimotos:
### What's this patch?

This is a minor patch that adds documentation for each configurable
option of Fluent::Plugin::MonitorAgentInput, which is currently
missing from the source code.


repeatedly:
Thx!

PR #2041, opened 2018-06-27
test_out_file.rb: Fix test failures due to Zlib::DataError.

fujimotos:
AppVeyor CI tests are constantly failing due to Zlib-related exceptions
like below:

    Error: test: append(FileOutputTest): Zlib::DataError: invalid
    distance too far back

This is a very weird error, because the file seems to contain a proper
gzip-compressed data stream. In fact, by inserting the following code
before the failing line, I can confirm that the file itself is NOT
broken:

```ruby
bytes = File.open(path).read   // Read the file being tested
puts(Zlib::GzipReader.new(StringIO.new(bytes)).read)
```

The strong suspicion is that Ruby's Zlib wrapper (or its build on
AppVeyor) has a serious bug that corrupts file pointers passed to
GzipReader, thus it ends up reading garbage data.

This patch is my attempt to mitigate the issue by modifying the test
functions to pass bytes (rather than file objects) to GzipReader, and
it should protect zlib from read-time data corruptions.

### DIFF
--- a/test/plugin/test_out_file.rb
+++ b/test/plugin/test_out_file.rb
@@ -304,7 +304,11 @@ class FileOutputTest < Test::Unit::TestCase
       assert_equal r4, d.formatted[3]
       assert_equal r5, d.formatted[4]
 
-      read_gunzip = ->(path){ File.open(path){|fio| Zlib::GzipReader.open(fio){|io| io.read } } }
+      read_gunzip = ->(path){
+        File.open(path){ |fio|
+          Zlib::GzipReader.new(StringIO.new(fio.read)).read
+        }
+      }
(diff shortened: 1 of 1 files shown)


fujimotos:
Related links for the actual error logs:

- https://ci.appveyor.com/project/fluent/fluentd/build/3021/job/6gp3tu2f3qimeyhy#L158
- https://ci.appveyor.com/project/fluent/fluentd/build/3020/job/k6ad5v0mwfdcmpem#L147
- https://ci.appveyor.com/project/fluent/fluentd/build/3019/job/pedwsa74ne4kjfhf#L275

repeatedly:
Thx for mitigating weird test failure.

PR #2045, opened 2018-06-28
ssl on fluentd.org does not appear to work

plicease:
The link in rubygems doesn't seem to work for me because `fluentd.org` isn't listening to the https port.  `fluentd.org:80` does forward on to `https://www.fluentd.org`, but that isn't what is in the gemspec file.  Could change it to use http in the gemspec file maybe but better to just https everything I think.


repeatedly:
Good catch. Thanks!

PR #2049, opened 2018-07-02
fixing regex of pid/procid field for conformance to RFC5424 spec

mdalley:
Per RFC5424, the field for PROCID is either a NILVALUE (-) or ASCII field from 1 to 128 characters.
https://tools.ietf.org/html/rfc5424#section-6

Valid RFC5424 formatted log using alphanumeric PROCID value of 'proc1':
<16>1 2013-02-28T12:00:00.003Z 192.168.0.1 fluentd __proc1__ ID24224 [exampleSDID@20224 iut="3" eventSource="Application" eventID="11211"] Hi, from Fluentd!

The regex in this PR commit will properly match the example above.

Also changed the _match_ field name from __pid__ to __procid__ in the REGEX to line up with RFC5424.


repeatedly:
Don't change capture name because it breaks existing users, e.g. schema mismatch.

mdalley:
Good point.  I will resubmit PR to include just the pattern match change and leave the match name 'pid' unchanged.

mdalley:
Replaced by #2051

PR #2051, opened 2018-07-03
update regex of pid field for conformance to RFC5424 spec

mdalley:
Per RFC5424, the field for PROCID is either a NILVALUE (-) or ASCII field from 1 to 128 characters.
https://tools.ietf.org/html/rfc5424#section-6

Valid RFC5424 formatted log using alphanumeric PROCID(PID) value of '__proc1__':
<16>1 2013-02-28T12:00:00.003Z 192.168.0.1 fluentd __proc1__ ID24224 [exampleSDID@20224 iut="3" eventSource="Application" eventID="11211"] Hi, from Fluentd!

The regex in this PR commit will properly match the example above.

Replacing PR #2049


repeatedly:
Merged. Thanks!

Next, please update commits with `push -f`. No need to create new PR for update.

PR #2054, opened 2018-07-05
Consider timezone when calculate timekey

### LINKED ISSUE #1986: file timezone problem
Fluentd version: td-agent 1.0.2

I want to transfer logfile from CentOS to CentOS by fluentd.
I could transfer file but file switches at incorrect time.
My timezone is +0900.

- output files (ls -l)
 rw-r--r-- 1 root root  34201618  5月 15 09:02 test-log.20180514
 rw-r--r-- 1 root root  40060437  5月 16 09:01 test-log.20180515
 rw-r--r-- 1 root root   5943936  5月 16 10:47 test-log.20180516

I want files switch at 00:00 but actually they switch at 09:00 everyday.
When I set time_slice_format = %Y%m%d_%H files switch correctly. (test.log.20180516_01 is created at 01:00 )

how can I solve it?

config (input) :
```
<source>
 type tail
 format none
 path /var/log/test-log
 pos_file /var/log/td-agent/test-log.pos
 tag hostfrom.test-log
</source>

<match *.*>
 type forward
 <server>
  host hostlog
  port 24224
 </server>
</match>
```

config (output) :
```
<source>
 @type forward
 bind 0.0.0.0
 port 24224
</source>

<match hostfrom.test-log>
 type file
 format single_value
 buffer_path /var/log/td-agent/buffer/test-log.buf
 path /log/test-log
 append true
 path_suffix
</match>
```

okkez:
Because the timestamp of the midnight cannot be divisible by 86400 in
localtime (JST). The timestamp of the midnight is divisible by 86400
only in UTC. Therefore we must consider offset from UTC in localtime.

For example, at `2018-07-04 01:23:23 +0900`:

If timekey is 86400 and path template is `/log/%Y%m%d.log`.
In previous version, extract path template to `/log/20180703.log`.
In this version, extract path template to `/log/20180704.log`.

Fix #1986

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -803,25 +803,32 @@ module Fluent
           if !@chunk_key_time && !@chunk_key_tag
             @buffer.metadata()
           elsif @chunk_key_time && @chunk_key_tag
-            time_int = time.to_i
-            timekey = (time_int - (time_int % @buffer_config.timekey)).to_i
+            timekey = calculate_timekey(time)
             @buffer.metadata(timekey: timekey, tag: tag)
           elsif @chunk_key_time
-            time_int = time.to_i
-            timekey = (time_int - (time_int % @buffer_config.timekey)).to_i
+            timekey = calculate_timekey(time)
             @buffer.metadata(timekey: timekey)
           else
             @buffer.metadata(tag: tag)
           end
         else
           timekey = if @chunk_key_time
-                      time_int = time.to_i
-                      (time_int - (time_int % @buffer_config.timekey)).to_i
+                      calculate_timekey(time)
                     else
                       nil
                     end
           pairs = Hash[@chunk_key_accessors.map { |k, a| [k, a.call(record)] }]
           @buffer.metadata(timekey: timekey, tag: (@chunk_key_tag ? tag : nil), variables: pairs)
+        end
+      end
+
+      def calculate_timekey(time)
+        time_int = time.to_i
+        if @buffer_config.timekey_use_utc
+          (time_int - (time_int % @buffer_config.timekey)).to_i
+        else
+          offset = Time.at(time_int).utc_offset
+          (time_int - ((time_int + offset)% @buffer_config.timekey)).to_i
         end
       end
 
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -829,7 +829,7 @@ module Fluent
         if @timekey_use_utc
           (time_int - (time_int % @timekey)).to_i
         else
-          offset = Time.at(time_int).utc_offset
+          offset = Fluent::Timezone.utc_offset(time, @timekey_zone)
           (time_int - ((time_int + offset)% @timekey)).to_i
         end
       end
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -304,6 +304,8 @@ module Fluent
             raise Fluent::ConfigError, "<buffer ...> argument includes 'time', but timekey is not configured" unless @buffer_config.timekey
             Fluent::Timezone.validate!(@buffer_config.timekey_zone)
             @timekey_zone = @buffer_config.timekey_use_utc ? '+0000' : @buffer_config.timekey_zone
+            @timekey = @buffer_config.timekey
+            @timekey_use_utc = @buffer_config.timekey_use_utc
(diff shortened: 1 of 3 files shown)


repeatedly:
Test failed. Could you check it?

test/plugin/test_out_file.rb
```diff
@@ -557,6 +557,42 @@ def parse_system(text)
...
+      formatted_lines = %[2011-01-02T03:14:15+09:00\ttest\t{"a":1}\n] + %[2011-01-02T03:14:15+09:00\ttest\t{"a":2}\n]
+
+      write_once = ->(){
+        d = create_driver %[
+          path #{TMP_DIR}/out_file_test
+          compress gz
+          append true
+          <buffer>
+            timekey_use_utc false
+            timekey_zone Asia/Tokyo
```
repeatedly: Could you use "+0900" like value?
I'm not sure but test on Windows fails.

repeatedly: Just commit the fix :)

lib/fluent/plugin/output.rb
```diff
@@ -825,6 +822,16 @@ def metadata(tag, time, record)
         end
       end
 
+      def calculate_timekey(time)
+        time_int = time.to_i
+        if @buffer_config.timekey_use_utc
+          (time_int - (time_int % @buffer_config.timekey)).to_i
+        else
+          offset = Time.at(time_int).utc_offset
```
repeatedly: Should consider `timekey_zone` value instead of `utc_offset`?

okkez: Maybe yes. How about the following function?

```diff
diff --git a/lib/fluent/timezone.rb b/lib/fluent/timezone.rb
index bb6fa868..0e79dd03 100644
--- a/lib/fluent/timezone.rb
+++ b/lib/fluent/timezone.rb
@@ -139,5 +139,17 @@ module Fluent
 
       return nil
     end
+
+    def self.utc_offset(time, timezone)
+      return 0 if timezone.nil?
+
+      case timezone
+      when NUMERIC_PATTERN
+        Time.zone_offset(timezone)
+      when NAME_PATTERN
+        tz = TZInfo::Timezone.get(timezone)
+        tz.period_for_utc(time).utc_total_offset
+      end
+    end
   end
 end
```

repeatedly: looks good

lib/fluent/plugin/output.rb
```diff
@@ -825,6 +822,16 @@ def metadata(tag, time, record)
         end
       end
 
+      def calculate_timekey(time)
+        time_int = time.to_i
+        if @buffer_config.timekey_use_utc
```
repeatedly: accessing object member is slower than accessing instance variable.
`@timekey_use_utc = @buffer_config.timekey_use_utc` in `configure` is better.

lib/fluent/plugin/output.rb
```diff
@@ -825,6 +824,16 @@ def metadata(tag, time, record)
         end
       end
 
+      def calculate_timekey(time)
+        time_int = time.to_i
+        if @timekey_use_utc
+          (time_int - (time_int % @timekey)).to_i
+        else
+          offset = Fluent::Timezone.utc_offset(time, @timekey_zone)
```
repeatedly: Check `@timekey_zone` format everytime is slower.
This code is called in each emit so need to avoid extra cost.
Cache the result is better.

test/plugin/test_out_file.rb
```diff
@@ -557,6 +557,42 @@ def parse_system(text)
     check_gzipped_result(path, formatted_lines * 3)
   end
 
+  test 'append when JST' do
```
repeatedly: Could you add more test with different time and timezone, e.g. `timekey_zone` is `+09:00` and actual time is `+02:00`

lib/fluent/plugin/output.rb
okkez: Hmm, we should consider DST. The naive cache is not good for DST.
What should we do?

repeatedly: I see.
One way is generating `calculate_timekey` method in `configure`.
For numeric pattern, the offset value can be embbed directly.
For name pattern, embed getting tz data in the body.
How about this?

repeatedly:
Thanks!

repeatedly:
I released v1.2.4.rc1 for testing this patch.

Issue #2035, opened 2018-06-25
Http input plugin doesn't hornor parser time_format in batched scenario


yantang-msft:
Fluentd Version: 1.2.1
Environment: Windows 10
Configuration: 
```
<source>
  @type http
  port 8887
  <parse>
    @type json
    time_format %iso8601
  </parse>
</source>

<match **>
  @type stdout
</match>
```
Problem:
Running this command `curl -X POST -d '[{"foo":"bar", "time": "2018-06-12T18:35:48.1488850Z"},{"abc":"def"},{"xyz":"123"}]' http://localhost:8887/app.log
` gives me this error `can't convert String into an exact number
`

It's very similar to this issue https://github.com/fluent/fluentd/issues/2018 I reported earlier. The difference is this time it's the time_format parameter instead of the time_key.
With this fix https://github.com/fluent/fluentd/pull/2025, I can workaround the issue by setting time_key to nil, so it's not an urgent for me. But still it would nicer if the time_format parameter is also supported in batched scenario.

bitc:
I am seeing this too. Sending a single JSON event works, but sending a JSON array with multiple events triggers this error

yantang-msft:
@fujimotos I confirm the patch fixes this issue. 😄

fujimotos:
Now it has been merged in to the master branch today.
Thanks for reporting the issue and testing the patch.

PR #2055, opened 2018-07-05
in_http: Fix "ignored parser options" problem on bulk insertion

fujimotos:
This patch fixes the bug reported by #2035 (and other related issues as well).

### Problem

What used to happen is that options in the `<parser>` section get ignored
when we send events as a JSON array. This occurs because parser plugins
are designed to handle a single hash value and do not generallly handle
an array input well.

Until now, we have tried to solve this issue by emulating the semantics
of parser plugins in Fluent::Plugin::HttpInput (see 1afbfb1, 39f3a0d
and f560017c). However, this approach turned out to be error prone and
rather tedious.

### More Generalized solution

This patch takes a different path:

 - Whenever `@parser` is available, reuse `@parser.convert_values()` by
   manually applying it to each record in an input array.
 - Otherwise, fall back to the current logic.

The good part of this approach is that we do not need to duplicate the same
code in `in_http` and `parser_json` anymore, and it fixes this class of related
issues.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -209,13 +209,15 @@ module Fluent::Plugin
               single_record['REMOTE_ADDR'] = params['REMOTE_ADDR']
             end
 
-            if single_record.has_key?(@parser_time_key)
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks!

Issue #2047, opened 2018-07-02
Compression support in http input/output plugin


kishoreck:
Summary:

When http is used as a medium of transport for large volumes of data, it will be helpful to have compression support in place. Data compression support in buffers and forward plugin is already made available. It will be a nice to have the compression support in the http plugins as well.

repeatedly:
So you want to send large data with gzip to in_http, right?

kishoreck:
Yup. That's correct.

repeatedly:
Merged #2060

kishoreck:
Awesome. Thanks👍

boumitran:
we are looking forward to have a fluentd release with this in_http compression support. When is the next release schedule?

rockliffelewis:
Hi,

Looking at the Pull Request #2060 , it would appear that this was only added to `in_http` and not the `out_http` plugin as well?

It would be most excellent if this could be added in, for http based fluentd to fluend setups.

daipom:
Looks like it is not added to the `out_http`.
Could you make a new issue for the feature request?

PR #2060, opened 2018-07-06
in_http: Allow users to send payload in compressed form

fujimotos:
This adds support to `in_http` for the 'Content-Encoding' header,
which allows HTTP clients to transfer payload more efficiently by
applying compression (Currently it supports `gzip` and `deflate`).

You can test it using the following command:

```bash
$ echo 'json={"foo":"bar"}' | gzip > json.gz
$ curl --data-binary @json.gz -H "Content-Encoding: gzip" \
       http://localhost:9880/debug.log
````

This feature should be useful especially for users who handle large
amounts of data.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -316,6 +316,7 @@ module Fluent::Plugin
         end
         @env = {}
         @content_type = ""
+        @content_encoding = ""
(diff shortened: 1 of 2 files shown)


fujimotos:
Note: this patch should solve the issue #2047.

repeatedly:
Looks good. Thanks!

Issue #2062, opened 2018-07-06
in_tail: Errno::ENOENT occurs at File.mtime()


t-osakada:
```
2018-06-11 20:30:04 +0900 [error]: #0 Unexpected error raised. Stopping the timer. title=:in_tail_refresh_watchers error_class=Errno::ENOENT error="No such file or directory @ rb_file_s_mtime - /var/log/app/201806042000.log"
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:221:in `mtime'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:221:in `block (2 levels) in expand_paths'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:218:in `select'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:218:in `block in expand_paths'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:215:in `each'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:215:in `expand_paths'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin/in_tail.rb:251:in `refresh_watchers'
  2018-06-11 20:30:04 +0900 [error]: #0 /var/opt/fluentd/vendor/bundle/ruby/2.3.0/gems/fluentd-1.1.3/lib/fluent/plugin_helper/timer.rb:80:in `on_timer'
[...]

PR #2063, opened 2018-07-06
fix to rescue Errno::ENOENT for File.mtime()

t-osakada:
This patch fixes the https://github.com/fluent/fluentd/issues/2062 issue.


repeatedly:
Could you recommit with `-s`?
We enabled DCO and we need your sign in commits.

t-osakada:
@repeatedly

I added Signed-off-by line at the commit log.
Could you check this again?

repeatedly:
Thanks!

Issue #2067, opened 2018-07-12
Input plugin amplifies events when buffer limit is exceeded


patrobinson:
- fluentd or td-agent version: 1.2 and 0.12.41
- Environment information, e.g. OS. Linux
- Your configuration See https://github.com/envato/reproduce-fluentd-cascading-buffer

## Behaviour

When using docker logs plugin with Fluentd the agent can get into an unrecoverable situation where the buffer is full and further events emitted cause the buffer to remain full and never empty. The repository linked above has a complete example which replicates the behaviour seen, run `docker-compose up` to see it in action. We are using the Kinesis Plugin, provided by Amazon, which in testing is able to flush about 2000 events per second. If this is unavailable for some time, due to say too many events being sent to the stream, then the internal Fluentd buffer becomes full and an exception is raised (as per the default behaviour) which is handled by the input plugin. The input plugin throws an error and closes the connection to the docker daemon. This results in 7-9 extra events (depending on the version of Fluentd) being emitted to the output plugin for every event that the docker daemon attempts to emit.

Fluentd 1.2 logs:
```
fluentd_1       | 2018-07-12 04:53:27 +0000 [error]: #0 failed to emit fluentd's log event tag="fluent.warn" event={"action"=>:throw_exception, "message"=>"failed to write data into buffer by buffer overflow action=:throw_exception", "time"=>"2018-07-12T04:53:26Z"} error_class=Fluent::Plugin::Buffer::BufferOverflowError error="buffer space has too many data"
fluentd_1       | 2018-07-12 04:53:27 +0000 [warn]: #0 failed to write data into buffer by buffer overflow action=:throw_exception
fluentd_1       | 2018-07-12 04:53:27 +0000 [warn]: #0 failed to write data into buffer by buffer overflow action=:throw_exception
fluentd_1       | 2018-07-12 04:53:27 +0000 [warn]: #0 emit transaction failed: error_class=Fluent::Plugin::Buffer::BufferOverflowError error="buffer space has too many data" location="/usr/lib/ruby/gems/2.4.0/gems/fluentd-1.2.3/lib/fluent/plugin/buffer.rb:269:in `write'" tag="send.logs"
fluentd_1       |   2018-07-12 04:53:27 +0000 [warn]: #0 suppressed same stacktrace
[...]

PR #2070, opened 2018-07-13
Emit less logs

patrobinson:
In an attempt to tackle the issues in #2067 this PR reduces the number of errors and warnings printed out when the output buffer becomes full and raises an exception.


patrobinson:
pingj @repeatedly

ganmacs:
Such an error log is pretty important for all developers to identify the error reason.
Why do you consider these changes to be needed?

patrobinson:
@ganmacs see the issue described in #2067 
Effectively during a downstream outage these error messages reduce the maximum throughput 10 fold and results in a cascading failure where the buffer is always full of error messages and is never able to recover

ganmacs:
Yes. but this cannot be the reason to remove error logs.
If they aren't necessary,  I'll approve of this change. but they look necessary.

The expected behavior you mentioned in #2067 seems the correct solution to tackle this issue.

patrobinson:
@ganmacs That's fair. I'm happy to close that in favour of that solution, but I don't have the ability to work on it right now

PR #2087, opened 2018-07-26
server-helper: Fix connection leak by close timing issue. fix #2086

### LINKED ISSUE #2086: Memory leaks when running fluentd in TLS mode
- Fluentd  version: version 1.1
- Environment information : Ubuntu 16.01
- Fluentd configuration used below

---- fluentd config -----
```
<source>
  @type monitor_agent
  bind 0.0.0.0
  port 24220
</source>

<source>
  @type forward
  @id forward
  port 24225
  bind 0.0.0.0

  @log_level debug
  <security>
    self_hostname myhost
    shared_key veryverysecret
  </security>
  <transport tls>
    version TLSv1_2
    ca_path  /home/ubuntu/client_auth/certs/ca.crt.pem
    cert_path  /home/ubuntu/client_auth/certs/server.crt.pem
    private_key_path  /home/ubuntu/client_auth/private/server.key.pem
    client_cert_auth true
  </transport>
</source>

 <filter logs.**>
     @type record_transformer
     @id record_transformer
     enable_ruby true
     <record>
         # the timestamp field has be named as @timestamp for kibana to work properly.
         @timestamp ${time.strftime('%Y %b %d %H:%M:%S')}
     </record>
 </filter>

<match logs.**>
  @type copy
  @id copy
  <store>
   @type stdout
   @id stdout
  </store>
</match>
```
- When I run fluentd with the above configs and every time I hit port 24225 with nc command as follows
  **$ nc -vz localhost 24225**
  I can see server connections piling up under **_server_connections**. I verify that using the following command
**$ curl -s "localhost:24220/api/plugins.json?debug=1"|jq .|grep -A 5 "_server_connections"**

 We use tcp health checks on our AWS NLB load balancers which sends tcp connections similar to what I did above manually which I think is resulting in a lot of dangling tcp connections. That in turn is leaking memory.

repeatedly:
If clients shutdown the connection immediately, `close` callback is
sometimes called before `connect` callback. It causes connection object leak.

### DIFF
--- a/lib/fluent/plugin_helper/server.rb
+++ b/lib/fluent/plugin_helper/server.rb
@@ -600,7 +600,7 @@ module Fluent
           rescue => e
             @log.error "unexpected error on reading data", host: @callback_connection.remote_host, port: @callback_connection.remote_port, error: e
             @log.error_backtrace
-            close(true) rescue nil
+            close rescue nil
             raise if @under_plugin_development
           end
 
@@ -609,7 +609,7 @@ module Fluent
           rescue => e
             @log.error "unexpected error on reading data", host: @callback_connection.remote_host, port: @callback_connection.remote_port, error: e
             @log.error_backtrace
-            close(true) rescue nil
+            close rescue nil
             raise if @under_plugin_development
           end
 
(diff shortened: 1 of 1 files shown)


repeatedly:
@cosmo0920 Thanks for the review. I noticed addtional bug and add more commit. Could you re-check again?

cosmo0920:
Oops, I didn't notice about `#close` arity mistake.
It also looks good to me.

PR #2117, opened 2018-08-31
output: Add disable_chunk_backup for ignore broken chunks. fix #2112

### LINKED ISSUE #2112: Add option to send bad chunks to /dev/null or disable backups at all
- fluentd version: 1.2.4
- Environment information, e.g. OS: Docker / Kubernetes / GKE

We're using Fluentd inside a docker container to collect all Docker logs in our legacy environment and also inside Kubernetes. After updating to 1.2.0 container crashed after a while as /tmp/fluentd used all diskspace inside the container, which was limited to 3 gb. 

Via env var FLUENT_BACKUP_DIR we could modify the directory but still the only way to get around the problem is to mount the used directory to the host, and delete the files there from time to time.

Imho an option to send backups to /dev/null or to disable creation of this backups is missing currently.

Would be nice it could be added.

repeatedly:


repeatedly:
@okkez Could you review this patch?

repeatedly:
@cosmo0920 okkez seems busy for fluentd-ui. Could you check this patch?

PR #2139, opened 2018-09-25
in_http: Allow specifying the wildcard '*' as the CORS domain 

fujimotos:
This allows users to set up the configuration like below:

    <match debug.**>
      @type http
      cors_allow_origins ["*"]
    </match>

In this case, any origin can make a request to Fluentd; Plugin users
do not need to enumerate all the possible origins explicitly, so it
should be useful for, say, collecting data from a group of websites.

See [MCD "Access-Control-Allow-Origin" header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)


repeatedly:
Thanks!

Issue #2136, opened 2018-09-22
Oj is not installed, and failing back to Yajl for json parser even if oj gem is installed


alexander-stepanov:
I can see this warning in the log when we have migrated from alpine 3.7 to 3.8 in our base Docker container with ruby 2.5.1, fluentd 1.2.5, and installed oj-3.6.10:
```
"#0 [http_input] Oj is not installed, and failing back to Yajl for json parser"
```
Steps to reproduce:
Run an official Alpine Docker container `docker run -it --rm alpine:3.8` and execute the following commands:
```
apk add --update ruby ruby-dev ruby-irb build-base
gem install etc json oj webrick fluentd -N
fluentd -s
fluentd
```
An example log:
```
2018-09-22 08:21:33 +0000 [info]: starting fluentd-1.2.5 pid=1423 ruby="2.5.1"
2018-09-22 08:21:33 +0000 [info]: spawn command to main:  cmdline=["/usr/bin/ruby", "-Eascii-8bit:ascii-8bit", "/usr/bin/fluentd", "--under-supervisor"]
2018-09-22 08:21:33 +0000 [info]: gem 'fluentd' version '1.2.5'
2018-09-22 08:21:33 +0000 [info]: adding match pattern="debug.**" type="stdout"
2018-09-22 08:21:33 +0000 [info]: adding match pattern="system.**" type="forward"
2018-09-22 08:21:33 +0000 [info]: #0 adding forwarding server '192.168.0.12:24224' host="192.168.0.12" port=24224 weight=60 plugin_id="object:2aed9b79cc38"
2018-09-22 08:21:33 +0000 [info]: #0 [forward_output] adding forwarding server '192.168.0.11:24224' host="192.168.0.11" port=24224 weight=60 plugin_id="forward_output"
2018-09-22 08:21:33 +0000 [info]: adding source type="forward"
2018-09-22 08:21:33 +0000 [info]: adding source type="http"
2018-09-22 08:21:33 +0000 [info]: #0 [http_input] Oj is not installed, and failing back to Yajl for json parser
2018-09-22 08:21:33 +0000 [info]: adding source type="monitor_agent"
2018-09-22 08:21:33 +0000 [info]: adding source type="debug_agent"
2018-09-22 08:21:33 +0000 [info]: #0 starting fluentd worker pid=1427 ppid=1423 worker=0
2018-09-22 08:21:33 +0000 [info]: #0 delayed_commit_timeout is overwritten by ack_response_timeout
[...]

PR #2140, opened 2018-09-25
Handle LoadError properly

okkez:
Sometimes LoadError is raised while loading oj gem, even if oj gem is
installed. Therefore we want to see the correct message of LoadError.

For example, Alpine Linux provides the package ruby which does not
include bigdecimal and provides ruby-bigdecimal.

See #2136

Signed-off-by: Kenji Okimoto <okimoto@clear-code.com>


repeatedly:
Thanks!

PR #2144, opened 2018-10-05
in_http: Implement support for CORS preflight requests

fujimotos:
This patch should fix the issue #1701.

To receive a 'application/json' request from another origin, we need
to be able to handle CORS preflight requests. Otherwise, browsers will
refuse to interact with us.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests

This patch adds support for the feature.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -342,6 +342,10 @@ module Fluent::Plugin
             # For multiple X-Forwarded-For headers. Use first header value.
             v = v.first if v.is_a?(Array)
             @remote_addr = v.split(",").first
+          when /Access-Control-Request-Method/i
+            @access_control_request_method = v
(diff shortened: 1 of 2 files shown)


fujimotos:
I confirmed this patch works with Fluentd v1.2.6 and Firefox 60.2.1.

**Test Code (JavaScript)**

```javascript
// Post a JSON record using XMLHttpRequest
var req = new XMLHttpRequest();
req.open('POST', 'http://fluentd.local/debug.log');
req.setRequestHeader('Content-Type', 'application/json')
req.send(JSON.stringify({'foo': 'bar'}))
```

**Result**

2018-10-05-172738_597x68_scrot

Also, this patch should fix the issue #1701.

repeatedly:
Thanks!

PR #2147, opened 2018-10-15
in_syslog: Support octet counting frame. fix #1679

### LINKED ISSUE #1679: in_syslog support for RFC5424 Octet Counting / MSG-LEN header for tcp connections
While testing the in_syslog plugin to receive syslog messages generated by syslog-ng via a TCP connection I stuck with "that stupid integer at beginning of each message".

Example log message
<pre>
119 <139>1 2017-08-30T14:09:39+00:00 my-host app-name 34 - [meta sequenceId=\"1\"] [autoindex:error] Some logging message...
</pre>

After some reading I discovered that the leading integer is described here: https://tools.ietf.org/id/draft-gerhards-syslog-plain-tcp-12.html#rfc.section.3.4.1

What do you think about adding this (optional?) detail to the syslog parser?

repeatedly:
Add `frame_type` parameter to accept octet counting.
If set `frame_type octet_count` in the configuration, `in_syslog` can accept octet counting messag from senders.

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -75,8 +75,8 @@ module Fluent::Plugin
     config_param :tag, :string
     desc 'The transport protocol used to receive logs.(udp, tcp)'
     config_param :protocol_type, :enum, list: [:tcp, :udp], default: :udp
-    desc 'The message frame type.(normal, octet_count)'
-    config_param :frame_type, :enum, list: [:normal, :octet_count], default: :normal
+    desc 'The message frame type.(traditional, octet_count)'
+    config_param :frame_type, :enum, list: [:traditional, :octet_count], default: :traditional
 
     desc 'If true, add source host to event record.'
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_syslog.rb
```diff
@@ -75,6 +75,8 @@ class SyslogInput < Input
     config_param :tag, :string
     desc 'The transport protocol used to receive logs.(udp, tcp)'
     config_param :protocol_type, :enum, list: [:tcp, :udp], default: :udp
+    desc 'The message frame type.(normal, octet_count)'
+    config_param :frame_type, :enum, list: [:normal, :octet_count], default: :normal
```
repeatedly: I want better name for `normal`.

cosmo0920: I think so, too.
How about `older` or `traditional`?

repeatedly: `traditional` seems good. Will change soon.

PR #2165, opened 2018-11-05
Fix deadlock between on_writable and close

repeatedly:
TLSServer acquires lock in `on_writable` and `on_write_complete` callback is called inside its lock via cool.io.
To avoid this problem, callback order is moved to after cool.io's `on_writable`.

This patch includes out_forward tiny change to show the destination.


dpgaspar:
I've tested it and it works! Thank you.

PR #2170, opened 2018-11-08
Remove `Thread#run` to avoid unexpected interruption.

repeatedly:
Thread#run causes unexpected interruption and it affects
the behaviour of some libraries, e.g. DataError in zlib.
Use condition variable instead to handle thread execution.

This change is based on jl2005's #1984 patch.

Ref https://github.com/fluent/fluentd/issues/1847 / https://github.com/fluent/fluentd/issues/1903 

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -154,7 +154,7 @@ module Fluent
       end
 
       # Internal states
-      FlushThreadState = Struct.new(:thread, :next_clock)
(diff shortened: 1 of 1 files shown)


repeatedly:
@cosmo0920 ji2005's test case is passed with old implementation because the problem is rare timing issue with ruby's VM and C extension implementation.
So writing test is hard. We need feedback from users with fluentd v1.3.

PR #2173, opened 2018-11-09
Use flush_thread_count value for queued_chunks_limit_size when queued_chunks_limit_size is not specified

repeatedly:
Unlimited queued_chunks_limit_size causes unexpected resource consumption, e.g. lots of smaller files with file buffer, when
the destination has a problem with smaller flush_interval.
The default behaviour should be safe so use flush_thread_count by default.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -372,7 +372,7 @@ module Fluent
       end
 
       def queue_full?
-        @queued_chunks_limit_size && (synchronize { @queue.size } >= @queued_chunks_limit_size)
+        synchronize { @queue.size } >= @queued_chunks_limit_size
       end
 
       def queued_records
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -354,6 +354,10 @@ module Fluent
               log.warn "'flush_interval' is ignored because default 'flush_mode' is not 'interval': '#{@flush_mode}'"
             end
           end
+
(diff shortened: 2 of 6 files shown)


lib/fluent/plugin/buffer.rb
```diff
@@ -372,7 +372,7 @@ def write(metadata_and_data, format: nil, size: nil, enqueue: false)
       end
 
       def queue_full?
-        @queued_chunks_limit_size && (synchronize { @queue.size } >= @queued_chunks_limit_size)
+        synchronize { @queue.size } >= @queued_chunks_limit_size
```
cosmo0920: buffer test complains `@queued_chunks_limit_sise` is nil and not to be able to compare.
Could you check it?
https://travis-ci.org/fluent/fluentd/jobs/452714881#L547

lib/fluent/plugin/output.rb
```diff
@@ -354,6 +354,10 @@ def configure(conf)
               log.warn "'flush_interval' is ignored because default 'flush_mode' is not 'interval': '#{@flush_mode}'"
             end
           end
+
+          if @buffer.queued_chunks_limit_size.nil?
+            @buffer.queued_chunks_limit_size = @buffer_config.flush_thread_count
```
cosmo0920: `@i.buffer.queue.size` always seems to be assumed 1 in test cases.
Could you check it?
https://travis-ci.org/fluent/fluentd/jobs/452714881#L654

repeatedly:
I forgot to change tests. Just push the fix.

Issue #2175, opened 2018-11-10
Can't see fields on Kibana after parsing log.


spkesan:
Hi,

I was trying to use regexp to parse my logs to extract some of the fields, and I cannot see them on Kibana. I'm using fluentd container to achieve this. But there's no indication from docker logs to know what's wrong.

Here's my fluentd.conf:

```
<source>
        @type tail
        path /newfluentd/logs/asd.log
        pos_file /newfluentd/logs/asd.log.pos
        tag spk.aeroapp
        read_from_head true
        <parse>
                @type regexp
                expression /^(?<logtime>[a-zA-Z]* \d* \d* \d*:\d*:\d*) GMT\D\d*: INFO \(info\): \(ticker.c:\d*\) {(?<namespace>\w*)} client: tsvc \((?<client_tsvc_error>\d*),(?<client_tsvc_timeout>\d*)\) proxy \((?<client_proxy_complete>\d*),(?<client_proxy_error>\d*),(?<client_proxy_timeout>\d*)\) read \((?<client_read_success>\d*),(?<client_read_error>\d*),(?<client_read_timeout>\d*),(?<client_read_not_found>\d*)\) write \((?<client_write_success>\d*),(?<client_write_error>\d*),(?<client_write_timeout>\d*)\) delete \((?<client_delete_success>\d*),(?<client_delete_error>\d*),(?<client_delete_timeout>\d*),(?<client_delete_not_found>\d*)\) udf \((?<client_udf_complete>\d*),(?<client_udf_error>\d*),(?<client_udf_timeout>\d*)\) lang \((?<client_lang_read_success>\d*),(?<client_lang_write_success>\d*),(?<client_lang_delete_success>\d*),(?<client_lang_error>\d*)\)
                time_key logtime
                time_format %b %d %Y %H:%M:%S
        </parse>
</source>

<match spk.**>
        @type elasticsearch
        host 172.18.0.2
        port 9200
        index_name fluentd_asp
        flush_interval 10s
</match>
```

Here's my application log line (/newfluentd/logs/asd.log):
```
Oct 13 2018 23:59:53 GMT+0800: INFO (info): (ticker.c:585) {starhub_kvs} client: tsvc (0,0) proxy (0,0,0) read (1903439312,0,0,2110808) write (345776441,5944686,0) delete (0,0,0,0) udf (0,0,0) lang (0,0,0,0)
```

Here's my container logs:
```
[...]

PR #2176, opened 2018-11-12
Check right slash position

okkez:
This change can raise Fluent::ConfigError when right slash is missing.
This helps users to notice the weird behavior like #2175.

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -74,6 +74,9 @@ module Fluent
       return nil unless str
       return Regexp.compile(str) unless str.start_with?("/")
       right_slash_position = str.rindex("/")
+      if right_slash_position < str.size - 3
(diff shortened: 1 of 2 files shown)


lib/fluent/config/types.rb
```diff
@@ -74,6 +74,9 @@ def self.regexp_value(str)
       return nil unless str
       return Regexp.compile(str) unless str.start_with?("/")
       right_slash_position = str.rindex("/")
+      if right_slash_position < str.size - 3
+        raise ConfigError, "invalid regexp: missing right slash: #{str}"
```
repeatedly: Could you add `Fluent::`?
We want to use `Fluent::ConfigError` for new code.

okkez: Just added.

repeatedly:
Thanks!

PR #2180, opened 2018-11-16
Add connection check at startup when out_forward.

bootjp:
Added an option to check the out_forward node and communication when starting fluentd.

It is a matter with the following mailing list.
https://groups.google.com/forum/#!topic/fluentd/2ITnHbveYIM

> Hello
> 
> I am using fluentd v1.2.6 in TLS encryption mode when forwarding in out_forward -> in_forward.
> 
> At fluentd startup I'd like to check if the out_forward process can connection with the in_forward node before listening for the forward port.
> Does anyone know a good way about this?
> 
> The use case is below.
> Existing applications are built across AWS and GCP, and the analysis base is in GCP.
> I use out_forward TCP encryption for AWS and GCP connections.
> 
> The out_forward on the AWS side is redundant with the Active / Standby configuration.
> When fluentd is deployed Switch from Active to Standby and return to the original Active / Standby when deployment is completed.
> 
> I think that it is desirable that fluentd gets an error as it is when the out_forward -> in_forward connection can not be performed by the new deployment so that the crash occurs.
> 
> The assumed error is a certificate validation failure or SHARD key mismatch.
> * I recognize that these errors are not checked at startup in fluentd now
> 
> See you.

I am not familiar with Ruby, and since I read the Fluentd code for the first time this time, please let me know if there is a better implementation.

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -74,6 +74,9 @@ module Fluent
       return nil unless str
       return Regexp.compile(str) unless str.start_with?("/")
       right_slash_position = str.rindex("/")
+      if right_slash_position < str.size - 3
+        raise Fluent::ConfigError, "invalid regexp: missing right slash: #{str}"
+      end
       options = str[(right_slash_position + 1)..-1]
       option = 0
       option |= Regexp::IGNORECASE if options.include?("i")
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks. Will check after flight.

Issue #1879, opened 2018-03-05
Forward output with mutual TLS auth


martinssipenko:
I just tested #1861  this with fluent-bit and works as expected, but things don't look that good with flunetd. Looking at Fluentd's configuration options there is a [`tls_cert_path`](https://docs.fluentd.org/v1.0/articles/out_forward#tls_cert_path) property, which can be used to specify Certificate Authority certificates, not client certificates (I find the property name really misleading).

In order to add Mutual TLS auth there should be new configuration options added that would allow to specify client certificate and client key that should be used for TLS communication with another flunetd instance.

repeatedly:
> I find the property name really misleading

For better name, one way is mark `tls_cert_path` as deprecated and adding two options, `tls_ca_cert_path` and `tls_client_cert_path`.
out_forward sets `tls_cert_path` value to `tls_ca_cert_path` in `configure` for backward compatibility.

krysennn:
Hi, i'm facing to same problem.

Here my configuration:
```
# fluentd sender
<match **>
  @type forward
  transport tls
  tls_cert_path /root/ca.crt
  tls_verify_hostname false # for debugging
  <server>
    host X.X.X.X
    port 24224
  </server>
</match>
```

```
# fluentd receiver
<source>
  @type forward
  port 24224
  bind 0.0.0.0
  <transport tls>
    cert_path /root/cert.crt
    private_key_path /root/cert.key
    client_cert_auth true
    ca_cert_path /root/ca.crt
  </transport>
</source>
```

Error is `2018-08-03 13:26:18 +0000 [trace]: #0 unexpected error before accepting TLS connection error_class=OpenSSL::SSL::SSLError error="SSL_accept returned=1 errno=0 state=SSLv3 read client certificate B: peer did not return a certificate"`

I'm using `td-agent 1.2.2`
I've got my own CA with there client and server certificate signed.

I searched in documentation, code and issues but i didn't found field for set client certificate path in settings of `in_forward` input plugin.

Do you have an idea ?

krysennn:
@repeatedly Hi, do you have more information about it ^ ? Thanks.

joetym:
hi, I'm having problems with this as well. Part of the problem I think is that I'm trying to make it work with a certificate bundle used for a website, which explains why I can't find the client certificates (because there aren't any). I should be using Mutual TLS which can be self-generated what I gather. Is this functionality actually in a working state, or would you recommend using the plugin instead ?

[...]

PR #2181, opened 2018-11-16
out_forward: Rename parameter tls_cert_path to tls_ca_cert_path

okkez:
Clarify the meaning of the parameter.

See #1879 for more details

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -167,6 +167,10 @@ module Fluent::Plugin
       end
 
       if @transport == :tls
+        # for backward compatibility
+        if @tls_cert_path && !@tls_cert_path.empty?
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -166,8 +167,8 @@ def configure(conf)
       end
 
       if @transport == :tls
-        if @tls_cert_path && !@tls_cert_path.empty?
```
repeatedly: To avoid existing environment breakage, if `tls_cert_path` is not `nil`, assign `tls_cert_path` to `tls_ca_cert_path` is better.

okkez: Add code to keep compatibility at 71d202f.

repeatedly:
Thx!

PR #2184, opened 2018-11-21
Add connection check at startup when out_forward.

bootjp:
Added an option to check the out_forward node and communication when starting fluentd.

It is a matter with the following mailing list.
https://groups.google.com/forum/#!topic/fluentd/2ITnHbveYIM

Hello

I am using fluentd v1.2.6 in TLS encryption mode when forwarding in out_forward -> in_forward.

At fluentd startup I'd like to check if the out_forward process can connection with the in_forward node before listening for the forward port.
Does anyone know a good way about this?

The use case is below.
Existing applications are built across AWS and GCP, and the analysis base is in GCP.
I use out_forward TCP encryption for AWS and GCP connections.

The out_forward on the AWS side is redundant with the Active / Standby configuration.
When fluentd is deployed Switch from Active to Standby and return to the original Active / Standby when deployment is completed.

I think that it is desirable that fluentd gets an error as it is when the out_forward -> in_forward connection can not be performed by the new deployment so that the crash occurs.

The assumed error is a certificate validation failure or SHARD key mismatch.

I recognize that these errors are not checked at startup in fluentd now
See you.

I am not familiar with Ruby, and since I read the Fluentd code for the first time this time, please let me know if there is a better implementation.

This pull request is the same as below. #2180
I seemed to be merged because I was force pusn by the bot set in my account and the difference disappeared.
Bot has already been lifted.

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -262,7 +262,7 @@ module Fluent::Plugin
           begin
             node.verify_connection
           rescue StandardError => e
-            log.fatal e.message
+            log.fatal "forward's connection setting error: #{e.message}"
             raise Fluent::UnrecoverableError, e.message
           end
         end
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -77,6 +77,9 @@ module Fluent::Plugin
     desc 'Ignore DNS resolution and errors at startup time.'
     config_param :ignore_network_errors_at_startup, :bool, default: false
 
+    desc 'Verify that a connection can be made with one of out_forward nodes at the time of startup.'
+    config_param :verify_connection_at_startup, :bool, default: false
+
     desc 'Compress buffered data.'
     config_param :compress, :enum, list: [:text, :gzip], default: :text
 
@@ -252,6 +255,17 @@ module Fluent::Plugin
         @sock_ack_waiting_mutex = Mutex.new
         @sock_ack_waiting = []
         thread_create(:out_forward_receiving_ack, &method(:ack_reader))
+      end
+
+      if @verify_connection_at_startup
+        @nodes.each do |node|
+          begin
+            node.verify_connection
+          rescue StandardError => e
+            log.fatal e.message
+            raise Fluent::UnrecoverableError, e.message
+          end
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


test/plugin/test_out_forward.rb
```diff
@@ -217,6 +217,18 @@ def read_ack_from_sock(sock, unpacker)
     assert_equal 2, d.instance.ack_response_timeout
   end
 
+  test 'verify_connection_at_startup is disabled in default' do
+    @d = d = create_driver(CONFIG)
+    assert_equal false, d.instance.verify_connection_at_startup
```
repeatedly: `assert_false` is better.

test/plugin/test_out_forward.rb
```diff
@@ -217,6 +217,18 @@ def read_ack_from_sock(sock, unpacker)
...
+  end
+
+  test 'verify_connection_at_startup can be enabled' do
+    @d = d = create_driver(CONFIG + %[
+      verify_connection_at_startup true
+    ])
+    assert_equal true, d.instance.verify_connection_at_startup
```
repeatedly: `assert_true` is better

test/plugin/test_out_forward.rb
```diff
@@ -774,4 +786,112 @@ def create_target_input_driver(response_stub: nil, disconnect: false, conf: TARG
       i.configure(conf)
     end
   end
+
+  test 'verify_connection_at_startup_nodes_are_not_available' do
```
repeatedly: You can use `sub_test_case` for grouping related tests.

```
sub_test_case 'verify_connection_at_startup' do
  test 'nodes are not available' do
    # test body
  end
end
```

lib/fluent/plugin/out_forward.rb
```diff
@@ -253,6 +256,17 @@ def start
         @sock_ack_waiting = []
         thread_create(:out_forward_receiving_ack, &method(:ack_reader))
       end
+
+      if @verify_connection_at_startup
+        @nodes.each do |node|
+          begin
+            node.verify_connection
+          rescue StandardError => e
+            log.fatal e.message
```
repeatedly: log.fatal "forward's connection setting error: #{e.message}"

bootjp:
Thank you for code review.
Several Ruby knowledge was deepened and it was very meaningful.

repeatedly:
Merged. Thanks!

bootjp:
@repeatedly 
Thanks for merge.
When will this change be released?

PR #2187, opened 2018-11-22
out_forward: Support mutual TLS. fix #1879

### LINKED ISSUE #1879: Forward output with mutual TLS auth

repeatedly:


repeatedly:
@okkez Yes but we need feedback from other users.

repeatedly:
No feedback but okkez confirmed the behaviour and this change doesn't break existing behavior.
So I will merge this patch and release v1.3.2.

Issue #2109, opened 2018-08-22
record_transformer remove_keys not working as expected


aleks-mariusz:
I have a pipeline set up to get kubernetes based log messages into elasticsearch using fluentd (which originate from docker containers using the systemd logging driver, most examples online have the separate json files in /var/log/containers which docker seems to have moved away from recently).  

The messages arrive in elasticsearch fine, but along the path the log takes, i am trying to get fluentd to delete some keys that are excessive and unnecessary.  After using field_map in the systemd_entry <filter> block, I am using the record_transformer's remove_keys option inside a <filter> block, however certain keys do not get deleted and i'm wondering if this is a bug or am i just using this functionality incorrectly.

fluentd version: 1.2.4 running inside docker (1.13.1) container deployed by kubernetes (1.11.2):

the Gemfile used to install Fluentd
```
source 'https://rubygems.org'

gem 'fluentd', '<=1.2.4'
gem 'activesupport', '~>5.2.1'
gem 'fluent-plugin-kubernetes_metadata_filter', '~>2.1.2'
gem 'fluent-plugin-elasticsearch', '~>2.11.5'
gem 'fluent-plugin-systemd', '~>1.0.1'
gem 'fluent-plugin-detect-exceptions', '~>0.0.11'
gem 'fluent-plugin-prometheus', '~>1.0.1'
gem 'fluent-plugin-multi-format-parser', '~>1.0.0'
gem 'fluent-plugin-rewrite-tag-filter', '~>2.1.0'
gem 'oj', '~>3.6.5'
```

the Dockerfile used to create the container:
```
FROM debian:stretch-slim

ARG DEBIAN_FRONTEND=noninteractive

COPY clean-apt /usr/bin
COPY clean-install /usr/bin
COPY Gemfile /Gemfile

# 1. Install & configure dependencies.
# 2. Install fluentd via ruby.
# 3. Remove build dependencies.
# 4. Cleanup leftover caches & files.
RUN BUILD_DEPS="make gcc g++ libc6-dev ruby-dev libffi-dev" \
    && clean-install $BUILD_DEPS \
                     ca-certificates \
                     libjemalloc1 \
                     ruby \
    && echo 'gem: --no-document' >> /etc/gemrc \
    && gem install --file Gemfile \
    && apt-get purge -y --auto-remove \
                     -o APT::AutoRemove::RecommendsImportant=false \
                     $BUILD_DEPS \
    && clean-apt \
    # Ensure fluent has enough file descriptors
[...]

PR #2192, opened 2018-11-27
Delete top level using bracket style properly

okkez:
In previous version, following configuration does not work properly:

```
<source>
  @type dummy
  dummy [
    {"foo.bar": "test1234", "message": "Hello"}
  ]
  tag dummy
</source>

<filter dummy>
  @type record_transformer
  remove_keys "$['foo.bar']"
</filter>

<match dummy>
  @type stdout
</match>
```

This shows like following:

```
2018-11-27 15:19:18 +0900 [info]: #0 fluentd worker is now running worker=0
2018-11-27 15:19:19.008586045 +0900 dummy: {"foo.bar":"test1234","message":"Hello"}
2018-11-27 15:19:20.009721132 +0900 dummy: {"foo.bar":"test1234","message":"Hello"}
2018-11-27 15:19:21.010784035 +0900 dummy: {"foo.bar":"test1234","message":"Hello"}
```

In this version, it works well.

See also #2109

### DIFF
--- a/lib/fluent/plugin_helper/record_accessor.rb
+++ b/lib/fluent/plugin_helper/record_accessor.rb
@@ -40,8 +40,14 @@ module Fluent
           if @keys.is_a?(Array)
             @last_key = @keys.last
             @dig_keys = @keys[0..-2]
-            mcall = method(:call_dig)
-            mdelete = method(:delete_nest)
+            if @dig_keys.empty?
+              @keys = @keys.first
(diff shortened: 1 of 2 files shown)


repeatedly:
Good catch! Thanks!

zffocussss:
how about add a new key

<record>
  app "$['kubernetes.container_name']"
  app ${record["kubernetes.container_name"]}. # it does not work
</record>

PR #2193, opened 2018-11-27
fluent-cat: Handle retry limit

### LINKED ISSUE #2182: fluent-cat return zero exit code when error occurs
Version
```
$ fluent-cat --version
fluent-cat 1.3.0
```
Environment Version
```
docker run -it --rm fluent/fluentd /bin/sh
```

The problem:
`fluent-cat` return zero exit code when error occurs
```
$ echo '{"message":"foo"}' | fluent-cat --json --host asdfasdfasdfasdfasdf --port 12345 foo
connect failed: getaddrinfo: Name does not resolve
$ echo $?
0
```

okkez:
Fix #2182

### DIFF
--- a/lib/fluent/command/cat.rb
+++ b/lib/fluent/command/cat.rb
@@ -107,6 +107,8 @@ require 'msgpack'
 class Writer
   include MonitorMixin
 
+  RetryLimitError = Class.new(StandardError)
+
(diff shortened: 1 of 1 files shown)


repeatedly:
Thx!

PR #2198, opened 2018-11-29
varying status codes wrt path

rbkasat:
@okkez 
HTTP input plugin returns a status code 200 on success but if I want the plugin to return different status code based upon different ur's, it's not possible. This PR is to fill this gap.
My current usecase is, I want to use fluentd to act as a proxy for one of our API's


okkez:
@rbkasat Could you describe the purpose of this PR and use cases?

ganmacs:
Sorry but it's not reasonable for me. I'm closing.

Issue #1862, opened 2018-02-19
Robust RFC 5424, RFC 3339, ISO 8601, timestamp formatting


apolloclark:
### Expected Behavior

Fluentd syslog collector can collect RFC 5424 formatted syslog messages from both the Rsyslog Windows Agent, and Rsyslog Linux Agent, with or without fractional seconds, per RFC 5424, RFC 3339, and ISO 8601.

### Actual Behavior

Rsyslog Linux Agent timestamp (with microseconds) is accepted, but the Rsyslog Windows Agent timestamp (without microseconds) causes errors.

### Error Log

```
2018-02-17 23:20:23 +0000 [error]: #0 invalid input data="<134>1 2018-02-17T15:20:23-07:00 vagrant-2008R2 EvntSLog - - - Provider \"FileSystem\" is Started. Details: ProviderName=FileSystem NewProviderState=Started SequenceNumber=4 HostName=ConsoleHost HostVersion=2.0 HostId=06dbbd91-401c-47e7-b5d5-39658de1ae24 EngineVersion= RunspaceId= PipelineId= CommandName= CommandType= ScriptName= CommandPath= CommandLine=" error_class=Fluent::TimeParser::TimeParseError error="invalid time format: value = 1 2018-02-17T15:20:23-07:00 vagrant-2008R2, error_class = ArgumentError, error = string doesn't match"
```

### Environment

- RHEL 7.2, fluentd syslog collector, td-agent 1.0.2
- RHEL 7.2, rsyslog forwarder 8.24.0
- Windows 2008 R2 SP1, rsyslog forwarder 4.3.0.255

### Configuration

**td-agent**
```
<source>
  @type syslog
  port 5140
  bind 0.0.0.0
  protocol_type tcp
  tag  rsyslog
  <parse>
    message_format rfc5424
  </parse>
</source>
```

**Rsyslog Linux Agent**
```
*.* @@127.0.0.1:5141;RSYSLOG_SyslogProtocol23Format
```

**Rsyslog Windows Agent**
```
$szCustomSyslogHeader <%syslogprifac%>%syslogver% %timereported:::date-rfc3339% %source% %syslogappname% %syslogprocid% %syslogmsgid% %syslogstructdata% 
```

### Details

The Rsyslog Windows Agent, when configured to send RFC 5424, does not include fractional seconds / microseconds in the timestamp. I found a fluentd git commit which defines the default filter, and only supports fractional seconds:

[...]

PR #2239, opened 2018-12-20
Use String#squeeze instead of String#gsub

okkez:
Because String#squeeze is about 7x faster than String#gsub.
See also https://gist.github.com/okkez/6e199011505c6db7e2e1003cc94f8cbc


okkez:
I will send a PR for #1862 after this PR is merged.

repeatedly:
I first knew this method :)

okkez:
Thank you!

PR #2291, opened 2019-02-08
Reduce delay before flush when large timekey and small timekey_wait a…

iwasakims:
When large timekey and relatively small timekey_wait are given, the delay before flush could be too long. The cause is running interval of enqueue thread which is set as timekey / 11.0. If timekey is 86400 (1 day), the interval is 7854 seconds. Even if timekey_wait is set to 600 seconds in this case, the actual delay before flush could be up to 600+7854 seconds. Taking timekey_wait into account for deciding enqueue thread interval would mitigate this.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1336,7 +1336,7 @@ module Fluent
         end
         if @chunk_key_time
           if !value_for_interval || @buffer_config.timekey < value_for_interval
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks. Will check soon.

repeatedly:
Looks good. Thanks!

iwasakims:
Thanks for reviewing and merging.

PR #2295, opened 2019-02-14
Support embedded Ruby code in section attributes. Fix #2273

### LINKED ISSUE #2273: The “match” directive does not support embedded Ruby
Using the official `fluent/fluentd:v1.3.3-onbuild-1.0` docker image:

```
fluentd-1.3.3
ruby="2.5.2"
fluent-plugin-ec2-metadata version '0.1.2'
fluent-plugin-multi-format-parser version '1.0.0'
fluent-plugin-s3 version '1.1.7'
```
I confugure fluentd using environment variables and noticed that the `match` directive does not expand the match pattern:

```
# FLUENT_TAG is set dynamically at deployment time to match
# the application being deployed e.g. `prod.project1.web`


<source>
 @type tail
   ....
  tag "docker.#{ENV['FLUENT_TAG']}"
</source>

<match "docker.#{ENV['FLUENT_TAG']}" >
  @type ec2_metadata
  output_tag ${tag}.${instance_id}
</match>

<match docker.**>
  @type s3
  ...
</match>
 ....

```

Is this intended behaviour? If yes - what are the workarounds?

cosmo0920:

### DIFF
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -82,7 +82,7 @@ module Fluent
           elsif skip(/\</)
             e_name = scan(ELEMENT_NAME)
             spacing
-            e_arg = scan_nonquoted_string(/(?:#{ZERO_OR_MORE_SPACING}\>)/)
+            e_arg = scan_string(/(?:#{ZERO_OR_MORE_SPACING}\>)/)
             spacing
             unless skip(/\>/)
               parse_error! "expected '>'"
--- a/test/config/test_config_parser.rb
+++ b/test/config/test_config_parser.rb
@@ -329,6 +329,36 @@ module Fluent::Config
       end
     end
 
+    sub_test_case "Embedded Ruby Code in section attributes" do
+      setup do
+        ENV["EMBEDDED_VAR"] = "embedded"
+        ENV["NESTED_EMBEDDED_VAR"] = "nested-embedded"
+        @hostname = Socket.gethostname
+      end
+
+      teardown do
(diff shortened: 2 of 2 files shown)


cosmo0920:
But this implementation does not permit the following configuration:

```aconf
<source>
  @type tail
  path temp.log
  tag "docker.#{ENV['FLUENT_TAG']}"
  <parse>
    @type none
  </parse>
</source>

<match "docker.#{ENV['FLUENT_TAG']}".**> " Mixed case is NG
  @type stdout
</match>

<match docker.**>
  @type null
</match>
```

Should we also permit the above configuration? (Embedded Ruby code mixed attributes)

The below is OK scenario:

```aconf
```aconf
<source>
  @type tail
  path temp.log
  tag "docker.#{ENV['FLUENT_TAG']}"
  <parse>
    @type none
  </parse>
</source>

<match "docker.#{ENV['FLUENT_TAG']}.**"> # Entire Embedded Ruby code is OK
  @type stdout
</match>

<match docker.**>
  @type null
</match>
```

repeatedly:
> Should we also permit the above configuration? (Embedded Ruby code and plain text mixed attributes)

I think no need support because current `"` quoted parameter doesn't allow mixied style.

cosmo0920:
>> > Should we also permit the above configuration? (Embedded Ruby code and plain text mixed attributes)

> I think no need support because current `"` quoted parameter doesn't allow mixied style.

Hmm, I see.

repeatedly:
Thanks!

PR #2323, opened 2019-03-05
in_syslog - Check message lenght when read from buffer

### LINKED ISSUE #2298: in_syslog with tcp doesn't check message length when read from buffer
Currently I'm using syslog to send log to fluentd, my configuration is below:
- td-agent v3.3.0
- Amazon Linux 2
- My config file is

```
<source>
  @type syslog
  port 25140
  bind 0.0.0.0
  protocol_type tcp
  frame_type octet_count
  tag app1
  <parse>
    message_format rfc5424
  </parse>
</source>

<match app1.**>
  @type file
  @id output_file
  path /var/log/td-agent/app1
</match>
```

fluentd works well but sometimes in td-agent.log I saw the error like this

```
2019-02-17 18:05:00 +0000 [warn]: #0 failed to parse message data="<190>1 2019-02-17T18:05:00.02336"
2019-02-17 18:05:01 +0000 [error]: #0 unexpected error on reading data host="x.x.x.x" port=24241 error_class=ArgumentError error="invalid value for Integer(): \"2+00:00 \""
```

After tweaking around I figured out that fluentd doesn't check message length when read from buffer, as you could see in this line:

https://github.com/fluent/fluentd/blob/e0ac111800a5c9e5a0e1bbc4c765615094daeb54/lib/fluent/plugin/in_syslog.rb#L171

Due to the nature of TCP, fluentd might get only first halt of an message then try to parse the octet count on the second half

My propose is that we need to implement a check then if message length is smaller than octet count then do not process it and get new data from socket

```
--- in_syslog.rb        2019-02-17 16:50:40.071929637 +0000
+++ /opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.3.3/lib/fluent/plugin/in_syslog.rb        2019-02-17 18:17:54.548634558 +0000
@@ -169,6 +169,10 @@
               num = Integer(buffer[pos..idx])
               pos = idx + num
               msg = buffer[idx + 1...pos]
+              if msg.size < num - 1
+                pos = pos - num - num.to_s.size
+                break
+              end
               message_handler(msg, conn)
             end
           else
```

I could make a PR if you think this is good enough.

Cheers.

alexnguyen91:
Fixes #2298 

To correctly parge message when using in_syslog plugin

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -169,6 +169,10 @@ module Fluent::Plugin
               num = Integer(buffer[pos..idx])
               pos = idx + num
               msg = buffer[idx + 1...pos]
+              if msg.size < num - 1
+                pos = pos - num - num.to_s.size
+                break
+              end
               message_handler(msg, conn)
             end
           else


repeatedly:
Thx!

Issue #2330, opened 2019-03-12
Regexp does not return expected result on my elastic search


neilsrg:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

- fluentd or td-agent version. 0.14
- Environment information:
	- Operating system: MacOS 10.14.3
	- Kernel version: 18.2.10

Hi,

I have configured my regexp on td-agent.conf as below:

  <parse>

    @type regexp

    expression /{.*/

  </parse>

the input is this:

     [java] {"@timestamp":"2019-03-08T14:17:33.819+11:00","@version":1,"message":"Cannot put the message header key=standardHeader, value={\"esb_tracking_id\":\"aeaf67cb-bdfb-49cd-acc7-e9a30190c4bf\",\"esb_integration_id\":\"INT035\",\"esb_interface_id\":\"INF0177\",\"esb_interface_step\":\"Send To sqs-inf0178\",\"esb_interface_step_source\":\"STL\",\"esb_interface_step_dest\":\"sqs-inf0178\"} into Sqs MessageAttribute","logger_name":"org.apache.camel.component.aws.sqs.SqsProducer","thread_name":"qtp1370131471-73","level":"WARN","level_value":30000}

I want to remove the "     [java] " from the string it will return the json portion of it

{"@timestamp":"2019-03-08T14:17:33.819+11:00","@version":1,"message":"Cannot put the message header key=standardHeader, value={\"esb_tracking_id\":\"aeaf67cb-bdfb-49cd-acc7-e9a30190c4bf\",\"esb_integration_id\":\"INT035\",\"esb_interface_id\":\"INF0177\",\"esb_interface_step\":\"Send To sqs-inf0178\",\"esb_interface_step_source\":\"STL\",\"esb_interface_step_dest\":\"sqs-inf0178\"} into Sqs MessageAttribute","logger_name":"org.apache.camel.component.aws.sqs.SqsProducer","thread_name":"qtp1370131471-73","level":"WARN","level_value":30000}

in my regex I use  which /{.*/ I have tested on the regex editor

but when run this it seems like not processing it. It only return this

  |  
-- | --
  | March 8th 2019, 14:17:07.896 | @timestamp:March 8th 2019, 14:17:07.896 @log_name:nav.log _id:4jZNW2kBXLPO_FFV_aRA _type:fluentd _index:logstash-2019.03.08 _score: -

[...]

PR #2331, opened 2019-03-15
parser_regexp: Check named capture. ref #2330

repeatedly:

Avoid unexpected behaviour like #2330

multiline parser does named capture check so regexp parser should also check.

**Docs Changes**:
No need

**Release Note**: 
regexp parser raises configuration error when no named capture in expression.


repeatedly:
Ah, existence tests are broken ;p
Just fixed. After passed CI,  I will merge the patch.

Issue #2336, opened 2019-03-19
http_in plugin cors_allow_origins be able to specify wildcard subdomains


hirokikana:
`cors_allow_origins` is support * or the exact matching. HTTP request from many subdomain case is need to add many domain in `cors_allow_origins` setting.

I want to specify wildcard subdomain as a solution. But, according to `Access-Control-Allow-Origin` header specification unusable( https://www.w3.org/TR/cors/#access-control-allow-origin-response-header ).

May I add this feature? If allowed, I can create PullRequest.

repeatedly:
> May I add this feature? If allowed, I can create PullRequest.

yes please :)

repeatedly:
Patch is merged. Forgot to close.

PR #2337, opened 2019-03-19
in_http: Add support for subdomain in CORS domain

hirokikana:
Signed-off-by: Hiroki Takayasu <hiroki.kana@gmail.com>

#2336 

Currently `cors_allow_origins` could only use `*` or exact matching.  I modified able to could use wildcard in domain setting(ex: `http://*.example.com` )

**Docs Changes**:
No need

**Release Note**: 
See PR title

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -518,7 +518,10 @@ module Fluent::Plugin
           return true
         end
         filtered_cors_allow_origins = @cors_allow_origins.select {|origin| origin != ""}
-        return filtered_cors_allow_origins.find {|origin| (start_str,end_str) = origin.split("*",2); @origin.start_with?(start_str) and @origin.end_with?(end_str)} != nil
+        return filtered_cors_allow_origins.find do |origin|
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_http.rb
```diff
@@ -512,6 +512,14 @@ def send_response_nobody(code, header)
         data << "\r\n"
         write data
       end
+
+      def include_cors_allow_origin
+        if @cors_allow_origins.include?(@origin)
+          return true
+        end
+        filtered_cors_allow_origins = @cors_allow_origins.select {|origin| origin != ""}
+        return filtered_cors_allow_origins.find {|origin| (start_str,end_str) = origin.split("*",2); @origin.start_with?(start_str) and @origin.end_with?(end_str)} != nil
```
repeatedly: Could you separate this one liner block into multiple lines?
By default, fluentd code doesn't use `;` to force one line.

hirokikana: Thanks for your review. 
The contents pointed out were fixed. Do I need to `git rebase -i` ?

repeatedly:
Thanks!

PR #2343, opened 2019-03-22
Expose the current list of timekeys as a buffer metric

stevenjm:

This is useful for monitoring things like:

 - How old is the current oldest log message still in the buffer?
 - Equivalently, what is the maximum latency of the logging system as a whole? This might be interesting to track over time.
 - How new is the *newest* log message in the buffer? This might indicate a delay upstream of fluentd.

**Docs Changes**:

None of the existing `buffer_*` metrics appear to be documented, so there doesn't seem to be a sensible place to document this.

**Release Note**: 

A new `buffer_timekeys` metric provides a list of timekeys currently in the buffer when time-based chunking is enabled.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -154,6 +154,7 @@ module Fluent
         @dequeued_num = {} # metadata => int (number of dequeued chunks)
 
         @stage_size = @queue_size = 0
+        @timekeys = Hash.new(0)
         @metadata_list = [] # keys of @stage
       end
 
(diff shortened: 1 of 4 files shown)


stevenjm:
The initial test failure doesn't look related to my change. I've force pushed the same commit with an updated timestamp to re-run the CI.

repeatedly:
Looks good for me.

@cosmo0920 How about this?

cosmo0920:
Also looks good to me. :smile:

repeatedly:
Thx!

PR #2356, opened 2019-03-29
Fix plugin log time format if format is JSON

### LINKED ISSUE #2355: Plugin log time format not in sync with system config
### Front matter

- fluentd 1.4.1 (tested via Docker: [fluentd/fluentd:1.4-onbuild](https://github.com/fluent/fluentd-docker-image/blob/master/v1.4/alpine-onbuild/Dockerfile))
- OS: Alpine 3.8 (see above)
- Ruby: 2.5.2 (via Alpine 3.8 per)

### Configuration

This is a simplification of my configuration, enough I hope to illustrate the problem I'm having.

```
<source>
  @type                 my_plugin
  @id                   my_plugin
  @log_level            debug
  tag                   my_plugin
</source>

<system>
  <log>
    format json
    time_format %FT%T.%6N%:z
  </log>
</system>

<match my_plugin>
  @type copy
  <store>
    @type elasticsearch
    ...
  </store>
</match>
```

### Description

I am writing a custom plugin to pull logs from S3 and publish them to Elasticsearch. That processing works well! However, the internal logs generated by my plugin, and by the fluentd system, have inconsistent formats of timestamps in them that confuse our downstream log aggregators (ironically, also fluentd instances :-) ).

Examples:

A. This is a system logging statement with the timestamp format I want – good old RFC3339 date/time.

```
{"time":"2019-03-28T21:11:56.827082+00:00","level":"info","message":"parsing config file is succeeded path=\"/fluentd/etc/fluent.conf\""}
```

B. This is a log statement from my plugin:

```
{"time":"2019-03-28 21:11:58 +0000","level":"debug","message":"[my_plugin] log_bucket: logs"}
```

It is generated by code that looks like this:

```
def configure(conf)
  super
  log.debug("log_bucket: logs")
end
```

Note the default timestamp format, instead of the system-configured time format that I want. I was expecting my specified time format to be used.

Interestingly, if I do this from within my plugin:

```
log.time_format = '%FT%T.%6N%:z'
log.debug("log_bucket: logs")
```

the time format doesn't change!

### Analysis

I have determined the root cause, which I believe to be a bug. (That being said, I'm a newcomer to the fluentd code base, so do double-check my analysis.)

The fundamental issue, I think, is that `Fluent::PluginLogger` both inherits from `Fluent::Log` (i.e. contains internal log state of its own) _and_ wraps around a "parent" logger (the `@logger` instance variable). This creates lots of opportunities for mischief.

Some, but not all, methods delegate to the parent logger. In particular, `#puts` and `#<<` delegate to the parent logger, but `#debug`, for example, does not. This means that all the formatting of log messages is handled by the `PluginLogger`, _not_ by the parent logger.

Some attempt is made to make sure that a `PluginLogger` uses the same format as its parent logger, by synchronizing the state between the two. `@format` is synced whenever `#format=` is called, for example. However, `@time_format` is not synced, and it retains its default value whenever a `PluginLogger` is constructed. I believe this is why I'm seeing an inconsistent time format coming from my plugin.

Now, the reason that `#time_format=` has no effect is that both `#time_format` and `#time_format=` are delegated to the parent logger – they do not read from or modify the plugin logger's internal state. This would be fine if `PluginLogger` were using `#time_format` to format time objects instead of using its own `@time_format` instance variable. However, its definition of `#debug` et al are inherited from `Fluent::Log`, and they use `#format_time` to do the formatting – which, unsurprisingly, uses `@time_format` under the hood.

Syncing `@time_format` is not sufficient, I discovered. If the time format is compatible with the performance-optimized `strptime` library, `Fluent::Log` will create a `@time_formatter` object to format times instead of Ruby's `Time#strftime` implementation. Thus, we must synchronize `@time_formatter` as well.

I think the best fix, although perhaps the most dangerous from a backwards compatibility standpoint, might be to remove the inheritance relationship between `PluginLogger` and `Fluent::Log`, and just have `PluginLogger` delegate everything to its parent logger. At least, the mixture of inheritance, composition, and partial delegation really ought to be revisited.

The most immediate fix, though, would be to ensure that `@time_format` and `@time_formatter` are synchronized properly with the parent logger.

_Edit:_ I discovered another nuance: the issue only arises in the case of JSON output. Why? Because the plugin logger's text formatter doesn't format time. Instead, it calls `#caller_line`, which is in fact delegated to the parent logger. By comparison, the JSON formatter calls `#format_time` itself.

Here's a test case that reproduces the situation. It works on the "text" case but fails on the "json" one:

```
    data(
      text: [:text, "2016 [info]: yaaay\n"],
      json: [:json, %Q({"time":"2016","level":"info","message":"yaaay"}\n)],
    )
    def test_time_format(data)
      fmt, expected_log_line = data
      @log.format = fmt
      @log.time_format = "%Y"
      @log.info "yaaay"
      assert{ @log_device.logs.include? expected_log_line }
    end
```

ods94065:
Fixes #2355 

The time format for a plugin logger was out of sync with the parent logger, and the delegation of `#time_format` and `#time_format=` made it impossible to fix the plugin logger's time format without hacks.

This problem was not noticeable if the log format was text, because time formatting was delegated via `#caller_line` to the parent logger. However, in the case of JSON output, the plugin logger's `#format_time` method was called, yielding incorrectly-formatted timestamps.

Added tests for both text and JSON, and added tests for format changes in general while I was in there.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -532,6 +532,7 @@ module Fluent
       end
 
       self.format = @logger.format
+      self.time_format = @logger.time_format
       enable_color @logger.enable_color?
     end
 
@@ -540,11 +541,17 @@ module Fluent
     end
 
     alias orig_format= format=
+    alias orig_time_format= time_format=
     alias orig_enable_color enable_color
 
     def format=(fmt)
       self.orig_format = fmt
       @logger.format = fmt
+    end
+
+    def time_format=(fmt)
+      self.orig_time_format = fmt
+      @logger.time_format = fmt
     end
 
     def enable_color(b = true)
@@ -554,9 +561,9 @@ module Fluent
 
     extend Forwardable
     def_delegators '@logger', :get_worker_id, :enable_color?, :enable_debug, :enable_event,
-      :disable_events, :log_event_enabled, :log_event_enabled=, :time_format, :time_format=,
-      :time_formatter, :time_formatter=, :event, :caller_line, :puts, :write, :<<, :flush,
-      :reset, :out, :out=, :optional_header, :optional_header=, :optional_attrs, :optional_attrs=
+      :disable_events, :log_event_enabled, :log_event_enabled=, :event, :caller_line, :puts, :write,
+      :<<, :flush, :reset, :out, :out=, :optional_header, :optional_header=, :optional_attrs,
+      :optional_attrs=
   end
 
 
--- a/test/test_log.rb
+++ b/test/test_log.rb
@@ -744,12 +744,6 @@ class PluginLoggerTest < Test::Unit::TestCase
       @log.disable_events(Thread.current)
     end
 
-    def test_time_format
-      assert_equal(@log.time_format, @logger.time_format)
-      @log.time_format = "time_format"
-      assert_equal(@log.time_format, @logger.time_format)
-    end
-
     def test_event
       mock(@logger).event(Fluent::Log::LEVEL_TRACE, { key: "value" })
       @log.event(Fluent::Log::LEVEL_TRACE, { key: "value" })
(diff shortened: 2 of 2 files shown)


repeatedly:
Thanks for the patch. Will check.

repeatedly:
I confirmed this patch works. Thanks!

PR #2372, opened 2019-04-07
drop dig_rb and update required ruby version to 2.3

henrich:
ruby2.3 or higher version has dig_rb gem function as default, see top of https://github.com/jrochkind/dig_rb/blob/master/README.md

Now distros support ruby2.3 or higher version, bump up it would be okay, IMO.
 - [Debian9](https://packages.debian.org/source/stable/ruby2.3) or newer (Debian10 will be released 2 or 3 months later)
 - [Ubuntu16.04](https://packages.ubuntu.com/xenial/ruby2.3) or newer
 - [RHEL7 and its clone CentOS has ruby2.0](https://developers.redhat.com/articles/using-native-ruby-rhel-6-or-rhel-7/), so out of the scope already ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,6 @@ cache: bundler
 # See here for osx_image -> OSX versions: https://docs.travis-ci.com/user/languages/objective-c
 matrix:
   include:
-    - rvm: 2.1.10
-      os: linux
-    - rvm: 2.2.10
-      os: linux
     - rvm: 2.4.6
       os: linux
     - rvm: 2.5.5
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks for the patch.

The problem is we can't drop older ruby support without annoucement.
One idea is drop ruby 2.1/2.2 support in this year. In addition, drop one old ruby version in each year.
2020 drops ruby 2.3. 2012 drops ruby 2.4, ...

henrich:
Make sense, I got it.
How about raise another issue to drop old ruby version and depends on it?

ganmacs:
Now, we can drop dig-rb. could you update this branch?

repeatedly:
https://github.com/fluent/fluentd/pull/2750 includes these changes.
So we can close.

Thanks for the patch and suggestion!

PR #2376, opened 2019-04-10
V0.12

renzheng1987:
**Which issue(s) this PR fixes**: 
Fixes #

**What this PR does / why we need it**: 

**Docs Changes**:

**Release Note**:

PR #2377, opened 2019-04-10
Update buffer.rb to recovery memory leak

paassdc:
Update buffer.rb to recovery memory leak when using @copy

Fixes #
   memory leak in my environment
    when new emit requests come,first judge if queue size is less than limit. if exceed limit, raise error.
**Docs Changes**:
     No
**Release Note**: 
     can release new version for v0.12


ganmacs:
dup https://github.com/fluent/fluentd/pull/2376

PR #2378, opened 2019-04-10
support config delimiter in syslog plugin

abenbachir:
There are softwares out there that still use NUL-terminating characters when sending events through TCP. Firewalls for example are used a lots, and it's impossible to change their behavior because they are closed source. Most Firewall implement RFC3164 or CEF protocol, so using in_syslog plugin comes very usefull than in_tcp plugin.
I'm adding delimiter we can edit the default delimiter ("\n") from configuration, similar to in_tcp plugin.

I made the change in v0.12, I can do the same for v1.x If this change is good to merge.

Something important to know is that NUL(\0) or newline(\n) characters must be used with double-quote in configuration:
```
<source>
    type syslog
    protocol_type tcp
    delimiter "\0"
</source>
```
**Docs Changes**:
I will update documentation when this PR make it

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -99,6 +99,9 @@ module Fluent::Plugin
 
     config_param :blocking_timeout, :time, default: 0.5
 
+    desc 'The delimiter value "\n"'
+    config_param :delimiter, :string, default: "\n" # syslog family add "\n" to each message
+
     config_section :parse do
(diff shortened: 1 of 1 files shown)


repeatedly:
v0.12 will merge only security fix so please send this patch to master.

abenbachir:
So you mean, this change will only make to  v1.x  ?

repeatedly:
Yes.

abenbachir:
Done

repeatedly:
Thanks!

PR #2381, opened 2019-04-14
parser_json: Add stream_buffer_size config param

### LINKED ISSUE #609: in_exec json streams does not work as expected
ruby-yajl reads 8k from the input stream before parsing so input is delayed until the buffer is full (which can take forever if no new input is generated).

This can be reproduced with the following config:

    <source>
      type exec
      format tsv
      command while sleep 1; do date; done
      keys tsv_date
      tag all
    </source>

    <source>
      type exec
      format json
      command while sleep 1; do echo '{"json_date": "'$(date)'"}'; done
      tag all
    </source>

    <match all>
      type stdout
    </match>

tsv_date is emitted each time but json_date is not emitted until the buffer is full (~5 minutes in this case).

I have created a workaround https://github.com/serbaut/fluentd/commit/c99bcf47a3695cd39aa211ccb2b0bd8bac5a704c that parses line by line but in the end ruby-yajl has to be fixed so is parses one json document at a time.

benwh:
Fixes #609 

Allow configuration of the size of the buffer that Yajl uses when parsing streaming input.

The advantage of this is that when using `out_exec_filter`, and parsing
as JSON, it's now possible to configure this plugin to avoid having to
wait for 8192 bytes of data to be parsed before events are emitted.

**Docs Changes**: fluent/fluentd-docs#631

**Release Note**: 
Use PR title.

### DIFF
--- a/lib/fluent/plugin/out_exec_filter.rb
+++ b/lib/fluent/plugin/out_exec_filter.rb
@@ -95,6 +95,7 @@ module Fluent::Plugin
     COMPAT_PARSE_PARAMS = {
       'out_format' => '@type',
       'out_keys' => 'keys',
+      'out_stream_buffer_size' => 'stream_buffer_size',
     }
     COMPAT_EXTRACT_PARAMS = {
       'out_tag_key' => 'tag_key',
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -30,6 +30,12 @@ module Fluent
       desc 'Set JSON parser'
       config_param :json_parser, :enum, list: [:oj, :yajl, :json], default: :oj
 
+      # The Yajl library defines a default buffer size of 8KiB when parsing
+      # from IO streams, so maintain this for backwards-compatibility.
+      # https://www.rubydoc.info/github/brianmario/yajl-ruby/Yajl%2FParser:parse
+      desc 'Set the buffer size that Yajl will use when parsing streaming input'
+      config_param :stream_buffer_size, :integer, default: 8192
+
       config_set_default :time_type, :float
 
       def configure(conf)
@@ -81,7 +87,7 @@ module Fluent
         y.on_parse_complete = ->(record){
           block.call(parse_time(record), record)
         }
-        y.parse(io)
+        y.parse(io, @stream_buffer_size)
       end
     end
   end
--- a/test/plugin/test_parser_json.rb
+++ b/test/plugin/test_parser_json.rb
@@ -111,4 +111,28 @@ class JsonParserTest < ::Test::Unit::TestCase
       assert_equal text, record['time']
     end
   end
+
+  def test_yajl_parse_io_with_buffer_smaller_than_input
+    parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin::JSONParser)
+    parser.configure(
+                     'keep_time_key' => 'true',
+                     'json_parser' => 'yajl',
+                     'stream_buffer_size' => 1,
+                     )
+    text = "100"
+
+    waiting(5) do
+      rd, wr = IO.pipe
+      wr.write "{\"time\":\"#{text}\"}"
+
+      parser.instance.parse_io(rd) do |time, record|
+        assert_equal text.to_i, time.sec
+        assert_equal text, record['time']
+
+        # Once a record has been received the 'write' end of the pipe must be
+        # closed, otherwise the test will block waiting for more input.
+        wr.close
+      end
+    end
+  end
 end
--- a/test/plugin/test_out_exec_filter.rb
+++ b/test/plugin/test_out_exec_filter.rb
@@ -328,6 +328,7 @@ class ExecFilterOutputTest < Test::Unit::TestCase
     </format>
     <parse>
       @type json
+      stream_buffer_size 1
     </parse>
     <extract>
       tag_key tag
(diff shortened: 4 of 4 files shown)


lib/fluent/plugin/out_exec_filter.rb
```diff
@@ -95,6 +95,7 @@ class ExecFilterOutput < Output
     COMPAT_PARSE_PARAMS = {
       'out_format' => '@type',
       'out_keys' => 'keys',
+      'out_stream_buffer_size' => 'stream_buffer_size',
```
benwh: The only reason that this was added was to shave another 30s off the tests

test/plugin/test_parser_json.rb
```diff
@@ -111,4 +111,29 @@ def test_parse_with_keep_time_key_without_time_format(data)
       assert_equal text, record['time']
     end
   end
+
+  data('yajl' => 'yajl')
+  def test_yajl_parse_io_with_buffer_smaller_than_input(data)
```
benwh: I was considering adding a test case for the inverse of this (input smaller than buffer) but didn't do this because it seemed like it would be quite awkward to implement - due to needing to assert on the parser *not* emitting an event - and probably not provide much value.
Happy to reconsider this if anyone has a suggestion on how this could be done neatly though!

lib/fluent/plugin/parser_json.rb
```diff
@@ -30,6 +30,11 @@ class JSONParser < Parser
       desc 'Set JSON parser'
       config_param :json_parser, :enum, list: [:oj, :yajl, :json], default: :oj
 
+      # The Yajl library defines a default buffer size of 8092 when parsing
+      # from IO streams, so maintain this for backwards-compatibility.
+      # https://www.rubydoc.info/github/brianmario/yajl-ruby/Yajl%2FParser:parse
+      config_param :stream_buffer_size, :integer, default: 8092
```
repeatedly: Please add `:desc` parameter to explain the parameter.  This is only for `yajl`.

test/plugin/test_parser_json.rb
```diff
@@ -111,4 +111,29 @@ def test_parse_with_keep_time_key_without_time_format(data)
       assert_equal text, record['time']
     end
   end
+
+  data('yajl' => 'yajl')
```
repeatedly: Don't use `data` for only 1 pattern.

lib/fluent/plugin/parser_json.rb
benwh: Done!

test/plugin/test_parser_json.rb
benwh: Good point, sorted :)

benwh:
Hi @repeatedly, those issues are fixed so could you take another look when you get some time? Thanks!

lib/fluent/plugin/parser_json.rb
repeatedly: If you use `desc` method, not `:desc` parameter, it should be put on top of `config_param`.
Currently, `desc 'Set the buffer size that Yajl will use when parsing streaming input'` is for `time_tupe`, not `stream_buffer_size`.
See https://docs.fluentd.org/v1.0/articles/api-plugin-base

benwh: Ah, that was a silly mistake, thanks for the pointer! Fixed.

lib/fluent/plugin/parser_json.rb
```diff
@@ -30,6 +30,12 @@ class JSONParser < Parser
       desc 'Set JSON parser'
       config_param :json_parser, :enum, list: [:oj, :yajl, :json], default: :oj
 
+      # The Yajl library defines a default buffer size of 8092 when parsing
```
ganmacs: [nits] It may be better a default buffer size is 8192  because of https://github.com/brianmario/yajl-ruby/commit/9d4cca0175142c2308746d7c1d40d89b682b168f

test/plugin/test_parser_json.rb
```diff
@@ -111,4 +111,28 @@ def test_parse_with_keep_time_key_without_time_format(data)
...
+  def test_yajl_parse_io_with_buffer_smaller_than_input
+    parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin::JSONParser)
+    parser.configure(
+                     'keep_time_key' => 'true',
+                     'json_parser' => 'yajl',
+                     'stream_buffer_size' => 1,
+                     )
+    text = "100"
+
+    waiting(5) do
```
ganmacs: Does it need? this test suit seems not to block since L134.

lib/fluent/plugin/parser_json.rb
ganmacs: FYI: https://github.com/brianmario/yajl-ruby/pull/192

benwh: Definitely not a nit; thanks for picking up on this! I did think that 8092 bytes was a bit of a strange size for a buffer, but didn't look into it 🤦‍♂️. 
Fixed and I'll update the docs PR too.

test/plugin/test_parser_json.rb
benwh: Yes it's not actually required for the test to work, I put it there to protect against any regression in the code being tested (or the test scaffolding) that might cause the test to fail, because if it does then it will block forever which would be quite annoying when it runs in CI, if nothing else.

repeatedly:
Thx!

PR #2384, opened 2019-04-17
Remove duplicated method

ganmacs:
L195 overwrite Fluentd::Log#time_format= which is defined by `attr_accessor`.
This has been introduced since https://github.com/fluent/fluentd/commit/80e45ea2db99b2e3624e8ea9dde6b1707162416e

nope

To fix the warning from ruby such as `warning: method redefined; discarding old time_format=`.

**Docs Changes**:

No need.

**Release Note**: 

No need.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -148,10 +148,10 @@ module Fluent
     end
 
     attr_reader :format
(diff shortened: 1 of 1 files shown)


repeatedly:
Okay. Thanks!

PR #2396, opened 2019-04-23
in_forward: Add tag and add_tag_prefix parameters

repeatedly:

none

I sometimes receive this request from users for aggregator.
There are several reasons:

- Port based tag handling

```
<source>
  @type forward
  port 24224
  add_tag_prefix cluster1
</source>

<source>
  @type forward
  port 24225
  add_tag_prefix cluster2
</source>
```

We can simulate this by adding new field for cluster detection in forwarder side and check it in aggregator side. But it makes configuration complicated.

- Make tag handling cost lower

We can rewrite tag with `route` or similar plugins but it has performance penalty for large traffic.

The implementation is very simple and it resolves above cases. So supporting these parameters is reasonable.

**Docs Changes**:
Will send a patch to fluentd-docs after merged.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -62,6 +62,11 @@ module Fluent::Plugin
     desc "The field name of the client's hostname."
     config_param :source_hostname_key, :string, default: nil
 
+    desc "New tag instead of incoming tag"
+    config_param :tag, :string, default: nil
+    desc "Add prefix to incoming tag"
+    config_param :add_tag_prefix, :string, default: nil
+
     config_section :security, required: false, multi: false do
       desc 'The hostname'
       config_param :self_hostname, :string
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_forward.rb
```diff
@@ -293,6 +298,9 @@ def on_message(msg, chunk_size, conn)
         log.warn "Input chunk size is larger than 'chunk_size_warn_limit':", tag: tag, host: conn.remote_host, limit: @chunk_size_warn_limit, size: chunk_size
       end
 
+      tag = @tag.dup if @tag
+      tag = "#{@add_tag_prefix}.#{tag}" if @add_tag_prefix
```
ganmacs: Is it ok without checking tag is an empty string?

repeatedly:
Yeah, label is one good approach but label itself doesn't change tag.
So this patch is also good for label combo, e.g. some output plugins use tag placeholder.

lib/fluent/plugin/in_forward.rb
repeatedly: Ah, good catch. I will add empty check on `configure`.

PR #2398, opened 2019-04-25
out_forward: do not use SO_LINGER on Windows

### LINKED ISSUE #1968: Encryption support on out_forwarder plugin in Windows (Fluentd 1.0.2)
Hi,
is TLS/SSL supported by Fluentd embedded forwarder plugin in Windows?

I'm trying to push logs from Windows and Linux machines to a Fluentd server (Linux).
Linux machines can successfully send logs to the server with TLS.
Windows machines, instead, cannot send logs to the server, when configured with TLS. Both client and server say "connection established" when a buffer is flushed to the server, the buffer on the client gets erased, but the server receives nothing. I've tried to disable TLS on both client and server and communication works properly. I've tried to provide the client a forged certificate and it gives an OpesSSL error. But in my case there is no error and the server cannot receive logs.

I'm using Fluentd v1.0.2 for server and Linux client, td-agent v.3.1.1 for Windows clients (which is based on Fluentd v1.0.2).

Server configuration is:
```
<source>
  @type forward
  port 24224
  bind 0.0.0.0
  <transport tls>
    cert_path /certs/fluentd.crt
    private_key_path /certs/fluentd.key
    private_key_passphrase xxxx
  </transport>
</source>
```
Windows client configuration is:
```
<match xxx.**>
  @type forward
  transport tls
  tls_cert_path C:\certs\fluentd.crt
  tls_verify_hostname false
  tls_allow_self_signed_cert true
  <server>
    host xxxx
    port 24224
  </server>
</match>
```

Linux clients configuration is the same (except `tls_cert_path`).

Options
```
  tls_verify_hostname false
  tls_allow_self_signed_cert true
```
are necessary in my local (test) environment.

fujimotos:
Fixes #1968 

There is an obscure bug in out_forward that prevents events from
being sent to a remote node. It occures only on Windows, and is
reproducable only when SSL/TLS is enabled.
    
The root cause is not fully investigated yet, but disabling
SO_LINGER is confirmed to fix the issue. So this implements it.

**Docs Changes**:

Not required

**Release Note**: 

"out_forward: fix data loss when SSL/TLS is enabled on Windows "

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -354,7 +354,10 @@ module Fluent::Plugin
           cert_path: @tls_client_cert_path,
           private_key_path: @tls_client_private_key_path,
           private_key_passphrase: @tls_client_private_key_passphrase,
-          linger_timeout: @send_timeout,
+
+          # Enabling SO_LINGER causes data loss on Windows
+          # https://github.com/fluent/fluentd/issues/1968
+          linger_timeout: Fluent.windows? ? nil : @send_timeout,
           send_timeout: @send_timeout,
           recv_timeout: @ack_response_timeout,
           &block


repeatedly:
Thanks!

michaelsteiner:
This fix is included in Fluentd 1.5.0 according to https://github.com/fluent/fluentd/blob/master/CHANGELOG.md.
According to https://support.treasuredata.com/hc/en-us/articles/360001479187-The-td-agent-ChangeLog, td-agent (.msi package for Windows) 3.4.1 is still based on Fluentd 1.4.2.

=> When will an .msi package for Windows be available that includes this fix? (Say, newer than td-agent-3.4.1-0-x64.msi that is currently available on https://td-agent-package-browser.herokuapp.com/3/windows )

repeatedly:
Next td-agent, 3.5.0, will be released late july or early august.
If you want to use newer version in msi package, gem update is one way.

michaelsteiner:
> 
> 
> Next td-agent, 3.5.0, will be released late july or early august.
> If you want to use newer version in msi package, gem update is one way.

Thanks a lot for your fast response. The following update worked for me:

Download https://rubygems.org/downloads/fluentd-1.5.2-x64-mingw32.gem and install it with
`fluent-gem install --force --local *.gem`

PR #2399, opened 2019-04-25
Support tls in syslog

ganmacs:

none

support tls for in_syslog

**Docs Changes**:

needed. I'll change docs.

**Release Note**: 

same as title


repeatedly:
All test failed. Could you check it?

lib/fluent/plugin/in_syslog.rb
```diff
@@ -74,7 +74,7 @@ class SyslogInput < Input
     desc 'The prefix of the tag. The tag itself is generated by the tag prefix, facility level, and priority.'
     config_param :tag, :string
     desc 'The transport protocol used to receive logs.(udp, tcp)'
-    config_param :protocol_type, :enum, list: [:tcp, :udp], default: :udp
+    config_param :protocol_type, :enum, list: [:tcp, :udp], default: nil
```
repeatedly: For deprecated parameter, use `deprecated:`

lib/fluent/plugin/in_syslog.rb
```diff
@@ -141,12 +145,13 @@ def multi_workers_ready?
     def start
       super
 
-      log.info "listening syslog socket on #{@bind}:#{@port} with #{@protocol_type}"
-      case @protocol_type
+      log.info "listening syslog socket on #{@bind}:#{@port} with #{@protocol_type || @transport_config.protocol}"
+      case @protocol_type || @transport_config.protocol
```
repeatedly: Default `@transport_config.protocol` is `tcp`. Change default to `udp`.

ganmacs: good catch 🙏 
 https://github.com/fluent/fluentd/pull/2399/commits/f23566a50b6300bab4d029e9b0ad26db36eb036a

lib/fluent/plugin/in_syslog.rb
```diff
@@ -107,6 +107,11 @@ class SyslogInput < Input
       config_param :with_priority, :bool, default: true
     end
 
+    # overwrite server plugin to change default to :udp
+    config_section :transport, required: false, multi: false, init: true, param_name: :transport_config do
+      config_argument :protocol, :enum, list: [:tcp, :tls], default: :udp
```
cosmo0920: Should we use the following line?

```ruby
config_argument :protocol, :enum, list: [:tcp, :udp, :tls], default: :udp
```

ganmacs: Right. Thank you. https://github.com/fluent/fluentd/pull/2399/commits/bf319e95b3f101843c7cfd143c80b7ce2d8df38b

repeatedly:
I confirmed this works with following configuration with rsyslog

- fluentd conf

```
<source>
  @type syslog
  port 5140
  bind 0.0.0.0
  <transport tls>
    ca_path /etc/pki/ca.pem
    cert_path /etc/pki/cert.pem
    private_key_path /etc/pki/key.pem
    private_key_passphrase PASSPHRASE
  </transport>
  tag system
</source>

<match system.**>
  @type stdout
</match>
```

- rsyslog conf

```
$DefaultNetstreamDriverCAFile /home/repeatedly/pki/ca.pem

# set up the action
$DefaultNetstreamDriver gtls
$ActionSendStreamDriverMode 1
$ActionSendStreamDriverAuthMode anon
*.* @@127.0.0.1:5140
```

PR #2409, opened 2019-05-10
Improve storage local plugin

### LINKED ISSUE #2373: on XFS might not be able to read local storage pos file after crash
td-agent 2.3.6
rhel 7.6, XFS filesystem

```
    <source>
      @type systemd
      <storage>
        @type local
        path /var/log/journald.pos
      </storage>
      tag journald
    </source>
```

After a kernel panic and reboot, td-agent refused to start with:

`2019-04-07 15:55:50 +0000 [error]: fluent/log.rb:362:call: config error file="/td-agent/config/fluent.conf" error_class=Fluent::ConfigError error="Unexpected error: failed to read data from plugin storage file: '/var/log/journald.pos/worker0/storage.json'"`

/var/log/journald.pos/worker0/storage.json existed but was 0 bytes long.

It is good practice to use fdatasync() or fsync(); on XFS it is essential:
* https://stackoverflow.com/questions/47077828/xfs-rhel7-3-cold-reboot-file-truncate
* https://bugzilla.redhat.com/show_bug.cgi?id=845233

I recommend an fdatasync or fsync call is added around here:
https://github.com/fluent/fluentd/blob/0eec5542bbf2ef1e796a5b4a02acd4b5cf6b2b3e/lib/fluent/plugin/storage_local.rb#L131

It may also be worth considering logging a warning in the event of a 0-length file at https://github.com/fluent/fluentd/blob/0eec5542bbf2ef1e796a5b4a02acd4b5cf6b2b3e/lib/fluent/plugin/storage_local.rb#L94 but continuing startup as if the file did not exist.

repeatedly:
Fixes #2373 

- Ignore empty file
- Add `fsync` call in `save`

**Docs Changes**:
No need

**Release Note**: 
storage_local: fluentd starts even if existing file is empty.
storage_local: Call `fsync` in `save` for XFS

### DIFF
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -133,7 +133,7 @@ module Fluent
         tmp_path = @path + '.tmp'
         begin
           json_string = Yajl::Encoder.encode(@store, pretty: @pretty_print)
-          open(tmp_path, 'w:utf-8', @mode){ |io| io.write json_string }
+          open(tmp_path, 'w:utf-8', @mode) { |io| io.write json_string; io.fsync }
           File.rename(tmp_path, @path)
         rescue => e
           log.error "failed to save data for plugin storage to file", path: @path, tmp: tmp_path, error: e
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks for the reviews

PR #2418, opened 2019-05-26
in_tail: increase read block size

### LINKED ISSUE #2416: in_tail: readpartial maxlen is too small
We use fluentd to transfer line based large log file; about 4GiB / hour and 1495 byte / line.

```
$ wc apache2.acc.21.log
   2949500  242309651 4410569207 apache2.acc.21.log
```

CPU usage of fluentd is very high.  We tried sigdump and it shows this line every time:

https://github.com/fluent/fluentd/blob/31cd8cab4d8581699adfb54b563c38316f8faac6/lib/fluent/plugin/in_tail.rb#L735

I feel maxlen=2048 is too small, at least on our case.  It reads only one or two lines on each call.

FYI, tail command in coreutils uses `BUFSIZ` in `stdio.h`, and `BUFSIZ` in Debian and Ubuntu is 8192.

In case of ruby, overhead of calling `io.readpartial` will be higher than `read` in C.  Would you change the maxlen to at least 8192?

---

- fluentd or td-agent version: 1.3.3
- Environment information:
  - Operating system: `cat /etc/os-release`: "Debian GNU/Linux 9 (stretch)"
  - Kernel version: `uname -r`: 4.9.0-5-amd64
- Your configuration

```
<source>
  @type tail
  path /.../web/%Y-%m-%d/apache2.acc.%H.log
  pos_file /.../td-agent/web.acc.pos
  read_from_head
  tag web.acc
  format none
  enable_stat_watcher false
</source>
```

methane:
2048 was too small for large logs.
8192 is chosen because BUFSIZ in stdio.h is 8192
on Ubuntu and Debian.

Signed-off-by: Inada Naoki <songofacandy@gmail.com>

Fixes #2416

**Docs Changes**:

**Release Note**: 
in_tail: increased read block size from 2048 to 8192.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -732,7 +732,7 @@ module Fluent::Plugin
               if !io.nil? && @lines.empty?
                 begin
                   while true
-                    @fifo << io.readpartial(2048, @iobuf)
+                    @fifo << io.readpartial(8192, @iobuf)
                     while (line = @fifo.next_line)
                       @lines << line
                     end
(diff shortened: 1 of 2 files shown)


repeatedly:
Thx!

PR #2421, opened 2019-05-28
Fix error if timezone with name is used in parser

jonasrutishauser:
Fixes a `TypeError: nil can't be coerced into Integer` when a timezone name is used in a parser configuration.

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -201,14 +201,18 @@ module Fluent
       format_with_timezone = format && (format.include?("%z") || format.include?("%Z"))
 
       # unixtime_in_expected_tz = unixtime_in_localtime + offset_diff
-      offset_diff = ->(t){
-                      case
-                      when format_with_timezone then nil
-                      when timezone  then Time.now.localtime.utc_offset - offset_utc(timezone, t)
-                      when localtime then 0
-                      else Time.now.localtime.utc_offset # utc
+      offset_diff = case
+                    when format_with_timezone then nil
+                    when timezone  then
+                      offset = Fluent::Timezone.utc_offset(timezone)
+                      if offset.respond_to?(:call)
+                        ->(t) { Time.now.localtime.utc_offset - offset.call(t) }
+                      else
+                        ->(t) { Time.now.localtime.utc_offset - offset }
                       end
-                    }
+                    when localtime then 0
+                    else Time.now.localtime.utc_offset # utc
+                    end
 
       strptime = format && (Strptime.new(format) rescue nil)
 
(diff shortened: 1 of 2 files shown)


repeatedly:
Hmm... Could you show the actual example?
`timezone Europe/Zurich` has this problem?

jonasrutishauser:
Every zone name has this problem (see https://makandracards.com/makandra/67330-why-you-can-t-use-timezone-codes-like-pst-or-bst-for-time-objects).
With daylight saving time it is needed to use a zone name and the validation on line 155 accepts it whilst the parsing implementation on line 206 is not able to handle it.
And yes it is this config:
```
[...]
<parse>
  [...]
  timezone Europe/Zurich
</parse>
[...]
```

repeatedly:
Thanks. I will check it.

repeatedly:
Your patch has performance penalty and it is not good for existing users.
So need to rewrite the patch like below:

```
when timezone
  offset = Fluent::Timezone.utc_offset(timezone)
  if offset.respond_to?(:call)
    ->(t) { Time.now.localtime.utc_offset + offset.call(t) }
  else
    ->(t) { Time.now.localtime.utc_offset + offset }
  end
```
```
when strptime
  if offset_diff.respond_to?(:call)
    ->(v){ t = strptime.exec(v);  Fluent::EventTime.new(t.to_i + offset_diff.call(t), t.nsec) }
  else
    ->(v){ t = strptime.exec(v);  Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
  end
# same as format
```

repeatedly:
@ganmacs Could you also check this patch? I want to release v1.5.1 with this patch.

lib/fluent/time.rb
```diff
@@ -214,7 +220,12 @@ def initialize(format = nil, localtime = true, timezone = nil)
...
                when format_with_timezone             then ->(v){ Fluent::EventTime.from_time(Time.strptime(v, format)) }
                when format == '%iso8601'             then ->(v){ Fluent::EventTime.from_time(Time.iso8601(v)) }
-               when strptime then ->(v){ t = strptime.exec(v);         Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
+               when strptime then
+                 if offset_diff.respond_to?(:call)
+                   ->(v) { t = strptime.exec(v); Fluent::EventTime.new(t.to_i + offset_diff.call(t), t.nsec) }
+                 else
+                   ->(v) { t = strptime.exec(v); Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
+                 end
                when format   then ->(v){ t = Time.strptime(v, format); Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
```
ganmacs: Should this line be changed the same as the above lines(L223-L228)?

jonasrutishauser: You are right.
I have changed this line too.

repeatedly:
Thanks!

PR #2422, opened 2019-05-29
Refactor in monitor

ganmacs:

I'm thinking of replacing in_monitor's Webrick base HTTP server with [async](https://github.com/socketry/async) base HTTP server. In preparation for this, I cleaned up the in_monitor. 

**Docs Changes**:

not needed

**Release Note**: 

not needed


repeatedly:
I checked github repositories and changing method argument seems no problem. Merged.

PR #2424, opened 2019-05-29
Fix return value when not found path

ganmacs:

nothing

* endpoint api/config checks if the path is valid.
* return status code 404 if needed
* match returned error message with https://github.com/fluent/fluentd/blob/0f3704f7f9e4b5334eca643568eb515ce6daaa3e/lib/fluent/plugin/in_monitor_agent.rb#L55-L59

**Docs Changes**:

not needed

**Release Note**: 
not needed

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -68,10 +68,6 @@ module Fluent::Plugin
       end
 
       def build_object(req, opts)
-        unless req.path_info == ""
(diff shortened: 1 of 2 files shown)


repeatedly:
patch itself looks good but I'm now considering merge timing.
This patch changes the behaviour so v1.6 is better for me.

ganmacs:
> This patch changes the behaviour so v1.6 is better for me.

reasonable for me 👍

repeatedly:
Thanks

PR #2427, opened 2019-05-30
Fix log rotete age error handling

ganmacs:
nothing

Kernel.#Integer can raise ArgumentError https://docs.ruby-lang.org/ja/2.5.0/method/Kernel/m/Integer.html  (Sorry. but I wasn't able to find the English document...  
English document which I found(https://ruby-doc.org/core-2.5.1/Kernel.html#method-i-Integer)  doesn't refer to ArgumentError. )
And I also noticed the variable name  (`rotate_ages`) is used.

https://ruby-doc.org/core-2.5.1/Kernel.html

**Docs Changes**:

not needed

**Release Note**: 

not needed


repeatedly:
Thx!

PR #2434, opened 2019-05-31
Fix dec_ref bug in out_forwarder plugin

### LINKED ISSUE #2433: "Not found key for dec_ref" when using forward withkeepalive
I recently upgrade to v1.5 and enabled keepalive.

I'm using the `fluent/fluentd:v1.5-debian` Docker image and my config looks like this:

```
<match **>
	@type forward
	transport tls
	keepalive true

	<server>
		host "#{ENV["HOST"]}"
		port 24284
	</server>

	<security>
    self_hostname "#{Socket.gethostname}"
		shared_key "#{ENV["SHARED_KEY"]}"
	</security>

	<buffer>
		flush_interval 5s
		total_limit_size 64m
	</buffer>
</match>
```

When running it, nothing comes across to the destination and on the source I get a continuous stream of errors:  `[warn]: #0 Not found key for dec_ref: #<Fluent::PluginHelper::Socket::WrappedSocket::TLS:0x00007f64bffed208>`

Disabling keepalive resolves the issue.

luhn:
Fixes #2433

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -1084,10 +1084,10 @@ module Fluent::Plugin
           begin
             yield(sock)
           rescue
-            @socket_cache.revoke(sock) if @keepalive
+            @socket_cache.revoke if @keepalive
             raise
           else
-            @socket_cache.dec_ref(sock) if @keepalive
+            @socket_cache.dec_ref if @keepalive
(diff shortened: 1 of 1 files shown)


repeatedly:
@ganmacs Could you check this patch?

repeatedly:
Thanks for quick patch!

PR #2445, opened 2019-06-05
Match result contents

ganmacs:

nothing

Match the response form of `/api/config` and `/api/plugins`.
`/api/plugins` doesn't include the value which is the instance of Hash or Array class at [here](https://github.com/fluent/fluentd/blob/b3abefcd166aa99bcaded1cf52f8b71177e3af3d/lib/fluent/plugin/in_monitor_agent.rb#L168-L170) into the response body .
But `/api/plugins` includes  such a value[*](https://github.com/fluent/fluentd/blob/b3abefcd166aa99bcaded1cf52f8b71177e3af3d/lib/fluent/plugin/in_monitor_agent.rb#L206-L208).

I think it's safe that this diff is released as v1.6 since it includes breaking change which changes the form of the response body.

**Docs Changes**:

not needed

**Release Note**: 
not needed

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -55,8 +55,8 @@ module Fluent::Plugin
           code, header, body = render_error_json(
             code: 500,
             msg: 'Internal Server Error',
-            'error' => "#{$!}",
-            'backtrace' => $!.backtrace,
+            error: "#{$!}",
+            backtrace: $!.backtrace,
           )
         end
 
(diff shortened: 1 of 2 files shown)


repeatedly:
How about adding parameter to handle this?

ganmacs:
>  adding parameter to handle this?

Like `flatten_array_value=true`? seems not bad, but isn't it a bit complex?

repeatedly:
@ganmacs Could you resolve conflicts?

ganmacs:
This patch became obsolete after merged https://github.com/fluent/fluentd/pull/2447. so I'll close this.

PR #2447, opened 2019-06-06
Use async-http in monitor plugin

ganmacs:
depends on https://github.com/fluent/fluentd/pull/2424, last 2 commits are for this patch.

no

in_monitor_agent: Replace webrick base HTTP server with [async-http](https://github.com/socketry/async-http) base HTTP server.

Simple benchmark: https://gist.github.com/ganmacs/3f7984930a3cd9109bfbb8a2ca5d6991

**Docs Changes**:

~~no~~
Add http_server helper document

**Release Note**: 

monitor_agent uses async-http base HTTP server  instead of webrick base HTTP server


repeatedly:
Hmm... async doesn't support 2.2 or earlier but fluentd still supports these version in this year.
http helper looks good. How about using webrick when async-http is not available?

repeatedly:
The smart way is http helper implements compatibility layer.
The simple way is check require and choose `in_monitr_agent_async`/`in_monitor_agent_webrick`.

ganmacs:
I did the former way! https://github.com/fluent/fluentd/pull/2447/commits/c8f751b52b82c95feaa7461f753c9c358626f01c

cosmo0920:
Hi, these patches seems to be reasonable for me. :+1:
Thank you for your hard work. :smile:

I have a nitpick question.
Should we add test cases for http plugin helper?

ganmacs:
I'm going to add more tests later :) (So I changed this PR title into the one  which starts with [WIP] )

ganmacs:
Ready to be reviewed! @cosmo0920 @repeatedly could you review this?

lib/fluent/plugin_helper/http_server/server.rb
```diff
@@ -0,0 +1,87 @@
...
+            Async::HTTP::Endpoint.parse(@uri)
+          )
+
+          if block_given?
+            yield(self)
+          end
+        end
+
+        def start(notify = nil)
+          @logger.debug("Start HTTP server listening #{@uri}")
```
repeatedly: For debug, I want async-http in the log like below

```
Start async HTTP server listening #{@uri}"
```

lib/fluent/plugin_helper/http_server/compat/server.rb
```diff
@@ -0,0 +1,79 @@
...
+            # @example ["/example.json", :get, handler object]
+            @methods = []
+
+            if block_given?
+              yield(self)
+            end
+          end
+
+          def start(notify = nil)
+            build_handler
```
repeatedly: `logger.debug` with `Start webrick HTTP server listening`

lib/fluent/plugin/in_monitor_agent.rb
```diff
@@ -43,29 +43,80 @@ class MonitorAgentInput < Input
...
+      end
+
+      private
+
+      def render_error_json(code:, msg:, pretty_json: nil, **additional_params)
+        resp = additional_params.merge('message' => msg)
+        render_json(resp, code: code, pretty_json: pretty_json)
+      end
+
+      def render_json(obj, code: 200, **opts)
```
repeatedly: `render_json` accepts only `pretty_json` so how about replace `**opts` with `pretty_json: false`?

lib/fluent/plugin_helper/http_server/compat/server.rb
ganmacs: https://github.com/fluent/fluentd/pull/2447/commits/4b48997523c0555517b0dbc47f52642739a0b511 fixed

lib/fluent/plugin/in_monitor_agent.rb
ganmacs: right https://github.com/fluent/fluentd/pull/2447/commits/42fd38f8ab12276cf03da8ebc2e64f832cf9aafd

repeatedly:
Thanks!

repeatedly:
@ganmacs Could you send the patch to add http_server article to gitbook?

ganmacs:
will do!

PR #2450, opened 2019-06-11
Add more metrics to in_monitor_agent

ganmacs:

no

I added metrics which output and buffer have to the response of in_monitor_agent plugin to fetch the fluentd's inner stats.

**Docs Changes**:

Need to revise examples https://docs.fluentd.org/input/monitor_agent#output-example

**Release Note**: 

Same as the title

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -740,6 +740,28 @@ module Fluent
         enqueue_chunk(metadata) if enqueue_chunk_before_retry
         retry
       end
+
+      def statistics
+        buffer_space = 1.0 - ((@stage_size + @queue_size * 1.0) / @total_limit_size).round
+        stats = {
+          'stage_length' => @stage.size,
+          'stage_byte_size' => @stage_size,
+          'queue_length' => @queue.size,
+          'queue_byte_size' => @queue_size,
+          'available_buffer_space_ratios' => buffer_space * 100,
+          'total_queued_size' => @stage_size + @queue_size,
+        }
+
+        if (m = timekeys.min)
+          stats['oldest_timekey'] = m
+        end
+
+        if (m = timekeys.max)
+          stats['newest_timekey'] = m
+        end
+
+        { 'buffer' => stats }
+      end
     end
   end
 end
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -184,6 +184,8 @@ module Fluent
         @emit_records = 0
         @write_count = 0
         @rollback_count = 0
+        @flush_time_count = 0
+        @slow_flush_count = 0
 
         # How to process events is decided here at once, but it will be decided in delayed way on #configure & #start
         if implement?(:synchronous)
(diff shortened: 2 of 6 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -1457,6 +1457,26 @@ def flush_thread_run(state)
...
+          # Respect original name
+          # https://github.com/fluent/fluentd/blob/45c7b75ba77763eaf87136864d4942c4e0c5bfcd/lib/fluent/plugin/in_monitor_agent.rb#L284
+          'retry_count' => @num_errors,
+          'emit_count' => @emit_count,
+          'write_count' => @write_count,
+          'rollback_count' => @rollback_count,
+        }
+
+        if @buffer && @buffer.respond_to?(:statistics)
+          (@buffer.statistics && @buffer.statistics['buffer'] || {}).each do |k, v|
```
repeatedly: When `@buffer.statistics` reutern false/nil?

lib/fluent/plugin/buffer.rb
```diff
@@ -740,6 +740,28 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
+
+        if (m = timekeys.min)
+          stats['oldest_timekey'] = m
+        end
+
+        if (m = timekeys.max)
+          stats['newest_timekey'] = m
+        end
+
+        { 'buffer' => stats }
```
repeatedly: What is the merit of `{'buffer' => stats}` rather than `stats`?

lib/fluent/plugin/output.rb
ganmacs: looks unnecessary. thank you!
https://github.com/fluent/fluentd/pull/2450/commits/7e0941a73b8c0703757dfe6d91a5b12b6fa8b697

bai:
👋Semi-related question: do you think it would make sense to add `flush_time` metric as well, that'd answer "how long it takes for Output to perform write/try_write"?

ganmacs:
> do you think it would make sense to add flush_time metric as well

Looks neat for me.  I added `flush_time_count` and `slow_flush_count` https://github.com/fluent/fluentd/pull/2450/commits/63c006e31a8dd301e42c3f076e069d010eccca84 !

bai:
Fantastic, thank you.

repeatedly:
> flush_time_count

Is this useful? This value seems the total of flush time. How to use this value in monitoring?

ganmacs:
I think it is useful for getting the value which is like average flush_time in 5 mins.
(in [promql](https://prometheus.io/docs/prometheus/latest/querying/basics/) query is like `increase(flush_time_count[5m] / write_count[5m])` )

bai:
> I think it is useful for getting the value which is like average flush_time in 5 mins.
> (in promql query is like increase(flush_time_count[5m] / write_count[5m]) )

This is exactly what I've meant 👍 🎉

repeatedly:
Okay, I understood. `elapsed_time` is Float type. How to handle overflow? Reset to 0?
Some fluentd users many flush thread and it increases flush_time_count quickly.

ganmacs:
> elapsed_time is Float type.

Good catch… Since flush_time_count is counter, I think this value should increase monotonically. I think it's good to use Integer type (to calling `#to_i` before adding elapsed_time to flush_time ) since most of flush_time is smaller than 1.

bai:
@ganmacs Do we need to multiply `elapsed_time` by `1000` before calling `to_i` to get millisecond precision?

lib/fluent/plugin/output.rb
```diff
@@ -1202,7 +1204,10 @@ def backup_chunk(chunk, using_secondary, delayed_commit)
 
       def check_slow_flush(start)
         elapsed_time = Fluent::Clock.now - start
+        # millsec precision
+        @counters_monitor.synchronize { @flush_time_count += (elapsed_time * 1000).to_i }
```
repeatedly: How about `(elapsed_time * 1000).to_i`  calculation outside of synchronize?

ganmacs: https://github.com/fluent/fluentd/pull/2450/commits/05d842e944f43d9f7610db4d7ed62ea3399273ff fixed

repeatedly:
Thx!

PR #2453, opened 2019-06-12
command: Add --conf-encoding option to accept non utf-8 configuration file

repeatedly:

None

Some environment doesn't use utf-8 for text file and convert encoding is not easy for non technical users. This option reduces the encoding trouble.

**Docs Changes**:
I will add option to command line option article.

**Release Note**: 
Same as title.


ganmacs:
ah, tests failed. can you check this?

It may be caused by passing an empty string or nil.
`warning: Unsupported encoding :utf-8 ignored`

Issue #2451, opened 2019-06-11
out_foward with keepalive and shared_key don't work together


luhn:
When using out_forward with keepalive and shared_key, no data is sent.  There is no error.  Everything works as expected when either shared_key or keepalive is disabled.

I've created a docker-compose which reproduces the issue:  https://gist.github.com/luhn/af36a17b46a3eb7640ccb7c5bacea9bc

repeatedly:
Hmm... I can't reproduce this problem on my environment.
Did you see the error in fluentd log?

luhn:
There is no error that I can see. 

You should be seeing load averages show up after a few seconds in the dst container. Nothing ever comes across. 

— Theron

repeatedly:
Ah wait. My mistake. I forgot to rebuild src image with `keepalive true`.
This problem is reproduced with docker-compose.
In addition, I tested this config without docker-compose and it worked.
The weird thing is setup socket is same between `keepalive false` and `keepalive true`.
Need to investigate.

ganmacs:
Thank you for reporting! @luhn .
I confirmed this is a keepalive bug. #2456 should fix this issue.

luhn:
Thanks!

PR #2456, opened 2019-06-13
Fix keepalive and shared key bug

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2451

If the socket which already handshaked tries to handshake, it will be
blocked and raise IO::WaitReadable [here](https://github.com/fluent/fluentd/blob/12ffa0afe313737bf09900efce8c8656baaf77bc/lib/fluent/plugin/out_forward.rb#L787).
So, I changed the code to avoid invoking handshake twice on the same socket.

**Docs Changes**:

no

**Release Note**: 

no

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -770,8 +770,7 @@ module Fluent::Plugin
       end
 
       def verify_connection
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks. I confirmed this works.

PR #2459, opened 2019-06-17
Treat empty hash_value_field as nil

torbjorn2000:
No issue connected to this.

When configuring `hash_value_field` in a parser filter, it will treat empty values as `nil`, as if it was not configured at all.

This enables you to configure it conditionally. For example, you only want the `hash_value_field` if the environment variable "HASH" is set to true. Then you can configure the parser with: 

```
<filter *.*.*.*.*.*.raw_json>
  @type parser
  ...
  hash_value_field "#{'parsed' if ENV['HASH']=='TRUE'}"
  ...
</filter>
```
With current code (before this PR), this would generate an empty value for `hash_value_field`, and not `nil`. The result would be an empty parent field like this:  `{ "": {"key1":"value1"} }`. 

**Docs Changes**:
No changes

**Release Note**: 
* An empty `hash_value_field` will be treated as `nil` (not configured).

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -120,7 +120,7 @@ module Fluent::Plugin
       if values && @inject_key_prefix
         values = Hash[values.map { |k, v| [@inject_key_prefix + k, v] }]
       end
-      r = @hash_value_field ? {@hash_value_field => values} : values
+      r = @hash_value_field && @hash_value_field != '' ? {@hash_value_field => values} : values
       if @reserve_data
         r = r ? record.merge(r) : record
       end
--- a/test/plugin/test_filter_parser.rb
+++ b/test/plugin/test_filter_parser.rb
@@ -498,6 +498,34 @@ class ParserFilterTest < Test::Unit::TestCase
     assert_equal({"data.xxx"=>"first","data.yyy"=>"second"}, record['parsed'])
   end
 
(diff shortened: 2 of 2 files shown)


ganmacs:
Although The code looks good, [IMO] I think this behavior is not expected by users. 
it's better to raise an error when receiving an empty string as `hash_value_field` directive in configure method.

torbjorn2000:
> Although The code looks good, [IMO] I think this behavior is not expected by users.
> it's better to raise an error when receiving an empty string as `hash_value_field` directive in configure method.

I see, but in my case I really want to have the same behaviour for an empty string as for nil, so that if a certain environment variable is not set, the parsed log will not be nested under hash field. But maybe that's a special use case that makes no sense for others. If the suggested behaviour may break it for others, I'd rather prefer rollback the PR.

repeatedly:
Maybe, we need to consider general approaches like below:

- If "#{}" evaluated value is `nil`, use `nil` instead of type converted value. This may break existing environment
- support `<if ruby-code>` like configuration section to support condition.

```
<match foo>
  @type foo
  <if ENV['cond'] == 'hey'>
    # parameters
  </if>
</match>
```

torbjorn2000:
I really like the second option, the `<if ruby-code>` style. Makes sense. However, my Ruby knowledge is at a copy paste-level at its best, so I cannot really help with that..

PR #2466, opened 2019-06-19
Fix cert match pattern

### LINKED ISSUE #2458: Start even if failed to parse certificate file
- fluentd version: 1.4.2
- Environment information:
  - Operating system: `16.04.6 LTS (Xenial Xerus)`
  - Kernel version: `4.4.0-145-generic`
- Your configuration
```
<source>
  @type forward
  bind 0.0.0.0
  port 24224

  <transport tls>
    cert_path /fluentd/etc/fluentd.pem
    private_key_path /fluentd/etc/fluentd.pem
  </transport>
</source>

<match **>
  @type null
</match>
```

It seems that fluentd start even if failed to parse certificate file in forward input with TLS.

If we use a certificate file that does not end with a newline (`\n`), it will not match Fluentd's internal pattern[^1]. However Fluentd seems to start and listen on TLS connection. If you connect to Fluentd with TLS in this state, all handshakes will fail.

[^1]: https://github.com/fluent/fluentd/blob/ea651827782d7ec92c3cd09cb6ada65e32bfab44/lib/fluent/plugin_helper/cert_option.rb#L171

summerwind:
Fixes #2458

This PR fixes regex pattern for PEM encoded certificate files to allow parsing files that do not ended with newline. This will also add a test for the fix.

**Docs Changes**:
No.

**Release Note**: 
- Allow parsing of certificate files that do not end with a newline.


test/plugin_helper/data/cert/cert-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
ganmacs: Is this file used?

summerwind: No. Should I remove this? I put this file for future use due to the certificate file is always pair with the key file.

repeatedly: I think no problem.

ganmacs: My bad. no problem. 
I thought that other cert files already existed for the test. So this patch's cert files are only used for checking regex.

summerwind: No problem. Thank you for your review!

repeatedly:
Thanks!

PR #2467, opened 2019-06-19
socket helper: Support connect_timeout for TCP/TLS

repeatedly:
out_forward also supports connect_timeout parameter

None

**Docs Changes**:
After release v1.6.0, need to add connect_timeout parameter to socket helper/out_forward article.

**Release Note**: 
socket helper/out_forward: Support connect_timeout parameter

### DIFF
--- a/lib/fluent/plugin_helper/socket.rb
+++ b/lib/fluent/plugin_helper/socket.rb
@@ -55,8 +55,14 @@ module Fluent
         end
       end
 
-      def socket_create_tcp(host, port, resolve_name: false, **kwargs, &block)
-        sock = WrappedSocket::TCP.new(host, port)
+      def socket_create_tcp(host, port, resolve_name: false, connect_timeout: nil, **kwargs, &block)
+        sock = if connect_timeout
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin_helper/socket.rb
```diff
@@ -55,8 +55,14 @@ def socket_create(proto, host, port, **kwargs, &block)
         end
       end
 
-      def socket_create_tcp(host, port, resolve_name: false, **kwargs, &block)
-        sock = WrappedSocket::TCP.new(host, port)
+      def socket_create_tcp(host, port, resolve_name: false, connect_timeout: nil, **kwargs, &block)
+        sock = if connect_timeout
+                 s = ::Socket.tcp(host, port, connect_timeout: connect_timeout)
```
ganmacs: when and who will close this socket if it is set false to autoclose?

repeatedly:
TLS supports connect_timeout via TCP socket creation.
udp is connectionless. what is the timeout for udp?

PR #2469, opened 2019-06-20
time: Add EventTime#to_time for faster Time creation

repeatedly:

None

Fluent::EventTime can be converted into Time via Time.at but
this approach is slower. Output plugins sometimes require Time
for client libraries so adding faster method.

Here is benchmark code:

```
require 'benchmark/ips'
require 'fluent/time'

ot = Time.now
et = Fluent::EventTime.from_time(ot)

Benchmark.ips do |x|
  x.report "to_time" do
    et.to_time
  end

  x.report "Time.at" do
    Time.at(et)
  end
end
```

```
             to_time   171.865k i/100ms
             Time.at    92.733k i/100ms
             to_time      2.861M (± 5.1%) i/s -     14.265M in   5.002172s
             Time.at      1.184M (± 4.1%) i/s -      5.935M in   5.022106s
```

**Docs Changes**:
No need for now.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -69,6 +69,19 @@ module Fluent
       @sec.to_s
     end
 
+    begin
+      # ruby 2.5 or later
+      Time.at(0, 0, :nanosecond)
+
+      def to_time
+        Time.at(@sec, @nsec, :nanosecond)
+      end
+    rescue
+      def to_time
+        Time.at(@sec, @nsec / 1000.0)
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


ganmacs:
the change looks good but 2.4 or before tests failed.
https://travis-ci.org/fluent/fluentd/jobs/548400039

repeatedly:
That's weird. This test passed with ruby 2.4.6 on my Mac...

repeatedly:
Test now passed on older environment

PR #2479, opened 2019-07-04
Update tzinfo to v2

repeatedly:
This is for v1.7.0.

None

Update tzinfo to latest v2. The merit removes deprecated `thread_safe` gem dependency.
Here is the performance comparison.

- with v2

```
      timezone fixed   147.900k i/100ms
     timezone region   148.310k i/100ms
      timezone fixed      2.259M (± 2.7%) i/s -     11.388M in   5.045857s
     timezone region      2.312M (± 1.5%) i/s -     11.568M in   5.005012s
```

- with v1

```
      timezone fixed   150.174k i/100ms
     timezone region   151.443k i/100ms
      timezone fixed      2.324M (± 1.2%) i/s -     11.714M in   5.041740s
     timezone region      2.316M (± 2.9%) i/s -     11.661M in   5.040527s
```

tzinfo v2 is bit slower but acceptable.

```
require 'benchmark/ips'
require 'fluent/time'

formatter1 = Fluent::TimeFormatter.new("%Y%m%d %H%M%z", false, "+01:30")
formatter2 = Fluent::TimeFormatter.new("%Y%m%d %H%M%z", false, "Asia/Tokyo")
time = Fluent::EventTime.from_time(Time.new(2014, 9, 27, 0, 0, 0, 0))

Benchmark.ips do |x|
  x.report "timezone fixed" do
    formatter1.format(time)
  end

  x.report "timezone region" do
    formatter2.format(time)
  end
end
```

**Docs Changes**:
No need

**Release Note**: 
Update tzinfo dependency to v2.

### DIFF
--- a/lib/fluent/timezone.rb
+++ b/lib/fluent/timezone.rb
@@ -18,6 +18,15 @@ require 'tzinfo'
 
 require 'fluent/config/error'
 
+# For v0.12. Will be removed after v2
+module IntegerExt
+  refine Integer do
+    def to_time
+      Time.at(self)
+    end
+  end
+end
+
 module Fluent
   class Timezone
     # [+-]HH:MM, [+-]HHMM, [+-]HH
@@ -84,6 +93,8 @@ module Fluent
       end
     end
 
+    using IntegerExt
+
     # Create a formatter for a timezone and optionally a format.
     #
     # An Proc object is returned. If the given timezone is invalid,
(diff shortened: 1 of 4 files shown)


ganmacs:
The cause of CI fail in ruby 2.1 and 2.2(https://travis-ci.org/fluent/fluentd/jobs/554194714) seems this PR.

lib/fluent/timezone.rb
```diff
@@ -100,15 +111,15 @@ def self.formatter(timezone = nil, format = nil)
         case
         when format.is_a?(String)
           return Proc.new {|time|
-            Time.at(time).localtime(offset).strftime(format)
```
ganmacs: If refinements is going to be removed, I think these lines(L103, L107, L111, L127, L131, and L135) should be left as it is.
so when removing refinements in the future, we need not change these lines (and it'll be able to narrow down the scope of refinements)

repeatedly: fluentd v2 will drop Integer support for event time, so we can call `EventTime#to_time` without Integer ext.

ganmacs: I see.

bai:
👋Friendly observation: Looks like this PR made things a bit more complex for fluentd plugins that depend on ActiveSupport (e.g. `fluent-plugin-splunk-hec`). Because Rails and ActiveSupport have hard dependency on tzinfo v1.1 (see https://github.com/rails/rails/pull/34799) it conflicts with v2.0 requirements that this PR added.

PR #2482, opened 2019-07-09
Avoid zero division error when there are no available nodes

ganmacs:
no

I added checking of the size of `regular_nodes` to avoid to raise ZeroDivisionError when there are no available nodes.

https://github.com/fluent/fluentd/blob/2ec546695a843a5751a1456ed6cdcdf22b300b8f/lib/fluent/plugin/out_forward.rb#L439

**Docs Changes**:

No need

**Release Note**: 

No need


repeatedly:
Good catch!

PR #2487, opened 2019-07-09
socket: Support all private keys OpenSSL supports

### LINKED ISSUE #2486: Support abitrary private keys for TLS sockets

While RSA is popular, other key types (such as EC) can be more secure and faster.

Support arbitrary private keys.

None

**Additional context**

About to submit PR, this issue was created to document the process.

mmussomele:
When creating a TLS socket, the helpers only attempted to parse an RSA
key with `OpenSSL::PKey::RSA.new`. By using the higher level
`OpenSSL::PKey::read`, OpenSSL will determine and parse the proper key
type, adding support for EC and DSA keys.

Signed-off-by: Matthew Mussomele <matt@nefeli.io>

Fixes #2486

This adds support for using any private keys that OpenSSL supports to the socket helper. I looked for existing tests around the TLS socket to augment, but failed to find any, but offline tested that the changes snippet properly loads other key types.

**Docs Changes**:
None

**Release Note**:

### DIFF
--- a/lib/fluent/command/ca_generate.rb
+++ b/lib/fluent/command/ca_generate.rb
@@ -91,7 +91,7 @@ HELP
     def self.generate_server_pair(opts={})
       key = OpenSSL::PKey::RSA.generate(opts[:private_key_length])
 
-      ca_key = OpenSSL::PKey::RSA.new(File.read(opts[:ca_key_path]), opts[:ca_key_passphrase])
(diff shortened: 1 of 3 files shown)


mmussomele:
I looked at the failing tests but they seem unrelated to my changes. I may be mistaken, as I am not experienced with ruby though. Any help would be much appreciated.

repeatedly:
Thanks for the patch. Other ruby developer, naruse-san, also confirmed this patch looks good :)

PR #2488, opened 2019-07-09
Add HTTP output plugin

repeatedly:

None

Add HTTP output plugin for v1.7. I got several requests from different users in kubecon. Users have the endpoint to accept logs via HTTP/HTTPs. So generic HTTP output plugin is useful for typical usecases.

Discussion point:
- Support in_http format. fluentd has forward protocol so out_http -> in_http usecase is very rare. I think no need for now
- Support `[{}, {},..., {}]` like json/msgpack. Current implementation uses ndjson. I'm not sure one large array payload is useful or not.

**Docs Changes**:
Need to add article to output section. I will write it after ready to merge.

**Release Note**: 
Add `http` output plugin

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -130,17 +130,17 @@ module Fluent::Plugin
       when 'out_file', 'single_value', 'stdout', 'hash'
         'text/plain'
       else
-        raise Fluent::ConfigError, "can't determind Content-Type from formatter type. Set content_type parameter explicitly"
+        raise Fluent::ConfigError, "can't determine Content-Type from formatter type. Set content_type parameter explicitly"
       end
     end
 
     def setup_http_option
       use_ssl = URI.parse(@endpoint).scheme == 'https'
       opt = {
-        :open_timeout => @open_timeout,
-        :read_timeout => @read_timeout,
-        :ssl_timeout => @ssl_timeout,
-        :use_ssl => use_ssl
+        open_timeout: @open_timeout,
+        read_timeout: @read_timeout,
+        ssl_timeout: @ssl_timeout,
+        use_ssl: use_ssl
       }
 
       if use_ssl
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -1,3 +1,19 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
 require 'net/http'
 require 'uri'
 require 'openssl'
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -8,6 +8,8 @@ module Fluent::Plugin
   class HTTPOutput < Output
     Fluent::Plugin.register_output('http', self)
 
+    class RetryableResponse < StandardError; end
+
     helpers :formatter
 
     desc 'The endpoint for HTTP request, e.g. http://example.com/api'
@@ -45,6 +47,8 @@ module Fluent::Plugin
 
     desc 'Raise UnrecoverableError when the response is non success, 4xx/5xx'
     config_param :error_response_as_unrecoverable, :bool, default: true
+    desc 'The list of retryable response code'
+    config_param :retryable_response_codes, :array, value_type: :integer, default: [503]
 
     config_section :format do
       config_set_default :@type, 'json'
@@ -190,9 +194,13 @@ module Fluent::Plugin
             end
 
       if res.is_a?(Net::HTTPSuccess)
-        log.debug { "#{res.code} #{res.message} #{res.body}" }
+        log.debug { "#{res.code} #{res.message}#{res.body}" }
       else
-        msg = "#{res.code} #{res.message} #{res.body}"
+        msg = "#{res.code} #{res.message}#{res.body}"
+
+        if @retryable_response_codes.include?(res.code.to_i)
+          raise RetryableResponse, msg
+        end
 
         if @error_response_as_unrecoverable
           raise Fluent::UnrecoverableError, msg
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -117,6 +117,8 @@ module Fluent::Plugin
       send_request(uri, req)
     end
 
+    private
+
     def setup_content_type
       case @formatter_configs.first[:@type]
       when 'json'
@@ -135,7 +137,7 @@ module Fluent::Plugin
     end
 
     def setup_http_option
-      use_ssl = URI.parse(@endpoint).scheme == 'https'
+      use_ssl = @endpoint.start_with?('https')
       opt = {
         open_timeout: @open_timeout,
         read_timeout: @read_timeout,
@@ -154,7 +156,7 @@ module Fluent::Plugin
         end
         if @tls_private_key_path
           raise Fluent::ConfigError, "tls_private_key_path is wrong: #{@tls_private_key_path}" unless File.file?(@tls_private_key_path)
-          opt[:key] = OpenSSL::PKey::read(File.read(@tls_private_key_path), @tls_private_key_passphrase)
+          opt[:key] = OpenSSL::PKey.read(File.read(@tls_private_key_path), @tls_private_key_passphrase)
         end
         opt[:verify_mode] = case @tls_verify_mode
                             when :none
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -1,0 +1,205 @@
+require 'net/http'
+require 'uri'
+require 'openssl'
+require 'fluent/plugin/output'
+require 'fluent/plugin_helper/socket'
+
+module Fluent::Plugin
+  class HTTPOutput < Output
+    Fluent::Plugin.register_output('http', self)
+
+    helpers :formatter
+
+    desc 'The endpoint for HTTP request, e.g. http://example.com/api'
+    config_param :endpoint, :string
+    desc 'The method for HTTP request'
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+        'application/x-msgpack'
+      when 'out_file', 'single_value', 'stdout', 'hash'
+        'text/plain'
+      else
+        raise Fluent::ConfigError, "can't determind Content-Type from formatter type. Set content_type parameter explicitly"
+      end
+    end
+
+    def setup_http_option
+      use_ssl = URI.parse(@endpoint).scheme == 'https'
```
ganmacs: if a user sets URI which includes a placeholder (e.g. `http://127.0.0.1/${path}`) to endpoint directive,  does this line raise `URI::InvalidURIError`?

```rb
require 'net/http'
require 'uri'
 
URI.parse('http://127.0.0.1/${path}') #=> raise 'bad URI(is not URI?): "http://127.0.0.1/${path}" (URI::InvalidURIError)'
```

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+      opt = {
+        :open_timeout => @open_timeout,
```
ganmacs: To use newer version hash is better. ` { open_timeout: @open_timeout, .... }`

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+    d.run(default_tag: 'test.http', shutdown: false) do
+      test_events.each { |event|
+        d.feed(event)
+      }
+    end
+    assert_match(/got error response from.*404 Not Found Not Found/, d.instance.log.out.logs.first)
+    d.instance_shutdown
+  end
+
+  sub_test_case 'basie auth' do
```
cosmo0920:
```diff
-basie
+basic
```
nit: Typo?

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+          raise Fluent::ConfigError, "tls_ca_cert_path is wrong: #{@tls_ca_cert_path}" unless File.file?(@tls_ca_cert_path)
+          opt[:ca_file] = @tls_ca_cert_path
+        end
+        if @tls_client_cert_path
+          raise Fluent::ConfigError, "tls_client_cert_path is wrong: #{@tls_client_cert_path}" unless File.file?(@tls_client_cert_path)
+          opt[:cert] = OpenSSL::X509::Certificate.new(File.read(@tls_client_cert_path))
+        end
+        if @tls_private_key_path
+          raise Fluent::ConfigError, "tls_private_key_path is wrong: #{@tls_private_key_path}" unless File.file?(@tls_private_key_path)
+          opt[:key] = OpenSSL::PKey::read(File.read(@tls_private_key_path), @tls_private_key_passphrase)
```
ganmacs: [nits] OpenSSL::PKey.read

cosmo0920: :+1:

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+                            end
+        opt[:ciphers] = @tls_ciphers
+        opt[:ssl_version] = @tls_version
+      end
+
+      opt
+    end
+
+    def parse_endpoint(chunk)
+      endpoint = extract_placeholders(@endpoint, chunk)
```
cosmo0920: :+1:

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+              Net::HTTP.start(uri.host, uri.port, @proxy_uri.host, @proxy_uri.port, @proxy_uri.user, @proxy_uri.password, @http_opt) { |http|
+                http.request(req)
+              }
+            else
+              Net::HTTP.start(uri.host, uri.port, @http_opt) { |http|
+                http.request(req)
+              }
+            end
+
+      if res.is_a?(Net::HTTPSuccess)
```
ganmacs: This code treats response code 1xx( Net::HTTPInformation) and 3xx (Net::HTTPRedirection) as an error. Is it okay?

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
+require 'net/http'
+require 'uri'
+require 'openssl'
+require 'fluent/plugin/output'
+require 'fluent/plugin_helper/socket'
```
ganmacs: unnecessary require.

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+    end
+
+    result = @@result
+    assert_equal method.upcase, result.method
+    assert_equal 'application/x-ndjson', result.content_type
+    assert_equal test_events, result.data
+    assert_not_empty result.headers
+  end
+
+  def test_write_with_single_value_format
```
cosmo0920: test_write_with_single_value_format is failing in Ruby 2.1 and 2.2 tasks.
Could you check them?
https://travis-ci.org/fluent/fluentd/jobs/563321723#L1909

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+require 'webrick'
+require 'webrick/https'
+require 'net/http'
+require 'uri'
+require 'json'
+
+# WEBrick's ProcHandler doesn't handle PUT by default
+module WEBrick::HTTPServlet
+  class ProcHandler < AbstractServlet
+    alias do_PUT  do_GET
```
ganmacs: [nits] there are 2 spaces between `do_PUT` and `do_GET`

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+
+    def shutdown
+      @@http_server_thread.kill
+      @@http_server_thread.join
+    rescue
+    end
+  end
+
+  def server_port
+    19880
```
ganmacs: unused_port is better?

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+  end
+
+  def base_endpoint
+    "http://127.0.0.1:#{server_port}"
+  end
+
+  def server_config
+    config = {:BindAddress => '127.0.0.1', :Port => server_port}
```
ganmacs: [nits]`{ 'BindAddress': '127.0.0.1' ... }` ?

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+
+  def config
+    %[
+      endpoint #{base_endpoint}/test
+    ]
+  end
+
+  def test_events
+    [
+      {"message" => "hello", "num" => 10, "bool" => true},
```
ganmacs: [nits]To use a newer version hash is better.

test/plugin/test_out_http.rb
```diff
@@ -0,0 +1,351 @@
...
+      d.run(default_tag: 'test.http', shutdown: false) do
+        test_events.each { |event|
+          d.feed(event)
+        }
+      end
+    end
+    d.instance_shutdown
+  end
+
+  # basic auth test includes "error_response_as_unrecoverable true" case
```
ganmacs: unnecessary comment?

lib/fluent/plugin/out_http.rb
```diff
@@ -0,0 +1,213 @@
...
+              }
+            end
+
+      if res.is_a?(Net::HTTPSuccess)
+        log.debug { "#{res.code} #{res.message}#{res.body}" }
+      else
+        msg = "#{res.code} #{res.message}#{res.body}"
+
+        if @retryable_response_codes.include?(res.code.to_i)
+          raise RetryableResponse, msg
```
ganmacs: Who catches this error and does retry?

test/plugin/test_out_http.rb
repeatedly: What does "newer version" mean?
`{message: "hello"}` creates symbol key, not string key.

test/plugin/test_out_http.rb
repeatedly: I think this is needed because I don't add explicit standalone test for `true` version

lib/fluent/plugin/out_http.rb
repeatedly: output plugin. This is standard error so output plugin retry it.

lib/fluent/plugin/out_http.rb
repeatedly: This is needed because out_http uses several constants from this module.

test/plugin/test_out_http.rb
cosmo0920: `{"message": "hello", "num": 10, "bool": true}` style hash does not handle Ruby 2.1.
This feature had been provided Ruby 2.2 or later.
For now, we shouldn't drop Ruby 2.1 support.
So, hash rocket notation is needed here.

lib/fluent/plugin/out_http.rb
repeatedly: I think ok because it means sending data is not succeeded.

test/plugin/test_out_http.rb
ganmacs: Ah, string key is needed here. I see 👌

test/plugin/test_out_http.rb
repeatedly: I'm not sure which is the better. This is fixed value, not dynamic value like helper's unused_port.
So this naming is also ok for me.

test/plugin/test_out_http.rb
ganmacs: > I think this is needed because I don't add explicit standalone test for `true` version

Yes. but 

> basic auth test includes "error_response_as_unrecoverable true" case

what does this comment mean?
This test is not for basic auth, right? and basic auth test looks not having `error_response_as_unrecoverable true` case. why do we need this comment?

lib/fluent/plugin/out_http.rb
ganmacs: sorry… I missed it…

repeatedly:
out_http doesn't cover all cases of fluent-plugin-out-http so the successor is not the correct.
For example, this plugin doesn't support non-buffer mode and form format.
But for buffering case, I think this plugin is better than fluent-plugin-out-http so user can choose the plugin for the usecases.

cosmo0920:
> out_http doesn't cover all cases of fluent-plugin-out-http so the successor is not the correct.
> For example, this plugin doesn't support non-buffer mode and form format.
> But for buffering case, I think this plugin is better than fluent-plugin-out-http so users can choose the plugin for their usecases.

I see. Thanks for the information. :smiley:

PR #2490, opened 2019-07-10
Separate a connection with keepalive code from  normal connection

ganmacs:

I created a new method(`connect_keepalive`) to remove too many `if/unless @keepalive` since it's too complicated.
For the same reason as above, I changed send_data to use connect.

**Docs Changes**:
no needed

**Release Note**: 

no needed

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -1095,7 +1095,8 @@ module Fluent::Plugin
         begin
           yield(socket, request_info)
         ensure
-          socket.close
+          socket.close_write rescue nil
+          socket.close rescue nil
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -1079,38 +1065,54 @@ def on_read(sock, ri, data)
...
+          s = @sender.create_transfer_socket(host || resolved_host, port, @hostname)
+          request_info = RequestInfo.new(@sender.security ? :helo : :established) # overwrite if new connection
+          s
+        end
+
+        unless block_given?
+          return [socket, request_info]
+        end
+
+        ret =
```
repeatedly: This `ret` seems to not need.

ganmacs: I changed to return the value which `yield` returns because of following the default behavior. 
https://github.com/fluent/fluentd/blob/9dcf9488c3df269db37917ff09b7b91860d46fac/lib/fluent/plugin/out_forward.rb#L822-L834

https://github.com/fluent/fluentd/pull/2490/commits/60f324947f8d5450011c64ae8f2f73be0ced20fe

PR #2493, opened 2019-07-11
Add title to have unique thread name

ganmacs:
no

According to https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-thread#methods, create_thread receives `title` which must be unique as an argument and http_server_helper uses thread_helper internally.
There are no problems now since http_server is used only by in_monitor plugin.
But we need to change http_server can change their name to have a unique name.

**Docs Changes**:

yes. need to add the description about `title`.

**Release Note**: 

no

### DIFF
--- a/lib/fluent/plugin_helper/http_server.rb
+++ b/lib/fluent/plugin_helper/http_server.rb
@@ -34,11 +34,12 @@ module Fluent
       # close    : correct stopped threads
       # terminate: kill thread
 
+      # @param title [Symbol] the thread name. this value should be unique.
       # @param addr [String] Listen address
       # @param port [String] Listen port
       # @param logger [Logger] logger used in this server
       # @param default_app [Object] This method must have #call.
-      def create_http_server(addr:, port:, logger:, default_app: nil)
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin_helper/http_server.rb
```diff
@@ -34,11 +34,12 @@ module HttpServer
       # close    : correct stopped threads
       # terminate: kill thread
 
+      # @param title [Symbol] the thread name. this value should be unique.
       # @param addr [String] Listen address
       # @param port [String] Listen port
       # @param logger [Logger] logger used in this server
       # @param default_app [Object] This method must have #call.
-      def create_http_server(addr:, port:, logger:, default_app: nil)
+      def create_http_server(title:, addr:, port:, logger:, default_app: nil)
```
repeatedly: Normal argument is better for `title` argument. `server` and other helpers use normal argument in 1st position.

ganmacs: fixed https://github.com/fluent/fluentd/pull/2493/commits/2373dc30b30740b0469865c8fc93ad5cd542e748

PR #2499, opened 2019-07-15
in_syslog: add 'emit_unmatched_lines' option

candlerb:
Fixes #2497

Signed-off-by: Brian Candler <b.candler@pobox.com>

**What this PR does / why we need it**: 

Many devices emit syslog messages which don't match expected format, and are otherwise thrown away.  Add `emit_unmatched_lines` as per the [equivalent in_tail option](https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/plugins/input/tail.md#emit_unmatched_lines).

**Docs Changes**:

Add new parameter to https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/plugins/input/syslog.md

**Release Note**: 

```
### Enhancement

* in_syslog: Add `emit_unmatched_lines` to capture lines which don't parse
```

PR #2500, opened 2019-07-15
in_syslog: add option 'original_message_key'

candlerb:
(Note: this applies on top of PR #2499 )

N/A

Some log systems like [loki](https://github.com/grafana/loki) natively work on raw text lines, and are more natural to use in this form as opposed to a JSON collection of parsed fields.

Loki can accept certain parsed fields out-of-band as tags, but only if low-cardinality.  syslog parsing can generate high cardinality values, in particular `pid`, so all these values cannot be mapped to tags.  But if they are not mapped to tags, and the original line is not kept, then these values are lost.

I think it would also be useful with out_file to record the entire original syslog line.

**Docs Changes**:

Add new setting to https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/plugins/input/syslog.md

**Release Note**: 

```
### Enhancement

in_syslog: add `original_message_key` to retain the entire original message
```

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -82,6 +82,8 @@ module Fluent::Plugin
     config_param :include_source_host, :bool, default: false, deprecated: 'use "source_hostname_key" or "source_address_key" instead.'
     desc 'Specify key of source host when include_source_host is true.'
     config_param :source_host_key, :string, default: 'source_host'.freeze
+    desc 'Enable the option to emit unmatched lines.'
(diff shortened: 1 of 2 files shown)


repeatedly:
This is done by meta parser like multi-format-parser and it is more general approach.

```
<parse>
  @type keep_original  # wrap other parser and assign original `text` into parsed result.
  <parser>
    @type json
  </parser>
</parse>
```

candlerb:
For benefit of readers here is a link to [multi-format-parser](https://github.com/repeatedly/fluent-plugin-multi-format-parser)

PR #2501, opened 2019-07-16
Change SocketCache to be shared between nodes

ganmacs:

no

Old SocketCache isn't shared between other nodes since node's `port` and `host` is fixed at start time. 
In the situation nodes' port and host change dynamically, sharing SocketCache between nodes makes more efficient. 
And recent refactoring can make SocketCache more simple.

**Docs Changes**:

no need

**Release Note**: 

no need

### DIFF
--- a/lib/fluent/plugin/out_forward/socket_cache.rb
+++ b/lib/fluent/plugin/out_forward/socket_cache.rb
@@ -19,135 +19,118 @@ require 'fluent/plugin/output'
 module Fluent::Plugin
   class ForwardOutput < Output
     class SocketCache
-      TimedSocket = Struct.new(:timeout, :sock, :ref)
+      TimedSocket = Struct.new(:timeout, :key, :sock)
 
       def initialize(timeout, log)
         @log = log
         @timeout = timeout
-        @active_socks = {}
-        @inactive_socks = {}
+        @available_sockets = Hash.new { |obj, k| obj[k] = [] }
+        @inflight_sockets = {}
+        @inactive_sockets = []
         @mutex = Mutex.new
       end
 
-      def revoke(key = Thread.current.object_id)
+      def checkout_or(key)
         @mutex.synchronize do
-          if @active_socks[key]
-            @inactive_socks[key] = @active_socks.delete(key)
-            @inactive_socks[key].ref = 0
+          tsock = pick_socket(key)
+
+          unless tsock
+            val = yield
+            new_tsock = TimedSocket.new(timeout, key, val)
+            @log.debug("connect new socket #{new_tsock}")
+
+            @inflight_sockets[val] = new_tsock
+            return new_tsock.sock
+          end
+          tsock.sock
+        end
+      end
+
+      def checkin(sock)
+        @mutex.synchronize do
+          if (s = @inflight_sockets.delete(sock))
+            @available_sockets[s.key] << s
+          else
+            @log.debug("there is no socket #{sock}")
           end
         end
       end
 
-      def clear
+      def revoke(sock)
         @mutex.synchronize do
-          @inactive_socks.values.each do |s|
-            s.sock.close rescue nil
+          if (s = @inflight_sockets.delete(sock))
+            @inactive_sockets << s
+          else
+            @log.debug("there is no socket #{sock}")
           end
-          @inactive_socks.clear
-
-          @active_socks.values.each do |s|
-            s.sock.close rescue nil
-          end
-          @active_socks.clear
         end
       end
 
       def purge_obsolete_socks
+        sockets = []
+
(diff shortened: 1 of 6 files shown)


lib/fluent/plugin/out_forward/socket_cache.rb
```diff
@@ -19,135 +19,118 @@
...
+
+            @inflight_sockets[val] = new_tsock
+            return new_tsock.sock
           end
+          tsock.sock
         end
       end
 
-      def clear
+      def checkin(sock)
```
repeatedly: Hmm... `checkin` is unclear for me. After `checkin`, we are inflight for airplane but this method changes inflight to available. `checkout` too. Maybe, we don't use `checkout` for flight.

lib/fluent/plugin/out_forward/socket_cache.rb
```diff
@@ -19,135 +19,118 @@
...
-          @active_socks.keys.each do |k|
-            if expired?(k) && @active_socks[k].ref <= 0
-              @inactive_socks[k] = @active_socks.delete(k)
-            end
-          end
+          sockets += @inactive_sockets
+          @inactive_sockets.clear
+        end
+
+        while (s = sockets.pop)
```
repeatedly: `each` is more lightweight due to no `sockets` update.

lib/fluent/plugin/out_forward/socket_cache.rb
```diff
@@ -19,135 +19,118 @@
...
-      def dec_ref(key = Thread.current.object_id)
-        @mutex.synchronize do
-          if @active_socks[key]
-            @active_socks[key].ref -= 1
-          elsif @inactive_socks[key]
-            @inactive_socks[key].ref -= 1
-          else
-            @log.warn("Not found key for dec_ref: #{key}")
-          end
+        while (s = sockets.pop)
```
repeatedly: ditto

lib/fluent/plugin/out_forward/socket_cache.rb
```diff
@@ -19,135 +19,118 @@
...
-            @inactive_socks[key].ref = 0
+          tsock = pick_socket(key)
+
+          unless tsock
+            val = yield
+            new_tsock = TimedSocket.new(timeout, key, val)
+            @log.debug("connect new socket #{new_tsock}")
+
+            @inflight_sockets[val] = new_tsock
+            return new_tsock.sock
```
repeatedly: To reduce `return` by `if`

```
if tsock
  tsock.sock
else
  # same
end
```

lib/fluent/plugin/out_forward/socket_cache.rb
ganmacs: I referred connection_pool gem to determine the name of them.
https://github.com/mperham/connection_pool/blob/99b81aa0a9dccb3b423260740ce2d603f1479356/lib/connection_pool.rb#L80

Any name is okay for me if it's clear for readers. Do you have any suggestions?

lib/fluent/plugin/out_forward/socket_cache.rb
ganmacs: fixed https://github.com/fluent/fluentd/pull/2501/commits/5dd10e6f630f3b90f4911798387d144c1411e9bf

lib/fluent/plugin/out_forward/socket_cache.rb
ganmacs: right……… fixed https://github.com/fluent/fluentd/pull/2501/commits/eda267cee62e3230d700037fef0fd90a3076c5fa

lib/fluent/plugin/out_forward/socket_cache.rb
repeatedly: Hmm.... I see. If this naming is popular in ruby, it is okay for me.

lib/fluent/plugin/out_forward/socket_cache.rb
```diff
@@ -19,135 +19,119 @@
...
         @mutex.synchronize do
-          if @active_socks[key]
-            @inactive_socks[key] = @active_socks.delete(key)
-            @inactive_socks[key].ref = 0
+          tsock = pick_socket(key)
+
+          if tsock
+            tsock.sock
+          else
+            val = yield
```
repeatedly: `sock` or `new_sock` is better than `val`

ganmacs: fixed https://github.com/fluent/fluentd/pull/2501/commits/14985b352e909c984f26e44253bcdff13116a0ff

repeatedly:
looks good.

PR #2502, opened 2019-07-16
Pass `@nodes` to rebuild_weight_array

ganmacs:
Follow up https://github.com/fluent/fluentd/pull/2483/

i forgot to pass `@node` to rebuild_weight_array.

**Docs Changes**:
no needed

**Release Note**: 
no needed

cc @repeatedly


repeatedly:
good catch

PR #2514, opened 2019-07-22
Make LoadBalancer thread safe

ganmacs:

no

LoadBalancer was thread-unsafe.
LoadBalancer#select_healthy_node  and LoadBalancer#rebuild_weight_array are called in other threads. If context switch happens [this line](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward/load_balancer.rb#L36), and do [this line](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward/load_balancer.rb#L97) in another thread.  [node]( https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_forward/load_balancer.rb#L36) can be `nil`.

**Docs Changes**:

no need

**Release Note**: 

same as the title.

### DIFF
--- a/lib/fluent/plugin/out_forward/load_balancer.rb
+++ b/lib/fluent/plugin/out_forward/load_balancer.rb
@@ -25,15 +25,21 @@ module Fluent::Plugin
         @weight_array = []
         @rand_seed = Random.new.seed
         @rr = 0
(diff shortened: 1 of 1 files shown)


repeatedly:
Looks good.

PR #2516, opened 2019-07-22
Ack handler

ganmacs:
no

Extract the code around ack_response  as  AckHandler class.
It aims to simplify out_forward and add more test about it.

**Docs Changes**:

no needed

**Release Note**: 

no needed


ganmacs:
@repeatedly ping?

lib/fluent/plugin/out_forward.rb
```diff
@@ -428,95 +414,40 @@ def on_purge_obsolete_socks
...
-            commit_write(chunk_id) if chunk_id
+          case result
+          when AckHandler::Result::SUCCESS
+            commit_write(chunk_id)
+          when AckHandler::Result::FAILED
+            node.disable!
+            rollback_write(chunk_id, update_retry: false)
+          when AckHandler::Result::CHUNKID_UNMATCHED
+            rollback_write(chunk_id, update_retry: false)
+          else
```
repeatedly: result is 3 types so this `else` flow is for catching implementation bug, right?

lib/fluent/plugin/out_forward/ack_handler.rb
```diff
@@ -0,0 +1,161 @@
...
+              yield info.chunk_id, info.node, info.sock, ret
+            end
+          end
+        rescue => e
+          @log.error 'unexpected error while receiving ack', error: e
+          @log.error_backtrace
+        end
+      end
+
+      ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :chunk_id_base64, :node, :time, :timeout) do
```
repeatedly: I don't remember why this struct has `time` and `timeout` fields and calculate expired time in `expired?` call.
Could we change this struct like below?

```
ACKWaitingSockInfo = Struct.new(:sock, :chunk_id, :chunk_id_base64, :node, :expired_time) do
  def expired?(now)
    expired_time < now
  end
end
```

lib/fluent/plugin/out_forward.rb
ganmacs: Yes. this condition is for the case I don't expect now happen.

repeatedly: If so,  add `BUG:` to log message is better for notification.

https://github.com/fluent/fluentd/blob/73de02a559a2f51371a985bf907a98ab2e2333c2/lib/fluent/plugin_helper/socket.rb#L160

ganmacs:
Thank you for the review!

PR #2523, opened 2019-07-24
Fix warning printed when chunk key placeholder not replaced

dobesv:
This used `$1` to print the key name.  However, there are no capture groups in `CHUNK_KEY_PLACEHOLDER_PATTERN`, so `$1` would not be set.  According to the Ruby docs I looked up, `$&` can be used to get the part of the string that matched, which I think will give the correct output in this case.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -758,7 +758,7 @@ module Fluent
             rvalue = rvalue.gsub(CHUNK_KEY_PLACEHOLDER_PATTERN, hash)
           end
           if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
-            log.warn "chunk key placeholder '#{$1}' not replaced. template:#{str}"
+            log.warn "chunk key placeholder '#{$&}' not replaced. template:#{str}"
           end
           rvalue.sub(CHUNK_ID_PLACEHOLDER_PATTERN) {
             if chunk_passed


ganmacs:
> there are no capture groups in CHUNK_KEY_PLACEHOLDER_PATTERN, so $1 would not be set.

Right. but I think changing the regex pattern of  `CHUNK_KEY_PLACEHOLDER_PATTERN` from `/\$\{[-_.@$a-zA-Z0-9]+\}/` to `/\$\{([-_.@$a-zA-Z0-9]+)\}/` is better since we need to capture only the value of tag, not a whole of a tag.

an example using `$&`
```rb
CHUNK_KEY_PLACEHOLDER_PATTERN = /\$\{[-_.@$a-zA-Z0-9]+\}/

rvalue = "${a}"

if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
  p $&                          # "${a}"
end
```

an example using `/\$\{([-_.@$a-zA-Z0-9]+)\}/` for regex pattern
```rb
CHUNK_KEY_PLACEHOLDER_PATTERN = /\$\{([-_.@$a-zA-Z0-9]+)\}/

rvalue = "${a}"

if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
  p $1                          #=> "a"
end

```

dobesv:
Personally I would prefer to see the whole `${a}` but it's totally unimportant, TBH. Either would be a tremendous improvement.

dobesv:
You are welcome to make the change your way and make another PR, I won't be updating this PR.

ganmacs:
it's not for my preference. A current implementation which does the similar thing returns the value of the tag (https://github.com/fluent/fluentd/pull/2523/files#diff-7a3f304568ba6143aaf2b7e78011d044R749).  I think it's important to have consistency with other warnings.

PR #2524, opened 2019-07-25
buf_file: Add path_suffix parameter. ref #2236

### LINKED ISSUE #2236: Fluentd memory leak in Kubernetes with "buffer_type file"
[Fluentd version]
v1.2.5

[Environment]
Kubernetes

[Configuration]
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-gcp/fluentd-gcp-configmap.yaml

[Problems]
We are seeing Fluentd memory leaks with Kubernetes versions >= v1.10. After some investigation, we realized the log rotation mechanism changed from Kubernetes v1.9 and v1.10, which might trigger the memory leak. Guess the logs are rotated in a way that Fluentd does not support / handle well.

Some experiment with exactly the same Fluentd version, output plugin version but different Kubernetes versions are shown below:
image

Is there any known limitation for log rotation mechanism that might trigger this? We report a [similar issue](https://github.com/fluent/fluentd/issues/1941#issuecomment-412689268) before, but I've verified that we are using a Fluentd version with that fix.

[The log rotation manager]
In case it helps, the log rotation mechanism is https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/pkg/kubelet/logs/container_log_manager.go#L150.

repeatedly:

Fixes #2236 

Add `path_suffix` parameter to change suffix of buffer files.
This is mainly for k8s environment

**Docs Changes**:
Add path_suffix parameter to buf_file article.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -34,6 +34,8 @@ module Fluent
 
       desc 'The path where buffer chunks are stored.'
       config_param :path, :string, default: nil
+      desc 'The suffix of buffer chunks'
+      config_param :path_suffix, :string, default: '.log'
 
       config_set_default :chunk_limit_size, DEFAULT_CHUNK_LIMIT_SIZE
       config_set_default :total_limit_size, DEFAULT_TOTAL_LIMIT_SIZE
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/buf_file.rb
```diff
@@ -78,23 +80,23 @@ def configure(conf)
 
         if specified_directory_exists || unexisting_path_for_directory # directory
           if using_plugin_root_dir || !multi_workers_configured
-            @path = File.join(@path, 'buffer.*.log')
```
ganmacs: Probably it's better creating this string object at #configure since `@path_suffix` doesn't change.

PR #2525, opened 2019-07-26
in_tail: Add switch_io_after_read parameter. fix #2478

### LINKED ISSUE #2478: in_tail take long time to found rotated files
related to issue https://github.com/fluent/fluentd/issues/2477

Will there be any risk if I change read_more = true to read_more = false, such as losing record or other issues, just in my copy of in_tail source code?

We came across a problem here. We have ten log files, our log-generator writes record(0.5KB per record) to them by 1000 lines/s each and rotate it by renaming the file to another name(the inode changed but filename stay the same) if the size of the file reaches 100MB. We use in_tail to collect these files.

We found two problems.

1. when the file is rotated, it takes a long time for in_tail to discover the new inode of the logfile, and after several rotate, it seems in_tail just lost some of the rotated files(rotate is quite fast)
2. when starting the fluentd, I found it takes a long time for in_tail to start_watchers for these files when I look at the log of fluentd(I add some debug log after setup_watcher). And with the rotate issues, it seems fluentd finally lost some of the log files.

I'm not familiar with in_tail's performance, so maybe it's not in_tail's problem but the way I use it. The usage of CPU is not high and the flush speed is ok.

For the moment, I solve these two problems with change read_more = true to read_more = false(it take effects), but I'm not sure if there will be any risk with this change.

It would be very kind of you if you can give me some advice with these two problems.

repeatedly:

Fixes #2478 

Add `switch_io_after_read` parameter to control IO read operation.
I tested several cases and no problem. But I can't say all environment have no problem with this change so I add parameter for smooth migration. 

**Docs Changes**:
Will submit a PR to add parameter for v1.6.3.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -92,6 +92,8 @@ module Fluent::Plugin
     config_param :open_on_every_update, :bool, default: false
     desc 'Limit the watching files that the modification time is within the specified time range (when use \'*\' in path).'
     config_param :limit_recently_modified, :time, default: nil
+    desc 'After read lines, switch IO operation to other file'
+    config_param :switch_io_after_read, :bool, default: false  # TODO: set true in fluentd v2
     desc 'Enable the option to skip the refresh of watching list on startup.'
     config_param :skip_refresh_on_startup, :bool, default: false
     desc 'Ignore repeated permission error logs'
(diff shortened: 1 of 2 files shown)


ganmacs:
According to this comment https://github.com/fluent/fluentd/issues/2478#issuecomment-512086345, I think `@fifo` and the contents of a file should be flushed not to lost the data when detaching.

>  resolve this by setting rotate_wait to a bigger value

I think this is good for a workaround. but not for the final solution.

ashie:
It seems that #3185 is now going to include this fix.
I'll close this if #3185 is landed with similar implementation with this.
On the other hand, it doesn't yet resolve problems mentioned in https://github.com/fluent/fluentd/issues/2478#issuecomment-512086345, so further work is needed.

ashie:
#2478 has been merged.
Setting `read_bytes_limit_per_second` to `8192` (or less than it) will take same effect with this.

PR #2527, opened 2019-07-26
Replace in_tail with a faster implementation.

ganmacs:
no

`slice!` is very slow. so I replaced in_tail implementation with `slice` + `freeze`.
This change makes FIFO of in_tail about 100x faster in ruby 2.6's simple benchmark.

benchmark script & result: https://gist.github.com/ganmacs/3f517c79881f56547cde305ae7a86b35

**Docs Changes**:

no

**Release Note**: 

Replace in_tail with faster implementation.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -700,7 +700,14 @@ module Fluent::Plugin
 
         def next_line
           idx = @buffer.index(@eol)
-          convert(@buffer.slice!(0, idx + 1)) unless idx.nil?
+
+          unless idx.nil?
+            # Using freeze and slice is faster than slice!
+            @buffer.freeze
+            rbuf = @buffer.slice(0, idx + 1)
+            @buffer = @buffer.slice(idx + 1, @buffer.size)
+            convert(rbuf)
+          end
         end
 
         def bytesize
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -698,15 +698,17 @@ module Fluent::Plugin
           s.encode!(@encoding, @from_encoding, :invalid => :replace, :undef => :replace)
         end
 
-        def next_line
+        def read_lines(lines)
           idx = @buffer.index(@eol)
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Hmm... why `slice!` is so slow?

repeatedly:
How about reducing method call by passing `@lines`?

```
# call side
@fifo.parse_lines(@lines)

# implementation
def parse_lines(lines)
  idx = @buffer.index(@eol)

  until idx.nil?
    @buffer.freeze
    rbuf = @buffer.slice(0, idx + 1)
    @buffer = @buffer.slice(idx + 1, @buffer.size)
    idx = @buffer.index(@eol)
    lines << convert(rbuf)
  end
end
```

ganmacs:
According to the [profiling](https://gist.github.com/ganmacs/f099e9f95c1c0191de65616a1bbf463f), `slice` + `freeze` version allocate less memory than `slice!` one.

`slice!` seems to create a new string object per each call and copy whole value given as an argument.

ganmacs:
> How about reducing method call by passing @lines?

looks good. I confirmed this way is faster. 
https://github.com/fluent/fluentd/pull/2527/commits/0b17c0441dc787ab9096051dceba69aa64edddaf

repeatedly:
Good.

repeatedly:
Note: fluentd uses `slice!` in several code.

```
% grep -rne 'slice!' lib/fluent/                                                                   (git)-[master]
lib/fluent//plugin_helper/server.rb:753:                @_handler_write_buffer.slice!(0, written_bytes)
lib/fluent//compat/socket_util.rb:85:          @buffer.slice!(0, pos) if pos > 0
lib/fluent//plugin/in_tcp.rb:86:          conn.buffer.slice!(0, pos) if pos > 0
lib/fluent//plugin/in_syslog.rb:195:          buffer.slice!(0, pos) if pos > 0
lib/fluent//plugin/buffer/memory_chunk.rb:51:          @chunk.slice!(@chunk_bytes, @adding_bytes)
```

ganmacs:
I will replace them in my spare time.

PR #2551, opened 2019-08-12
Not to call IO#truncate to avoid too many IO flush

ganmacs:
Fixes #

IO#truncate call IO#flush internally. this cause heavy IO in case file buffer of fluentd.
so I change file buffer metadata format not to call IO#truncate.

**Docs Changes**:

*Release Note**: 

same as title.


repeatedly:
Good.

PR #2553, opened 2019-08-12
add support for source_hostname_key and source_address_key on umatched syslog messages

### LINKED ISSUE #2552: "emit_unmatched_lines" ommits source_hostname_key or source_address_key
**Describe the bug**
https://github.com/fluent/fluentd/pull/2499 introduced the option `emit_unmatched_lines` to the syslog input to keep messages in the pipeline even if parsing fails.
Due to the way this was implemented `source_hostname_key` or `source_address_key` do not get included in the emitted message even though this information is provided by the underlying udp/tcp server.

https://github.com/fluent/fluentd/blob/9e881035bdfb4b65e320d0521b3ddca783a021ea/lib/fluent/plugin/in_syslog.rb#L208-L210

https://github.com/fluent/fluentd/blob/9e881035bdfb4b65e320d0521b3ddca783a021ea/lib/fluent/plugin/in_syslog.rb#L220-L222

https://github.com/fluent/fluentd/blob/9e881035bdfb4b65e320d0521b3ddca783a021ea/lib/fluent/plugin/in_syslog.rb#L240-L242

**To Reproduce**
* Create syslog input with `emit_unmatched_lines true` and `source_hostname_key source_hostname` and inject some unparseable message.
* the emitted message only contains the key `unmatched_line` and no `source_hostname`

**Expected behavior**
`source_hostname_key` and `source_address_key` should be honored even for unmatched lines.

**Your Environment**

- Fluentd or td-agent version: fluentd 1.6.3
- Operating system: Debian 9

aeber:
Fixes #2552 

Unmatched messages from the syslog input should contain `source_hostname_key` or
`source_address_key` if configured as this information is provided by the
underlying udp oder tcp server and is available even if the message
could not be parsed.

This PR will add support for the expected behaviour of honoring `source_address_key` and `source_hostname_key` on unmatched lines.

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -199,6 +199,13 @@ module Fluent::Plugin
 
     private
 
+    def emit_unmatched(data, sock)
+      record = {"unmatched_line" => data}
+      record[@source_address_key] = sock.remote_addr if @source_address_key
+      record[@source_hostname_key] = sock.remote_host if @source_hostname_key
+      emit("#{@tag}.unmatched", Fluent::EventTime.now, record)
+    end
+
     def message_handler(data, sock)
       pri = nil
       text = data
@@ -206,7 +213,7 @@ module Fluent::Plugin
         m = SYSLOG_REGEXP.match(data)
         unless m
           if @emit_unmatched_lines
-            emit("#{@tag}.unmatched", Fluent::EventTime.now, {"unmatched_line" => data})
+            emit_unmatched(data, sock)
           end
           log.warn "invalid syslog message: #{data.dump}"
           return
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_syslog.rb
```diff
@@ -199,14 +199,22 @@ def start_tcp_server(tls: false)
 
     private
 
+    def emit_unmatched(data, sock)
+      record = {}
```
repeatedly: `record = {"unmatched_line" => data}` is better. No need line 206

aeber: Yip. Changed and force pushed.

repeatedly:
Looks good. Thanks!

PR #2556, opened 2019-08-13
socket_util.rb: correct misspelled word

ethan-daocloud:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>

Fixes #

make message to end user more meaningfull and understandable

**Docs Changes**:

**Release Note**:


repeatedly:
Ah, good catch!

PR #2557, opened 2019-08-13
Use faster way to get sec and nsec

ganmacs:
no

`Process.clock_gettime` is faster than `Time.now` if getting sec and nsec.
and `Time.now` is called in `EventTime.now`.
EventTime.now is called per data in [parser_none](https://github.com/fluent/fluentd/blob/785fe3df5df76ae7110ea226c506131d8e607aa6/lib/fluent/plugin/parser_none.rb#L31) and per [chunk creation](https://github.com/fluent/fluentd/blob/785fe3df5df76ae7110ea226c506131d8e607aa6/lib/fluent/plugin/buffer/chunk.rb#L60-L61). so it's better to replace them with faster version's method.

```ruby
def now1
  now = Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
  sec = now / 1_000_000_000
  nsec = now % 1_000_000_000
  [sec, nsec]
end

def now2
  t = Time.now
  sec = t.to_i
  nsec = t.nsec
  [sec, nsec]
end

require 'benchmark/ips'

Benchmark.ips do |x|
  x.report("Time.now") do
    now2
  end

  x.report("Process.clock_gettime") do
    now1
  end
end
```

```
            Time.now   191.241k i/100ms
Process.clock_gettime
                       291.915k i/100ms
            Time.now      2.956M (± 1.9%) i/s -     14.917M in   5.047422s
Process.clock_gettime
                          5.563M (± 4.5%) i/s -     28.024M in   5.047899s
```

**Docs Changes**:

no need

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -111,7 +111,8 @@ module Fluent
     end
 
     def self.now
-      now = Fluent::Clock.real_now(:nanosecond)
+      # This method is called many time. so call Process.clock_gettime directly instead of Fluent::Clock.real_now
+      now = Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
       Fluent::EventTime.new(now / 1_000_000_000, now % 1_000_000_000)
     end
 
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -111,7 +111,8 @@ module Fluent
     end
 
     def self.now
-      from_time(Time.now)
+      now = Fluent::Clock.real_now(:nanosecond)
+      Fluent::EventTime.new(now / 1_000_000_000, now % 1_000_000_000)
     end
 
     def self.parse(*args)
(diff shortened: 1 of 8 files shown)


ganmacs:
memo: timecop's issue. https://github.com/travisjeffery/timecop/issues/220

lib/fluent/time.rb
```diff
@@ -111,7 +111,8 @@ def self.eq?(a, b)
     end
 
     def self.now
-      from_time(Time.now)
+      now = Fluent::Clock.real_now(:nanosecond)
```
repeatedly: self.now is called frequently so how about call `Process.clock_gettime` directly to reduce function call?

ganmacs: fixed https://github.com/fluent/fluentd/pull/2557/commits/809af50a26d8477d5081908cd604694667734111

PR #2559, opened 2019-08-14
Cache msgpack packer/unpacker

ganmacs:

cache msgpack packer/unpacker not to create unnecessary objects.
https://gist.github.com/ganmacs/a620be919851d49bff3bf2de58da2052 shows decreasing CPU time of ` Fluent::EventStream#to_msgpack_stream` and ` Fluent::MessagePackFactory::Mixin#msgpack_packer`.

**Docs Changes**:

no need

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/event_router.rb
+++ b/lib/fluent/event_router.rb
@@ -17,6 +17,7 @@
 require 'fluent/match'
 require 'fluent/event'
 require 'fluent/filter'
+require 'fluent/msgpack_factory'
 
 module Fluent
   #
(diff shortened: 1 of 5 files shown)


lib/fluent/event_router.rb
```diff
@@ -182,7 +184,7 @@ def filter_stream(tag, es)
 
         def optimized_filter_stream(tag, es)
           new_es = MultiEventStream.new
-          es.each do |time, record|
+          es.each(unpacker: @unpacker) do |time, record|
```
repeatedly: Is this safe? event_router can be called from multiple inputs.

ganmacs: good catch! 
https://github.com/fluent/fluentd/pull/2559/commits/b8021e9857a7815ea92faaa0acef11dbfe0a240d

PR #2561, opened 2019-08-14
Use Mutex instead of Monitor

ganmacs:

Fixes #

I used Mutex when no need reentrant lock since Mutex is faster than Montior.

```rb
require 'benchmark/ips'

mo = Monitor.new
mu = Mutex.new

Benchmark.ips do |x|
  x.report("monitor") do
    mo.synchronize {}
  end

  x.report("mutex") do
    mu.synchronize {}
  end
end
```

```
             monitor   224.461k i/100ms
               mutex   371.210k i/100ms
             monitor      3.987M (± 2.4%) i/s -     19.977M in   5.013755s
               mutex     10.221M (± 2.5%) i/s -     51.227M in   5.015317s
```

**Docs Changes**:

no need

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -352,7 +352,7 @@ module Fluent
           write_guard do
             @buffer.write({meta => data}, format: ->(_data){ _data }, size: ->(){ size }, enqueue: enqueue)
           end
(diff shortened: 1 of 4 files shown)


repeatedly:
Looks good.

PR #2563, opened 2019-08-15
Remove `@metadata_list` from buffer.rb

ganmacs:
Fixes #

I removed `@metada_list` from buffer.rb since this ivar has almost no meaning. Moreover, this is an overhead since Buffer#add_metadata is called many times.

**Docs Changes**:

no need

**Release Note**: 

same as title


repeatedly:
I noticed `metadata` calls `new_metadata` but calling keyword argument function is slow.
How abut creating `Metadata` directly instead of `new_metadata`?

https://github.com/fluent/fluentd/blob/f2b24f509d4c5f3c123ae3a1377eb36fa72f5a3b/lib/fluent/plugin/buffer.rb#L265

lib/fluent/plugin/buffer.rb
```diff
@@ -532,13 +491,16 @@ def purge_chunk(chunk_id)
...
-            @metadata_list.delete(metadata)
             @queued_num.delete(metadata)
             @dequeued_num.delete(metadata)
-            del_timekey(metadata)
           end
           log.trace "chunk purged", instance: self.object_id, chunk_id: dump_unique_id_hex(chunk_id), metadata: metadata
         end
+
+        if metadata && metadata.timekey
+          @mutex.synchronize { del_timekey(metadata.timekey) }
```
repeatedly: `@mutex.synchronize` is called in `del_timekey`

ganmacs: I forgot about it! thanks.
 .https://github.com/fluent/fluentd/pull/2563/commits/5c81b40df6586e0ad659ef045ea0536b969e13b2

repeatedly:
Looks good.

PR #2565, opened 2019-08-15
OneEvent isn't always empty

ganmacs:

Fixes #

```rb
require 'benchmark/ips'

class A
  def empty?
    size == 0
  end
end

class B < A
  def size
    1
  end
end

class C < A
  def empty?
    false
  end
end


b = B.new
c = C.new

Benchmark.ips do |x|
  x.report("b") do
    b.empty?
  end

  x.report("c") do
    c.empty?
  end
end
```

```
                   b   437.425k i/100ms
                   c   454.017k i/100ms
                   b     16.370M (± 6.5%) i/s -     81.798M in   5.021556s
                   c     22.938M (± 3.4%) i/s -    114.866M in   5.013438s
```

**Docs Changes**:

no need

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -87,6 +87,10 @@ module Fluent
       OneEventStream.new(@time, @record.dup)
     end
 
+    def empty?
+      false
+    end
+
     def size
       1
     end


repeatedly:
Good catch!

PR #2567, opened 2019-08-15
in_tcp: Improve the performance for multiple events case

repeatedly:
None

Use `MultiEventStream` to improve the performance for multiple events in receive data.

**Docs Changes**:
No need.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/in_tcp.rb
+++ b/lib/fluent/plugin/in_tcp.rb
@@ -61,13 +61,15 @@ module Fluent::Plugin
     def start
       super
 
+      del_size = @delimiter.length
       if @_extract_enabled && @_extract_tag_key
         server_create(:in_tcp_server_single_emit, @port, bind: @bind, resolve_name: !!@source_hostname_key) do |data, conn|
           conn.buffer << data
+          buf = conn.buffer
           pos = 0
-          while i = conn.buffer.index(@delimiter, pos)
-            msg = conn.buffer[pos...i]
-            pos = i + @delimiter.length
+          while i = buf.index(@delimiter, pos)
+            msg = buf[pos...i]
+            pos = i + del_size
 
             @parser.parse(msg) do |time, record|
               unless time && record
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tcp.rb
```diff
@@ -83,7 +85,32 @@ def start
               router.emit(tag, time, record)
             end
           end
-          conn.buffer.slice!(0, pos) if pos > 0
+          buf.slice!(0, pos) if pos > 0
+        end
+      else
+        server_create(:in_tcp_server_batch_emit, @port, bind: @bind, resolve_name: !!@source_hostname_key) do |data, conn|
+          conn.buffer << data
```
ganmacs: Is it needed if using a local variable(`buf`)? #buffer is just a string. It is used only for buffering https://github.com/fluent/fluentd/blob/5a7d94941371dddcb7150a73ddbe142102a68964/lib/fluent/plugin_helper/server.rb#L448.

repeatedly: `buf = conn.buffer` is for avoiding `buffer` accessor call in `while`.

ganmacs: My question is `conn.buffer << data` is needed? `conn.buffer` will not be used in the following logic.

repeatedly: Yes. `conn.buffer` may have previous data so need to concat before use.

ganmacs: ah, I see 👍

lib/fluent/plugin/in_tcp.rb
```diff
@@ -83,7 +85,32 @@ def start
...
+                next
+              end
+
+              time ||= extract_time_from_record(record) || Fluent::EventTime.now
+              record[@source_address_key] = conn.remote_addr if @source_address_key
+              record[@source_hostname_key] = conn.remote_host if @source_hostname_key
+              es.add(time, record)
+            end
+          end
+          router.emit_stream(@tag, es)
```
ganmacs: This PR has already been merged. but I found a considering point.

if `es` becomes bigger than [Buffer#chunk_size_over?](https://github.com/fluent/fluentd/blob/2ad5b6de80cdd64d2ba27526f806e91323e12a68/lib/fluent/plugin/buffer.rb#L564), fluentd shows warning log and does rollback constantly.
so I think it needs to add a configuration parameter to set the size of stream or something.

ganmacs: > so I think it needs to add a configuration parameter to set the size of streams or something.

re-think carefully, In this situation,  we can set`chunk_limit_size`. so it's not problem.

PR #2568, opened 2019-08-16
Update in_tail.rb

Projoke:
Change replace character from "?" to "" when encoding errors happened.

Fixes # Change replace character from default "?" to "" when encoding error happend.

None
**Docs Changes**: None

**Release Note**:  None

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -695,7 +695,7 @@ module Fluent::Plugin
             s
           end
         rescue
-          s.encode!(@encoding, @from_encoding, :invalid => :replace, :undef => :replace)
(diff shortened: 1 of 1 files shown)


ganmacs:
This is breaking change. could you tell us why you need this change? 

And could you follow DCO https://github.com/fluent/fluentd/pull/2568/checks?check_run_id=194934513 ?

Projoke:
> This is breaking change. could you tell us why you need this change?
> 
> And could you follow DCO https://github.com/fluent/fluentd/pull/2568/checks?check_run_id=194934513 ?

In our situation, character "" is a more transparent solution. Or, parameters are needed here. Otherwise, just expose the exception string.

ganmacs:
> In our situation, character "" is a more transparent solution.

Why? Could you tell us more specific?

ganmacs:
I'm closing this because we can get why this feature is needed.

Issue #2555, opened 2019-08-13
Cannot set utc in <parse> section


skaji:
**Describe the bug**

Cannot set utc in `<parse>` section:
```
❯ cat test.conf
<source>
  @type tail
  path test.log
  pos_file pos
  tag test
  <parse>
    @type json
    utc true
  </parse>
</source>

❯ fluentd --dry-run -c test.conf
2019-08-13 15:38:44 +0900 [info]: parsing config file is succeeded path="test.conf"
2019-08-13 15:38:44 +0900 [info]: starting fluentd-1.6.3 as dry run mode ruby="2.6.3"
2019-08-13 15:38:44 +0900 [error]: config error file="test.conf" error_class=Fluent::ConfigError error="both of utc and localtime are specified, use only one of them"
```

**To Reproduce**

* Prepare test.conf in "Describe the bug"
* Execute fluentd.

**Expected behavior**

Use UTC, and be able to run fluentd

**Your Environment**

- Fluentd or td-agent version
    ```
    ❯ fluentd --version
    fluentd 1.6.3
    ```
- Operating system
    ```
    ❯ sw_vers
    ProductName:	Mac OS X
    ProductVersion:	10.14.6
    BuildVersion:	18G84
    ```
- Kernel version
    ```
    ❯ uname -a
    Darwin MBP-19-002 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin
    ```

**Your Configuration**

```
<source>
  @type tail
  path test.log
  pos_file pos
  tag test
  <parse>
    @type json
    utc true
  </parse>
</source>
```

**Your Error Log**

```
2019-08-13 15:38:44 +0900 [error]: config error file="test.conf" error_class=Fluent::ConfigError error="both of utc and localtime are specified, use only one of them"
```

**Additional context**

It seems that this bug has been fixed in https://github.com/fluent/fluentd/pull/1319
Does it becomes a problem again?

repeatedly:
Looks so. The problem line is here https://github.com/fluent/fluentd/blob/5a7d94941371dddcb7150a73ddbe142102a68964/lib/fluent/time.rb#L158 and in_tail creates `parser_create` with `conf` argument.
[...]

PR #2569, opened 2019-08-16
Not to call Parser#configure is called twice

ganmacs:

Fixes https://github.com/fluent/fluentd/issues/2555

First time is in parser helper
https://github.com/fluent/fluentd/blob/5a7d94941371dddcb7150a73ddbe142102a68964/lib/fluent/plugin_helper/parser.rb#L82 (and the `localtime` is set in the first time. https://github.com/fluent/fluentd/issues/2555#issuecomment-521802165)
Second time is parser_create in TailInput
https://github.com/fluent/fluentd/blob/5a7d94941371dddcb7150a73ddbe142102a68964/lib/fluent/plugin/in_tail.rb#L151

**Docs Changes**:

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -97,6 +97,10 @@ module Fluent::Plugin
     desc 'Ignore repeated permission error logs'
     config_param :ignore_repeated_permission_error, :bool, default: false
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Confirmed this patch works.

PR #2574, opened 2019-08-16
Set thread title

jkohen:
This helps understand thread usage with commands like `top -H` and `ps -e -T`

Before this change the thread name shows up as `thread.rb:70`. After, it may show up as whatever title was given. The exact behavior will depend on the implementation of Ruby's `native_set_another_thread_name`.

### DIFF
--- a/lib/fluent/plugin_helper/thread.rb
+++ b/lib/fluent/plugin_helper/thread.rb
@@ -92,7 +92,7 @@ module Fluent
           end
         end
         thread.abort_on_exception = true
-        thread.name = title.to_s
+        thread.name = title.to_s if thread.respond_to?(:name)
         @_threads_mutex.synchronize do
           @_threads[thread.object_id] = thread
         end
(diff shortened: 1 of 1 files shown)


jkohen:
It looks like the `Thread.name=` method was introduced in Ruby 2.3.0, hence the build failures. Any suggestions on how to get that to work? We are investigating what is causing a high number of threads in fluentd, which affects the memory usage in GKE.

FWIW, Ruby 2.3.8 is already EOL and Travis doesn't seem to be testing against 2.3.x. When are you planning to stop supporting 2.1.10?

repeatedly:
> Any suggestions on how to get that to work?

How about `thread.name = title.to_s if thread.respond_to?(:name)` ?

jkohen:
> > Any suggestions on how to get that to work?
> 
> How about `thread.name = title.to_s if thread.respond_to?(:name)` ?

Cool, that seems to have worked! Thanks, I'm all new to Ruby.

ganmacs:
Thanks!

PR #2577, opened 2019-08-17
in_tail: Dump a file name in addition to a malformed line

fujimotos:
Fixes #

This patch is based on a feature request from one of our customers.

Right now, in_tail emits the following warning message when it reads
a malformed line.

    2019-08-18 00:20:03 +0900 [warn]: #0 "a,b\",c" error="Illegal quoting in line 1."

However, since we use in_tail extensively (e.g. tracking lots of files
at once), it is sometimes not easy to determine WHICH file caused the
error just by seeing this message.

With this patch, in_tail shows:

    2019-08-18 00:27:42 +0900 [warn]: #0 invalid line found file="test.csv"
    line="a,b\",c" error="Illegal quoting in line 1."

... which should be useful for debugging.

**Docs Changes**:

No need to update the document.

**Release Note**: 

Use the PR title.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -436,7 +436,7 @@ module Fluent::Plugin
           end
         }
       rescue => e
-        log.warn line.dump, error: e.to_s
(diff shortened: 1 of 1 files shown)


repeatedly:
Looks good. Thanks!

Issue #2153, opened 2018-10-23
Greatly increased file buffer write i/o activity moving from 0.12.x to 1.2.x


mffiedler:
- fluentd 1.2.6 file buffer behavior vs. fluentd 0.12.43
- RHEL 7.5.x, fluentd configured to read Docker json-file container logs
```aconf
      <buffer>
        @type file
        path '/var/lib/fluentd/buffer-output-es-config'
        flush_interval 1s
        flush_thread_count 8
        flush_at_shutdown 'false'
        retry_max_interval 300
        retry_forever true
        queued_chunks_limit_size 32
        chunk_limit_size 8m
        overflow_action block
      </buffer>
```

- Identical workload for both fluentd versions
- constant 1000 1KB messages per second per fluentd to be forwarded to elasticsearch.   
- 1.2.6 uses much less cpu than 0.12.43 (50% as much) 
- 1.2.6 has much higher write I/O activity to the file buffer directory (verified with sysdig).   
- eventually, 1.2.6 is throttled on iowait when higher message rates (2000 1KB messages/sec) are attempted - the nvme is provisioned for 3000 IOPS
- attached are graphs with a) write IOPS activity comparison from iostat and b) block device write activity comparison from vmstat.   
- Both fluentd systems are on identical AWS m5.xlarge instances with nvme storage for /var/lib/fluentd.

At 1000 IOPS for a workload of 1000 messages/second it seems 1.2.6 is performing a unique IO for each message.   I've tried adjusting the flush_interval, but still see consistent 1000 write iops for this workload.

![block_dev_write](https://user-images.githubusercontent.com/1133529/47385311-a1765680-d6d7-11e8-86f2-e756435bd4e7.png)
![write_iops](https://user-images.githubusercontent.com/1133529/47385312-a1765680-d6d7-11e8-8f13-381099f817f0.png)

Any thoughts on tuning this to lower the disk activity?  Other info that would help?

repeatedly:
fluentd v1 has metadata for buffer chunks so the number of IO operation is increased.
fluentd buffers handles these operations per event stream.
Do you send 1 event x 1000 / sec or 1000 msg x 1 / sec?

richm:
In this test, input is coming from in_tail for kubernetes/docker json-file logs: https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/lib/generate_throttle_configs/lib/generate_throttle_configs.rb#L183
```
<source>
  @type tail
[...]

PR #2579, opened 2019-08-19
Add buf_file_single plugin

repeatedly:
Ref https://github.com/fluent/fluentd/issues/2153

`buf_file` works fine for typical cases, e.g. `in_tail`/`in_forward`, but it is not good for lots of small emit cases. It issues lots of IO operation.
So to reduce IO tps on high traffic environment, introduce `buf_file_single`. This is similar to v0.12 buf_file and this is lightweight. 

`buf_file_single` has one limitation that supports only tag or one field chunk key, `<buffer tag>` or `<buffer key>`. This limitation will be relaxed by supporting metadata header in the file.

**Docs Changes**:
Add `buf_file_single` article. Will send a patch later.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/buf_file_single.rb
+++ b/lib/fluent/plugin/buf_file_single.rb
@@ -70,6 +70,8 @@ module Fluent
         else
           if owner.chunk_key_tag
             raise Fluent::ConfigError, "chunk keys must be tag or one field"
+          elsif owner.chunk_keys.size > 1
+            raise Fluent::ConfigError, "2 or more chunk keys is not allowed"
           else
             @key_in_path = owner.chunk_keys.first.to_sym
           end
@@ -110,8 +112,9 @@ module Fluent
           @multi_workers_available = true
         else # specified path is file path
           if File.basename(@path).include?('.*.')
-            log.warn "file_single doesn't allow user specified 'prefix.*.suffix' style path. Use 'fsb.*#{PATH_SUFFIX}' instead: #{@path}"
-            @path = "fsb.*#{PATH_SUFFIX}"
+            new_path = File.join(File.dirname(@path), "fsb.*#{PATH_SUFFIX}")
+            log.warn "file_single doesn't allow user specified 'prefix.*.suffix' style path. Use '#{new_path}' for file instead: #{@path}"
+            @path = new_path
           elsif File.basename(@path).end_with?('.*')
             @path = @path + PATH_SUFFIX
           else
@@ -131,7 +134,7 @@ module Fluent
       # This method is called only when multi worker is configured
       def multi_workers_ready?
         unless @multi_workers_available
-          log.error "file buffer with multi workers should be configured to use directory 'path', or system root_dir and plugin id"
+          log.error "file_single buffer with multi workers should be configured to use directory 'path', or system root_dir and plugin id"
         end
         @multi_workers_available
       end
--- a/lib/fluent/plugin/buf_file_single.rb
+++ b/lib/fluent/plugin/buf_file_single.rb
@@ -43,8 +43,10 @@ module Fluent
       config_set_default :chunk_limit_size, DEFAULT_CHUNK_LIMIT_SIZE
       config_set_default :total_limit_size, DEFAULT_TOTAL_LIMIT_SIZE
 
+      desc 'The permission of chunk file. If no specified, <system> setting or 0644 is used'
       config_param :file_permission, :string, default: nil # '0644'
-      config_param :dir_permission,  :string, default: nil # '0755'
+      desc 'The permission of chunk directory. If no specified, <system> setting or 0644 is used'
+      config_param :dir_permission, :string, default: nil # '0755'
 
       @@buffer_paths = {}
 
--- a/lib/fluent/plugin/buf_file_single.rb
+++ b/lib/fluent/plugin/buf_file_single.rb
@@ -160,8 +160,7 @@ module Fluent
 
           log.debug { "restoring buffer file: path = #{path}" }
 
-          m = new_metadata() # this metadata will be overwritten by resuming .meta file content
-                             # so it should not added into @metadata_list for now
+          m = new_metadata() # this metadata will be updated in FileSingleChunk.new
           mode = Fluent::Plugin::Buffer::FileSingleChunk.assume_chunk_state(path)
           if mode == :unknown
             log.debug "unknown state chunk found", path: path
@@ -169,7 +168,7 @@ module Fluent
           end
 
           begin
-            chunk = Fluent::Plugin::Buffer::FileSingleChunk.new(m, path, mode, @key_in_path) # file chunk resumes contents of metadata
+            chunk = Fluent::Plugin::Buffer::FileSingleChunk.new(m, path, mode, @key_in_path)
             chunk.restore_size(@chunk_format) if @calc_num_records
           rescue Fluent::Plugin::Buffer::FileSingleChunk::FileChunkError => e
             handle_broken_files(path, mode, e)
@@ -184,7 +183,7 @@ module Fluent
           end
         end
 
-        queue.sort_by! { |chunk| chunk.modified_at }
+        queue.sort_by!(&:modified_at)
 
         return stage, queue
       end
@@ -203,7 +202,7 @@ module Fluent
       end
 
       def handle_broken_files(path, mode, e)
-        log.error "found broken chunk file during resume. Delete corresponding files:", :path => path, :mode => mode, :err_msg => e.message
+        log.error "found broken chunk file during resume. Delete corresponding files:", path: path, mode: mode, err_msg: e.message
         # After support 'backup_dir' feature, these files are moved to backup_dir instead of unlink.
         File.unlink(path) rescue nil
       end
--- a/lib/fluent/plugin/buf_file_single.rb
+++ b/lib/fluent/plugin/buf_file_single.rb
@@ -70,7 +70,7 @@ module Fluent
         using_plugin_root_dir = false
         unless @path
           if root_dir = owner.plugin_root_dir
-            @path = File.join(root_dir, 'fsb')
+            @path = File.join(root_dir, 'buffer')
             using_plugin_root_dir = true # plugin_root_dir path contains worker id
           else
             raise Fluent::ConfigError, "buffer path is not configured. specify 'path' in <buffer>"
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/buffer/file_single_chunk.rb
```diff
@@ -0,0 +1,315 @@
...
+require 'fluent/msgpack_factory'
+
+module Fluent
+  module Plugin
+    class Buffer
+      class FileSingleChunk < Chunk
+        class FileChunkError < StandardError; end
+
+        ## buffer path user specified : /path/to/directory
+        ## buffer chunk path          : /path/to/directory/sfb.b513b61c9791029c2513b61c9791029c2.buf
```
ganmacs: `fsb` ?

lib/fluent/plugin/buffer/file_single_chunk.rb
```diff
@@ -0,0 +1,315 @@
...
+            :unknown
+          end
+        end
+
+        def self.unique_id_and_key_from_path(path)
+          base = File.basename(path)
+          res = PATH_REGEXP =~ base
+          return nil unless res
+
+          key = base[4..res - 1]  # remove 'sfb.' and '.'
```
ganmacs: fsb ?

lib/fluent/plugin/buffer/file_single_chunk.rb
```diff
@@ -0,0 +1,315 @@
...
+          chunk_id = Fluent::UniqueId.hex(unique_id)
+          staged_path = ".b#{chunk_id}."
+          if path.index(staged_path)
+            path.sub(staged_path, ".q#{chunk_id}.")
+          else # for unexpected cases (ex: users rename files while opened by fluentd)
+            path + ".q#{chunk_id}.chunk"
+          end
+        end
+
+        def restore_metadata
```
ganmacs: Doesn't file_single_chunk support file_chunk's metadata restore?(this is just confirmation)

test/plugin/test_buffer_file_single_chunk.rb
```diff
@@ -0,0 +1,621 @@
+require_relative '../helper'
+require 'fluent/plugin/buffer/file_single_chunk'
+require 'fluent/plugin/compressable'
+require 'fluent/unique_id'
+
+require 'fileutils'
+require 'msgpack'
+require 'time'
+require 'timecop'
```
ganmacs: where do you use timecop?

test/plugin/test_buf_file_single.rb
```diff
@@ -0,0 +1,678 @@
...
+      assert_equal 4, queue[0].size
+
+      assert_equal @c2id, queue[1].unique_id
+      assert_equal :queued, queue[1].state
+      assert_equal 'testing', queue[1].metadata.tag
+      assert_nil queue[1].metadata.variables
+      assert_equal 3, queue[1].size
+    end
+
+    test '#resume returns staged/queued chunks but skip size calculation by calc_num_records' do
```
ganmacs: [nits] skips

test/plugin/test_buffer_file_single_chunk.rb
```diff
@@ -0,0 +1,621 @@
...
+      if @c
+        @c.purge rescue nil
+      end
+      if File.exist?(@chunk_path)
+        File.unlink(@chunk_path)
+      end
+    end
+
+    test 'creates new files for chunk and metadata with specified path & permission' do
+      assert { @c.unique_id.size == 16 }
```
ganmacs: assert_equal is better

test/plugin/test_buffer_file_single_chunk.rb
```diff
@@ -0,0 +1,621 @@
...
+      assert_equal 4, @c.size
+
+      @c.rollback
+
+      assert_equal first_bytesize, @c.bytesize
+      assert_equal 2, @c.size
+      assert_equal (d1.to_json + "\n" + d2.to_json + "\n"), File.read(@c.path)
+    end
+
+    test 'can #rollback to revert non-committed data from #concat' do
```
ganmacs: I think it's better to add a test case for 'can not #rollback a committed data'

lib/fluent/plugin/buf_file_single.rb
```diff
@@ -0,0 +1,209 @@
...
+        end
+
+        @key_in_path = nil
+        if owner.chunk_keys.empty?
+          log.debug "use event tag for buffer key"
+        else
+          if owner.chunk_key_tag
+            raise Fluent::ConfigError, "chunk keys must be tag or one field"
+          else
+            @key_in_path = owner.chunk_keys.first.to_sym
```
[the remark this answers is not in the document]
repeatedly: Just raise `ConfigError` now.

lib/fluent/plugin/buffer/file_single_chunk.rb
repeatedly: We can support it with metadata header in the future.
Just now, support only one field key or tag.

lib/fluent/plugin/buf_file_single.rb
```diff
@@ -0,0 +1,209 @@
...
+      config_param :path, :string, default: nil
+      desc 'Calculate the number of record in chunk during resume'
+      config_param :calc_num_records, :bool, default: true
+      desc 'The format of chunk. This is used to calculate the number of record'
+      config_param :chunk_format, :enum, list: [:msgpack, :text, :auto], default: :auto
+
+      config_set_default :chunk_limit_size, DEFAULT_CHUNK_LIMIT_SIZE
+      config_set_default :total_limit_size, DEFAULT_TOTAL_LIMIT_SIZE
+
+      config_param :file_permission, :string, default: nil # '0644'
```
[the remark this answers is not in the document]
repeatedly: Add `desc` for it

lib/fluent/plugin/buf_file_single.rb
```diff
@@ -35,6 +35,10 @@ class FileSingleBuffer < Fluent::Plugin::Buffer
 
       desc 'The path where buffer chunks are stored.'
       config_param :path, :string, default: nil
+      desc 'Calculate the number of record in chunk during resume'
+      config_param :calc_num_records, :bool, default: true
```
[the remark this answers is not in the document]
repeatedly: Some plugins don't use it chunk's size for data flush, e.g. kafka, S3 and more.
So users can choose it to speedup restart time.

lib/fluent/plugin/buf_file_single.rb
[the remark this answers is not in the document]
ganmacs: why don't you use `default` parameter?

test/plugin/test_buf_file_single.rb
```diff
@@ -55,16 +71,36 @@ def metadata(timekey: nil, tag: 'testing', variables: nil)
...
-    test 'path has "fsb" prefix and "buf" suffix' do
+  sub_test_case 'configuration' do
+    test 'path has "fsb" prefix and "buf" suffix by default' do
       @d = create_driver
       p = @d.instance.buffer
       assert_equal File.join(@bufdir, 'fsb.*.buf'), p.path
     end
+
+    data('text based chunk' => [FluentPluginFileSingleBufferTest::DummyOutputPlugin, :text],
+        'msgpack based chunk' => [FluentPluginFileSingleBufferTest::DummyOutputMPPlugin, :msgpack])
```
ganmacs: [nits] indent

lib/fluent/plugin/buf_file_single.rb
```diff
@@ -0,0 +1,211 @@
...
+      desc 'The format of chunk. This is used to calculate the number of record'
+      config_param :chunk_format, :enum, list: [:msgpack, :text, :auto], default: :auto
+
+      config_set_default :chunk_limit_size, DEFAULT_CHUNK_LIMIT_SIZE
+      config_set_default :total_limit_size, DEFAULT_TOTAL_LIMIT_SIZE
+
+      desc 'The permission of chunk file. If no specified, <system> setting or 0644 is used'
+      config_param :file_permission, :string, default: nil # '0644'
+      desc 'The permission of chunk directory. If no specified, <system> setting or 0644 is used'
+      config_param :dir_permission, :string, default: nil # '0755'
```
ganmacs: L48 says `0644` but this line says `0755`. which is one correct?

lib/fluent/plugin/buffer/file_single_chunk.rb
```diff
@@ -0,0 +1,314 @@
...
+          @chunk = File.open(@path, 'rb')
+          @chunk.set_encoding(Encoding::ASCII_8BIT)
+          @chunk.binmode
+          @chunk.seek(0, IO::SEEK_SET)
+
+          restore_metadata
+
+          @state = :queued
+          @bytesize = @chunk.size
+          @commit_position = @chunk.size
```
ganmacs: Are there any reasons not to use `@chunk.pos`?

lib/fluent/plugin/buffer/file_single_chunk.rb
```diff
@@ -0,0 +1,314 @@
...
+          end
+        end
+
+        def self.unique_id_and_key_from_path(path)
+          base = File.basename(path)
+          res = PATH_REGEXP =~ base
+          return nil unless res
+
+          key = base[4..res - 1]  # remove 'fsb.' and '.'
+          hex_id = base[res + 2..-5] # remove '.' and '.buf'
```
ganmacs: `$2` is the same result?

lib/fluent/plugin/buf_file_single.rb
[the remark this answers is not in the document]
repeatedly: If use default, hard to check user specified or not. We need to use `<system>` setting when no specified.

test/plugin/test_buf_file_single.rb
```diff
@@ -0,0 +1,678 @@
...
+    test '#start creates directory for buffer chunks' do
+      @d = create_driver
+      @p = @d.instance.buffer
+
+      FileUtils.rm_r(@bufdir) if File.exist?(@bufdir)
+      assert !File.exist?(@bufdir)
+
+      @p.start
+      assert File.exist?(@bufdir)
+      assert { File.stat(@bufdir).mode.to_s(8).end_with?('755') }
```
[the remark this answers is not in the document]
repeatedly: assert_equal is not fit because result is `40755` like string. This equal seems good for test.

lib/fluent/plugin/buf_file_single.rb
repeatedly: ah, forgot it. Will fix.

PR #2584, opened 2019-08-21
Use size instead of 

ganmacs:

follow up https://github.com/fluent/fluentd/pull/2579

File#pos returns 0 since File#seek(0, IO::SEEK_SET) is called before

**Docs Changes**:

no need

**Release Note**: 

Fix bug `@commit_position` is always 0


repeatedly:
Good catch!

PR #2601, opened 2019-09-03
Load certificates from windows certstore

cosmo0920:
None.

Currently, Fluentd does not refer Windows Certstore as OpenSSL certificates store.
This PR makes to be able to refer it and import certificates which are stored in Windows certstore.

**Docs Changes**:
~~Add `Load certificates from windows certstore` article in out_forward document. I will send a patch later.~~
https://github.com/fluent/fluentd-docs-gitbook/pull/102

**Release Note**: 
Same as a title.


repeatedly:
Thanks!

PR #2604, opened 2019-09-06
Set encoding

ganmacs:

none

for the env which `Encoding.default_external` returns not `ASCII-8BIT`

```rb
s = StringIO.new
p s.external_encoding # => #<Encoding:UTF-8>
s.write('text'.force_encoding(Encoding::ASCII_8BIT))
p s.string.encoding # => #<Encoding:UTF-8>

s = StringIO.new('text'.force_encoding(Encoding::ASCII_8BIT))
p s.external_encoding # => #<Encoding:ASCII-8BIT>
p s.string.encoding # => #<Encoding:ASCII-8BIT>
```

**Docs Changes**:

no need

**Release Note**: 

no need

cc. @repeatedly


repeatedly:
looks good.

PR #2606, opened 2019-09-06
Increase minimum msgpack version

jonahb:
9aca889cd started using `MessagePack::Packer#full_pack`, which was introduced in msgpack 1.2.0.


repeatedly:
Could you follow DCO?

https://github.com/fluent/fluentd/pull/2606/checks?check_run_id=214610500

jonahb:
@repeatedly I have signed off

repeatedly:
Thanks! Good catch 👍

Issue #2613, opened 2019-09-10
Defining a `send` method on a plugin causes failure on shutdown


ccheek21:
**Describe the bug**
When a plugin defines a `send` method, it causes a failure during plugin lifecycle operations because it overrides [`Object.send`](https://ruby-doc.org/core-2.4.5/Object.html#method-i-send) which is used in [`root_agent.rb`](https://github.com/fluent/fluentd/blob/a2a164e782d919da5044cc2fd8235e7d9dc12780/lib/fluent/root_agent.rb#L280). 

This caused an issue in the New Relic output plugin ([link](https://github.com/newrelic/newrelic-fluentd-output/pull/9))

This should be fairly easily fixable by using `Object.__send__` instead of `Object.send` (as suggested by the Ruby docs).

**To Reproduce**
1) Create a plugin with a `send` method like the New Relic output plugin did that takes a non-symbol argument
2) Shut down fluentd
3) Examine logs for stacktrace

**Expected behavior**
Shutdown still works correctly

**Your Environment**

- Fluentd or td-agent version: Tested on Fluentd 1.3.3 and 1.7.1
- Operating system: CentOS 7
- Kernel version: 3.10.0-862.11.6.el7.x86_64

**Your Configuration**

```
 <ROOT>
  <source>
    @type syslog
    bind "0.0.0.0"
    port 5140
    tag "e8475b22-48cf-4ae9-a92c-854960028d31.syslog"
    protocol_type tcp
  </source>
  <match e8475b22-48cf-4ae9-a92c-854960028d31.**>
    @type record_reformer
    tag "${tag_suffix[1]}"
    <record>
      bindplane_source_name Test Newrelic Syslog
      bindplane_source_id e8475b22-48cf-4ae9-a92c-854960028d31
      bindplane_source_type syslog
    </record>
  </match>
  <match **>
    @type newrelic
    api_key "<redacted>"
  </match>
</ROOT>

```

**Your Error Log**

```
2019-09-10 10:32:26 -0400 [info]: starting fluentd-1.7.1 pid=5463 ruby="2.4.5"
[...]

PR #2614, opened 2019-09-11
Use Object.__send__ to avoid conflicting method name

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2613

Use `Object.__send__` to avoid conflicting method name

**Docs Changes**:

no need

**Release Note**: 

same as the title


repeatedly:
All test failed. Could you check it?

Issue #2607, opened 2019-09-07
error_class=ZeroDivisionError error="divided by 0"


luthienberen:
**Describe the issue**

I've deployed fluentd daemonset and now fluentd logs returns following errors

**Your Error Log**

```
kubectl logs fluentd-xzgzv -f

2019-09-07 13:23:35 +0000 [info]: parsing config file is succeeded path="/fluentd/etc/fluent.conf"
2019-09-07 13:23:36 +0000 [warn]: [out_es] 'flush_interval' is ignored because default 'flush_mode' is not 'interval': 'lazy'
2019-09-07 13:23:36 +0000 [warn]: [out_es] Detected ES 7.x or above: `_doc` will be used as the document `_type`.
2019-09-07 13:23:37 +0000 [info]: using configuration file: <ROOT>
  <source>
    @id bridge-containers.log
    @type tail
    path "/var/log/name0-changed/*/*.log,/var/log/name0-changed/name1-changed/*/*.log"
    pos_file "/var/log/name0-changed/name2-changed.log.pos"
    tag "name-changed0"
    read_from_head true
    <parse>
      @type "json"
    </parse>
  </source>
  <source>
    @type tail
    @id in_tail_container_logs
    path "/var/log/containers/*.log"
    pos_file "/var/log/fluentd-containers.log.pos"
    tag "kubernetes"
    read_from_head true
    <parse>
      @type "json"
      time_format "%Y-%m-%dT%H:%M:%S.%NZ"
      time_type string
    </parse>
  </source>
  <filter kubernetes.**>
    @type kubernetes_metadata
    @id filter_kube_metadata
  </filter>
  <source>
    @type tail
    @id in_tail_kube_apiserver_audit
    multiline_flush_interval 5s
    path "/var/log/kubernetes/kube-apiserver-audit.log"
    pos_file "/var/log/kube-apiserver-audit.log.pos"
    tag "kube-apiserver-audit"
    <parse>
      @type "multiline"
      format_firstline "/^\\S+\\s+AUDIT:/"
[...]

PR #2615, opened 2019-09-11
validate if timkey is less than equal 0

ganmacs:

Fixes https://github.com/fluent/fluentd/issues/2607

I made a change to raise an error when a user specifies `timekey`  less than equal 0.
when `timekey` is 0, fluentd will raise `ZeroDivisionError` at [here](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L1379).

**Docs Changes**:

no need

**Release Note**: 

same as the title


repeatedly:
Looks good.

The potential problem is fluentd uses non-strict conversion for parameters like https://github.com/fluent/fluentd/blob/07d9eecf0ce1413ab7873da84462b6f8fe9fae04/lib/fluent/config/types.rb#L101
We should improve this point to use `Integer()` or others.
This change has compatibility probolem so it will be implemented in v2 or via configuration.

PR #2620, opened 2019-09-17
Allow uncompressing buffer files during restore

### LINKED ISSUE #2618: Compression for buf_file plugin not used on restore (master)
**Describe the bug**

While buffering a chunk on disk it can be compressed using gzip, but when restoring it, is isn't uncompressed, leading to a broken chunk that's possibly send further on.

**To Reproduce**

```
  <buffer>
    @type file
    path /var/log/shopify-fluentd/fluentd/example.buffer
    chunk_limit_size 10M
    total_limit_size 1280M
    flush_interval 5s
    retry_max_interval 30
    retry_forever true
    flush_thread_count 4
    overflow_action block
    compress gzip
  </buffer>
```

Cause a chunk to be buffered, restart fluentd. The latter should cause the chunk to be unbuffered and uncompressed, but it's loaded in compressed form.

**Expected behavior**

Chunks should be buffered and unbuffered correctly regardless of compression setting.

berlincount:
Fixes #2618

Chunks buffered via buf_file where not uncompressed on restore.

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -146,7 +146,7 @@ module Fluent
           end
 
           begin
-            chunk = Fluent::Plugin::Buffer::FileChunk.new(m, path, mode) # file chunk resumes contents of metadata
+            chunk = Fluent::Plugin::Buffer::FileChunk.new(m, path, mode, compress: @compress) # file chunk resumes contents of metadata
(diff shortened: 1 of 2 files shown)


ganmacs:
Sorry. We don't have a plan to release the next version of 1.6.x. so could you please close PR https://github.com/fluent/fluentd/pull/2621 ?
cc @repeatedly

repeatedly:
Thanks for the patch!

And yes. We will release new v1.7 soon but no new v1.6 release.

PR #2640, opened 2019-10-08
Semantic Status Code - modify on_request response from 200 to 204 in plugin/in_http

achinchen:
**The reason about this PR**: 

I used fluentd service and figured that 200 response with empty string is a little bit not precise.
According to [rfc7231](https://tools.ietf.org/html/rfc7231#section-6.3.5), 204 is more proper status code.
> a 204 status code is commonly used with document editing interfaces corresponding to a "save" action

- For more semantic status code
- Add test for this change in test_in_http
- Fix typo in test_in_http

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -152,7 +152,7 @@ module Fluent::Plugin
           if @respond_with_empty_img
             return ["200 OK", {'Content-Type'=>'image/gif; charset=utf-8'}, EMPTY_GIF_IMAGE]
           else
-            return ["200 OK", {'Content-Type'=>'text/plain'}, ""]
+            return ["204 No Content", {}]
           end
         end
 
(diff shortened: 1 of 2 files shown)


repeatedly:
Yeah, we can't judge this change is safe for existing users.
So following step is better:

- Add bool option to change status code, like `use_204_response`.
- Change default behaviour and remove option at fluentd v2.

lib/fluent/plugin/in_http.rb
```diff
@@ -140,7 +140,7 @@ def close
       super
     end
 
-    def on_request(path_info, params)
+    def on_request(path_info, params, use_204_response=false)
```
ganmacs: How do users set `use_204_response`?

achinchen: Oops, I think that is my misunderstanding.
I fixed it and add config_param in in_http.

This is my first time to PR with Ruby.
let me know if anything not suitable.
Thanks for review.

repeatedly:
Thanks!

PR #2643, opened 2019-10-09
formatter_csv: support nested fields. fix #2630

### LINKED ISSUE #2630: Support nested flieds in CSV format
Consider the following record:
```
 {
    "hostname":"buster",
    "priority":"6",
    "selinux_context":"unconfined\n",
    "pid":"4866",
    "image_name":"nginx",
    "container_name":"fluentd-docker_nginx_1",
    "container_id":"109e52fb8de8",
    "message":{
       "remote":"172.20.0.1",
       "host":"-",
       "user":"-",
       "method":"GET",
       "path":"/",
       "code":"200",
       "size":"612",
       "referer":"-",
       "agent":"curl/7.64.0",
       "http_x_forwarded_for":"-"
    },
    "source_realtime_timestamp":"1569258331312517"
 }
```
I would be nice to access the flieds in `message` fields in CSV format
```
  <filter **>
    @type parser
    key_name message
    reserve_data true
    hash_value_field message
    <parse>
      @type regexp
      expression /^(?<method>[^ ]*) (?<path>[^ ]*) (?<code>[0-9]*) (?<ms>[^ ]*) ms - (?<bytes_out>[0-9-]*)$/
    </parse>
  </filter>

  <match **>
    #@type file
    @type stdout

    #path /tmp/access.log
    <format>
      @type csv
      force_quotes false
      delimiter " "
      fields container_tag,message['method']
    </format>
  </match>
```
I have tried `message['method']` `message.method` and `message[:method]` none of them works

repeatedly:

Fixes #2630 

Support nested fields via `record_accessor`.

**Docs Changes**:
Add note for nested fields to `fields` parameter.

**Release Note**: 
same as title

### DIFF
--- a/lib/fluent/plugin/formatter_csv.rb
+++ b/lib/fluent/plugin/formatter_csv.rb
@@ -14,6 +14,7 @@
 #    limitations under the License.
 #
 
+require 'fluent/plugin_helper'
 require 'fluent/plugin/formatter'
 require 'csv'
 
@@ -21,6 +22,9 @@ module Fluent
   module Plugin
     class CsvFormatter < Formatter
       Plugin.register_formatter('csv', self)
+
+      include PluginHelper::Mixin
+      helpers :record_accessor
 
       config_param :delimiter, default: ',' do |val|
         ['\t', 'TAB'].include?(val) ? "\t" : val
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/formatter_csv.rb
```diff
@@ -37,6 +41,14 @@ def configure(conf)
         @fields = fields.select{|f| !f.empty? }
         raise ConfigError, "empty value is specified in fields parameter" if @fields.empty?
 
+        if @fields.any? { |f| record_accessor_nested?(f) }
+          @accessors = @fields.map { |f| record_accessor_create(f) }
```
ganmacs: How about this? it can avoid the needless call of `Accessor#call`. Not nested fields don't need to be called this method, right?

```rb
nested = @fileds.select { |f| record_accessor_nested?(f) }
unless nested.empty?
  @accessors = nested.map { |f| record_accessor_create(f) }

   ...
end
```

repeatedly: Your appoarch is bit complicated.
If fields are `a,$.b.c,d`, formatter_csv needs to remember the position when separate normal access and record_accessor.

PR #2649, opened 2019-10-11
child_process helper: fix stderr blocking for discard case. fix #2609

### LINKED ISSUE #2609: Writing to stderr from out_exec_filter plugin, blocks
**Describe the bug**

I'm using `exec_filter` to process some data using Python. My initial command looks like

```
<match logs.**>
    # https://docs.fluentd.org/output/exec_filter
    @type exec_filter

    command /path/to/python/script

    <format>
      @type json
    </format>
    <parse>
      @type json
    </parse>
</match>
```

For logging, my script is using the standard python3 logging library for debugging.
I noticed the script backing up and spent a while trying to debug it, but I now suspect the lower level `child_process_execute` in `fluentd/lib/fluent/plugin_helper/child_process.rb` to be suspect

Updating the above command to `command /path/to/python/script 2>> /var/log/td-agent/script.log` seems to suddenly unblock processing

**Expected behavior**

Within, out_exec_filter, a child process is called via 

```
execute_child_process = ->(index){
    child_process_execute("out_exec_filter_child#{index}".to_sym, @command, on_exit_callback: exit_callback) do |readio, writeio|
        child_process_callback.call(index, readio, writeio)
    end
}
```

which means the parameter defaults to `stderr: :discard` which then arrives at this block

```
if !mode.include?(:stderr) && !mode.include?(:read_with_stderr) && stderr != :discard # connect
    writeio, readio, wait_thread = *Open3.popen2(*spawn_args, spawn_opts)
elsif mode.include?(:read_with_stderr)
    writeio, readio, wait_thread = *Open3.popen2e(*spawn_args, spawn_opts)
else
    writeio, readio, stderrio, wait_thread = *Open3.popen3(*spawn_args, spawn_opts)
end
```

While I have not debugged any further than this, I would expect if fluentd's plugin is not handling stderr, then it would at least default to sending it to /dev/null or some other place. There was 

**Your Environment**

```
fluentd --version 
fluentd 1.7.0
python -m platform
Linux-3.10.0-957.21.3.el7.x86_64-x86_64-with-centos-7.6.1810-Core
```

repeatedly:

Fixes #2609 

reopen with IO::NULL is not correct way to discard message.
It replaces ruby side IO object, not process. Thanks nurse-san!
Need to specify IO::NULL when launch the process.

**Docs Changes**:
No need

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin_helper/child_process.rb
+++ b/lib/fluent/plugin_helper/child_process.rb
@@ -257,7 +257,8 @@ module Fluent
         readio = writeio = stderrio = wait_thread = nil
         readio_in_use = writeio_in_use = stderrio_in_use = false
 
-        if !mode.include?(:stderr) && !mode.include?(:read_with_stderr) && stderr != :discard # connect
+        if !mode.include?(:stderr) && !mode.include?(:read_with_stderr)
+          spawn_opts[:err] = IO::NULL if stderr == :discard
           writeio, readio, wait_thread = *Open3.popen2(*spawn_args, spawn_opts)
         elsif mode.include?(:read_with_stderr)
           writeio, readio, wait_thread = *Open3.popen2e(*spawn_args, spawn_opts)
@@ -281,7 +282,7 @@ module Fluent
           stderrio.set_encoding(external_encoding, internal_encoding, encoding_options)
           stderrio_in_use = true
         else
-          stderrio.reopen(IO::NULL) if stderrio && stderrio == :discard
+          stderrio.reopen(IO::NULL) if stderrio && stderr == :discard
         end
 
         pid = wait_thread.pid # wait_thread => Process::Waiter
--- a/test/plugin_helper/test_child_process.rb
+++ b/test/plugin_helper/test_child_process.rb
@@ -124,6 +124,39 @@ class ChildProcessTest < Test::Unit::TestCase
     assert_equal expected, ary
   end
 
+  test 'can execute external command at just once, which can handle both of read and write. Ignore stderr message/no block' do
+    m = Mutex.new
+    ary = []
+    Timeout.timeout(TEST_DEADLOCK_TIMEOUT) do
+      ran = false
+      # lots of stderr message should not be blocked and message should not be printed in test.
+      cmd = "ruby -e 'while !STDIN.eof? && line = STDIN.readline; STDERR.puts line.chomp * 1000; STDOUT.puts line.chomp; STDOUT.flush rescue nil; end'"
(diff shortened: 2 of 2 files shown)


test/plugin_helper/test_child_process.rb
```diff
@@ -124,6 +124,39 @@ def configure(conf)
     assert_equal expected, ary
   end
 
+  test 'can execute external command at just once, which can handle both of read and write. Ignore stderr message/no block' do
+    m = Mutex.new
+    ary = []
+    Timeout.timeout(TEST_DEADLOCK_TIMEOUT) do
+      ran = false
+      # lots of stderr message should not be blocked and message should not be printed in test.
+      cmd = "ruby -e 'while !STDIN.eof? && line = STDIN.readline; STDERR.puts line.chomp * 1000; STDOUT.puts line.chomp; STDOUT.flush rescue nil; end'"
```
ganmacs: I'm not sure but it needs to call`STDERR.flush`?

test/plugin_helper/test_child_process.rb
```diff
@@ -124,6 +124,39 @@ def configure(conf)
...
+        writeio.close
+
+        while line = readio.readline
+          ary << line
+        end
+        m.unlock
+      end
+      sleep TEST_WAIT_INTERVAL_FOR_BLOCK_RUNNING until m.locked? || ran
+      m.lock
+      m.unlock
```
ganmacs: Mutex#unlock should be called in `ensure` context for safe.

lib/fluent/plugin_helper/child_process.rb
```diff
@@ -281,7 +282,7 @@ def child_process_execute_once(
           stderrio.set_encoding(external_encoding, internal_encoding, encoding_options)
           stderrio_in_use = true
         else
-          stderrio.reopen(IO::NULL) if stderrio && stderrio == :discard
```
ganmacs: Just checking, This was a bug, right? `stderrio` cannot be a `:discard`.

test/plugin_helper/test_child_process.rb
repeatedly: STDERR writes lots of data so no flush call is not critical.

test/plugin_helper/test_child_process.rb
repeatedly: good catch.

lib/fluent/plugin_helper/child_process.rb
repeatedly: Yes. This is typo.

Issue #2624, opened 2019-09-20
Light reloading mechanism


ganmacs:
*Is your feature request related to a problem? Please describe.**

no

**Describe the solution you'd like**

Current implement,  fluentd just [kills worker](https://github.com/fluent/fluentd/blob/02c1e2ba2e46cea5b7b77c5366dfad3a709e9ae2/lib/fluent/supervisor.rb#L178) when receiving SIGHUP(reload signal).
This needs to re-spawn the worker process. so it takes not a short time (No downtime alright, but fluentd can't deal with messages at that time).

So it would be good if there is a more lightweight, safer reload mechanism.
I think switching `Fluent:Engine` to the new one when receiving the signal solves the problem.

**Describe alternatives you've considered**

I thought that blue-green deployment like reloading would work(starting new config one -> run two processes (old one and new one) at time time -> kill old one).
But it turned out that this kind of reloading mechanism is not well-matched for fluentd.
e.g. in\_tail plugin and file\_buffer plugin don't work properly when running two fluentd processes.

~~Use [Serverengine](https://github.com/treasure-data/serverengine)'s  [live restart]( https://github.com/treasure-data/serverengine#live-restart) to reload config setting.~~

**Additional context**

PR #2653, opened 2019-10-15
No need to use instance variable

ganmacs:
Ref https://github.com/fluent/fluentd/issues/2624

No need to use ivar since since https://github.com/fluent/fluentd/pull/496 .

**Docs Changes**:

no need

**Release Note**: 

no need


repeatedly:
Good

PR #2660, opened 2019-10-21
record_accessor: Make code simple and bit faster

repeatedly:

None

Make record_accessor code simple by moving `call_index` to default implementation.
New implementation is bit faster.

```
                base   317.029k i/100ms
                test   327.406k i/100ms
                base     10.600M (± 6.1%) i/s -     52.944M in   5.019043s
                test     10.797M (± 4.2%) i/s -     54.022M in   5.013882s
```

**Docs Changes**:
No need

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin_helper/record_accessor.rb
+++ b/lib/fluent/plugin_helper/record_accessor.rb
@@ -46,25 +46,19 @@ module Fluent
             @dig_keys = @keys[0..-2]
             if @dig_keys.empty?
               @keys = @keys.first
-              mcall = method(:call_index)
-              mdelete = method(:delete_top)
             else
               mcall = method(:call_dig)
               mdelete = method(:delete_nest)
+              singleton_class.module_eval do
+                define_method(:call, mcall)
+                define_method(:delete, mdelete)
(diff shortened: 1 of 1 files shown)


repeatedly:
Ah, I forgot to paste the code.

```
require 'benchmark/ips'

class Base
  def initialize
    @keys = 'k'
    mcall = method(:call_index)
    singleton_class.module_eval do
      define_method(:call, mcall)
    end
  end

  def call(r)
  end

  def call_dig(r)
    r.dig(*@keys)
  end

  def call_index(r)
    r[@keys]
  end
end

class Test
  def initialize
    @keys = 'k'
  end

  def call(r)
    r[@keys]
  end

  def call_dig(r)
    r.dig(*@keys)
  end
end

record = {'k' => 'foo'}
base = Base.new
test = Test.new

Benchmark.ips do |x|
  x.report "base" do
    base.call(record)
  end

  x.report "test" do
    test.call(record)
  end
end

```

ganmacs:
I see. made it faster when using a default implementation 👍

Issue #2659, opened 2019-10-18
Quick Start is broken


nokute78:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**

Quick Start is broken at git master.
out_forward with @id may not work.

It works on v1.7.3.

**To Reproduce**

```
$ mkdir tmp
$ bin/fluentd -s tmp
$ bin/fluentd -c tmp/fluent.conf
```

**Expected behavior**

Fluentd works.

**Your Environment**

- Fluentd or td-agent version: `fluentd --version` or `td-agent --version`
git master b1bd1dab520f4a4592e96e4ba64908a4e6aa6c65
- Operating system: `cat /etc/os-release`
Ubuntu 18.04
```
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```
- Kernel version: `uname -r`
5.0.0-31-generic

If you hit the problem with older fluentd version, try latest version first.

**Your Configuration**
https://github.com/fluent/fluentd/blob/b1bd1dab520f4a4592e96e4ba64908a4e6aa6c65/fluent.conf

**Your Error Log**

```
$ bin/fluentd -c tmp/fluent.conf 
2019-10-18 22:56:11 +0900 [info]: parsing config file is succeeded path="tmp/fluent.conf"
2019-10-18 22:56:11 +0900 [info]: adding forwarding server '192.168.0.12:24224' host="192.168.0.12" port=24224 weight=60 plugin_id="object:2b0a500b55c4"
2019-10-18 22:56:11 +0900 [error]: config error file="tmp/fluent.conf" error_class=Fluent::ConfigError error="Duplicated plugin id `forward_output`. Check whole configuration and fix it."
$ 
```

**Additional context**
It works on v1.7.3.
#2541 seems to break Quick Start.

[...]

PR #2661, opened 2019-10-21
copy setting sd plugin needs only

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2659

not to copy `@id` or other necessary things to sd plugin.

**Docs Changes**:

no need

**Release Note**: 

Fix sd plugin's bugs with `@id`

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -233,7 +233,9 @@ module Fluent::Plugin
       end
 
       unless conf.elements(name: 'service').empty?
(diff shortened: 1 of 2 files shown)


ganmacs:
@repeatedly 
Sorry :bow:. I changed this a bit. creating a new `element` is done in out_forward because doing it in PluginHelper::ServiceDiscovery::Manager doesn't work in some case. could you take a look 🙇

repeatedly:
> doing it in PluginHelper::ServiceDiscovery::Manager doesn't work in some case

Ah, I see. Good catch :)

PR #2664, opened 2019-10-24
Suppress warning about keyword argument introduced from ruby 2.7

ganmacs:
 no

For https://bugs.ruby-lang.org/issues/14183

there are a vast number of warnings in https://travis-ci.org/fluent/fluentd/jobs/602120485 which is tested in ruby 2.7 preview2.

See also https://www.ruby-lang.org/en/news/2019/10/22/ruby-2-7-0-preview2-released/
**Docs Changes**:

no need

**Release Note**: 

no need

### DIFF
--- a/lib/fluent/test/driver/base_owned.rb
+++ b/lib/fluent/test/driver/base_owned.rb
@@ -53,13 +53,26 @@ module Fluent
           @section_name = ''
         end
 
-        def configure(conf, syntax: :v1)
+        def configure(conf)
           if conf.is_a?(Fluent::Config::Element)
             @config = conf
           elsif conf.is_a?(Hash)
             @config = Fluent::Config::Element.new(@section_name, "", Hash[conf.map{|k,v| [k.to_s, v]}], [])
           else
-            @config = Fluent::Config.parse(conf, @section_name, "", syntax: syntax)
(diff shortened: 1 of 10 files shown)


ganmacs:
Warnings around keyword arguments look disappearing https://travis-ci.org/fluent/fluentd/jobs/602174792?utm_medium=notification&utm_source=github_status

lib/fluent/test/driver/base_owned.rb
```diff
@@ -53,13 +53,26 @@ def initialize(klass, opts: {}, &block)
           @section_name = ''
         end
 
-        def configure(conf, syntax: :v1)
+        def configure(conf)
```
repeatedly: this change doesn't impact existing test cases?

ganmacs: looks like so. this was introduced in https://github.com/fluent/fluentd/pull/1047 . but I can't get any more information about why this method needs this parameter. so leave it as it is. and I will delete it when fluentd becomes v2. or you think it should be deleted in this PR?

ganmacs: There is no problem for me because this is only for test, not for production code.
Failing test is not critical and easy to fix it.

PR #2667, opened 2019-10-25
Make engine lighter

ganmacs:
Ref https://github.com/fluent/fluentd/issues/2624

Make engine lighter for safe reloading.

**Docs Changes**:

no need

**Release Note**: 

no need

### DIFF
--- a/lib/fluent/fluent_log_event_router.rb
+++ b/lib/fluent/fluent_log_event_router.rb
@@ -17,9 +17,20 @@
 require 'fluent/log'
 
 module Fluent
+  # DO NOT write any logic here
+  class NullFluentLogEventRouter
+    def start; end
+
+    def stop; end
+
+    def graceful_stop; end
+
+    def emit_event; end
+  end
+
   # This class is for handling fluentd's inner log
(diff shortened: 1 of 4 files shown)


lib/fluent/fluent_log_event_router.rb
```diff
@@ -0,0 +1,137 @@
...
+module Fluent
+  # DO NOT write any logic here
+  class NullFluentLogEventRouter
+    def start; end
+
+    def stop; end
+
+    def graceful_stop; end
+
+    def emit_event; end
```
repeatedly: No argument. is it intended?

ganmacs: Good catch!  I missed adding an argument.
https://github.com/fluent/fluentd/pull/2667/commits/0970e78b9f17b5fcccc2524f529ceba573944b90

repeatedly: should add test for `NullFluentLogEventRouter` methods?

ganmacs: I added https://github.com/fluent/fluentd/pull/2667/commits/65b7429c03cc7fc9eead11a506338d9859c44f5e

PR #2668, opened 2019-10-25
Revive MsgpackPack for compat

ganmacs:

no 

Follow up https://github.com/fluent/fluentd/pull/2657
Revive `Engine.msgpack_*` for compat.

**Docs Changes**:

no need

**Release Note**: 

no need


repeatedly:
Is this for 3rd party plugins, right?

ganmacs:
> Is this for 3rd party plugins, right?

Yes. I forgot that… so revived it.

Issue #2645, opened 2019-10-09
Faster implementation of monitor than standard library


ganmacs:
I'll try https://github.com/nurse/ext_monitor

cc. @nurse

ganmacs:
oh… I made a mistake… transfered to fluent/fluentd.

repeatedly:
Here is my benchmark result of https://github.com/fluent/fluentd/pull/2561
`Monitor#synchronize` is still slower...

```
- ruby 2.6.3

Warming up --------------------------------------
             monitor   184.967k i/100ms
               mutex   289.882k i/100ms
Calculating -------------------------------------
             monitor      3.287M (± 2.2%) i/s -     16.462M in   5.011436s
               mutex      8.005M (± 1.6%) i/s -     40.294M in   5.034949s

- ruby 2.6.5
Warming up --------------------------------------
             monitor    56.806k i/100ms
               mutex   280.742k i/100ms
Calculating -------------------------------------
             monitor    678.492k (± 4.7%) i/s -      3.408M in   5.035571s
               mutex      8.054M (± 2.4%) i/s -     40.427M in   5.022535s

- ruby 2.6.5 with ext_monitor
Warming up --------------------------------------
             monitor    62.786k i/100ms
               mutex   292.756k i/100ms
Calculating -------------------------------------
             monitor    754.165k (± 2.2%) i/s -      3.830M in   5.080977s
               mutex      8.031M (± 1.4%) i/s -     40.400M in   5.031691s
```

ganmacs:
For my understanding, this gem is a faster version of the exising `Monitor` class, not `Mutex` class.
So we should compare Monitor library ruby provides and this gem.

```rb
require 'benchmark/ips'
# require 'ext_monitor'

mo = Monitor.new
mu = Mutex.new

p "Ruby #{RUBY_VERSION}"

Benchmark.ips do |x|
  x.report("monitor") do
    mo.synchronize {}
  end

  x.report("mutex") do
    mu.synchronize {}
  end
end
```

with ext_monitor
```
"Ruby 2.5.5"
Warming up --------------------------------------
             monitor   266.565k i/100ms
               mutex   351.016k i/100ms
Calculating -------------------------------------
             monitor      4.937M (± 4.7%) i/s -     24.791M in   5.036034s
               mutex      7.851M (± 4.7%) i/s -     39.314M in   5.018843s
```

without ext_monitor (commented out)
```
"Ruby 2.5.5"
Warming up --------------------------------------
[...]

Issue #2380, opened 2019-04-11
Drop older ruby 2.1/2.2/2.3 support


repeatedly:
Fluentd v1 now works on ruby 2.1 or later but it has several drawbacks.

- ruby 2.1/2.2 are dead version. ruby team doesn't maintain it.
- Need dig_rb gem for `dig` method: See https://github.com/fluent/fluentd/pull/2372

Removing 2.1/2.2 is good for the further development. td-agent and docker images use ruby 2.3 or later so it doesn't break popular environment.
However removing support without the transition period may break user environment. So support 2.1/2.2 until the end of 2019 is better.

UPDATE: Drop 2.3 together. See https://www.fluentd.org/blog/drop-schedule-announcement-in-2019

henrich:
Thanks for opening this issue.
Maybe it's also better to discuss future support roadmap, I mean "when we drop support with Ruby2.3, 2.4, 2.5... for fluentd in the future" policy. Not fixed one but a rough consensus is good, IMO.

repeatedly:
> Maybe it's also better to discuss future support roadmap, I mean "when we drop support with Ruby2.3, 2.4, 2.5... for fluentd in the future"

One year support after EOL is good for me but several popular distributions use EOL version with own patches. So fixed drop timeline is not good for community.

henrich:
Of course, we should care about distributions, and we can predict distros EOL thus suggest "plan" to the community, right?

distro | ruby version | date
------- | ----------------  | ------
Ubuntu 14.04 LTS | ruby 1.9.1 | EOL in this April
RHEL6 (CentOS6) | ruby 1.8.7  | November 2020
Ubuntu 16.04 LTS | ruby 2.3.1  | April 2021
Debian9 LTS | ruby 2.3.3  | June 2022
Ubuntu 18.04 LTS | ruby 2.5.1 | April 2023 
Debian10 LTS | ruby 2.5.5 | May? 2024
RHEL7 (CentOS7) | ruby 2.0.0 | June 2024
RHEL8 (CentOS8) | ? | ? 2029

So, I suggest you
- drop ruby2.1 and 2.2 support in this summer (2019)
- drop ruby2.3 and 2.4 support in July 2022
- drop ruby2.5 support - no plan ;)

[...]

PR #2670, opened 2019-10-28
Add ext_monitor gem to improve Monitor class's operations

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2645

From ruby 2.6.5, Monitor class became slower due to fixing a deadlock bug (https://bugs.ruby-lang.org/issues/15992). 
https://github.com/nurse/ext_monitor/ mitigates performance for ruby 2.6.5 or later and improves it for prior to 2.6.4  by 2x(https://github.com/fluent/fluentd/issues/2645#issuecomment-546817682).

**Docs Changes**:

no need

**Release Note**: 

same as the title


ganmacs:
I'll merge this after https://github.com/fluent/fluentd/issues/2380

repeatedly:
By introducing ext_monitor in ruby 2.7, how about changing to `if ext_monitor is installed, fluentd uses it`? It reduces the dependencies and fluentd works even if `ext_monitor` has a problem.

ganmacs:
Sorry, I had a mistake about a condition of installing.
Ruby 2.7 is bundled ext_monitor like mitigation code by default so 2.7 is fast enough without ext_monitor.

>  It reduces the dependencies and fluentd works even if ext_monitor has a problem.

Does it mean like below? and removing `gem.add_runtime_dependency("ext_monitor", [">= 0.1.1", "< 0.2"])` from gemspec?


```rb
if Gem::Version.create(RUBY_VERSION) >= Gem::Version.create('2.7.0')
  require 'monitor'
else
  begin
    require 'ext_monitor'
  rescue LoadError => _
    require 'monitor'
  end
end
```

repeatedly:
> Does it mean like below? 

Yes.

ganmacs:
fixed.
https://github.com/fluent/fluentd/pull/2670/commits/4ae541676e6dce64a2caa90bcc62700f62184638

This log always outputs to STDOUT without ext_monitor. it might be noisy. what do you think of that? @repeatedly

repeatedly:
Maybe, such message is not needed.
td-agent and docker images will install `ext_monitor`.
So most users don't have Monitor issue.

Add to instruction to gem install article is enough.

ganmacs:
> Maybe, such message is not needed.

:+1: 

fixed https://github.com/fluent/fluentd/pull/2670/commits/0a772f2a72b3f8141a762f7d70bea9d7742cd50e

PR #2671, opened 2019-10-28
Properly show class names

dlackty:

None

We recently encountered a weird error showing that:

`value must be a string or a number: (value.class)"`

After digging into the source code, I finally realized that I passed `nil` into the parser but the message was quite confusing.

From my perspective, I thought `(value.class)` should actually be `(#{value.class})`, which showed the class name of the value. In my case, it will be `(NilClass)`, which should be clear for users to understand.

**Docs Changes**:

None

**Release Note**: 

None


repeatedly:
Ah, good catch! Thanks.

PR #2673, opened 2019-10-29
Accept tzinfo not only 2.x but also 1.x

### LINKED ISSUE #2634: Need to downgrade tzinfo for fluent-pluigin that depends on ActiveSupport?
**This is just a reminder for fluentd team**.

ref: https://github.com/fluent/fluentd/pull/2479

>  Looks like this PR made things a bit more complex for fluentd plugins that depend on ActiveSupport (e.g. fluent-plugin-splunk-hec). Because Rails and ActiveSupport have hard dependency on tzinfo v1.1 (see rails/rails#34799) it conflicts with v2.0 requirements that this PR added.

ashie:
Fixes #2634

The pull request #2479 enforces using tzinfo ~> 2.0 but it conflicts
with plugins which depend on activesupport such as fluent-plugin-sql
since it requires tzinfo ~> 1.1.
Until activesupport upgrades it, it would be better to accept also
tzinfo 1.x to keep availability of such plugins. Fluentd can accept
both versions of tzinfo since #2479 absorb the difference of their
API.

**Docs Changes**:
None

**Release Note**: 
None

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("serverengine", [">= 2.0.4", "< 3.0.0"])
   gem.add_runtime_dependency("http_parser.rb", [">= 0.5.1", "< 0.7.0"])
   gem.add_runtime_dependency("sigdump", ["~> 0.2.2"])
-  gem.add_runtime_dependency("tzinfo", ["~> 2.0"])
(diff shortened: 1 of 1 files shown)


repeatedly:
Does fluentd with tzinfo v1.x passes all tests?
Current CI uses fluentd and tzinfo v2.

ashie:
Yes, all tests passed with tzinfo 1.x.

repeatedly:
Ok, Thanks!

Issue #2444, opened 2019-06-05
in_udp doesn't raise error when non-string is specified in port parameter but it listens port randomly


cosmo0920:
I've heard this issue from @yomi322.

---

Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

- fluentd or td-agent version.
Fluentd master (068a0045ef4109eb0de6a581fe91628ceb204b1b)
- Environment information:
	- Operating system: `cat /etc/os-release`
	- Kernel version: `uname -r`
```
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

	

```
4.9.0-8-amd64
```

- Your configuration

```aconf
<source>
  @type udp
  format none
  bind 0.0.0.0
  port invalid # Non string
  body_size_limit 4KB
  source_host_key "host"
  tag test
</source>

<match test>
  @type stdout
</match>
```
- Your problem explanation. If you have an error logs, write it together.

No error occurred, but `in_udp` launches with random listening port:

#### Booting Log

```log
2019-06-05 10:48:17 +0900 [info]: parsing config file is succeeded path="example/in_udp.conf"
2019-06-05 10:48:17 +0900 [warn]: 'source_host_key' parameter is deprecated: use source_hostname_key instead.
2019-06-05 10:48:17 +0900 [warn]: 'body_size_limit' parameter is deprecated: use message_length_limit instead.
2019-06-05 10:48:17 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type udp
    format none
    bind "0.0.0.0"
    port invalid
    body_size_limit 4KB
    source_host_key "host"
    tag "test"
    <parse>
      @type none
    </parse>
  </source>
  <match test>
    @type stdout
  </match>
</ROOT>
2019-06-05 10:48:17 +0900 [info]: starting fluentd-1.5.0 pid=3234 ruby="2.5.0"
[...]

PR #2685, opened 2019-11-06
Make config values more strictly and enable to set nil or default

### LINKED ISSUE #2616: Add option to make config value more stricitive

ref: https://github.com/fluent/fluentd/issues/2607

https://github.com/fluent/fluentd/pull/2615#issuecomment-530246816

> We should improve this point to use Integer() or others.
This change has compatibility probolem so it will be implemented in v2 or via configuration.

**Additional context**

ashie:
Fixes #2616

Add a new command line option `--strict-config-value` to parse `integer`, `float` and `bool` values in a config file more strictly. A system config item `strict_config_value` is also added.
Since non-numerical (or unrecognized values for `bool`) values for them in config values are implicitly parsed as `0` , sometimes it makes troubleshooting hard. This option will raise an error for such values to clarify such issues in a config file.

In addition, this PR also adds helper methods to set `nil` or default value to parameters in config files, since many users expect `nil` for `"#{ENV['SOME_ENV_VAR']}"` with empty `SOME_ENV_VAR` environment variable (but it's converted to an empty string in actual).

```
some_param "#{ENV['SOME_ENV_VAR'] || use_nil}"
```

```
some_param "#{ENV['SOME_ENV_VAR'] || use_default}"
```

**Docs Changes**:
* Add `--strict-config-value` to `Command Line Options` in `Deployment - System Configuration`.
* Add `strict_config_value` to system config in `Configuration - Config File Syntax` and `Deployment - System Configuration`.
* Add note for `use_nil` and `use_default` to `Embedding Ruby Expressions` in `Config File Syntax`.

**Release Note**: 
Same with the title.

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -182,24 +182,27 @@ module Fluent
         false
       end
 
-      def options(key)
-        return {} if @corresponding_proxies.empty?
+      def param_type(key)
+        return nil if @corresponding_proxies.empty?
 
         param_key = key.to_sym
         proxy = @corresponding_proxies.detect do |_proxy|
           _proxy.params.has_key?(param_key)
         end
-        return {} unless proxy
+        return nil unless proxy
         _block, opts = proxy.params[param_key]
-        opts
-      end
-
-      def param_type(key)
-        options(key)[:type]
+        opts[:type]
       end
 
       def default_value(key)
-        options(key)[:default]
+        return nil if @corresponding_proxies.empty?
+
+        param_key = key.to_sym
+        proxy = @corresponding_proxies.detect do |_proxy|
+          _proxy.params.has_key?(param_key)
+        end
+        return nil unless proxy
+        proxy.defaults[param_key]
       end
 
       def dump_value(k, v, nindent)
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -182,35 +182,45 @@ module Fluent
         false
       end
 
-      def param_type(key)
-        return nil if @corresponding_proxies.empty?
+      def options(key)
+        return {} if @corresponding_proxies.empty?
 
         param_key = key.to_sym
         proxy = @corresponding_proxies.detect do |_proxy|
           _proxy.params.has_key?(param_key)
         end
-        return nil unless proxy
+        return {} unless proxy
         _block, opts = proxy.params[param_key]
-        opts[:type]
+        opts
+      end
+
+      def param_type(key)
+        options(key)[:type]
+      end
+
+      def default_value(key)
+        options(key)[:default]
       end
 
       def dump_value(k, v, nindent)
-        if secret_param?(k)
-          "#{nindent}#{k} xxxxxx\n"
+        return "#{nindent}#{k} xxxxxx\n" if secret_param?(k)
+        return "#{nindent}#{k} #{v}\n" unless @v1_config
+
+        # for v1 config
+        if v.nil?
+          "#{nindent}#{k} \n"
+        elsif v == :default
+          "#{nindent}#{k} #{default_value(k)}\n"
         else
-          if @v1_config
-            case param_type(k)
-            when :string
-              "#{nindent}#{k} \"#{self.class.unescape_parameter(v)}\"\n"
-            when :enum, :integer, :float, :size, :bool, :time
-              "#{nindent}#{k} #{v}\n"
-            when :hash, :array
-              "#{nindent}#{k} #{v}\n"
-            else
-              # Unknown type
-              "#{nindent}#{k} #{v}\n"
-            end
+          case param_type(k)
+          when :string
+            "#{nindent}#{k} \"#{self.class.unescape_parameter(v)}\"\n"
+          when :enum, :integer, :float, :size, :bool, :time
+            "#{nindent}#{k} #{v}\n"
+          when :hash, :array
+            "#{nindent}#{k} #{v}\n"
           else
+            # Unknown type
             "#{nindent}#{k} #{v}\n"
           end
         end
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -142,7 +142,7 @@ module Fluent
         indent = "  " * nest
         nindent = "  " * (nest + 1)
         out = ""
-        if @arg.empty?
+        if @arg.nil? || @arg.empty?
           out << "#{indent}<#{@name}>\n"
         else
           out << "#{indent}<#{@name} #{@arg}>\n"
--- a/lib/fluent/config/error.rb
+++ b/lib/fluent/config/error.rb
@@ -24,9 +24,9 @@ module Fluent
   class ObsoletedParameterError < ConfigError
   end
 
-  class SetNil < ConfigError
+  class SetNil < Exception
   end
 
-  class SetDefault < ConfigError
+  class SetDefault < Exception
   end
 end
--- a/lib/fluent/config/error.rb
+++ b/lib/fluent/config/error.rb
@@ -23,4 +23,7 @@ module Fluent
 
   class ObsoletedParameterError < ConfigError
   end
+
+  class SetNil < ConfigError
+  end
 end
--- a/lib/fluent/config/error.rb
+++ b/lib/fluent/config/error.rb
@@ -26,4 +26,7 @@ module Fluent
 
   class SetNil < ConfigError
   end
+
+  class SetDefault < ConfigError
+  end
 end
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -52,6 +52,16 @@ module Fluent
       def initialize(strscan, eval_context)
         super(strscan)
         @eval_context = eval_context
+        unless @eval_context.respond_to?(:use_nil)
+          def @eval_context.use_nil
+            raise SetNil
+          end
+        end
+        unless @eval_context.respond_to?(:use_default)
+          def @eval_context.use_default
+            raise SetDefault
+          end
+        end
       end
 
       def parse_literal(string_boundary_charset = LINE_END)
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -81,7 +81,11 @@ module Fluent
         string = []
         while true
           if skip(/\"/)
-            return string.join
+            if string.include?(nil)
+              return nil
+            else
+              return string.join
+            end
           elsif check(/[^"]#{LINE_END_WITHOUT_SPACING_AND_COMMENT}/)
             if s = check(/[^\\]#{LINE_END_WITHOUT_SPACING_AND_COMMENT}/)
               string << s
@@ -168,7 +172,11 @@ module Fluent
 hostname = Socket.gethostname
 worker_id = ENV['SERVERENGINE_WORKER_ID'] || ''
 EOM
-        @eval_context.instance_eval(code)
+        begin
+          @eval_context.instance_eval(code)
+        rescue SetNil => e
+          nil
+        end
       end
 
       def eval_escape_char(c)
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -83,6 +83,8 @@ module Fluent
           if skip(/\"/)
             if string.include?(nil)
               return nil
+            elsif string.include?(:default)
+              return :default
             else
               return string.join
             end
@@ -176,6 +178,8 @@ EOM
           @eval_context.instance_eval(code)
         rescue SetNil => e
           nil
+        rescue SetDefault => e
+          :default
         end
       end
 
--- a/lib/fluent/config/section.rb
+++ b/lib/fluent/config/section.rb
@@ -142,18 +142,35 @@ module Fluent
         proxy.params.each_pair do |name, defval|
           varname = name.to_sym
           block, opts = defval
+          opts = opts.merge(strict: true) if strict_config_value
+
           if conf.has_key?(name.to_s) || opts[:alias] && conf.has_key?(opts[:alias].to_s)
             val = if conf.has_key?(name.to_s)
                     conf[name.to_s]
                   else
                     conf[opts[:alias].to_s]
                   end
-            opts = opts.merge(strict: true) if strict_config_value
-            begin
-              section_params[varname] = self.instance_exec(val, opts, name, &block)
-            rescue ConfigError => e
-              logger.error "config error in:\n#{conf}" if logger
-              raise e
+
+            if val == :default
+              # default value is already set if it exists
+              unless section_params.has_key?(varname)
+                logger.error "config error in:\n#{conf}" if logger
+                raise ConfigError, "'#{varname}' doesn't have default value"
+              end
+            else
+              begin
+                section_params[varname] = self.instance_exec(val, opts, name, &block)
+              rescue ConfigError => e
+                logger.error "config error in:\n#{conf}" if logger
+                raise e
+              end
+            end
+
+            if section_params[varname].nil?
+              unless proxy.defaults.has_key?(varname) and proxy.defaults[varname].nil?
+                logger.error "config error in:\n#{conf}" if logger
+                raise ConfigError, "'#{name}' parameter is required but nil is specified"
+              end
             end
 
             # Source of definitions of deprecated/obsoleted:
@@ -168,12 +185,6 @@ module Fluent
             if opts[:obsoleted]
               logger.error "config error in:\n#{conf}" if logger
               raise ObsoletedParameterError, "'#{name}' parameter is already removed: #{opts[:obsoleted]}" + section_stack
-            end
-            if section_params[varname].nil?
-              unless opts.has_key?(:default) and opts[:default].nil?
-                logger.error "config error in:\n#{conf}" if logger
-                raise ConfigError, "'#{name}' parameter is required but nil is specified"
-              end
             end
           end
           unless section_params.has_key?(varname)
--- a/lib/fluent/config/section.rb
+++ b/lib/fluent/config/section.rb
@@ -106,7 +106,7 @@ module Fluent
     end
 
     module SectionGenerator
-      def self.generate(proxy, conf, logger, plugin_class, strict_config_value = false, stack = [])
+      def self.generate(proxy, conf, logger, plugin_class, stack = [], strict_config_value = false)
         return nil if conf.nil?
 
         section_stack = ""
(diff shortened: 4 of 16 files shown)


ashie:
This PR adds a new option `:strict` to `config_param` to confirm the
concept. Now I've confirmed that it can address referred issues such as
#2607, #2616 and #2444 when I merge this option to all plugins by
default (it's not commited yet). But I think it's not a thing you want.
A global option is the required one, isn't it?

My first question is "where should I add the global option"?

  1. environment variable
  2. command line option
  3. [system config](https://docs.fluentd.org/deployment/system-config)

I think the command line option is the desired place.
What do you think about it?

The second question is: "The `:strict` option for `config_param` is usable?"
If it's not, I don't expose it to plugins and just switch the behavior
internally. If it's usable, I'll continue to implement this feature
(more tests are needed).

I have some more questions. I'll post it later.

cosmo0920:
> My first question is "where should I add the global option"?

> 1. environment variable
> 2. command line option
> 3. system config

IMO, 3 would be ideal. 2 would be better.

> The second question is: "The :strict option for config_param is usable?"

Looks workable in my environment:

```diff
diff --git a/lib/fluent/plugin/in_udp.rb b/lib/fluent/plugin/in_udp.rb
index c2d43611..ac3d0bc3 100644
--- a/lib/fluent/plugin/in_udp.rb
+++ b/lib/fluent/plugin/in_udp.rb
@@ -25,7 +25,7 @@ module Fluent::Plugin
     desc 'Tag of output events.'
     config_param :tag, :string
     desc 'The port to listen to.'
-    config_param :port, :integer, default: 5160
+    config_param :port, :integer, default: 5160, strict: true
     desc 'The bind address to listen to.'
     config_param :bind, :string, default: '0.0.0.0'
```

```aconf
<source>
  @type udp
  format none
  bind 0.0.0.0
  port invalid
  body_size_limit 4KB
  source_host_key "host"
  tag test
</source>

<match test>
  @type stdout
</match>
```

And it would be preferable just adding strict checking.

lib/fluent/config/types.rb
```diff
@@ -98,12 +98,24 @@ def self.regexp_value(str)
...
     }
-    INTEGER_TYPE = Proc.new { |val, opts| val.to_i }
-    FLOAT_TYPE = Proc.new { |val, opts| val.to_f }
-    SIZE_TYPE = Proc.new { |val, opts| Config.size_value(val) }
-    BOOL_TYPE = Proc.new { |val, opts| Config.bool_value(val) }
-    TIME_TYPE = Proc.new { |val, opts| Config.time_value(val) }
-    REGEXP_TYPE = Proc.new { |val, opts| Config.regexp_value(val) }
+    INTEGER_TYPE = Proc.new do |val, opts|
+      if opts[:strict]
+        Integer(val)
```
cosmo0920: Just dumping backtrace is not kindness for users.

How about dump error as follows?

```diff
diff --git a/lib/fluent/config/types.rb b/lib/fluent/config/types.rb
index 9c8b52fc..81cd7875 100644
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -100,7 +100,11 @@ module Fluent
     }
     INTEGER_TYPE = Proc.new do |val, opts|
       if opts[:strict]
-        Integer(val)
+        begin
+          Integer(val)
+        rescue ArgumentError
+          raise ConfigError, "'#{val}' is not valid value for Integer."
+        end
       else
         val.to_i
       end
```

lib/fluent/config/types.rb
```diff
@@ -98,12 +98,24 @@ def self.regexp_value(str)
...
+      else
+        val.to_i
+      end
+    end
+    FLOAT_TYPE = Proc.new do |val, opts|
+      if opts[:strict]
+        Float(val)
```
cosmo0920: Ditto:

```diff
diff --git a/lib/fluent/config/types.rb b/lib/fluent/config/types.rb
index 9c8b52fc..8e5d6527 100644
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -107,7 +107,11 @@ module Fluent
     end
     FLOAT_TYPE = Proc.new do |val, opts|
       if opts[:strict]
-        Float(val)
+        begin
+          Float(val)
+        rescue ArgumentError
+          raise ConfigError, "'#{val}' is not valid value for Float."
+        end
       else
         val.to_f
       end
```

test/config/test_types.rb
```diff
@@ -5,59 +5,95 @@ class TestConfigTypes < ::Test::Unit::TestCase
...
+    test 'not assumed case' do |(expected, val)|
+      assert_equal(expected, Config.size_value(val))
+    end
+
+    data("integer" => [6, 6],
+         "hoge" => [ArgumentError.new("invalid value for Integer(): \"hoge\""), "hoge"],
+         "empty" => [ArgumentError.new("invalid value for Integer(): \"\""), ""],
+         "nil" => [TypeError.new("can't convert nil into Integer"), nil])
+    test 'not assumed case with strict' do |(expected, val)|
+      if expected.kind_of? Exception
```
cosmo0920: Should we use more concrete exception class for the assertion?

test/config/test_types.rb
```diff
@@ -5,59 +5,95 @@ class TestConfigTypes < ::Test::Unit::TestCase
...
-      assert_equal(0.4, Config.time_value(0.4))
-      assert_equal(0.0, Config.time_value("hoge"))
-      assert_equal(0.0, Config.time_value(""))
-      assert_equal(0.0, Config.time_value(nil))
+    data("integer" => [6, 6],
+         "hoge" => [ArgumentError.new("invalid value for Float(): \"hoge\""), "hoge"],
+         "empty" => [ArgumentError.new("invalid value for Float(): \"\""), ""],
+         "nil" => [TypeError.new("can't convert nil into Float"), nil])
+    test 'not assumed case with strict' do |(expected, val)|
+      if expected.kind_of? Exception
```
cosmo0920: Ditto.

cosmo0920:
I've commented the my first impression.
Please take a look, @ashie.

lib/fluent/config/types.rb
ashie: > Just dumping backtrace is not kindness for users.

I think so too.
But I think dumping only `val` isn't still kindness. I want to dump also the value name for it.
I'm not sure how to do it at this moment but probably it can be realized by catching the error at a caller such as `SectionGenerator::generate`.

test/config/test_types.rb
ashie: The concrete exceptions are checked by the following `assert_raise`.

repeatedly:
I think `:strict` option in `config_param` is not good because it is unclear behaviour for users.
So command line option and system configuration are better.

lib/fluent/config/types.rb
ashie: I implemented it.

lib/fluent/config/types.rb
ashie: Implemented.

lib/fluent/config/literal_parser.rb
```diff
@@ -81,7 +91,13 @@ def scan_double_quoted_string
         string = []
         while true
           if skip(/\"/)
-            return string.join
+            if string.include?(nil)
```
ganmacs: If a user writes config like `"TEST#{ENV['A'] || use_nil}"`, the actual value becomes `nil`, not `"TEST"` (I mean, we can use `use_nil` or `use_default` even if there is other values).  Is it intentional?

lib/fluent/config/types.rb
```diff
@@ -20,7 +20,18 @@
 
 module Fluent
   module Config
-    def self.size_value(str)
+    def self.default_value(opts = {})
+      if opts.has_key?(:default)
+        opts[:default]
+      else
+        nil
```
ganmacs: Is there a reason why it returns `nil` if the default value is not defined? I think that raising some errors is fine.

lib/fluent/config/literal_parser.rb
ashie: Yes, it's intentional. But I'm not sure it's desired behavior or not. At first I implemented it as `string == [nil]` but replaced it with `string.include?(nil)` for considering some cases like following:

```
somepath "#{ENV['HOME'] || use_nil}/path/to/sub/dir"
```

In this case users might not want replacing it with "/path/to/sub/dir".

lib/fluent/config/types.rb
```diff
@@ -84,46 +104,106 @@ def self.regexp_value(str)
...
+        if val.nil?
+          nil
+        elsif val == :default
+          Config.default_value(opts)
+        elsif opts[:strict]
+          Integer(val)
+        else
+          val.to_i
+        end
+      rescue ArgumentError, TypeError => e
```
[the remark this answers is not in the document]
ashie: That's right. I've addressed it. Thanks!

lib/fluent/config/error.rb
```diff
@@ -23,4 +23,10 @@ class ConfigParseError < ConfigError
 
   class ObsoletedParameterError < ConfigError
   end
+
+  class SetNil < ConfigError
```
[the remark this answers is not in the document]
ashie: I agree with you, it's not an error.
At first I implemented it as `class SetNil < Exception` but replaced it with the current implementation to rescue it by default. But now I think it should be rescued intentionally.

lib/fluent/config/section.rb
```diff
@@ -106,7 +106,7 @@ def method_missing(name, *args)
     end
 
     module SectionGenerator
-      def self.generate(proxy, conf, logger, plugin_class, stack = [])
+      def self.generate(proxy, conf, logger, plugin_class, strict_config_value = false, stack = [])
```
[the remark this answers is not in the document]
ashie: Because I checked all callers in fluentd then I found that there is a case that `stack` is omitted and `strict_config_value` should be specified. The reversed case doesn't seem exist.
But I've notice that I didn't care the case that plugins may use it. I'll address it.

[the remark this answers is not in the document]
ashie: done.

lib/fluent/config/types.rb
ashie: I agree with you. And now I've noticed that we should handle also `nil` case more properly. The current code causes unhandled error when a plugin doesn't care `nil`.
In addition the current code breaks `Element::to_s`. We should add `nil` case for it.

ashie: I'm now adding more tests for `Configurable`.
While adding them I've noticed that the implementation for `:default` is wrong.
The default value should be taken from `ConfigProxy::defaults`, not `opts[:default]`.
Otherwise it cannot return a correct default value overridden by `config_set_default`.
I'm now addressing it.

ashie: I've fixed it.

ashie:
Oops, sorry I missed some comments.
I'll address them.

test/config/test_types.rb
```diff
@@ -5,59 +5,146 @@ class TestConfigTypes < ::Test::Unit::TestCase
...
+    test 'nil with strict' do
+      assert_equal(nil, Config.size_value(nil, { strict: true }))
+    end
+
+    test 'default' do
+      assert_equal(2, Config.size_value(:default, { default: 2 }))
+    end
+
+    test 'no default' do
+      assert_equal(nil, Config.size_value(:default))
```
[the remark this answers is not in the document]
ashie: I removed these tests because getting default value at `Config.*_TYPE` is wrong as I mentioned in https://github.com/fluent/fluentd/pull/2685#discussion_r352413304. Now getting default value is done at section.rb and I added tests for them to test_configurable.rb

test/config/test_configurable.rb
```diff
@@ -602,6 +602,20 @@ class TestConfigurable < ::Test::Unit::TestCase
             assert_equal([], x6a.obj2)
           end
         end
+
+        test 'strict value type' do
+          class InfFloatConfig
```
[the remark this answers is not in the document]
ashie: Probably its inpossible because the class name is referred while constructing the class.
I've moved it to `ConfigurableSpec` to respect the way of sibling tests.

lib/fluent/supervisor.rb
```diff
@@ -427,6 +427,7 @@ def self.default_options
         suppress_repeated_stacktrace: true,
         without_source: nil,
         use_v1_config: true,
+        strict_config_value: nil,
```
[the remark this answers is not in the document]
ashie: It should be `nil` to respect SystemConfig's one if the command line argument isn't specified.
`without_source` is also `bool` and set as `nil`. I think it's same reason.
If this value is false, SystemConfig's one isn't applied even if a user doesn't use the command line option. See [`Supervisor::build_system_config`](https://github.com/fluent/fluentd/blob/master/lib/fluent/supervisor.rb#L802).

lib/fluent/system_config.rb
```diff
@@ -95,13 +97,13 @@ def self.overwrite_system_config(hash)
...
+    def initialize(conf=nil, strict_config_value=false)
       super()
       conf ||= SystemConfig.blank_system_config
-      configure(conf)
+      configure(conf, strict_config_value: strict_config_value)
     end
 
-    def configure(conf)
+    def configure(conf, strict_config_value: false)
       super
```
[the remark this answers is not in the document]
ashie: > Does strict_config_value in system config file work correctly?

No. I just want to discuss the behavior for it.

Should `strict_config_value` in `SystemConfig` affect to parameters in `SystemConfig` itself?
To do it, `strict_config_value` should be parsed before parsing other parameters.

lib/fluent/system_config.rb
```diff
@@ -95,13 +97,13 @@ def self.overwrite_system_config(hash)
       end
     end
 
-    def initialize(conf=nil)
+    def initialize(conf=nil, strict_config_value=false)
```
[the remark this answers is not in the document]
ashie: When I use keyword argument for it, some tests are broken like this:

```
==============================================================================================================================================
/home/aho/Projects/fluent/fluentd/test/config/test_system_config.rb:100:in `block in <class:TestSystemConfig>'
/home/aho/Projects/fluent/fluentd/test/config/test_system_config.rb:100:in `new'
/home/aho/Projects/fluent/fluentd/lib/fluent/system_config.rb:103:in `initialize'
/home/aho/Projects/fluent/fluentd/lib/fluent/system_config.rb:107:in `configure'
/home/aho/Projects/fluent/fluentd/lib/fluent/configurable.rb:90:in `configure'
Error: test: accepts parameters[workers](Fluent::Config::TestSystemConfig): NoMethodError: undefined method `<<' for nil:NilClass
```

`nil` seems be passed for the first argument of `initialize` even if a caller passed a valid config.

It is resolved when I add the argument `strict_config_value: flase (or true)` to the caller, or remove the default value `nil` from `conf`.
I'm not sure Ruby's spec for this issue.

lib/fluent/system_config.rb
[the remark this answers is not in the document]
ashie: Ah, it's not difficult. Just only passing parsed `conf["strict_config_value"]` to `super`.
I'll address it.

lib/fluent/system_config.rb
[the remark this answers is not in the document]
ganmacs: > It is resolved when I add the argument strict_config_value: flase (or true) to the caller, or remove the default value nil from conf.

This is what I intended. the following two syntaxes should be almost the same. 
Or am I get wrong something?

```
def initialize(conf=nil, strict_config_value=false)
 ...
end

```

```
def initialize(conf=nil, strict_config_value: false)
 ...
end
```

lib/fluent/config/element.rb
```diff
@@ -142,7 +142,7 @@ def to_s(nest = 0)
         indent = "  " * nest
         nindent = "  " * (nest + 1)
         out = ""
-        if @arg.empty?
+        if @arg.nil? or @arg.empty?
```
ganmacs: [nits] I think Rubyist prefers `||` to `or`. (`and` is the same.  preferred `&&` to `and`)

```suggestion
        if @arg.nil? || @arg.empty?
```

lib/fluent/config/section.rb
```diff
@@ -122,9 +122,23 @@ def self.generate(proxy, conf, logger, plugin_class, stack = [])
         end
 
         if proxy.argument
-          unless conf.arg.empty?
+          unless conf.arg.nil? or conf.arg.empty?
```
ganmacs:
```suggestion
          unless conf.arg.nil? || conf.arg.empty?
```

lib/fluent/config/element.rb
```diff
@@ -182,35 +182,45 @@ def secret_param?(key)
         false
       end
 
-      def param_type(key)
-        return nil if @corresponding_proxies.empty?
+      def options(key)
```
ganmacs: How about `options_for(key)`(or other better name)? `options(key)` is a bit obscure for me.

ganmacs: And if this method is used by this class only, could you make it a private method?

lib/fluent/config/element.rb
ganmacs: `@arg` seems to be expected String type. when does `@arg.nil? #=> true` happen?

ashie: `nil` will be set by `use_nil` in embedded Ruby code which is introduced in this PR, like the following case:

```
<section>
  <subsection "#{ENV['SOME_ENV_VAR'] || use_nil}">
    ...
  </subsection>
<sectin>
```

lib/fluent/config/element.rb
ashie: Sorry, now I think that this method isn't needed anymore, I'll revert it.

This method is introduced for getting a default value:

```diff
+       def default_value(key)
+         options(key)[:default]
+       end
```

and unifying similar code in `param_type` method. But getting a default value from `options` is wrong as I mentioned in https://github.com/fluent/fluentd/pull/2685#discussion_r352413304.

I'll revert this code, and simply show `:default` for a default value since it can't determined in this context.

ashie: > I'll revert this code, and simply show a text :default for a default value since it can't determined in this context.

It's still wrong.
A default value can be taken from `@corresponding_proxies`.
I've fixed it.

lib/fluent/system_config.rb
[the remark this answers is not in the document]
ashie: Probably one of the cause is that `Fluent::Config::Element` inherits Hash class.
On calling `SystemConfig::new(conf)`, `conf`is evaluated as keyword arguments at first since its a Hash object (like `SystemConfig::new({a: 1, b: 2})`). The conf object seems be broken while processing arguments by Ruby (maybe).

To avoid such confusion, keyword arguments shouldn't be used in this case.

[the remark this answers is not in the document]
ashie: > why don't you use keyword arguments since you already did it at SystemConfig#configure and SystemConfig.create?

I've replace with a normal argument for them to unify the signature and to avoid confusion.

repeatedly:
Looks good for me. Does anyone have other concerns?

ashie:
Thanks for your review!
It's completed, I have no additional patch for this PR.
I'll add description for this feature to fluent-docs if there is no objection to its specificaion.

cosmo0920:
Looks good for me, too. I have no concern for this patches.

repeatedly:
Merged. Thanks!

PR #2688, opened 2019-11-07
Fix pitfall which is the order of filter and match

ganmacs:
no

https://docs.fluentd.org/configuration/config-file#note-on-match-order says as follows.

> The common pitfall is when you put a <filter> block after <match>. It will never work as supposed, since events never go through the filter for the reason explained above.

I think the behavior is a bit confusing. And I couldn't find any reasonable reasons for this behavior.
if there are reasons/pros to adopt this behavior, Could you tell me why it is? (@repeatedly ) or just a historical reason?

**Docs Changes**:

Delete the sentence about filter-match order in https://docs.fluentd.org/configuration/config-file#note-on-match-order

**Release Note**: 

same as the title

**TODO**

* [ ] add test 

before
```

2019-11-07 16:30:21 +0900 [info]: gem 'fluentd' version '1.8.0.rc2'
2019-11-07 16:30:22 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:30:22 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:30:22 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type dummy
    tag "test.test.test.test"
  </source>
  <match test.test.test.test>
    @type stdout
  </match>
  <filter test.**>
    @type record_transformer
    <record>
      test_rec test
    </record>
  </filter>
</ROOT>
2019-11-07 16:30:22 +0900 [info]: starting fluentd-1.8.0.rc2 pid=69911 ruby="2.6.5"
2019-11-07 16:30:22 +0900 [info]: spawn command to main:  cmdline=["ruby", "-Eascii-8bit:ascii-8bit", "-rbundler/setup", "bin/fluentd", "-c", "example/debug/r.conf", "--under-supervisor"]
2019-11-07 16:30:22 +0900 [info]: adding match pattern="test.test.test.test" type="stdout"
2019-11-07 16:30:22 +0900 [info]: adding filter pattern="test.**" type="record_transformer"
2019-11-07 16:30:22 +0900 [info]: adding source type="dummy"
2019-11-07 16:30:22 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:30:22 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:30:22 +0900 [info]: #0 starting fluentd worker pid=69925 ppid=69911 worker=0
2019-11-07 16:30:22 +0900 [info]: #0 fluentd worker is now running worker=0
2019-11-07 16:30:23.054363000 +0900 test.test.test.test: {"message":"dummy"}
2019-11-07 16:30:24.075024000 +0900 test.test.test.test: {"message":"dummy"}
2019-11-07 16:30:25.001102000 +0900 test.test.test.test: {"message":"dummy"}
```

after

```
2019-11-07 16:29:34 +0900 [info]: gem 'fluentd' version '1.8.0.rc2'
2019-11-07 16:29:34 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:29:34 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:29:34 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type dummy
    tag "test.test.test.test"
  </source>
  <match test.test.test.test>
    @type stdout
  </match>
  <filter test.**>
    @type record_transformer
    <record>
      test_rec test
    </record>
  </filter>
</ROOT>
2019-11-07 16:29:34 +0900 [info]: starting fluentd-1.8.0.rc2 pid=69824 ruby="2.6.5"
2019-11-07 16:29:34 +0900 [info]: spawn command to main:  cmdline=["ruby", "-Eascii-8bit:ascii-8bit", "-rbundler/setup", "bin/fluentd", "-c", "example/debug/r.conf", "--under-supervisor"]
2019-11-07 16:29:34 +0900 [info]: adding filter pattern="test.**" type="record_transformer"
2019-11-07 16:29:34 +0900 [info]: adding match pattern="test.test.test.test" type="stdout"
2019-11-07 16:29:34 +0900 [info]: adding source type="dummy"
2019-11-07 16:29:34 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:29:34 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2019-11-07 16:29:34 +0900 [info]: #0 starting fluentd worker pid=69838 ppid=69824 worker=0
2019-11-07 16:29:34 +0900 [info]: #0 fluentd worker is now running worker=0
2019-11-07 16:29:35.015223000 +0900 test.test.test.test: {"message":"dummy","test_rec":"test"}
2019-11-07 16:29:36.039565000 +0900 test.test.test.test: {"message":"dummy","test_rec":"test"}
2019-11-07 16:29:37.066552000 +0900 test.test.test.test: {"message":"dummy","test_rec":"test"}

```

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -61,19 +61,26 @@ module Fluent
       super
 
       # initialize <match> and <filter> elements
-      conf.elements('filter', 'match').each { |e|
+      conf.elements('filter').each do |e|
         if !Fluent::Engine.supervisor_mode && e.for_another_worker?
           next
         end
         pattern = e.arg.empty? ? '**' : e.arg
         type = e['@type']
         raise ConfigError, "Missing '@type' parameter on <#{e.name}> directive" unless type
-        if e.name == 'filter'
-          add_filter(type, pattern, e)
-        else
-          add_match(type, pattern, e)
+        add_filter(type, pattern, e)
+      end
+
+      conf.elements('match').each do |e|
+        if !Fluent::Engine.supervisor_mode && e.for_another_worker?
+          next
         end
-      }
+        pattern = e.arg.empty? ? '**' : e.arg
+        type = e['@type']
+        raise ConfigError, "Missing '@type' parameter on <#{e.name}> directive" unless type
+
+        add_match(type, pattern, e)
+      end
     end
 
     def lifecycle_control_list


repeatedly:
So you think following config should work?

```
<match pattern>
  @type foo
</match>

<filter pattern>
  @type bar
</filter>
```

ganmacs:
> So you think following config should work?

Yes. this is IMO but code position shouldn't affect semantics(as much as possible).

repeatedly:
Hmm... I think this is error prone because users need to check all configuration.
Currenlty, data flow is very clear because `<match>` is the end position.
But if we allow this order, the end positions is unclear.
For example, fluentd users use `@include` in many places.

```
@include a.conf  # <source>, <filter> and <match>
@include b.conf  # <source>, <filter> and <match>
@include c.conf  # <source>, <filter> and <match>
```

In current behaviour, users don't need to check b.conf and c.conf for a.conf <match>. But this change is not.

repeatedly:
I think if users need to add <filter> after <match>, it is something wrong.

ganmacs:
> Currenlty, data flow is very clear because <match> is the end position.

Current fluentd accepts the following config, right? (there is `source` directive after `match` directive). so, after all, users need to check all configs because there is no order restriction between `match` directive and `source` directive. I think it's Inconsistent that there is the order dependency between`match` directive and `filter` directive only. 

```
<filter test.**>
  @type record_transformer
  <record>
    test_rec test
  </record>
</filter>

<match test.test.test.test>
  @type stdout
</match>

<source>
  @type dummy
  tag test.test.test.test
</source>
```

ganmacs:
or it would be good that we introduce order dependency between `source` and  `match` directive.

repeatedly:
> it would be good that we introduce order dependency between source and match directive.

I think this is good. Current behaviour is mainly for `<match fluent.**>` because fluentd doesn't have label feature before.

ganmacs:
I raised an issue about it.

PR #2700, opened 2019-11-14
Allow cert chains

### LINKED ISSUE #2699: Fluentd does not support certificate chains

X509 certificate files often contain the leaf certificate and all intermediate CA certificates excluding the root. Fluentd should support such files, otherwise only certificates signed directly by the root CA can be used.

Load all certs in the provided file into the OpenSSL context.

The only alternative is to issue certificates directly from the root, which is bad practice.

**Additional context**

About to open a PR - this issue is mostly for documentation purposes.

mmussomele:
Fixes #2699 

Adds support for certificate chains to Fluentd sockets. Needed for good practice around issuing certificates.

The added test fails if the new line:
```
context.extra_chain_cert = client_certs[1..-1]
```
is remove (effectively reverting to the old behavior of only loading the leaf certificate).

**Docs Changes**: 

Unsure if needed.

**Release Note**:

### DIFF
--- a/lib/fluent/plugin_helper/socket.rb
+++ b/lib/fluent/plugin_helper/socket.rb
@@ -38,6 +38,7 @@ module Fluent
       ### follow httpclient configuration by nahi
       # OpenSSL 0.9.8 default: "ALL:!ADH:!LOW:!EXP:!MD5:+SSLv2:@STRENGTH"
       CIPHERS_DEFAULT = "ALL:!aNULL:!eNULL:!SSLv2" # OpenSSL >1.0.0 default
+      CERT_PEM_RE = Regexp.compile('-+BEGIN CERTIFICATE-+\n(?:[^-]*\n)+-+END CERTIFICATE-+\n', Regexp::MULTILINE)
 
       attr_reader :_sockets # for tests
 
@@ -155,8 +156,12 @@ module Fluent
           else
             context.verify_hostname = false
           end
-          context.cert = OpenSSL::X509::Certificate.new(File.read(cert_path)) if cert_path
-          context.key = OpenSSL::PKey::read(File.read(private_key_path), private_key_passphrase) if private_key_path
+
+          cert_data = File.read(cert_path)
+          client_certs = []
+          cert_data.scan(CERT_PEM_RE){|match| client_certs << OpenSSL::X509::Certificate.new(match)}
+          private_key = OpenSSL::PKey::read(File.read(private_key_path), private_key_passphrase) if private_key_path
+          context.add_certificate(client_certs[0], private_key, client_certs[1..-1])
         end
 
         tcpsock = socket_create_tcp(host, port, **kwargs)
--- a/lib/fluent/plugin_helper/socket.rb
+++ b/lib/fluent/plugin_helper/socket.rb
@@ -150,7 +150,11 @@ module Fluent
           context.ciphers = ciphers
           context.verify_mode = OpenSSL::SSL::VERIFY_PEER
           context.cert_store = cert_store
-          context.verify_hostname = true if verify_fqdn && fqdn && context.respond_to?(:verify_hostname=)
+          if verify_fqdn && fqdn && context.respond_to?(:verify_hostname=)
+            context.verify_hostname = true
+          else
+            context.verify_hostname = false
+          end
           context.cert = OpenSSL::X509::Certificate.new(File.read(cert_path)) if cert_path
           context.key = OpenSSL::PKey::read(File.read(private_key_path), private_key_passphrase) if private_key_path
         end
--- a/test/plugin/test_out_forward.rb
+++ b/test/plugin/test_out_forward.rb
@@ -1,3 +1,6 @@
+require 'openssl'
+require 'tempfile'
+
 require_relative '../helper'
 require 'fluent/test/driver/output'
 require 'fluent/plugin/out_forward'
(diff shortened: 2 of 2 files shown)


mmussomele:
seems AppVeyor doesn't like `ensure`, will have to just let the temporary files be deleted by the GC or on exit

mmussomele:
unsure why the builds failed - the tests passed for me locally

mtovino-nefeli:
👍 this would help me too

repeatedly:
I will check this patch in this week.

> unsure why the builds failed - the tests passed for me locally

This is travis-ci environment issue. It sometimes slow down and it causes unexpected test failure.

lib/fluent/plugin_helper/socket.rb
```diff
@@ -150,8 +150,19 @@ def socket_create_tls(
           context.ciphers = ciphers
           context.verify_mode = OpenSSL::SSL::VERIFY_PEER
           context.cert_store = cert_store
-          context.verify_hostname = true if verify_fqdn && fqdn && context.respond_to?(:verify_hostname=)
-          context.cert = OpenSSL::X509::Certificate.new(File.read(cert_path)) if cert_path
+          if verify_fqdn && fqdn && context.respond_to?(:verify_hostname=)
+            context.verify_hostname = true
+          else
+            context.verify_hostname = false
```
repeatedly: Why change this line to set `false` explicitly?

test/plugin/test_out_forward.rb
```diff
@@ -621,6 +624,105 @@ def try_write(chunk)
...
+    # No need to write root key to file
+
+    intermediate = create_certificate("intermediate-ca", 2, root)
+    # The intermediate_ca is not written to a file - it will be in the
+    # client and service certificates' chain.
+
+    # Test both (root->cert) and (root->intermediate->cert)
+    parents = [root, intermediate]
+    parents.each do |parent|
+      client = create_certificate("client-cert", 3, parent, false)
```
repeatedly: you can use following code:

```
client_cert, client_key = create_certificate("client-cert", 3, parent, false)
```

test/plugin/test_out_forward.rb
```diff
@@ -621,6 +624,105 @@ def try_write(chunk)
...
+      if parent != root
+        client_cert_file.write(parent[0].to_pem())
+      end
+      client_cert_file.close
+
+      client_key_file = Tempfile.new("client-key")
+      client_key_file.write(client_key.to_pem())
+      client_key_file.close
+
+      server = create_certificate("server-cert", 4, parent, false)
```
repeatedly: ditto

test/plugin/test_out_forward.rb
```diff
@@ -920,6 +1022,39 @@ def create_target_input_driver(response_stub: nil, disconnect: false, conf: TARG
...
+    cert.serial = serial
+    cert.subject = OpenSSL::X509::Name.parse "/CN="+name
+    cert.issuer = parent_cert.subject
+    cert.public_key = key.public_key
+    cert.not_before = Time.now
+    cert.not_after = cert.not_before + 60*60 # 1 hour
+    ef = OpenSSL::X509::ExtensionFactory.new
+    ef.subject_certificate = cert
+    ef.issuer_certificate = parent_cert
+    cert.add_extension(ef.create_extension("subjectKeyIdentifier","hash",false))
```
ganmacs: space after `,`.
```suggestion
    cert.add_extension(ef.create_extension("subjectKeyIdentifier", "hash", false))
```

test/plugin/test_out_forward.rb
```diff
@@ -920,6 +1022,39 @@ def create_target_input_driver(response_stub: nil, disconnect: false, conf: TARG
...
+    parent_cert = cert
+    parent_key = key
+    if parent
+      parent_cert = parent[0]
+      parent_key = parent[1]
+    end
+
+    cert.version = 2
+    cert.serial = serial
+    cert.subject = OpenSSL::X509::Name.parse "/CN="+name
```
ganmacs: how about this?
```suggestion
    cert.subject = OpenSSL::X509::Name.parse("/CN=#{name}")
```

test/plugin/test_out_forward.rb
```diff
@@ -920,6 +1022,39 @@ def create_target_input_driver(response_stub: nil, disconnect: false, conf: TARG
...
+    cert.issuer = parent_cert.subject
+    cert.public_key = key.public_key
+    cert.not_before = Time.now
+    cert.not_after = cert.not_before + 60*60 # 1 hour
+    ef = OpenSSL::X509::ExtensionFactory.new
+    ef.subject_certificate = cert
+    ef.issuer_certificate = parent_cert
+    cert.add_extension(ef.create_extension("subjectKeyIdentifier","hash",false))
+    if ca
+      cert.add_extension(ef.create_extension("basicConstraints","CA:TRUE",true))
```
ganmacs: space after ,. (same as L1048-L1052)
```suggestion
      cert.add_extension(ef.create_extension("basicConstraints", "CA:TRUE", true))
```

test/plugin/test_out_forward.rb
```diff
@@ -621,6 +624,105 @@ def try_write(chunk)
...
+      target_input_driver.run(expect_records: 2, timeout: 3) do
+        d.run(default_tag: 'test', wait_flush_completion: false, shutdown: false) do
+          records.each do |record|
+            d.feed(time, record)
+          end
+        end
+      end
+
+      events = target_input_driver.events
+      assert{ events != [] }
```
ganmacs: `assert` has little information when this case failed. so how about this?

```suggestion
      assert_not_equal(event, [])
```

lib/fluent/plugin_helper/socket.rb
```diff
@@ -150,8 +150,19 @@ def socket_create_tls(
...
+          end
+
+          cert_data = File.read(cert_path)
+          pattern = Regexp.compile('-+BEGIN CERTIFICATE-+\n(?:[^-]*\n)+-+END CERTIFICATE-+\n', Regexp::MULTILINE)
+          client_certs = []
+          cert_data.scan(pattern){|match| client_certs << OpenSSL::X509::Certificate.new(match)}
+          context.cert = client_certs[0]
+          context.extra_chain_cert = client_certs[1..-1]
```
ganmacs: these methods are deprecated https://ruby.github.io/openssl/OpenSSL/SSL/SSLContext.html#method-i-add_certificate

> cert=, key=, and extra_chain_cert= are old accessor methods for setting certificate and internally call this method.

how about #add_certificate?

test/plugin/test_out_forward.rb
```diff
@@ -621,6 +624,105 @@ def try_write(chunk)
     assert_equal(['test', time, records[1]], events[1])
   end
 
+  data('ack true' => true,
+       'ack false' => false)
+  test 'TLS transport client cert chain' do |ack|
+    omit "TLS and 'ack false' always fails. Need to debug" if !ack
```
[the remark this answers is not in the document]
mmussomele: This is an existing bug prior to my changes - see the [test above](https://github.com/fluent/fluentd/blob/4cd9d6b9137dd52ddb5f8f889a64f0fe4f26cbf6/test/plugin/test_out_forward.rb#L585).

test/plugin/test_out_forward.rb
mmussomele: will do - note that this is copied from the test above though

lib/fluent/plugin_helper/socket.rb
mmussomele: oh that's much better, will do

test/plugin/test_out_forward.rb
mmussomele: oh great - new to ruby ha, didn't know it could auto-unpack

lib/fluent/plugin_helper/socket.rb
mmussomele: see commit body of the first [commit](https://github.com/fluent/fluentd/pull/2700/commits/18d6846c046cbd081142444e3a5c0b3342fa3acd):
> Otherwise it seems to default to true, which results in logging of
>
>    warning: verify_hostname requires hostname to be set
>
> When running tests that use TLS without verifying the hostname.

mmussomele:
@repeatedly @ganmacs changes made, except for the `verify_hostname` one - just want to verify with discussion first. Thanks for the comments!

test/plugin/test_out_forward.rb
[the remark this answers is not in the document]
ganmacs: https://github.com/fluent/fluentd/blob/4cd9d6b9137dd52ddb5f8f889a64f0fe4f26cbf6/test/plugin/test_out_forward.rb#L585 is for AppVeyor environment which the test run Windows. It runs on Linux environment. below is right.

```suggestion
    omit "TLS and 'ack false' always fails. Need to debug" if Fluent.windows? && !ack
```

repeatedly:
Re-run test to check random test failures.

test/plugin/test_out_forward.rb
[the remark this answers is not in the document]
mmussomele: corrected omit check

mmussomele:
sorry for the delay, have been very busy. Just updated the omit check to specify windows

ganmacs:
This is done by https://github.com/fluent/fluentd/pull/2930 .

PR #2702, opened 2019-11-18
Add log throttling per file

rewiko:
Running in a big cluster with high volume of log, it would be nice to throttle the log shipping to avoid network saturation and make it easier to calculate the max throughput per node for example in a Kubernetes cluster.

Tail plugin is watching files and every second reading from the last pointer to the end of the file.
This change allow to stop reading the file after X number of logs lines read and update the pointer in the pos file as usual.

This [commit](https://github.com/fluent/fluentd/pull/2702/commits/2ce05b1aa46737409e95e9d9725386fd1e385804) adds log throttling per bytes for each files, should work only when `watch_timer` is enabled and the `stat_watcher` (inotify) is disabled. 

In order to have this feature for any watch configuration (timer or inotify..), I've updated to add a sleep when you have been reaching the bytes read limit, the sleep would block process and affect other file ingestion, I've added a basic thread array to have a multi-threading ingestion. 
However I've noticed you are relying on `cool.io`, and I was wondering if I should use this library instead.

Would you be interested in this feature?

Some [discussions](https://github.com/rewiko/fluentd/pull/1) before submitting the PR.

**Docs Changes**:

adding read_lines_limit_per_notify which by default is set to -1, so no throttling involve by default.

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -208,6 +208,7 @@ module Fluent::Plugin
       @threads['in_tail_refresh_watchers'] = Thread.new(@refresh_interval) do |refresh_interval|
         timer_execute(:in_tail_refresh_watchers, @refresh_interval, &method(:refresh_watchers))
       end
+      @threads['in_tail_refresh_watchers'].priority = 10 # Default is zero; higher-priority threads will run before lower-priority threads.
       
       @threads.each { |thr| 
         thr.join
@@ -289,6 +290,11 @@ module Fluent::Plugin
 
       stop_watchers(unwatched, immediate: false, unwatched: true) unless unwatched.empty?
       start_watchers(added) unless added.empty?
+
+      log.debug "Thread refresh_watchers"
+      @threads.each { |thr| 
+            log.debug "Thread #{thr[0]} #{thr[1].status}"
+      }
     end
 
     def setup_watcher(path, pe)
@@ -312,26 +318,37 @@ module Fluent::Plugin
 
     def start_watchers(paths)
       paths.each { |path|
-        @threads[path] = Thread.new(path) do |path|
-          pe = nil
-          if @pf
-            pe = @pf[path]
-            if @read_from_head && pe.read_inode.zero?
-              begin
-                pe.update(Fluent::FileWrapper.stat(path).ino, 0)
-              rescue Errno::ENOENT
-                $log.warn "#{path} not found. Continuing without tailing it."
+        unless @threads[path].nil?
+          log.debug "Check Thread #{path} #{@threads[path].status}"
+          if @threads[path].status != "sleep" and  @threads[path].status != "run"
+            log.debug "Stopping Thread #{path} #{@threads[path].status}"
+            @threads[path].exit
+            @threads.delete(path)
+          end
+        end
+        if @threads[path].nil?
+          log.debug "Add Thread #{path}"
+          @threads[path] = Thread.new(path) do |path|
+            pe = nil
+            if @pf
+              pe = @pf[path]
+              if @read_from_head && pe.read_inode.zero?
+                begin
+                  pe.update(Fluent::FileWrapper.stat(path).ino, 0)
+                rescue Errno::ENOENT
+                  $log.warn "#{path} not found. Continuing without tailing it."
+                end
               end
             end
-          end
 
-          begin
-            tw = setup_watcher(path, pe)
-          rescue WatcherSetupError => e
-            log.warn "Skip #{path} because unexpected setup error happens: #{e}"
-            next
+            begin
+              tw = setup_watcher(path, pe)
+            rescue WatcherSetupError => e
+              log.warn "Skip #{path} because unexpected setup error happens: #{e}"
+              next
+            end
+            @tails[path] = tw
           end
-          @tails[path] = tw
         end
       }
     end
@@ -375,6 +392,7 @@ module Fluent::Plugin
         end
       end
       rotated_tw = @tails[path]
+
       @tails[path] = setup_watcher(path, pe)
       detach_watcher_after_rotate_wait(rotated_tw) if rotated_tw
     end
@@ -781,6 +799,7 @@ module Fluent::Plugin
                     number_bytes_read += bytes_to_read 
                     limit_bytes_per_second_reached = (number_bytes_read >= @watcher.read_bytes_limit_per_second and @watcher.read_bytes_limit_per_second > 0)
 
+                    @watcher.log.debug("reading file: #{ @watcher.path}")
                     if @lines.size >= @watcher.read_lines_limit or limit_bytes_per_second_reached 
                       # not to use too much memory in case the file is very large
                       read_more = true
@@ -788,7 +807,7 @@ module Fluent::Plugin
                       if limit_bytes_per_second_reached 
                         # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
                         time_spent_reading = Time.new - start_reading 
-                        @watcher.log.debug("time_spent_reading: #{time_spent_reading}")
+                        @watcher.log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
                         if (time_spent_reading < 1)
                           debug_time = 1 - time_spent_reading
                           @watcher.log.debug("sleep: #{debug_time}")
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -201,7 +204,15 @@ def start
       end
 
       refresh_watchers unless @skip_refresh_on_startup
-      timer_execute(:in_tail_refresh_watchers, @refresh_interval, &method(:refresh_watchers))
+
+      @threads['in_tail_refresh_watchers'] = Thread.new(@refresh_interval) do |refresh_interval|
```
ganmacs: can you use https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-thread

lib/fluent/plugin/in_tail.rb
```diff
@@ -201,7 +204,15 @@ def start
...
+        timer_execute(:in_tail_refresh_watchers, @refresh_interval, &method(:refresh_watchers))
+      end
+      @threads['in_tail_refresh_watchers'].priority = 10 # Default is zero; higher-priority threads will run before lower-priority threads.
+      
+      @threads.each { |thr| 
+        thr.join
```
ganmacs: if it blocks here, all code after this is blocking.

lib/fluent/plugin/in_tail.rb
```diff
@@ -279,11 +290,16 @@ def refresh_watchers
 
       stop_watchers(unwatched, immediate: false, unwatched: true) unless unwatched.empty?
       start_watchers(added) unless added.empty?
+
+      log.debug "Thread refresh_watchers"
+      @threads.each { |thr| 
+            log.debug "Thread #{thr[0]} #{thr[1].status}"
```
ganmacs: indent

lib/fluent/plugin/in_tail.rb
ganmacs: `@threads` is hash. so `thr` is Array.

lib/fluent/plugin/in_tail.rb
```diff
@@ -356,6 +392,7 @@ def update_watcher(path, pe)
         end
       end
       rotated_tw = @tails[path]
+
```
ganmacs: unnecessary

lib/fluent/plugin/in_tail.rb
```diff
@@ -302,31 +318,49 @@ def setup_watcher(path, pe)
...
         end
-        @tails[path] = tw
       }
     end
 
     def stop_watchers(paths, immediate: false, unwatched: false, remove_watcher: true)
       paths.each { |path|
         tw = remove_watcher ? @tails.delete(path) : @tails[path]
+        if remove_watcher
+          @threads[path].exit
```
ganmacs: Thread#exit is dangerous. could you finish this thread in a proper way?

lib/fluent/plugin/in_tail.rb
```diff
@@ -747,16 +785,37 @@ def on_notify
...
               if !io.nil? && @lines.empty?
                 begin
                   while true
-                    @fifo << io.readpartial(8192, @iobuf)
+                    @fifo << io.readpartial(bytes_to_read, @iobuf)
                     @fifo.read_lines(@lines)
-                    if @lines.size >= @watcher.read_lines_limit
+
+                    number_bytes_read += bytes_to_read 
+                    limit_bytes_per_second_reached = (number_bytes_read >= @watcher.read_bytes_limit_per_second and @watcher.read_bytes_limit_per_second > 0)
```
ganmacs:
```suggestion
                    limit_bytes_per_second_reached = (number_bytes_read >= @watcher.read_bytes_limit_per_second && @watcher.read_bytes_limit_per_second > 0)
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -747,16 +785,37 @@ def on_notify
...
+                    @watcher.log.debug("reading file: #{ @watcher.path}")
+                    if @lines.size >= @watcher.read_lines_limit or limit_bytes_per_second_reached 
                       # not to use too much memory in case the file is very large
                       read_more = true
+
+                      if limit_bytes_per_second_reached 
+                        # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
+                        time_spent_reading = Time.new - start_reading 
+                        @watcher.log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
+                        if (time_spent_reading < 1)
```
ganmacs:
```suggestion
                        if time_spent_reading < 1
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -302,31 +318,49 @@ def setup_watcher(path, pe)
...
+              end
+            end
 
-        begin
-          tw = setup_watcher(path, pe)
-        rescue WatcherSetupError => e
-          log.warn "Skip #{path} because unexpected setup error happens: #{e}"
-          next
+            begin
+              tw = setup_watcher(path, pe)
```
ganmacs: It can be a race condition. before passing `pe` to setup_watcher, L334 should be called. but the current code does not ensure it.

lib/fluent/plugin/in_tail.rb
```diff
@@ -302,31 +318,49 @@ def setup_watcher(path, pe)
...
+          log.debug "Check Thread #{path} #{@threads[path].status}"
+          if @threads[path].status != "sleep" and  @threads[path].status != "run"
+            log.debug "Stopping Thread #{path} #{@threads[path].status}"
+            @threads[path].exit
+            @threads.delete(path)
           end
         end
+        if @threads[path].nil?
+          log.debug "Add Thread #{path}"
+          @threads[path] = Thread.new(path) do |path|
```
ganmacs: why did you change these codes to run on new thread?

lib/fluent/plugin/in_tail.rb
```diff
@@ -747,16 +785,37 @@ def on_notify
...
 
               if !io.nil? && @lines.empty?
                 begin
                   while true
-                    @fifo << io.readpartial(8192, @iobuf)
+                    @fifo << io.readpartial(bytes_to_read, @iobuf)
                     @fifo.read_lines(@lines)
-                    if @lines.size >= @watcher.read_lines_limit
+
+                    number_bytes_read += bytes_to_read
```
ganmacs: IO#readpartial does not alway read `bytes_to_read` bytes. Is this code ok?

rewiko:
> test fails. so could you fix the test first? 

Hello, 
I'm definitely add more test and fix those failing, I've created the PR to be able to discuss in term of design and making sure you would be interested by this kind of feature.

I'm gonna spend more time to add more tests and review your comments. 
Thanks

ganmacs:
Ok. I like this feature :)
but, There are some considering points.

Creating threads per file is not acceptable. Because if there are 1000 files to monitor, it creates 1000 threads. https://github.com/fluent/fluentd/pull/2702/files#diff-1da710c9dcc8d0fc57996df7a9d39695R331
This patch restricts the bytes size per file, right? I think that restring all data(not per file) could be better for the situation like "to avoid network saturation and make it easier to calculate the max throughput per node".

rewiko:
Hi @ganmacs, 

> Creating threads per file is not acceptable. Because if there are 1000 files to monitor, it creates 1000 threads. 

That's was one of my worries in term of performance, I've been testing with 200 files and a high number of bytes per file, the number of thread hasn't been an issue but I agree having one thread per file is node ideal, I was thinking about using a thread pool but that would need a bigger re-architecture and I'm not sure the inotify will work as expect. 

> I think that restring all data(not per file) could be better for the situation like "to avoid network saturation and make it easier to calculate the max throughput per node". 

I agree, however we are more interested to throttle log in a Kubernetes environment per container which means per file, to do not affect container which are sending at a decent rate.

 In the [commit](https://github.com/fluent/fluentd/pull/2702/commits/2ce05b1aa46737409e95e9d9725386fd1e385804), I've been implementing the log throttling per file but this implementation work nicely only with you use the timer because you it will stop reading and if it did not get notified again then some logs might not be read. 
Maybe I could reuse that concept and more breaking I can push the function with a sleep and a notify which will be consume by a thread pool. In that case only the file throttled will create thread.

repeatedly:
cosmo0920 is now working on this feature on #3185 with newer in_tail code.
Thanks for the idea!

PR #2706, opened 2019-11-25
Expose Fluentd version in config REST endpoints

### LINKED ISSUE #2705: Retrieving the Fluentd version remotely
Hi there!

As mentioned [here](https://docs.fluentd.org/deployment/monitoring-rest-api#datadog-dd-agent-integration), Datadog has a native integration with Fluentd (docs [here](https://docs.datadoghq.com/integrations/fluentd/)).

As part of evolving this integration, we'd like to forward the Fluentd version to Datadog so that users can view this information.

The `fluentd` CLI has a `--version` flag (although [not documented](https://docs.fluentd.org/deployment/command-line-option) in the CLI reference), which is enough for cases when Fluentd is running on the same host than the Datadog Agent.

However, if Fluentd is not available locally, then this method cannot be used.

For this reason, I'm popping by to discuss exposing the Fluentd version as part of the monitoring API used by Datadog, Prometheus, and probably other systems.

Add a `version` key to the `/config.json` endpoint, and by symmetry to the `/config` endpoint.

For the JSON endpoint, this would result in:

```js
{
  "pid": 123,
  "ppid": 1,
  "version": "1.8.0.rc2",
  // ...
}
```

I will submit a PR shortly with a working implementation.

I must say I'm not *too* familiar with how Fluentd is used in the wild. So my main question would be: can it actually be the case that a user has Fluentd monitoring enabled on another machine than the one where Fluentd is running?

**Additional context**

See https://github.com/DataDog/integrations-core/pull/5057

florimondmanca:
**Which issue(s) this PR fixes**: Fixes #2705 

**What this PR does / why we need it**: add a `version` key to the response returned by `/config.json` and `/config` endpoints from the REST monitoring APIs, so that monitoring systems can retrieve what version of Fluentd is running via a remote HTTP call (i.e. without having to run `fluentd --version` on the host machine).

**Docs Changes**: I haven't found existing documentation on the config endpoints, but if there are I'll gladly update them with this change.

**Release Note**: /

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -64,7 +64,8 @@ module Fluent::Plugin
       def config_ltsv(_req)
         obj = {
           'pid' => Process.pid,
-          'ppid' => Process.ppid
+          'ppid' => Process.ppid,
+          'version' => Fluent::VERSION,
         }.merge(@agent.fluentd_opts)
 
         render_ltsv([obj])
@@ -73,7 +74,8 @@ module Fluent::Plugin
       def config_json(req)
         obj = {
           'pid' => Process.pid,
-          'ppid' => Process.ppid
+          'ppid' => Process.ppid,
+          'version' => Fluent::VERSION,
         }.merge(@agent.fluentd_opts)
         opts = build_option(req)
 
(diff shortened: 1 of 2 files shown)


ganmacs:
Thanks!

PR #2708, opened 2019-11-27
inline_config is replaced with the input value when user specify `-`

ganmacs:
no

Current implementation calling `STDIN.read` in supervisors and workers.
it causes blocking in worker side.

**Docs Changes**:

no

**Release Note**: 

same as the title .


ganmacs:
could you review this patch? @cosmo0920 @ashie

ganmacs:
True. I think so. 
We can delete this confusing feature because fluentd document doesn't mention this feature (https://docs.fluentd.org/deployment/command-line-option#i-inline-config). But for now, I just fixed this feature's bug. And I created an issue about this. this should be discussed in it. https://github.com/fluent/fluentd/issues/new?template=feature_request.md

PR #2709, opened 2019-11-27
Remove unnecessary config reload

ganmacs:
ref https://github.com/fluent/fluentd/issues/2624

Current implementation loads system_config parameters(rpc_endpoint,
enable_get_dump, counter_server) in Supervisor.load_config
which is called at ServerEngine creation time.
And it can also reload config dynamically via serverengine's
mechanism(sending signal USR2 to fluentd).

These parameters can be reloadable but it doesn't make any changes
to fluentd itself. fluentd document also doesn't mention about signal USR2.
so it's no need to load in Supervisor.load_config.
it's enough to pass the value at start time.

**Docs Changes**:

no need

**Release Note**: 

no need


ganmacs:
@ashie @cosmo0920 could you review this patch?

cosmo0920:
nits: Which issue will be fixed by this PR?
I didn't find the concrete issue number on the pull request's description:

> **Which issue(s) this PR fixes:**
ref https://github.com/fluent/fluentd/issues

ganmacs:
> nits: Which issue will be fixed by this PR?

Good catch… this is related to safe config reloading mechanism.

PR #2716, opened 2019-12-09
Reload new config without restarting process

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2624

This change make fluentd be able to reload new config without restarting process with SIGUSR2 signal. 

it's very lighter and safer than existing reloading feature. but it has 2 limitations.

1.  A change to system_config is ignored because it needs to restart(kill/spawn) process. 
2. All plugins must not use class variable when restarting.

**Docs Changes**:

Add USR2 description.

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -157,14 +157,7 @@ module Fluent
         raise
       end
 
-      unless @log_event_verbose
-        $log.enable_event(false)
-        @fluent_log_event_router.graceful_stop
-      end
-      $log.info "shutting down fluentd worker", worker: worker_id
-      shutdown
-
-      @fluent_log_event_router.stop
+      stop_phase(@root_agent)
     end
 
     # @param conf [Fluent::Config]
@@ -190,17 +183,9 @@ module Fluent
         return
       end
 
-      # Stop phaze
-      unless @log_event_verbose
-        $log.enable_event(false)
-        @fluent_log_event_router.graceful_stop
-      end
-      $log.info 'shutting down fluentd worker', worker: worker_id
-      old_agent.shutdown      # Stop first but we can still accept sockets, thanks to serverengine SocketManager,
+      stop_phase(old_agent)
 
-      @fluent_log_event_router.stop
-
-      # Restart phaze
+      # Restart phase
       new_fleunt_log_event_router = FluentLogEventRouter.build(new_agent)
       if new_fleunt_log_event_router.emittable?
         $log.enable_event(true)
@@ -233,12 +218,20 @@ module Fluent
     end
 
     private
-    def start
-      @root_agent.start
+
+    def stop_phase(root_agent)
+      unless @log_event_verbose
+        $log.enable_event(false)
+        @fluent_log_event_router.graceful_stop
+      end
+      $log.info 'shutting down fluentd worker', worker: worker_id
+      root_agent.shutdown
+
+      @fluent_log_event_router.stop
     end
 
-    def shutdown
-      @root_agent.shutdown
+    def start
+      @root_agent.start
     end
   end
 
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -23,6 +23,7 @@ require 'fluent/time'
 require 'fluent/system_config'
 require 'fluent/plugin'
 require 'fluent/fluent_log_event_router'
+require 'fluent/static_config_analysis'
 
 module Fluent
   class EngineClass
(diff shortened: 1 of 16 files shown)


lib/fluent/engine.rb
```diff
@@ -157,14 +158,54 @@ def run
...
+    # @param supervisor [Bool]
+    # @reutrn nil
+    def reload_config(conf, supervisor: false)
+      # configure first to reduce down time while restarting
+      new_agent = RootAgent.new(log: log, system_config: @system_config)
+      ret = Fluent::StaticConfigAnalysis.call(conf, workers: system_config.workers)
+
+      ret.all_plugins.each do |plugin|
+        if plugin.respond_to?(:reloadable_plugin?) && !plugin.reloadable_plugin?
+          raise Fluent::ConfigError, "Unreloadable plugin: #{plugin.class}"
```
cosmo0920: This code seems good but users doesn't always concern about their using plugin implemented class.
Can we output `plugin_name` in Fluentd configuration(`@type plugin_name`)?

lib/fluent/engine.rb
```diff
@@ -157,14 +158,54 @@ def run
...
+      end
+
+      stop_phase(old_agent)
+
+      # Restart phase
+      new_fleunt_log_event_router = FluentLogEventRouter.build(new_agent)
+      if new_fleunt_log_event_router.emittable?
+        $log.enable_event(true)
+      end
+      @fluent_log_event_router = new_fleunt_log_event_router
```
cosmo0920: nits: Why not extract here as `restart_phase` method?
I'm slightly a bit confused here due to method implementation granularity glitch.

lib/fluent/engine.rb
ganmacs: fixed https://github.com/fluent/fluentd/pull/2716/commits/a78f7de7f4791faf97010be9c378f67d772b27b5

lib/fluent/engine.rb
ganmacs: fixed https://github.com/fluent/fluentd/pull/2716/commits/37f64c6ff2b4c3d0d87107cc7e59a43bc87daba1

ganmacs:
@repeatedly ?

lib/fluent/static_config_analysis.rb
```diff
@@ -0,0 +1,194 @@
...
+require 'fluent/config'
+require 'fluent/plugin'
+
+module Fluent
+  # Static Analysis means analysing all plugins and Fluent::Element without invokeing Plugin#configure
+  class StaticConfigAnalysis
+    module Elem
+      Input = Struct.new(:plugin, :config)
+      Output = Struct.new(:plugin, :config)
+      Format = Struct.new(:plugin, :config)
```
repeatedly: Why `Format`, not `Filter`?

test/test_config.rb
```diff
@@ -175,5 +175,27 @@ def create_warn_dummy_logger
     logger = ServerEngine::DaemonLogger.new(logdev, dl_opts)
     $log = Fluent::Log.new(logger)
   end
-end
 
+  sub_test_case '.build'do
```
repeatedly: Need space before `do`

lib/fluent/plugin/buf_file.rb
```diff
@@ -43,19 +44,20 @@ class FileBuffer < Fluent::Plugin::Buffer
...
         @multi_workers_available = false
         @additional_resume_path = nil
         @buffer_path = nil
+        @variable_store = nil
       end
 
       def configure(conf)
         super
 
+        @variable_store = Fluent::VariableStore.fetch_or_build(:buf_file_single)
```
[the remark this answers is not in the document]
ganmacs: fixed https://github.com/fluent/fluentd/pull/2716/commits/500b0254587b85a44c8af932642b7b094fd3a56d

test/test_config.rb
ganmacs: fixed https://github.com/fluent/fluentd/pull/2716/commits/500b0254587b85a44c8af932642b7b094fd3a56d

lib/fluent/static_config_analysis.rb
ganmacs: Good catch... https://github.com/fluent/fluentd/pull/2716/commits/60440d5fe931e3964b4bf9c05bb8ea8a70cfee75

repeatedly:
I noticed we need to change HTTP RPC for reload.
Should we change this to use USR2?

https://github.com/fluent/fluentd/blob/c8097882f6485e15da8b396c18f1b65a97c570e7/lib/fluent/supervisor.rb#L106

cosmo0920:
> Should we change this to use USR2 or add new path?

I believe that we should add a new path for reloading config.
Windows does not have `USR2` signal:
* https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2019

ganmacs:
> I noticed we need to update HTTP RPC for this feature.
Should we change this to use USR2 or add new path?

Good catch! I forgot about it. I added new endpoint `/api/config.gracefulReload`(https://github.com/fluent/fluentd/pull/2716/commits/d8b8bc6ad4aec438eba96a695a59266be03bcc34) for compatibility.

> Windows does not have USR2 signal:

I think this feature can't support windows like `USR1`  because a supervisor needs to send `USR2` signal to workers to do this. https://github.com/fluent/fluentd/pull/2716/files#diff-dbe0e1ec4079138e48ca6a4d7c7248f9R215
We need to introduce another way to communicate between supervisor and workers without using a signal to be able to use this feature in windows.

repeatedly:
I will release v1.9.1.rc1 soon.

PR #2719, opened 2019-12-10
format in_tail path with the specified timezone

tatsu-yam:
None

If the file path which in_tail plugin reads includes strftime format, it is determined by system timezone.
I want to determine this path in the specified timezone for each tail plugin object.
For example, a fluentd process reads a UTC file path ( `/path/to/data/test_utc/20191210_04.json` ) and a JST file path ( `/path/to/data/test_jst/20191210_13.json` ) in one conf.

```
$ TZ='Asia/Tokyo' ./bin/fluentd -c test.conf
2019-12-10 13:48:19 +0900 [info]: parsing config file is succeeded path="test.conf"
2019-12-10 13:48:19 +0900 [info]: using configuration file: <ROOT>
  <source>
    @type tail
    path "/path/to/data/test_utc/%Y%m%d_%H.json"
    pos_file "/path/to/pos/test_utc.pos"
    tag "test_utc"
    timezone "+00"
    <parse>
      @type "json"
    </parse>
  </source>
  <source>
    @type tail
    path "/path/to/data/test_jst/%Y%m%d_%H.json"
    pos_file "/path/to/pos/test_jst.pos"
    tag "test_jst"
    <parse>
      @type "json"
    </parse>
  </source>
  <match **>
    @type stdout
  </match>
</ROOT>
2019-12-10 13:48:19 +0900 [info]: starting fluentd-1.7.4 pid=64544 ruby="2.5.5"
...
2019-12-10 13:48:20 +0900 [info]: #0 following tail of /path/to/data/test_jst/20191210_13.json
2019-12-10 13:48:20 +0900 [info]: #0 following tail of /path/to/data/test_utc/20191210_04.json
```
I wrote a simple path, but is it correct to use `Fluent::Timezone` class?

**Docs Changes**:
None

**Release Note**: 
None

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -96,6 +96,8 @@ module Fluent::Plugin
     config_param :skip_refresh_on_startup, :bool, default: false
     desc 'Ignore repeated permission error logs'
     config_param :ignore_repeated_permission_error, :bool, default: false
+    desc 'Format path with the specified timezone'
+    config_param :path_timezone, :string, default: nil
 
     config_section :parse, required: false, multi: true, init: true, param_name: :parser_configs do
       config_argument :usage, :string, default: 'in_tail_parser'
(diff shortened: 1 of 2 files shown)


ganmacs:
It's reasonable for me to use `Fluent::Timezone` for such purpose.

 @repeatedly ?

repeatedly:
Yes. `Fluent::Timezone` is better to handle timezone task.

tatsu-yam:
Thank you for reply.
I'll continue as it is and remove "Draft" after adding the following:
- since option is only used to determine the path, change it from `timezone` to `path_timezone` or is there a better name? 
- add the same implementation to `exclude_path`
- add test

If there are any problems, please point them out.

ganmacs:
> since option is only used to determine the path, change it from timezone to path_timezone or is there a better name?

Sounds good. I've thought same thing. path_timezone is better.

Could you follow the DCO https://github.com/fluent/fluentd/pull/2719/checks?check_run_id=341323383 ?(I think you know. it's just confirming)

tatsu-yam:
> Could you follow the DCO https://github.com/fluent/fluentd/pull/2719/checks?check_run_id=341323383 ?(I think you know. it's just confirming)

Sure, I'll squash commits and sign off at that time.

ganmacs:
Thank you!

PR #2720, opened 2019-12-10
Accept localtime xor utc

ganmacs:
None

I think `localtime = true && utc =false` and `locatime= false && utc = true` should be valid. 

In the current implementation, stdout format plugin raises the error like below with `utc true` in format section.

```
2019-12-10 15:06:02 +0900 [error]: config error file="example/debug/nginx.conf" error_class=Fluent::ConfigError error="both of utc and localtime are specified, use only one of them"
```

example conf
```
<match *>
  @type stdout

  <format>
    @type stdout
    utc true
  </format>
</match>
```

This is because of calling [TimeParameters#configure](https://github.com/fluent/fluentd/blob/master/lib/fluent/time.rb#L153) twice.
but it seems to be inevitable since the stdout format plugin has sub_formater object internally.

**Docs Changes**:

No need

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/plugin_helper/compat_parameters.rb
+++ b/lib/fluent/plugin_helper/compat_parameters.rb
@@ -194,7 +194,9 @@ module Fluent
           hash['time_type'] = 'unixtime'
         end
         if conf.has_key?('localtime') || conf.has_key?('utc')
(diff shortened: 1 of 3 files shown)


repeatedly:
Looks good.

Issue #2718, opened 2019-12-09
Unable to send correct timestamp to ES


flowl:
Having times like those in the logs: `02/Dec/2019:12:59:56 +0100`
I am not able to turn it into a proper elastic search format.

According to [the docs](https://docs.fluentd.org/configuration/format-section), `time_format` is the input format and `types` with the declaration below is the paramter which fluentd converts to (output).

    # Apache log DateTime
    time_format %d/%b/%Y:%H:%M:%S %z

    # ES DateTime
    types @timestamp:time:%Y-%m-%dT%H:%M:%S.%L%z

Which produces wrong results in ES: 

    @timestamp | Jan 19, 1970 @ 06:34:24.410

Whereas this works:

    index_name fluentd.${tag}.%Y%m%d

The index are in the past (time from the logs), which is as expected.

Full config, using fluentd 1.7, ES7.x, elastic plugin and kibana in Docker

```
<source>
  @type tail
  path /logs/error.log
  pos_file /logs/error.log.pos
  tag apache.error
  <parse>
    @type regexp
    # IPv4 only!
    expression ^\[(?<@timestamp>.*?)\] \[(?<level>.*?)\] \[pid (?<pid>\d*?)\](?: \[client (?<client>.*?)\:.*?\])? (?<message>.*)$
    time_key @timestamp
    keep_time_key true
    time_format %a %b %d %H:%M:%S.%N %Y
    types message:string, pid:integer, @timestamp:time:%Y-%m-%dT%H:%M:%S.%L%z
  </parse>
</source>

<match apache.**>
  @type elasticsearch
  host elasticsearch
  port 9200
  user elastic
  password changeme

  # This works:
  index_name fluentd.${tag}.%Y%m%d

  include_tag_key true
  tag_key @log_name

   <buffer tag, time>
     @type memory
     timekey 3600
   </buffer>
</match>
```

starting fluentd-1.7.4 pid=6 ruby="2.6.5"
spawn command to main:  cmdline=["/usr/local/bin/ruby", "-Eascii-8bit:ascii-8bit", "/usr/local/bundle/bin/fluentd", "-c", "/fluentd/etc/fluent.conf", "-p", "/fluentd/plugins", "--under-supervisor"]
gem 'fluent-plugin-elasticsearch' version '3.7.1'
gem 'fluent-plugin-multi-format-parser' version '1.0.0'
gem 'fluentd' version '1.7.4'

flowl:
[...]

PR #2721, opened 2019-12-10
Make time_format configurable in stdout format

ganmacs:

ref https://github.com/fluent/fluentd/issues/2718#issuecomment-563372355

stdout format does not see the time_format even if user can set the value. 

**Docs Changes**:

no need(already mentioned but it doesn't work)

**Release Note**: 

same as the title.


repeatedly:
Ah, good catch!

PR #2730, opened 2019-12-12
Fix bug that log level in system config doesn't affect

ganmacs:

follow up https://github.com/fluent/fluentd/pull/2674

Fix bug that log level in system config doesn't affect

**Docs Changes**:

need to change [log_level](https://docs.fluentd.org/deployment/system-config#log_level) default value https://github.com/fluent/fluentd-docs-gitbook/pull/122

**Release Note**: 

same as the title


repeatedly:
Good catch!

PR #2732, opened 2019-12-12
LoggerInitializer updated to create directory when it doesn't exists

ganmacs:

https://github.com/fluent/fluentd/pull/1130

https://github.com/fluent/fluentd/pull/1130 is very old and the author hasn't replied. so renew it 

**Docs Changes**:

none

**Release Note**: 

same as the title.


repeatedly:
Looks good.

PR #2733, opened 2019-12-12
Fix regex and output warning log correctly

ganmacs:
ref: https://github.com/fluent/fluentd/pull/2523

Fixed regext to match with https://github.com/fluent/fluentd/blob/7482802d75ba7b61a2997b15a02c45ff6a376e61/lib/fluent/plugin/output.rb#L753.
Be able to output a warning log even if  chunk_key is passed.

**Docs Changes**:

none

**Release Note**: 

same as the title


repeatedly:
Looks good.

Issue #2712, opened 2019-12-03
stage size computation in buffer plugin is not thread safe


hlakshmi:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**
The computation of stage_size in buffer plugin is not thread safe. This is the line which can cause synchronization issues because the after the value `staged_bytesize` is computed, a chunk could be enqueued by the enqueue thread causing incorrect computation of `stage_size`

**To Reproduce**

**Expected behavior**

**Your Environment**

- Fluentd or td-agent version: 1.4.1, 1.6.3
- Operating system: CentOS 7
- Kernel version: 3.10.0-1062.4.3.el7.x86_64

If you hit the problem with older fluentd version, try latest version first.

**Your Configuration**

```
<system>
# process_name fluentd-data
# log_level info
  log_level debug
# log_level trace
  suppress_repeated_stacktrace
</system>

#
# datastore pipeline
#
<source>
  @type monitor_agent
  bind 0.0.0.0
  port 25220
</source>

<source>
  @type forward
  port 25224
</source>

<match fluent.**>
# @type stdout
  @type null
</match>

#
# references.*
# flows.*
#
<match *.*>
  @type file
  path /xx/xxx/../${tag[1]}/${tag[0]}/%Y/%m/%d/%H
  append false
  compress gzip
  <buffer tag,time>
    @type file
    path <path-to-buffer>
    flush_mode interval
    flush_interval 10m
    flush_at_shutdown false
    timekey 86400
    timekey_use_utc
    chunk_limit_size 10mb
    chunk_limit_records 100000
    flush_thread_count 5
  </buffer>
  <format>
    @type json
  </format>
</match>
```

**Your Error Log**
As the computation of stage_size is not thread safe, it results in `BufferOverflowError` after a while even though the actual buffer directory has not reached the limit.
```
3602  2019-11-09 14:10:29 +0000 [warn]: failed to write data into buffer by buffer overflow action=:throw_exception
    1 2019-11-09 14:10:29 +0000 [warn]: emit transaction failed: error_class=Fluent::Plugin::Buffer::BufferOverflowError error="buffer space has too many data" location="/var/external/lib/ruby/gems/2.4.0/gems/fluentd-1.4.1/lib/fluent/plugin/bu
[...]

PR #2734, opened 2019-12-12
Make stage_size & stage computation thread safe

hlakshmi:
https://github.com/fluent/fluentd/issues/2712
Fixes #

When writing chunks to buffer, the `stage_size` computation can be thread-unsafe as mentioned in the bug which can result in `BufferOverflowErrors` even though the buffer is not actually full. This happened on a live long running fluentd. The method `buffer.write` is supposed to write to a chunk atmost once and hence `stage_size` should be added for a chunk only once but due to `write_step_by_step` the block which updates the value of `staged_bytesize` may be called multiple times and the chunk can be rollbacked but the value of `staged_bytesize` is not reverted causing the `stage_size` to be more than the actual value.

This PR addresses this issue. 

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -283,8 +283,13 @@ module Fluent
               chunk.mon_enter # add lock to prevent to be committed/rollbacked from other threads
               operated_chunks << chunk
               if chunk.staged?
-                staged_bytesizes_by_chunk[chunk] = 0 if staged_bytesizes_by_chunk[chunk].nil?
-                staged_bytesizes_by_chunk[chunk] += adding_bytesize
+                #
+                # https://github.com/fluent/fluentd/issues/2712
+                # write_once is supposed to write to a chunk only once
+                # but this block **may** run multiple times from write_step_by_step and previous write may be rollbacked
+                # So we should be counting the stage_size only for the last successful write
+                #
+                staged_bytesizes_by_chunk[chunk] = adding_bytesize
               elsif chunk.unstaged?
                 unstaged_chunks[metadata] ||= []
                 unstaged_chunks[metadata] << chunk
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks!. We will check issue and patch later.

hlakshmi:
@repeatedly Can we get this merged into the upcoming release if it looks good as this is causing a critical issue of data loss. We have tested this in large scale environment and the fix seems to work. Thanks!

lib/fluent/plugin/buffer.rb
```diff
@@ -283,7 +283,13 @@ def write(metadata_and_data, format: nil, size: nil, enqueue: false)
...
               operated_chunks << chunk
               if chunk.staged?
-                staged_bytesize += adding_bytesize
+                #
+                # https://github.com/fluent/fluentd/issues/2712
+                # write_once is supposed to write to a chunk only once
+                # but this block **may** run multiple times from write_step_by_step and previous write may be rollbacked
+                # So we should be counting the stage_size only for the last successful write
+                #
+                staged_bytesizes_by_chunk[chunk] = adding_bytesize
```
ganmacs: I think the following situation can happen. what do you think of this?

1. `staged_bytesizes_by_chunk` stores chunk1 
2. rollback happens in write_step_by_step (ShouldRetry raises)
3. another thread euqueues chunk1 before this thread enters into https://github.com/fluent/fluentd/blob/065751f5c370b70ba8ab5e12aeced0c4a6138eb8/lib/fluent/plugin/buffer.rb#L673
4. create new chunk(chunk2) in https://github.com/fluent/fluentd/blob/065751f5c370b70ba8ab5e12aeced0c4a6138eb8/lib/fluent/plugin/buffer.rb#L663
5. `staged_bytesizes_by_chunk` stores chunk2 (still remtains chunk1 in `staged_bytesizes_by_chunk`)

hlakshmi: @ganmacs Steps (3) can't happen because `enqueue_chunk` can't get a lock and waits until the lock is released by the thread which is writing to the chunk.

The lock is acquired here: https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/buffer.rb#L279
And released here:
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/buffer.rb#L319

So `enqueue_chunk` can happen only after the commit is done. 
Hope this helps.

ganmacs: Make sense. You're right. other thread can get the chunk but can't enter https://github.com/fluent/fluentd/blob/065751f5c370b70ba8ab5e12aeced0c4a6138eb8/lib/fluent/plugin/buffer.rb#L674 .

repeatedly:
Thanks!

PR #2739, opened 2019-12-17
This class is no more used

ganmacs:
ref https://github.com/fluent/fluentd/issues/2624

introduce at
https://github.com/fluent/fluentd/commit/ed93293a0360d839bb7da0183e0d4a06e19cbc1d

Label class's error_collector can not be not nil.
because error_collector is assigned a value by root_agent. and Lable
inherits Agent not RootAgent.
https://github.com/fluent/fluentd/blob/fcef949ce40472547fde295ddd2cfe297e1eddd6/lib/fluent/label.rb

**Docs Changes**:

no

**Release Note**: 

no


ganmacs:
Thank you!

Issue #2711, opened 2019-11-30
Can we delete confusing feature about inline-config command line option?


ganmacs:
**Is your feature request related to a problem? Please describe.**

https://github.com/fluent/fluentd/pull/2708

> I hadn't use this this feature until now.
> I got confused about this feature.

> I expected that inline config is freshly and entirely used for Fluentd configuration.
> However, currently, inline config is appended loaded Fluentd configuration.

> This behavior got confuse for me.

**Describe the solution you'd like**

Just deleting the feature which reads an additional config from STDIN when passing the value `-` as inline-config command-line option to fluentd.

**Describe alternatives you've considered**

**Additional context**

Are there any backgrounds about this feature? @repeatedly

PR #2746, opened 2019-12-23
Add deprecated log for inline_config

ganmacs:
ref: https://github.com/fluent/fluentd/issues/2711

add log when inline_config is `-`. it's good to remove this feature in v1.20.0.

**Docs Changes**:

no need

**Release Note**: 

no need


repeatedly:
> it's good to remove this feature in v1.20.0.

v2?

ganmacs:
you're right. it's v2.

PR #2748, opened 2019-12-25
Prevent garbage from coming out of the LTSV parser

threewordphrase:
I have not opened one yet.

The LTSV parser does not check whether the delimiter is in the pair before it tries to parse it. With `delimiter_pattern /\s+/` and `label_delimiter =`,If I give it the value:
```
"Sylvanus_007               : ok=45   changed=6"
```
it parses it like this:
```
{ "Sylvanus_007":null,":":null,"ok":"45","changed":"6" }
```
Both `Sylvanus_007: null` and `":": null` were never specified by key=value. This PR prevents this drastically wrong behavior and parses it like this:
```
{ "ok":"45","changed":"6" }
```

**Docs Changes**:
None

**Release Note**: 
None

### DIFF
--- a/lib/fluent/plugin/parser_ltsv.rb
+++ b/lib/fluent/plugin/parser_ltsv.rb
@@ -38,8 +38,10 @@ module Fluent
       def parse(text)
         r = {}
         text.split(@delimiter).each do |pair|
-          key, value = pair.split(@label_delimiter, 2)
-          r[key] = value
+          if pair.include? @label_delimiter
+            key, value = pair.split(@label_delimiter, 2)
+            r[key] = value
+          end
         end
         time, record = convert_values(parse_time(r), r)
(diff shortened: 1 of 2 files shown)


threewordphrase:
Not sure what the CI failure is about - I ran that test module per the README, but locally for me `bundle exec rake test TEST=test/plugin/test_in_tail.rb` passes fine. I cloned the repo fresh just to work on this PR. Ideas?

repeatedly:
Thanks for the patch!

> Not sure what the CI failure is about 

This random failure is caused by travis-ci performance/machine. No problem.

threewordphrase:
Oh wow - I was not expecting this to be merged to master within 24 hours of starting on it. Thank you, fluentd maintainers! When will this version be available on rubygems?

PR #2750, opened 2019-12-26
Drop ruby 2.1/2.2/2.3 support

### LINKED ISSUE #2380: Drop older ruby 2.1/2.2/2.3 support

repeatedly:

Fixes #2380 
See also: https://www.fluentd.org/blog/drop-schedule-announcement-in-2019

This is for fluentd v1.9.0 which will be released in 2020

Drop older rubies and related gems.

**Docs Changes**:
Need to update supported version

**Release Note**: 
Same as title.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,7 @@ source 'https://rubygems.org/'
 
 gemspec
 
-# https://github.com/socketry/async-io/blob/v1.23.1/async-io.gemspec#L21
-if Gem::Version.create(RUBY_VERSION) >= Gem::Version.create('2.3.0')
-  gem 'async-http', '~> 0.42'
-end
+gem 'async-http', '~> 0.42'
 
 local_gemfile = File.join(File.dirname(__FILE__), "Gemfile.local")
 if File.exist?(local_gemfile)
--- a/lib/fluent/plugin_helper/http_server.rb
+++ b/lib/fluent/plugin_helper/http_server.rb
@@ -15,7 +15,6 @@
 #
 
 begin
-  # raise if RUBY_VERSION < 2.3.x. see Gemfile
   require 'async'
   require 'fluent/plugin_helper/http_server/server'
 rescue LoadError => _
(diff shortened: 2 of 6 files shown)


ganmacs:
ref https://github.com/fluent/fluentd/pull/2372

Gemfile
```diff
@@ -2,10 +2,7 @@ source 'https://rubygems.org/'
 
 gemspec
 
-# https://github.com/socketry/async-io/blob/v1.23.1/async-io.gemspec#L21
-if Gem::Version.create(RUBY_VERSION) >= Gem::Version.create('2.3.0')
-  gem 'async-http', '~> 0.42'
-end
+gem 'async-http', '~> 0.42'
```
ganmacs: this gem should be moved into fluentd.gemspec

lib/fluent/plugin_helper/http_server.rb
```diff
@@ -15,7 +15,6 @@
 #
 
 begin
-  # raise if RUBY_VERSION < 2.3.x. see Gemfile
```
ganmacs: `rescue` expression is also not necessary ( also Fluent::PluginHelper::HttpServer::Compat::Server is )

Gemfile
repeatedly: I think fluentd should work with minimal dependencies. So make async-http optional seems good for me. If we can replace cool.io with async gems, async-http should be default dependency.

ganmacs: If we can replace cool.io with async gems, async-http should be default dependency.

Make sense 👍

PR #2751, opened 2019-12-26
output: Show better message for different secondary type

repeatedly:

No issue in this repository.

Current message is bit misleading for users. Some output plugins work with primary.

**Docs Changes**:
No need for now.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -387,7 +387,7 @@ module Fluent
           @secondary.acts_as_secondary(self)
           @secondary.configure(secondary_conf)
           if (self.class != @secondary.class) && (@custom_format || @secondary.implement?(:custom_format))
-            log.warn "secondary type should be same with primary one", primary: self.class.to_s, secondary: @secondary.class.to_s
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -387,7 +387,7 @@ def configure(conf)
           @secondary.acts_as_secondary(self)
           @secondary.configure(secondary_conf)
           if (self.class != @secondary.class) && (@custom_format || @secondary.implement?(:custom_format))
-            log.warn "secondary type should be same with primary one", primary: self.class.to_s, secondary: @secondary.class.to_s
+            log.warn "Use different plugin for secondary. Check the plugin works with primary like secondary_file", primary: self.class.to_s, secondary: @secondary.class.to_s
```
ganmacs: I have two questions.

* Why this warning is needed? if it won't work, I think it should be error.
* How to check this? > Check the plugin works with primary like secondary_file

repeatedly: > Why this warning is needed? if it won't work, I think it should be error.
> How to check this? > Check the plugin works with primary like secondary_file

Hard to reject the plugin by fluentd side because it depends on implementation issue.
If users can't judge the plugin works as secondary, users should use primary plugin or secondary_file.

PR #2756, opened 2020-01-03
Use Ruby's builtin time parser.

jkohen:
This reduces CPU usage in micro and macro benchmarks of our GKE setup by 20-30%. I ran the macro benchmark with Ruby 2.6.3, and the micro benchmark with Ruby 2.5.7 and 2.6.3.

This replaces nurse/strptime, which is purported to be more efficient, but that no longer seems true on Ruby 2.5 and higher. I couldn't find any existing benchmarks, so I made my own micro and macro benchmarks: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/363

I see nurse/strptime hasn't gotten much love recently. E.g. https://github.com/nurse/strptime/issues/10

I could remove the dependency on strptime by replacing the usage of FORMATTER inside `inspect` with a Time.strftime call, but I don't know much about that impact.

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -16,6 +16,7 @@
 
 require 'time'
 require 'msgpack'
+require 'strptime'
 require 'fluent/timezone'
 require 'fluent/configurable'
 require 'fluent/config/error'
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -16,7 +16,6 @@
 
 require 'time'
 require 'msgpack'
-require 'strptime'
 require 'fluent/timezone'
 require 'fluent/configurable'
 require 'fluent/config/error'
@@ -231,18 +230,9 @@ module Fluent
                     else Time.now.localtime.utc_offset # utc
                     end
 
-      strptime = format && (Strptime.new(format) rescue nil)
-
       @parse = case
-               when format_with_timezone && strptime then ->(v){ Fluent::EventTime.from_time(strptime.exec(v)) }
                when format_with_timezone             then ->(v){ Fluent::EventTime.from_time(Time.strptime(v, format)) }
                when format == '%iso8601'             then ->(v){ Fluent::EventTime.from_time(Time.iso8601(v)) }
-               when strptime then
-                 if offset_diff.respond_to?(:call)
-                   ->(v) { t = strptime.exec(v); Fluent::EventTime.new(t.to_i + offset_diff.call(t), t.nsec) }
-                 else
-                   ->(v) { t = strptime.exec(v); Fluent::EventTime.new(t.to_i + offset_diff, t.nsec) }
-                 end
                when format   then
                  if offset_diff.respond_to?(:call)
                    ->(v){ t = Time.strptime(v, format); Fluent::EventTime.new(t.to_i + offset_diff.call(t), t.nsec) }


repeatedly:
This seems to depend on system performance.
I tested your code and strptime gem is still faster.

- Mac (2.9 GHz Intel Core i7)

```
% ruby --version
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]
% time ruby strptime-builtin.rb
ruby strptime-builtin.rb  7.08s user 0.08s system 98% cpu 7.253 total
% time ruby strptime-nurse.rb
ruby strptime-nurse.rb  4.39s user 0.07s system 97% cpu 4.569 total
```

- Arm server (c2.large.arm on packet.net)

```
# time ruby strptime-builtin.rb 

real	0m19.572s
user	0m16.247s
sys	0m3.313s
# time ruby strptime-nurse.rb 

real	0m16.395s
user	0m7.131s
sys	0m9.261s
```

strptime consumes system resource. So is this the problem on your environment?

jkohen:
> This seems to depend on system performance.
> I tested your code and strptime gem is still faster.
> 
> * Mac (2.9 GHz Intel Core i7)

Good point. I ran the micro and macro benchmarks on GCE (Google Compute Engine) VM instances. These run a lot of FluentDs in production and the change is very significant, so we'll probably want to keep it because it translates directly into savings for our users.

Are you open to switching back to the Ruby's implementation of `strptime`? How would you like to make the decision?

Some facts we've discovered:
- Performance varies by system.
- Ruby's implementation seems to be faster in the default machines behind GCE VMs.
- Nurse's implementation seems to be faster in Mac.
- Ruby's implementation is maintained by the community and the authoritative implementation.
- Nurse's implementation has one maintainer and is missing at least one feature: https://github.com/nurse/strptime/issues/10
- Ruby's implementation is (in my opinion) fairly readable Ruby code, which in my opinion makes it easier to maintain.
- Nurse's implementation is (in my opinion) a hard-to-read manually tuned FSM, which in my opinion makes it harder to maintain. And I'm more fluent in C than in Ruby.

> 
> ```
> % ruby --version
> ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]
> % time ruby strptime-builtin.rb
> ruby strptime-builtin.rb  7.08s user 0.08s system 98% cpu 7.253 total
> % time ruby strptime-nurse.rb
> ruby strptime-nurse.rb  4.39s user 0.07s system 97% cpu 4.569 total
> ```
> 
> * Arm server (c2.large.arm on packet.net)
> 
> ```
> # time ruby strptime-builtin.rb 
> 
> real	0m19.572s
> user	0m16.247s
> sys	0m3.313s
> # time ruby strptime-nurse.rb 
> 
> real	0m16.395s
> user	0m7.131s
> sys	0m9.261s
> ```
> 
> strptime consumes system resource. So is this the problem on your environment?

What kind of problem do you have in mind? The results that I posted for this benchmark look similar, just slightly different proportions of usage.

/cc @nurse

nurse:
I think you hit a performance problem of glibc's `mktime(3)`: https://github.com/ruby/ruby/commit/4d7a6d04b2c71aabb9d6e619f4405887806a5be8
You can confirm it with strace and checking read operation to `/etc/localtime`.
You can avoid the problem with setting TZ environment variable if my idea is correct.

Note that @repeatedly is also a maintainer of strptime.gem.

jkohen:
> I think you hit a performance problem of glibc's `mktime(3)`: [ruby/ruby@4d7a6d0](https://github.com/ruby/ruby/commit/4d7a6d04b2c71aabb9d6e619f4405887806a5be8)
> You can confirm it with strace and checking read operation to `/etc/localtime`.
> You can avoid the problem with setting TZ environment variable if my idea is correct.

@nurse Thanks for the pointer. Setting TZ makes both implementations faster, and now yours is faster in my benchmark. I confirm the same improvement with Ruby 2.7 without having to set `TZ`, but only for the Ruby strptime; your implementation still requires TZ to be set to avoid the system CPU usage. Is this something you intend to fix? At that point, we would be happy to switch back to it.

Alternatively, could FluentD set TZ safely?

> 
> Note that @repeatedly is also a maintainer of strptime.gem.

nurse:
Thank you for so quick confirmation!
The source of speed of strptime is byte code VM and fast pass (shortcut).
TZ is just a workaround and I'll implement a shortcut for strptime.

PR #2765, opened 2020-01-09
Try GitHub actions

ganmacs:
none

try github actions

**Docs Changes**:

none

**Release Note**: 

none

### DIFF
--- a/.github/workflows/run_test.yml
+++ b/.github/workflows/run_test.yml
@@ -1,0 +1,26 @@
+name: Run tests
+
+on: [push]
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


ganmacs:
It turned out that we can't replace Travis and appvoyor with GitHub actions now.
I might be going to try it when GitHub  actions become more mature.

ruby 2.7 + windows env does not work correctly https://github.com/ganmacs/fluentd/commit/8bdcb1c4a66b0727dc1d01e07c20ad269e8df986/checks?check_suite_id=393400732

repeatedly:
Yes. In addition, we need to consider non-amd64 platform.

PR #2767, opened 2020-01-09
Add multiline RFC5424 support

onet-git:
Fixes #

RFC5424 (https://tools.ietf.org/html/rfc5424) supports multi-line syslog messages. Specifically, all octets are supposed to be accepted rather than treating \n as a end of message marker, see page 8 and 9 of the RFC. Proper framing and unframing of messages should be handled by the transport layer. Note, support for multi-line log is an explicit and intentional change in RFC5424 compared to the legacy syslog message format.

More importantly the existing implementation will reject any FRC5424 message that ends in a \n. Which has resulted in all fluentd syslog input plugins (both core and 3rd party) silently truncating all log messages. Sometimes with chomp, sometimes with rstrip, and sometimes just dropping the last character. The last case has causes real world issues with mangled log messages because not all syslog implementations add a \n at the end of each message.

Note, this implementation calls chomp on the syslog message before adding it to the record. This will allow input plugins to stop truncating messages while maintaining compatibility with existing user visible behavior.

**Docs Changes**:

None

**Release Note**: 

Added support for multi-line RFC5424 messages.

### DIFF
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -27,8 +27,8 @@ module Fluent
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = /\A^(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?$\z/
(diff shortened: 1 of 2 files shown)


ganmacs:
Does this patch work with in_syslog which uses TCP protocol like the following config?
When in_syslog tcp protocol, it uses `\n` as the delimiter of each message.
https://github.com/fluent/fluentd/blob/afdd90780a12a0010775743bb24651b271e5b54a/lib/fluent/plugin/in_syslog.rb#L105
https://github.com/fluent/fluentd/blob/afdd90780a12a0010775743bb24651b271e5b54a/lib/fluent/plugin/in_syslog.rb#L182

```
<source>
  @type syslog


  tag test
  message_format auto
  with_priority true

  protocol_type tcp # here
</source>
```

onet-git:
> ```
> <source>
>   @type syslog
> 
> 
>   tag test
>   message_format auto
>   with_priority true
> 
>   protocol_type tcp # here
> </source>
> ```

It won't change the existing behavior, just enable new behavior.

To have multi-line syslog support your message delimiter must be excluded from the characters allowed in the message. The @delimiter flag is there so you can do something like use `\0` as the delimiter so that you can include `\n` in log messages. However, without the change I am proposing the syslog parser would fail to parse such messages as the regex will choke on any `\n` included in the message.

For multi-line syslog you actually want to use octet-counted encoding rather than delimited encoding. See https://www.rsyslog.com/doc/v8-stable/configuration/modules/omfwd.html under TCP_Framing for the historical background. Note, there is a off-by one bug in in_syslog.rb for octet-counted transport.

https://github.com/fluent/fluentd/blob/afdd90780a12a0010775743bb24651b271e5b54a/lib/fluent/plugin/in_syslog.rb#L184

The off-by one bug likely hasn't been noticed because the default config for many syslog implementations append a `\n` to every log message, even when using octet-counted encoding for transport. The correct behavior would be to replace the off-by-one with an explicit `chomp`.

onet-git:
See https://github.com/fluent/fluentd/pull/2771

ganmacs:
I see. Looks good. Could you follow the DCO https://github.com/fluent/fluentd/pull/2767/checks?check_run_id=382280745 ?

onet-git:
After updating to head on master, CI tests are failing on tests not changed in this pull. If there anything I can do to resolve this?

repeatedly:
It seems travis-ci issue. Tests on almost all versions are succeeded. So merged.

PR #2771, opened 2020-01-10
Fix off-by one for octet-counting code in in_sysylog.rb

onet-git:
Fixes #

There is an off-by-one counting issue in octet-transported syslog messages. In most case syslog transport clients will append a `\n` to the end of each line, hiding this bug. The bug gets triggered whenever a syslog client doesn't append a `\n`.

**Docs Changes**:

None

**Release Note**: 

Support octet-counted syslog transport when the last character in each message is not a \n.

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -181,12 +181,14 @@ module Fluent::Plugin
           if octet_count_frame
             while idx = buffer.index(delimiter, pos)
               num = Integer(buffer[pos..idx])
-              pos = idx + num
-              msg = buffer[idx + 1...pos]
-              if msg.size < num - 1
-                pos = pos - num - num.to_s.size
+              idx = idx + delimiter_size
+              msg = buffer[idx, num]
+              if msg.size != num
                 break
               end
+              pos = idx + msg.size
+              # syslog might append a \n as an extra frame delimiter
+              msg.chomp!
               message_handler(msg, conn)
             end
           else
--- a/test/plugin/test_in_syslog.rb
+++ b/test/plugin/test_in_syslog.rb
@@ -321,11 +321,11 @@ EOS
   end
 
   sub_test_case 'octet counting frame' do
-    def test_msg_size_with_tcp
+    def test_octet_counting_with_tcp
       d = create_driver([CONFIG, "<transport tcp> \n</transport>", 'frame_type octet_count'].join("\n"))
       tests = create_test_case
 
-      d.run(expect_emits: 2) do
+      d.run(expect_emits: 1) do
         tests.each {|test|
           TCPSocket.open('127.0.0.1', PORT) do |s|
             s.send(test['msg'], 0)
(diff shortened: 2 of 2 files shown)


lib/fluent/plugin/in_syslog.rb
```diff
@@ -181,12 +181,14 @@ def start_tcp_server(tls: false)
...
-              pos = idx + num
-              msg = buffer[idx + 1...pos]
-              if msg.size < num - 1
-                pos = pos - num - num.to_s.size
+              idx = idx + delimiter_size
+              msg = buffer[idx, num]
+              if msg.size != num
                 break
               end
+              pos = idx + msg.size
```
ganmacs: [nit] `msg.size` is  `num`

lib/fluent/plugin/in_syslog.rb
```diff
@@ -181,12 +181,14 @@ def start_tcp_server(tls: false)
...
+              # syslog might append a \n as an extra frame delimiter
+              msg.chomp!
```
ganmacs: How can we handle if the user adds `\n` at the end of msg intentionally? 
In the situation, it shouldn't trim `\n` from msg.

test/plugin/test_in_syslog.rb
```diff
@@ -321,11 +321,11 @@ def compare_test_result(events, tests, options = {})
   end
 
   sub_test_case 'octet counting frame' do
-    def test_msg_size_with_tcp
+    def test_octet_counting_with_tcp
       d = create_driver([CONFIG, "<transport tcp> \n</transport>", 'frame_type octet_count'].join("\n"))
       tests = create_test_case
 
-      d.run(expect_emits: 2) do
+      d.run(expect_emits: 1) do
```
ganmacs: Is there a reason you changed this line? I think it shouldn't be changed since two messages are sent.

ganmacs:
Sorry for the delay! I left some comments. 

BTW, Could you follow DCO? https://github.com/fluent/fluentd/pull/2771/checks?check_run_id=384067806

lib/fluent/plugin/in_syslog.rb
onet-git: I was limiting this change to preventing the removal of characters other than `\n`. I could add that as a configuration option to control if the chomp should be called.

onet-git: I looked into this more and currently in the default parser configuration chomping at the in_syslog level has no impact because parse_syslog strips off trailing `\n`.

However, that is only part of the story. Until https://github.com/fluent/fluentd/pull/2767 lands if RFC5424 is selected as the format for parser_syslog messages it will fail to parse unless in_syslog does a chomp on the incoming messages.

There also is a second issue, without this change octet counted transport always strips off the last character of a message (incorrectly presumed to be a delimiter) while the delimited transport does not strip off the delimiter! So if chomp is off by default then it will be a behavior change for octect counted transport. If chomp is on by default then it will be a behavior change for delimited transport. In light of that I would personally would keep chomp off by default as it because it will impact fewer users and this change is supposed to address that fact that octect counted transport is currently incorrect.

ganmacs:
@onet-git  friendly  ping

ganmacs:
done in #2942. thanks!

PR #2775, opened 2020-01-14
Support Ruby's Time class in msgpack serde

repeatedly:

No issue.

Recent msgpack-ruby can (de)serialize Time class.
Add `enable_msgpack_time_support` parameter to system config for this feature.
This is useful for multiple time fields.

See also: https://github.com/msgpack/msgpack-ruby#serializing-and-deserializing-time-instances

fluentd v2 will enable this feature by default.

**Docs Changes**:
Will add `enable_msgpack_time_support` to system config article after merged the patch.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/msgpack_factory.rb
+++ b/lib/fluent/msgpack_factory.rb
@@ -70,9 +70,15 @@ module Fluent
       factory.unpacker(*args)
     end
 
-    def self.init
+    def self.init(enable_time_support: false)
       factory = MessagePack::Factory.new
       factory.register_type(Fluent::EventTime::TYPE, Fluent::EventTime)
+      if enable_time_support
+        factory.register_type(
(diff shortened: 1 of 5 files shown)


lib/fluent/msgpack_factory.rb
```diff
@@ -70,9 +70,15 @@ def self.unpacker(*args)
       factory.unpacker(*args)
     end
 
-    def self.init
+    def self.init(enable_time_support: false)
```
cosmo0920: Should we register `MessagePack::Timestamp::TYPE` in `self.factory` as well? Or it needn't to be registered there? :thinking:

repeatedly: Ah, I forgot `self.factory` because it is used in only tests...

repeatedly:
Applied reviews.

PR #2780, opened 2020-01-17
Support @include/include directive for spaces included path. Fix #2485

### LINKED ISSUE #2485: include directive does not support Windows directory with spaces
On Windows, if using an include directive with spaces in the path, the only way to load the file is to use old DOS short path notation.

**To Reproduce**
- In Windows, create an external config file with any valid configuration in a path similar to "C:\Folder With Spaces\fluentd.conf"
- In the root config file (e.g. td-agent.conf), add an include directive similar to "@include C:/Folder With Spaces/fluentd.conf"
- Start the fluentdwinsvc per normal, and the logs (e.g. td-agent.log) will contain no information

The only variant of the above example path that will work is "C:/FOLDER~1/fluentd.conf".  

Note that URI encoding the path (i.e. using %20 for each space) allows fluentd to start correctly, but still does not load the external configuration file. 

**Expected behavior**
The include directive should support directories with spaces, or at least document how to escape them, etc.  Additionally, some better logging around this would have been extremely helpful.

**Your Environment**
- td-agent v3.4.1
- Windows 2012 R2 (recreated on other Windows too)

**Additional context**
Note that I believe I understand the issue by reviewing the eval_include function in lib/fluent/config/v1_parser.rb.  It appears that any path given in the include directive is first passed into URI.parse, which will fail if it has spaces.  If you encode the path, URI.parse will succeed, but then when the path is extracted out of the URI, it will have the form "/Folder%20With%20Spaces/fluentd.conf", and that will not work with Dir.glob.  Not sure exactly what fix should be used, but I find it interesting that all paths must go through URI.parse, when file paths would not need to.

cosmo0920:
Based on @repeatedly's patch:
https://github.com/fluent/fluentd/issues/2485#issuecomment-510105857

Fixes #2485 

Support @include/include directive for spaces included path.

**Docs Changes**:
Needed?
**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -147,8 +147,8 @@ module Fluent
       end
 
       def eval_include(attrs, elems, uri)
-        u = URI.parse(URI.encode_www_form_component(uri))
-        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == URI.encode_www_form_component(uri) # file path
+        u = URI.parse(uri.gsub(/ /, '+')) # replace space(s)(' ') with '+' to prevent invalid uri due to space(s).
+        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == uri.gsub(/ /, '+') # file path
           # When the Windows absolute path then u.scheme.length == 1
           # e.g. C:
           path = URI.decode_www_form_component(u.path)
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -147,11 +147,11 @@ module Fluent
       end
 
       def eval_include(attrs, elems, uri)
-        u = URI.parse(uri)
-        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == uri # file path
+        u = URI.parse(URI.encode_www_form_component(uri))
+        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == URI.encode_www_form_component(uri) # file path
           # When the Windows absolute path then u.scheme.length == 1
           # e.g. C:
-          path = u.path
+          path = URI.decode_www_form_component(u.path)
           if path[0] != ?/
             pattern = File.expand_path("#{@include_basepath}/#{path}")
           else
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -147,7 +147,9 @@ module Fluent
       end
 
       def eval_include(attrs, elems, uri)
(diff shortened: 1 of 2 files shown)


cosmo0920:
Umm..., tests which are affected this changes are failed...

cosmo0920:
I used `uri.gsub(/ /, '+')` instead of URI.encode_www_form_component(uri)/URI.encode(uri).
Because we need to handle spaces in `URI.parse(uri)`.

ganmacs:
> Because we need to handle spaces in URI.parse(uri).

Both methods(`URI.encode_www_form_component(uri)`/`URI.encode(uri)`) can replace space with `+`. Are there any reasons not to use them? I think using them express intentional clearly (it takes me a little time to get why `+` is used as the replaced char).

```
irb(main):003:0> URI.encode('/test with ')
=> "/test%20with%20"
irb(main):004:0> URI.encode_www_form_component('/test with ')
=> "%2Ftest+with+"
```

cosmo0920:
> Both methods(`URI.encode_www_form_component(uri)`/`URI.encode(uri)`) can replace space with `+`. Are there any reasons not to use them? I think using them express intentional clearly (it takes me a little time to get why `+` is used as the replaced char).

URI.encode and URI.encode_www_form_components will escape drive schema like as `C:`.

```irb
irb>  URI.encode_www_form_component('C:/test with ')
=> "C%3A%2Ftest+with+"
```

URI.parse does not handle "C%3A%2F" as a schema.

This causes unintentional behavior on this clause:

https://github.com/fluent/fluentd/blob/master/lib/fluent/config/v1_parser.rb#L151

URI.encode has right behavior for this PR but this method is marked as obsoleted:
https://docs.ruby-lang.org/ja/latest/method/URI/s/encode.html
Should we use this method as in the past?

lib/fluent/config/v1_parser.rb
```diff
@@ -147,11 +147,11 @@ def parse_include(attrs, elems)
       end
 
       def eval_include(attrs, elems, uri)
-        u = URI.parse(uri)
-        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == uri # file path
+        u = URI.parse(uri.gsub(/ /, '+')) # replace space(s)(' ') with '+' to prevent invalid uri due to space(s).
```
ganmacs: Can you please put a comment like what you said in https://github.com/fluent/fluentd/pull/2780#issuecomment-576081212 or just put the link here.

cosmo0920: OK. I'll put it on here.

cosmo0920: Fixed https://github.com/fluent/fluentd/pull/2780/commits/cd67858eef00ae317885935d69910553bf5edbce.

cosmo0920:
@repeatedly ?

cosmo0920:
I'll merge this PR after finishing CI.

PR #2787, opened 2020-01-21
HTTP server helper supports https

ganmacs:

no

http_server helper start supporting https server.

**Docs Changes**:

Needed https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-http_server

**Release Note**: 
same as the title


ganmacs:
> This PR is the first step to support https in plugins such as in_http, right?

Right. but in_http plugin needs a performance so I'm not sure we can replace in_http with this helper. 
In my first intention, this is mainly for fluent-prometheus-plugin( https://github.com/fluent/fluent-plugin-prometheus/pull/136)

lib/fluent/plugin_helper/http_server/compat/server.rb
```diff
@@ -24,16 +25,26 @@ module Compat
...
+            config = {
               BindAddress: @addr,
               Port: @port,
               Logger: WEBrick::Log.new(STDERR, WEBrick::Log::FATAL),
               AccessLog: [],
-            )
+            }
+            if tls_context
+              require 'webrick/https'
+              @logger.warn('Webrick ignores given TLS version')
```
repeatedly: Does this bug of webrick or spec?

ganmacs: Spec. webrick does not provide an interface of setting the TLS version.
See https://github.com/ruby/webrick/blob/c5635fa5e2d4b7d5ca2e08e4c1ae6ec13c7237ec/lib/webrick/ssl.rb#L180-L205

lib/fluent/plugin_helper/http_server.rb
```diff
@@ -64,6 +114,16 @@ def stop
 
       private
 
+      def _http_server_overwrite_config(config, opts)
+        conf = config
```
repeatedly: Is this for short-cut or forget to call `.dup`?

ganmacs: forgot to call dup..
 https://github.com/fluent/fluentd/pull/2787/commits/abb50a853bc309dfa40f546c0f5416e72adb2e8a

ganmacs: @repeatedly https://github.com/fluent/fluentd/pull/298/files#diff-d34ab903425415e82b8cd4d891adadddR41
I've found all tests failed in travis because they expected Section class not to have #dup. 
Do you know why it does?

repeatedly: Hmm... I don't remember. If all tests are succeded after removed it, we can remove it.

PR #2796, opened 2020-01-27
Add path_delimiter to split with any char

ganmacs:

Fixes https://github.com/fluent/fluentd/pull/2001

Add path_delimiter to split with any char
**Docs Changes**:

needed?

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -35,6 +35,8 @@ module Fluent::Plugin
 
     helpers :timer, :event_loop, :parser, :compat_parameters
 
+    RESERVED_CHARS = ['/', '*', '%'].freeze
+
     class WatcherSetupError < StandardError
       def initialize(msg)
         @message = msg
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -35,6 +35,8 @@ class TailInput < Fluent::Plugin::Input
 
     helpers :timer, :event_loop, :parser, :compat_parameters
 
+    RESERVED_CHARS = ['/', '*', '%'].freeze
```
cosmo0920: These characters should be reserved characters for Unix-like systems.
MSDN says that the following characters are reserved characters in NTFS:

* `< (less than)`
* `> (greater than)`
* `: (colon)`
* `" (double quote)`
* `/ (forward slash)`
* `\ (backslash)`
* `| (vertical bar or pipe)`
* `? (question mark)`
* `* (asterisk)`


Should we also consider them here?

ref: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

ganmacs: `RESERVED_CHARS` is fluentd's reserved chars, not UNIX-like system's or MSDN's one. 
I think it's good that users can use as many chars as possible on their own responsibility. 
WDYT?

cosmo0920: I'm thinking of these reserved characters handling....
At least, I want to handle `\ (backslash)` as reserved character on Windows, if we can.

cosmo0920: > `RESERVED_CHARS` is fluentd's reserved chars, not UNIX-like system's or MSDN's one.
> I think it's good that users can use as many chars as possible on their own responsibility.
> WDYT?

 I see. It's reasonable for me. :+1:

PR #2802, opened 2020-01-29
Add TLS module

repeatedly:
- Fix bug of https://groups.google.com/forum/#!topic/fluentd/B818cI-OHdU

Add TLS module to unify fluentd code.
In addition, support recent SSLContext changes about TLS version

TODO: Add test soon

**Docs Changes**:
Add `min_version`/`max_version` to TLS transport.

**Release Note**: 
- Add TLS module to unify code
- Allow TLS v1.3 when available
- Fix bug of TLS version setting

### DIFF
--- a/lib/fluent/tls.rb
+++ b/lib/fluent/tls.rb
@@ -25,30 +25,39 @@ module Fluent
     # OpenSSL 0.9.8 default: "ALL:!ADH:!LOW:!EXP:!MD5:+SSLv2:@STRENGTH"
     CIPHERS_DEFAULT = "ALL:!aNULL:!eNULL:!SSLv2" # OpenSSL >1.0.0 default
 
-    METHODS_MAP = {
-      TLSv1: OpenSSL::SSL::TLS1_VERSION,
-      TLSv1_1: OpenSSL::SSL::TLS1_1_VERSION,
-      TLSv1_2: OpenSSL::SSL::TLS1_2_VERSION,
-    }.freeze
-    METHODS_MAP_FOR_VERSION = {
-      TLS1: :'TLSv1',
-      TLS1_1: :'TLSv1_1',
-      TLS1_2: :'TLSv1_2',
-    }.freeze
+    METHODS_MAP = begin
+                    map = {
+                      TLSv1: OpenSSL::SSL::TLS1_VERSION,
+                      TLSv1_1: OpenSSL::SSL::TLS1_1_VERSION,
+                      TLSv1_2: OpenSSL::SSL::TLS1_2_VERSION,
+                    }.freeze
+                    MIN_MAX_AVAILABLE = true
+                    map
+                  rescue NameError
+                    # ruby 2.4 doesn't have OpenSSL::SSL::TLSXXX constants and min_version=/max_version= methods
+                    map = {
+                      TLS1: :'TLSv1',
+                      TLS1_1: :'TLSv1_1',
+                      TLS1_2: :'TLSv1_2',
+                    }.freeze
+                    MIN_MAX_AVAILABLE = false
+                    map
+                  end
+    private_constant :METHODS_MAP
 
     # Helper for old syntax/method support:
     # ruby 2.4 uses ssl_version= but this method is now deprecated.
     # min_version=/max_version= use 'TLS1_2' but ssl_version= uses 'TLSv1_2'
     def set_version_to_context(ctx, version, min_version, max_version)
-      if ctx.respond_to?(:'min_version=')
+      if MIN_MAX_AVAILABLE
         case
         when min_version.nil? && max_version.nil?
-          min_version = METHODS_MAP[version]
-          max_version = METHODS_MAP[version]
+          min_version = METHODS_MAP[version] || version
+          max_version = METHODS_MAP[version] || version
         when min_version.nil? && max_version
-          raise Fluentd::ConfigError, "When you set max_version, must set min_version together"
+          raise Fluent::ConfigError, "When you set max_version, must set min_version together"
         when min_version && max_version.nil?
-          raise Fluentd::ConfigError, "When you set min_version, must set max_version together"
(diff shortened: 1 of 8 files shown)


repeatedly:
I searched the 3rd party plugins and `Fluent::PluginHelper::Socket::TLS_SUPPORTED_VERSION`/other constants are not used. We can remove it.

lib/fluent/tls.rb
```diff
@@ -0,0 +1,68 @@
...
+module Fluent
+  module TLS
+    DEFAULT_VERSION = :'TLSv1_2'
+    SUPPORTED_VERSIONS = [:'TLSv1_1', :'TLSv1_2', :'TLS1_1', :'TLS1_2']
+    ### follow httpclient configuration by nahi
+    # OpenSSL 0.9.8 default: "ALL:!ADH:!LOW:!EXP:!MD5:+SSLv2:@STRENGTH"
+    CIPHERS_DEFAULT = "ALL:!aNULL:!eNULL:!SSLv2" # OpenSSL >1.0.0 default
+
+    METHODS_MAP = {
+      TLSv1: OpenSSL::SSL::TLS1_VERSION,
```
repeatedly: ruby 2.4 doesn't have these constants. Fix soon.

lib/fluent/tls.rb
```diff
@@ -0,0 +1,68 @@
...
+        case
+        when min_version.nil? && max_version.nil?
+          min_version = METHODS_MAP[version]
+          max_version = METHODS_MAP[version]
+        when min_version.nil? && max_version
+          raise Fluentd::ConfigError, "When you set max_version, must set min_version together"
+        when min_version && max_version.nil?
+          raise Fluentd::ConfigError, "When you set min_version, must set max_version together"
+        else
+          min_version = METHODS_MAP[min_version] || min_version
```
ganmacs: if the `METHODS_MAP[min_version]` returns `nil`,  a user might make a typo or another mistake. 
How about warning it?

repeatedly: SSLContext raising an error for such cases.

repeatedly:
Test passed

repeatedly:
@ganmacs Check this again ;)

test/test_tls.rb
```diff
@@ -0,0 +1,63 @@
...
+    'New TLS v1.1' => :'TLS1_1',
+    'Old TLS v1.1' => :'TLSv1_1',
+  }
+  TEST_TLS1_2_CASES = {
+    'New TLS v1.2' => :'TLS1_2',
+    'Old TLS v1.2' => :'TLSv1_2'
+  } 
+  TEST_TLS_CASES = TEST_TLS1_1_CASES.merge(TEST_TLS1_2_CASES)
+
+  sub_test_case 'constants' do
```
ganmacs: Do these tests need? It looks that they do nothing...

test/test_tls.rb
```diff
@@ -0,0 +1,63 @@
...
+      }
+    end
+
+    data(TEST_TLS_CASES)
+    test 'can specify old/new syntax to min_version/max_version' do |ver|
+      omit "min_version=/max_version= is not supported" unless Fluent::TLS::MIN_MAX_AVAILABLE
+
+      assert_nothing_raised {
+        Fluent::TLS.set_version_to_context(@ctx, Fluent::TLS::DEFAULT_VERSION, ver, ver)
+      }
```
ganmacs: I think it needs to assert `@ctx.min_version` and `@ctx.max_version` is set the given value(Does what the method does set min and max version of TSL?).
The test is just checking this test does not raise an error and it doesn't check the behaivor of Fluent::TLS.set_version_to_context.

lib/fluent/tls.rb
```diff
@@ -0,0 +1,81 @@
...
+      else
+        ctx.ssl_version = METHODS_MAP[version] || version
+      end
+
+      ctx
+    end
+    module_function :set_version_to_context
+  end
+end
+
```
ganmacs: [nits] unnecessary line.

test/test_tls.rb
repeatedly: > I think it needs to assert @ctx.min_version and @ctx.max_version is set the given value(Does what the method does set min and max version of TSL?).

I agree. But the problem is `SSLContext` doesn't provide accessor for `min_version`/`max_version`. If openssl gem implements it, we can add it.

test/test_tls.rb
repeatedly: These constants are used by plugins.
These tests are for preventing unexpected value changes.

lib/fluent/tls.rb
```diff
@@ -0,0 +1,81 @@
...
+                      TLSv1_1: OpenSSL::SSL::TLS1_1_VERSION,
+                      TLSv1_2: OpenSSL::SSL::TLS1_2_VERSION
+                    }
+                    map[:'TLSv1_3'] = OpenSSL::SSL::TLS1_3_VERSION if defined?(OpenSSL::SSL::TLS1_3_VERSION)
+                    MIN_MAX_AVAILABLE = true
+                    map.freeze
+                  rescue NameError
+                    # ruby 2.4 doesn't have OpenSSL::SSL::TLSXXX constants and min_version=/max_version= methods
+                    map = {
+                      TLS1: :'TLSv1',
```
cosmo0920: There is map keys glitch between min max available case and unavailable cases.
Is it intended?
If this glitch is intended, we might have to write documentation about it.

repeatedly: > we might have to write documentation about it.

This `METHODS_MAP` is private constant, so no public usecase.
Plugin developers don't need the details.

cosmo0920: Oh! I see.

repeatedly:
Add comment for current SSLContext test

repeatedly:
If no other comment, I will merge the patch and release fluentd v1.9.1.

Issue #1126, opened 2016-07-27
Add a feature to clean up position informations on in_tail 


thkoch2001:
- fluentd version: 0.12.20 - commit aee8086e9fc
- environment: fluentd pods running on Google Container Engine / kubernetes

Google Container Engine uses fluentd pods to collect container log files via the in_tail plugin and forward the logs to Stackdriver logging.

When a container is deleted, kubernetes also deletes the containers log file and there will never be a log file at this filesystem path again.

However the position file will never clean up the obsolete line in the position file although the position value is ffffffffffffffff.

We see production clusters with position files of over 10000 lines.

Can this cause performance problems with fluentd? Should this be fixed?

The config stanza for the containers log files is:

```
<source>
  type tail
  format json
  time_key time
  path /var/log/containers/*.log
  pos_file /var/log/gcp-containers.log.pos
  time_format %Y-%m-%dT%H:%M:%S.%NZ
  tag reform.*
  read_from_head true
</source>
```

repeatedly:
in_tail removes untracked file positions at start phase.
So if you restart fluentd, pos_file is updated.

> Can this cause performance problems with fluentd?

I'm not sure. I didn't receive any report of pos_file releated performance issue.

tagomoris:
The file which grows infinitely sound terrible in production environment (even if that growing speed is slow, or if there are no reports about performance regression).
I think we should add any features to clean it up to tail plugin.

But currently, we have a plan to switch to use storage plugin from pos file for that purpose.
I think we can add this cleanup feature at the same time with switching storage plugins.

repeatedly:
I added note for this issue on http://docs.fluentd.org/articles/in_tail.

piosz:
cc @Crassirostris

mrkstate:
How do we get this feature (defect) fixed?   This is creating a situation for us - as the POS file is growing rather quickly and the process has to continue to comb thru this file.  We see constant CPU cycles even after no additional data has been written to the tailed log.

roffe:
@mrkstate Run a cron that kills fluentd every 30 minutes https://github.com/roffe/kube-gelf/blob/master/cron.yaml

Krishna1408:
hi @roffe  If i kiil fluentd won't it create problem for logging ? e.g. fluentd was down for 2 minutes so in that case won't I loose the log data for those 2 minutes ?

roffe:
@Krishna1408 no, since the position file will let it know where to pickup from last time

[...]

PR #2805, opened 2020-01-30
Impl posfile auto-compaction in in_tail

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/1126

note: merge after https://github.com/fluent/fluentd/pull/2803

**Docs Changes**:

needed

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -213,11 +213,11 @@ module Fluent::Plugin
         FileUtils.mkdir_p(pos_file_dir) unless Dir.exist?(pos_file_dir)
         @pf_file = File.open(@pos_file, File::RDWR|File::CREAT|File::BINARY, @file_perm)
         @pf_file.sync = true
-        @pf = PositionFile.load(@pf_file, logger: $log)
+        @pf = PositionFile.load(@pf_file, logger: log)
 
         if @pos_file_compaction_interval
           timer_execute(:in_tail_refresh_compact_pos_file, @pos_file_compaction_interval) do
-            $log.info('Clean up the pos file')
+            log.info('Clean up the pos file')
             @pf.try_compact
           end
         end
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -211,7 +213,14 @@ def start
         FileUtils.mkdir_p(pos_file_dir) unless Dir.exist?(pos_file_dir)
         @pf_file = File.open(@pos_file, File::RDWR|File::CREAT|File::BINARY, @file_perm)
         @pf_file.sync = true
-        @pf = PositionFile.parse(@pf_file)
+        @pf = PositionFile.load(@pf_file, logger: $log)
```
cosmo0920: Can we use `log` method as logger here?
I want to switch log level within PositionFile class.

ganmacs: fixed https://github.com/fluent/fluentd/pull/2805/commits/0ffdd845d39d6f6d1fe28638edc474fa7177768b

ganmacs:
@repeatedly ping

lib/fluent/plugin/in_tail/position_file.rb
```diff
@@ -48,56 +55,89 @@ def unwatch(path)
...
+      def try_compact
+        last_modified = nil
+        size = nil
+
+        @file_mutex.synchronize do
+          last_modified = @file.mtime
+          size = @file.size
+        end
+
+        entries = fetch_compacted_entries
```
juliantaylor: having this function outside of the lock is probably a race as it reads the file without a lock and as far as I can tell the file writes are not atomic
so this can read a file currently being modified by another thread leading to corrupted data

probable cause of: https://github.com/fluent/fluentd/issues/2918

ganmacs: This code seems not to have nothing to do with #2918 because [here](https://github.com/fluent/fluentd/pull/2805/files/8946e536e48c8bda125f7fd075f7a0c388b8850a#diff-0f4f3ff9be203ad2654dd6d9ee3a1f7bR93) verify that the file content doesn't change.
So there is no need to lock here.

PR #2809, opened 2020-01-31
Add warning about retryable_response_codes

ganmacs:
ref https://github.com/fluent/fluentd/pull/2779#discussion_r368345580

Added warning message when a user uses a default `retryable_response_codes`  to remove 503 from `retryable_response_codes` next release(fluentd 2.0?).

**Docs Changes**:

no need

**Release Note**: 

same as the title.

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -64,7 +64,7 @@ module Fluent::Plugin
     desc 'Raise UnrecoverableError when the response is non success, 4xx/5xx'
     config_param :error_response_as_unrecoverable, :bool, default: true
     desc 'The list of retryable response code'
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_http.rb
```diff
@@ -90,6 +90,11 @@ def initialize
     def configure(conf)
       super
 
+      if @retryable_response_codes.nil?
+        log.warn('Status code 503 is going to be removed from default `retryable_response_codes`. Please add it by yourself if you wish')
```
repeatedly: Add `since fluentd v2`.

ganmacs: fixed https://github.com/fluent/fluentd/pull/2809/commits/611db3e8e5deb4a8ff627c57afd1586e498f34a1

Issue #2799, opened 2020-01-28
failed fluentd with RUBYOPT has "-E" option


tdtds:
**Describe the bug**
When RUBYOPT environment variable has "-E" option, fluentd fails to start supervisor and into infinite loop.

**To Reproduce**
```
% RUBYOPT=-Eutf-8 fluentd                                                                                           [basic_input_plugin]
2020-01-28 11:35:17 +0900 [info]: parsing config file is succeeded path="/etc/fluent/fluent.conf"
2020-01-28 11:35:17 +0900 [info]: gem 'fluentd' version '1.9.0'
2020-01-28 11:35:17 +0900 [info]: using configuration file: <ROOT>
</ROOT>
2020-01-28 11:35:17 +0900 [info]: starting fluentd-1.9.0 pid=31909 ruby="2.5.7"
2020-01-28 11:35:17 +0900 [info]: spawn command to main:  cmdline=["/home/sho/.rbenv/versions/2.5.7/bin/ruby", "-Eascii-8bit:ascii-8bit", "-Eutf-8", "/home/sho/.rbenv/versions/2.5/bin/fluentd", "--under-supervisor"]
Traceback (most recent call last):
/home/sho/.rbenv/versions/2.5.7/bin/ruby: default_external already set to ascii-8bit (RuntimeError)
2020-01-28 11:35:17 +0900 [info]: Worker 0 finished unexpectedly with status 1
:
:
(repeat infinite...)
```

**Expected behavior**
fluentd stops with clearly error messages.
Or fluentd runs without "-Eascii-8bit:ascii-8bit" is better.

**Your Environment**
* fluentd 1.9.0
* ruby 2.5.7p206
* Debian GNU/Linux 9 (on Windows WSL)

PR #2814, opened 2020-02-03
Use given encoding when RUBYOPT has -E

ganmacs:
merge after https://github.com/fluent/fluentd/pull/2813
Fix https://github.com/fluent/fluentd/issues/2799

note: 
~~I decided that fluentd uses the last option of `-E`, but I'm not sure it's the best way.
Any suggestions are welcome!~~
Use the encoding options(`-E`, `--encoding`,  `--external-encoding`, `--internal-encoding`) in RUBYOPT env var as it is if exists. if not fluentd `-Eascii-8bit:ascii-8bit`. https://github.com/fluent/fluentd/pull/2814/commits/f51154699ca30485473c4097208148e91d15e3ab#diff-dbe0e1ec4079138e48ca6a4d7c7248f9R884

**Docs Changes**:

no need

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -871,28 +871,18 @@ module Fluent
       system_config
     end
 
+    RUBY_ENCODING_OPTIONS_REGEX = %r{\A(-E|--encoding=|--internal-encoding=|--external-encoding=)}.freeze
+
     def build_spawn_command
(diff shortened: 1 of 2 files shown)


ganmacs:
good catch added it https://github.com/fluent/fluentd/pull/2814/commits/f51154699ca30485473c4097208148e91d15e3ab

Issue #2815, opened 2020-02-03
rfc5424 syslog not parsed properly


BananaWanted:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**

rfc5424 syslog can't be parsed with only regex. 

**Examples for incorrect behavior**

1.  `[]` characters in message:
Give this log message:
```
[e@123 ...][meta ...] [this is message]
```
because fluentd parse *extradata* with `(\[(.*)\]` , it will consider the real message part as extra data as well, which is incorrect

source at https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/parser_syslog.rb#L30
rfc reference at https://tools.ietf.org/html/rfc5424#section-6.3

2. can't deal with unicode BOM
as docmented here https://tools.ietf.org/html/rfc5424#section-6.4 the *MSG* 

> If a syslog application encodes MSG in UTF-8, the string MUST start
   with the Unicode byte order mask (BOM), which for UTF-8 is ABNF
   %xEF.BB.BF.  

regex just can't deal with it.

**Your Environment**
fluentd v1.9.1

**Proposals for fixing**

For issues like given in the examples, I can fix it by slightly enhance the regex. However, since `rfc5424` is indeed a binary-based protocol, regex is definitely not the way that we should go at the end of the day.

Instead, we need to find a way to properly implement rfc5424 parsing, probably by integrate with a 3rd party library. However, I can't help with this part in a reasonable short term since I don't know ruby at all.

ganmacs:
2nd issue makes sense to me. and I've confirmed it doesn't work.
but it's unclear for me about 1st issue(`[]` characters in message).

These two test works without any change in my local env. Do you have any example to reproduce it?
https://github.com/fluent/fluentd/compare/master...ganmacs:syslog-regex-test?expand=1#diff-19d720eba498373bb2a1e9e85b5b2466L331
https://github.com/fluent/fluentd/compare/master...ganmacs:syslog-regex-test?expand=1#diff-19d720eba498373bb2a1e9e85b5b2466L297

BananaWanted:
@ganmacs thank you for quick responding.

for the second issue, could you explain more about why it doesn't work?

[...]

PR #2816, opened 2020-02-03
fix for RFC5424 parsing

BananaWanted:
Fixes https://github.com/fluent/fluentd/issues/2815

as described in the issue, updated the regex expression to cover the listed cases.

**Docs Changes**:
It's a bug fix, nothing doced should change

**Release Note**: 
Improved for parsing RFC5424 structured data in `parser_syslog`

### DIFF
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -27,7 +27,7 @@ module Fluent
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = "(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\\-|(?:\\[.*?(?<!\\\\)\\])+))(?: (?:\xef\xbb\xbf)?(?<message>.+))?"
+      REGEXP_RFC5424 = "(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\\-|(?:\\[.*?(?<!\\\\)\\])+))(?: (?<message>.+))?"
       REGEXP_RFC5424_NO_PRI = Regexp.new('\\A' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
       REGEXP_RFC5424_WITH_PRI = Regexp.new('\\A<(?<pri>[0-9]{1,3})\\>[1-9]\\d{0,2} ' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
       REGEXP_DETECT_RFC5424 = /^\<.*\>[1-9]\d{0,2}/
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -30,8 +30,12 @@ module Fluent
       REGEXP_RFC5424 = <<~'EOS'.chomp
         (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|(?:\[.*?(?<!\\)\])+))(?: (?<message>.+))?
       EOS
-      REGEXP_RFC5424_NO_PRI = Regexp.new('\\A' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
-      REGEXP_RFC5424_WITH_PRI = Regexp.new('\\A<(?<pri>[0-9]{1,3})\\>[1-9]\\d{0,2} ' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
+      REGEXP_RFC5424_NO_PRI = Regexp.new(<<~'EOS'.chomp % REGEXP_RFC5424, Regexp::MULTILINE)
+        \A%s\z
+      EOS
+      REGEXP_RFC5424_WITH_PRI = Regexp.new(<<~'EOS'.chomp % REGEXP_RFC5424, Regexp::MULTILINE)
+        \A<(?<pri>[0-9]{1,3})\>[1-9]\d{0,2} %s\z
+      EOS
       REGEXP_DETECT_RFC5424 = /^\<.*\>[1-9]\d{0,2}/
 
       config_set_default :time_format, "%b %d %H:%M:%S"
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -27,8 +27,9 @@ module Fluent
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = /\A(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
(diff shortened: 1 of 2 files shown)


ganmacs:
Could you follow the DCO? https://github.com/fluent/fluentd/pull/2816/checks?check_run_id=423745091
And tests are failing. could you check it?

lib/fluent/plugin/parser_syslog.rb
```diff
@@ -27,8 +27,9 @@ class SyslogParser < Parser
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = /\A(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
-      REGEXP_RFC5424_WITH_PRI = /\A\<(?<pri>[0-9]{1,3})\>[1-9]\d{0,2} (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
+      REGEXP_RFC5424 = "(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\\-|(?:\\[.*?(?<!\\\\)\\])+))(?: (?<message>.+))?"
```
ganmacs: Doesn't this suuport BOM?  (MSG-UTF8  in https://tools.ietf.org/html/rfc5424#section-6 )

test/plugin/test_parser_syslog.rb
```diff
@@ -294,14 +294,14 @@ def test_parse_with_rfc5424_structured_message
                         'message_format' => 'rfc5424',
                         'with_priority' => true,
                         )
-      text = '<16>1 2017-02-06T13:14:15.003Z 192.168.0.1 fluentd 11111 ID24224 [exampleSDID@20224 iut="3" eventSource="Application" eventID="11211"] Hi, from Fluentd!'
+      text = '<16>1 2017-02-06T13:14:15.003Z 192.168.0.1 fluentd 11111 ID24224 [exampleSDID@20224 iut="3" eventSource="Application" eventID="11211"] [Hi] from Fluentd!'
```
ganmacs: I looked over the RFC5424 ABFN.
it shows like below. This means that `[hi]` should be the `SD-ELEMENT`,  not `MSG`,  because `1*SD-ELEMENT` means at least one `SD-ELEMENT` is allowed.

```
 STRUCTURED-DATA = NILVALUE / 1*SD-ELEMENT
 SD-ELEMENT      = "[" SD-ID *(SP SD-PARAM) "]"
```

lib/fluent/plugin/parser_syslog.rb
```diff
@@ -27,8 +27,9 @@ class SyslogParser < Parser
...
+      REGEXP_RFC5424_NO_PRI = Regexp.new('\\A' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
```
ganmacs: `\` for escaping seems not to be necessary since `p Regexp.new('\\A', Regexp::MULTILINE) == Regexp.new('\A', Regexp::MULTILINE)` reutrns `true`

lib/fluent/plugin/parser_syslog.rb
BananaWanted: it does.
 I've removed the part that deals with BOM not only because it's not necessary but also for it making the test flaky

lib/fluent/plugin/parser_syslog.rb
BananaWanted: They work the same. I preferring always escape it for readability, so people who come from other languages (like me) can distinguish string literal from control chars easily. What do you think of it?

ganmacs: It's IMO but I prefer the one without escape because the current value is so.
I can't think that `(?:\\-|(?:\\[.*?(?<!\\\\)\\])+))` is readeable...

lib/fluent/plugin/parser_syslog.rb
```diff
@@ -27,8 +27,12 @@ class SyslogParser < Parser
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = /\A(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
-      REGEXP_RFC5424_WITH_PRI = /\A\<(?<pri>[0-9]{1,3})\>[1-9]\d{0,2} (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
+      REGEXP_RFC5424 = <<~'EOS'.chomp
+        (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|(?:\[.*?(?<!\\)\])+))(?: (?<message>.+))?
+      EOS
+      p(REGEXP_RFC5424)
```
cosmo0920: Could you remove debug print here if your debugging work is finished?

BananaWanted: sorry, forgot to remove it

lib/fluent/plugin/parser_syslog.rb
BananaWanted: Agreed. I tried to make it more readable by using Heredoc :)

ganmacs: Is there any reason to change `(?<extradata>(?:\-|(?:\[.*?(?<!\\)\])+))` ?
it makes sense to me for adding `+`  after `\])`. but I can't see the point of adding the part of `?(?<!\\)`.

lib/fluent/plugin/parser_syslog.rb
```diff
@@ -27,8 +27,11 @@ class SyslogParser < Parser
       REGEXP = /^(?<time>[^ ]*\s*[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
       # From in_syslog default pattern
       REGEXP_WITH_PRI = /^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[^ :\[]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
-      REGEXP_RFC5424 = /\A(?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
-      REGEXP_RFC5424_WITH_PRI = /\A\<(?<pri>[0-9]{1,3})\>[1-9]\d{0,2} (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|\[(.*)\]))(?: (?<message>.+))?\z/m
+      REGEXP_RFC5424 = <<~'EOS'.chomp
+        (?<time>[^ ]+) (?<host>[!-~]{1,255}) (?<ident>[!-~]{1,48}) (?<pid>[!-~]{1,128}) (?<msgid>[!-~]{1,32}) (?<extradata>(?:\-|(?:\[.*?(?<!\\)\])+))(?: (?<message>.+))?
+      EOS
+      REGEXP_RFC5424_NO_PRI = Regexp.new('\\A' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
+      REGEXP_RFC5424_WITH_PRI = Regexp.new('\\A<(?<pri>[0-9]{1,3})\\>[1-9]\\d{0,2} ' + REGEXP_RFC5424 + '\\z', Regexp::MULTILINE)
```
ganmacs: Can you change the same thing to https://github.com/fluent/fluentd/pull/2816/files#r375030986  (unnecessary escaping thing)?
Existing two types of a string is confusing. (also L33)

lib/fluent/plugin/parser_syslog.rb
BananaWanted: `+` applies for the whole group, because the message can carry mutiple `SD-ELEMENT` .

`(?<!...)` is a negative lookbehind to not match `\]` as an end of SD-ELEMENT since `]` is allowed in SD PARAM-VALUE by escaping with a `\` prefix

lib/fluent/plugin/parser_syslog.rb
BananaWanted: updated

BananaWanted:
Hi @ganmacs @cosmo0920  would you mind helping me take a look on the CI result https://ci.appveyor.com/project/fluent/fluentd/builds/30633930

I'm not sure what's going on there :(

ganmacs:
> Hi @ganmacs @cosmo0920 would you mind helping me take a look on the CI result

This seems not to be related to this issue. I'll merge later! Thank you!

Issue #2698, opened 2019-11-14
Restoring chunks with same metadata


fleprovost:
**Describe the bug**
When Fluentd restarts, it restores all chunks that are stored in buffer files and intends to flush them but only the last chunk restored is flushed.

**To Reproduce**
- Use any input that can ship thousands of events/s
- Use an output that supports file buffering
- Set chunk_limit_records to 10
- Send 1000* {"timestamp": 1} JSON events
(100 buffer files are created)
- Before the events are flushed, reload fluentd service

**Expected behavior**
All chunks are restored and flushed when Fluentd restarts

**Your Environment**

- Fluentd or td-agent version: fluentd 1.7.4
- Operating system: 
NAME="Alpine Linux"
VERSION_ID=3.9.4
- Kernel version: 4.4.0-154-generic

**Your Configuration**

```
<system>
  log_level trace
</system>

<source>
  @type tail
  path /fluentd/log/foo.log
  pos_file /fluentd/log/td-agent/foo.log.pos
  tag foo.bar
  <parse>
    @type json
  </parse>
</source>

<filter foo.bar>
    @type record_transformer
    enable_ruby
    <record>
      newtime ${record["timestamp"]}
    </record>
    renew_time_key newtime
</filter>

<match foo.bar>
  @type amqp
  exchange foo
  key ""
  addresses 172.18.0.2:5672
  user guest
  pass guest
  vhost /
  <buffer time>
    @type file
    timekey 30
    timekey_wait 0
    chunk_limit_records 10
    path /fluentd/buffer/
  </buffer>
</match>

```

**Your Error Log**

Before Fluentd reloading, all chunks are flushed correctly, but at restarts:

```
2019-11-14 14:42:44 +0000 [debug]: #0 fluent/log.rb:302:debug: restoring buffer file: path = /fluentd/buffer/buffer.b5974f78f0c58935521fbf76d5e691216.log
2019-11-14 14:42:44 +0000 [debug]: #0 fluent/log.rb:302:debug: restoring buffer file: path = /fluentd/buffer/buffer.b5974f78ecc0a41d98d0fccef5a906ab4.log
2019-11-14 14:42:44 +0000 [debug]: #0 fluent/log.rb:302:debug: restoring buffer file: path = /fluentd/buffer/buffer.b5974f78f0afa28e73d5e198227f834a6.log
[...]

PR #2824, opened 2020-02-13
Add seq to metadata that it can be unique

ganmacs:

Fixes https://github.com/fluent/fluentd/issues/2698

**Docs Changes**:

no need

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -60,7 +60,17 @@ module Fluent
       desc 'Compress buffered data.'
       config_param :compress, :enum, list: [:text, :gzip], default: :text
 
-      Metadata = Struct.new(:timekey, :tag, :variables) do
+      Metadata = Struct.new(:timekey, :tag, :variables, :seq) do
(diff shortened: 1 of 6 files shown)


ganmacs:
@repeatedly @cosmo0920 ?

lib/fluent/plugin/buffer.rb
```diff
@@ -60,7 +60,17 @@ class BufferChunkOverflowError < BufferError; end # A record size is larger than
       desc 'Compress buffered data.'
       config_param :compress, :enum, list: [:text, :gzip], default: :text
 
-      Metadata = Struct.new(:timekey, :tag, :variables) do
+      Metadata = Struct.new(:timekey, :tag, :variables, :seq) do
```
repeatedly: No need to use `seq` in `cmp_variables` and `<=>`?

ganmacs: Yes. `seq` is not related to an order. it's just used for the uniqueness of metadata.

Issue #2821, opened 2020-02-08
Backup directories created in world writable mode


saahn:
Our security scanner is detecting that backup directories created by fluentd are world writable.
I think they are getting created [here](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L1206): 

```ruby
          FileUtils.mkdir_p(backup_dir) unless Dir.exist?(backup_dir)
          File.open(backup_file, 'ab', system_config.file_permission || 0644) { |f|
            chunk.write_to(f)
          }
```
Is there any reason why those directories need to be world writable, rather than set to 755?

We are using fluentd version 1.7.0

ganmacs:
FileUtils.mkdir_p create the directory in 755. 

```rb
require 'fileutils'

path = './test'
p FileUtils.mkdir_p(path)
p File.stat(path).mode.to_s(8) #=> "40755"
```

BTW, please follow the issue template and https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md

saahn:
I tried to reproduce the above with the same ruby used by td-agent on our hosts, and I'm seeing inconsistencies in the default behavior of FileUtils.mkdir_p from what you posted above:

```bash
$ cat test.rb
require 'fileutils'

safe_plugin_id = 'some_plugin_id'
backup_base_dir = '/tmp'
p backup_base_dir
backup_file = File.join(backup_base_dir, 'backup', "foo_id", safe_plugin_id, "bar.log")
p backup_file
backup_dir = File.dirname(backup_file)
p backup_dir

FileUtils.mkdir_p(backup_dir) unless Dir.exist?(backup_dir)
$ rm -rf /tmp/backup/
$ /opt/td-agent/embedded/bin/ruby test.rb
"/tmp"
"/tmp/backup/foo_id/some_plugin_id/bar.log"
"/tmp/backup/foo_id/some_plugin_id"
$ ls -lah /tmp/ |  grep backup
drwx------     3 root      root      4.0K Feb 15 01:59 backup
$ ls -lah /tmp/backup/
total 628K
drwx------     3 root root 4.0K Feb 15 01:59 .
drwxrwxrwt 18770 root root 616K Feb 15 02:00 ..
drwx------     3 root root 4.0K Feb 15 01:59 foo_id
$ ls -lah /tmp/backup/foo_id/
total 12K
drwx------ 3 root root 4.0K Feb 15 01:59 .
[...]

PR #2827, opened 2020-02-17
Specify the dir mode explicitly

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2821

Specify the dir permission mode explicitly to avoid creating world writable dir in some env.

**Docs Changes**:

no need

**Release Note**: 

Specify the dir permission mode explicitly

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -49,6 +49,7 @@ module Fluent::Plugin
     end
 
     FILE_PERMISSION = 0644
+    DIR_PERMISSION = 0755
 
     def initialize
       super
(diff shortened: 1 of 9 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -49,6 +49,7 @@ def to_s
     end
 
     FILE_PERMISSION = 0644
+    DIR_PERMISSION = 0755
```
ganmacs: this value is just copied from https://github.com/fluent/fluentd/blob/9d113029d4550ce576d8825bfa9612aa3e55bff0/lib/fluent/plugin/buf_file.rb#L34

ganmacs:
> DIR_PERMISSION definitions will be unified? 

Sound good. I'll do this after merging this patch!

PR #2837, opened 2020-02-25
Fix the data loss in in_tail when emit_unmatched_lines  is enabled.

ganmacs:

no

Fix the data loss in in_tail when emit_unmatched_lines  is enabled.

**Docs Changes**:

Add unmatched_lines to multiline parser

**Release Note**: 

same as the title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -123,6 +123,8 @@ module Fluent::Plugin
         parser_config["format#{n}"] = conf["format#{n}"] if conf["format#{n}"]
       end
 
+      parser_config['unmatched_lines'] = conf['emit_unmatched_lines']
+
       super
 
       if !@enable_watch_timer && !@enable_stat_watcher
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -447,6 +447,16 @@ module Fluent::Plugin
           record[@path_key] ||= tw.path unless @path_key.nil?
           router.emit(tag, time, record)
         else
+          if @emit_unmatched_lines
+            record = { 'unmatched_line' => buf }
+            record[@path_key] ||= tail_watcher.path unless @path_key.nil?
+            tag = if @tag_prefix || @tag_suffix
+                    @tag_prefix + tw.tag + @tag_suffix
+                  else
+                    @tag
+                  end
+            router.emit(tag, Fluent::EventTime.now, record)
+          end
           log.warn "got incomplete line at shutdown from #{tw.path}: #{buf.inspect}"
         end
       }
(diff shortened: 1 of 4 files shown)


repeatedly:
Tests failed. Could you check the patch?

ganmacs:
I fixed tests.

ganmacs:
@repeatedly  ?

lib/fluent/plugin/parser_multiline.rb
```diff
@@ -24,9 +24,32 @@ class MultilineParser < Parser
...
+            end
+          end
+
+          time, record = convert_values(parse_time(r), r)
+
+          yield(time, record)
+          m
+        end
+      end
       def configure(conf)
```
repeatedly: putting empty line between `end` and `def` is better.

test/plugin/test_in_tail.rb
```diff
@@ -48,14 +48,14 @@ def cleanup_directory(path)
   MULTILINE_CONFIG = config_element(
     "", "", {
       "format" => "multiline",
-      "format1" => "/^s (?<message1>[^\\n]+)(\\nf (?<message2>[^\\n]+))?(\\nf (?<message3>.*))?/",
+      "format1" => "/^s (?<message1>[^\\n]+)(\\nf (?<message2>[^\\n]+))?(\\nf (?<message3>.[^\\n]+))?/",
```
repeatedly: This means this change has breaking compatibility?

lib/fluent/plugin/parser_multiline.rb
ganmacs: fixed https://github.com/fluent/fluentd/pull/2837/commits/237284a99707075603b9d46d6fcd36b963c99b8a

test/plugin/test_in_tail.rb
ganmacs: ah, good catch. added new const. https://github.com/fluent/fluentd/pull/2837/commits/e3079703f92a7c63d849950ce51ba2cc8d063a91

However, I think this change has a breaking change. 
This change can emit the data as unmatched_data which the previous version of fluentd ignores.

lib/fluent/plugin/in_tail.rb
```diff
@@ -451,6 +453,16 @@ def flush_buffer(tw)
...
+              record = { 'unmatched_line' => lb }
+              record[@path_key] ||= tail_watcher.path unless @path_key.nil?
+              tag = if @tag_prefix || @tag_suffix
+                      @tag_prefix + tw.tag + @tag_suffix
+                    else
+                      @tag
+                    end
+              router.emit(tag, Fluent::EventTime.now, record)
+            end
             log.warn "got incomplete line at shutdown from #{tw.path}: #{lb.inspect}"
```
repeatedly: This line should be in `else`?

ganmacs: I put it here because a similar code does the same thing https://github.com/fluent/fluentd/pull/2837/files#diff-1da710c9dcc8d0fc57996df7a9d39695R507.

cosmo0920:
This PR is conflicted. Could you solve the conflict?

PR #2852, opened 2020-03-03
log should be in else cond

ganmacs:
https://github.com/fluent/fluentd/pull/2843/files

follows up https://github.com/fluent/fluentd/pull/2843

I've noticed that I moved log.warn out of the else cond in https://github.com/fluent/fluentd/pull/2843/files#diff-1da710c9dcc8d0fc57996df7a9d39695L450-L451. so I reverted the change. 

**Docs Changes**:

no

**Release Note**: 

no


ganmacs:
thank you!

PR #2853, opened 2020-03-03
When chunks are queued, their `metadata.seq` should be 0

ganmacs:
Follow up https://github.com/fluent/fluentd/pull/2824

In #2824, I added `seq` to Metadata class so that it can identify the chunks which are created in Buffer#write_step_by_step for Buffer#resume.
When chunks are queued, their `metadata.seq` should be 0 for `@queued_num`. so I change seq to 0 when the chunks are queue. 

**Docs Changes**:

no

**Release Note**: 
same as the title


cosmo0920:
Ah, the testcase is failing:
https://travis-ci.org/fluent/fluentd/jobs/657615579#L1034

Could you check it?

ganmacs:
fixed! https://github.com/fluent/fluentd/pull/2853/commits/8881a9e9478b5f10da7c1ded4d80d049ac054550#diff-ae39384c4115511c1d63f4c22e1618e9L1075

ganmacs:
thank you!

Issue #2854, opened 2020-03-03
Fix failing AppVeyor CI


cosmo0920:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Is your feature request related to a problem? Please describe.**

Currently, AppVeyor CI is always failing.
We should pay attention to AppVeyor CI.

**Describe the solution you'd like**

Fix AppVeyor CI with several PRs.

**Describe alternatives you've considered**

We should mark omittable testcases for hard to fix.

**Additional context**

This makes obscure whether Fluentd on Windows works well or not.

cosmo0920:
https://ci.appveyor.com/project/fluent/fluentd/builds/31359133/job/9j76r04jtjfhclj5#L152 `@i.retry` is sometimes failing.
We should investigate it.

ganmacs:
It sometimes happens in not windows env https://travis-ci.org/fluent/fluentd/jobs/655185903

cosmo0920:
Hmm..., `@i.retry` assertion seems to be fragile....

cosmo0920:
Fragile testcases should cause failure in other platforms.
So, we can close this issue ticket due to fixing Windows related failures.

PR #2855, opened 2020-03-03
Fix appveyor ci round1

cosmo0920:

Related to #2854.

* Omit/Fix permission related testcases.
* Prevent raising error on `$log.out.reset` for non `Fluent::Test::DummyLogDevice` instance case.

We should take care of AppVeyor CI result.

**Docs Changes**:
No needed.

**Release Note**: 

None.

### DIFF
--- a/test/test_supervisor.rb
+++ b/test/test_supervisor.rb
@@ -161,7 +161,7 @@ class SupervisorTest < ::Test::Unit::TestCase
 
     assert{ $log.out.logs.first.end_with?(info_msg) }
   ensure
-    $log.out.reset
+    $log.out.reset unless $log.out.is_a?(Fluent::LogDeviceIO)
   end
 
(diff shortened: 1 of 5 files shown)


test/test_supervisor.rb
```diff
@@ -161,7 +161,7 @@ def test_rpc_server
 
     assert{ $log.out.logs.first.end_with?(info_msg) }
   ensure
-    $log.out.reset
+    $log.out.reset unless $log.out.is_a?(Fluent::LogDeviceIO)
```
ganmacs: This method(`reset`) is only implemented in  Fluent::Test::DummyLogDevice.  I think the following is better .

```suggestion
    $log.out.reset if $log.out.is_a?(Fluent::Test::DummyLogDevice)
```

PR #2858, opened 2020-03-04
Reduce TailWatcher argument and adding Io handler

ganmacs:
no

Reduce TailWatcher argument and adding Io handler

**Docs Changes**:

no

**Release Note**: 
no


ganmacs:
@repeatedly ?

PR #2862, opened 2020-03-05
Fix appveyor ci for round3

cosmo0920:

Related to #2854.

* Omit deleting ephemeral buffer file due to permission error.
  * Windows does not permit removing files which are used in another process even if administrator.
* Fix `linger_timeout` handling on Windows

We should take care of AppVeyor CI result.

**Docs Changes**:
No needed.

**Release Note**: 

None.

### DIFF
--- a/lib/fluent/plugin_helper/socket_option.rb
+++ b/lib/fluent/plugin_helper/socket_option.rb
@@ -21,6 +21,8 @@ require 'fcntl'
 module Fluent
   module PluginHelper
     module SocketOption
+      # ref: https://docs.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-linger
+      FORMAT_STRUCT_LINGER_WINDOWS  = 'S!S!' # { u_short l_onoff; u_short l_linger; }
       FORMAT_STRUCT_LINGER  = 'I!I!' # { int l_onoff; int l_linger; }
       FORMAT_STRUCT_TIMEVAL = 'L!L!' # { time_t tv_sec; suseconds_t tv_usec; }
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin_helper/socket_option.rb
```diff
@@ -49,9 +51,21 @@ def socket_option_set(sock, resolve_name: nil, nonblock: false, linger_timeout:
...
         end
-        if linger_timeout
+        if Fluent.windows?
+          # To prevent closing socket forcibly on Windows,
+          # this options shouldn't be set up when linger_timeout equals to 0 (including nil).
+          # This unintended behavior always ocurrs on Windows when linger_timeout.to_i == 0.
+          # This unintented behavior causes "Errno::ECONNRESET: An existing connection was forcibly
+          # closed by the remote host." on Windows.
+          if linger_timeout.to_i > 0
+            optval = [1, linger_timeout.to_i].pack(FORMAT_STRUCT_LINGER_WINDOWS)
```
ganmacs: It may be better to check `linger_timeout` should be less than 2^16? or this case won't happen?

cosmo0920: Hmm..., specifying `2**17` here does not cause `Errno::ECONNRESET`.
But it maybe reasonable to check greater than 65535(2**16).

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/2862/commits/b55e610909943cce71dda04c242a87f573fc1be9

cosmo0920:
I'll merge this PR after finishing CIs.

PR #2879, opened 2020-03-12
Feature/fix wrong syslog format detection

### LINKED ISSUE #2877: Bug RFC detection regex for syslog parser

**Describe the bug**
When RFC3164 formatted message contains '>' character followed by a number, it is incorrectly detected as RFC5425 message due to an issue with regex [here](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/parser_syslog.rb#L39) 

Simple fix is to make the detection regex match the parsing regex - i.e. 
instead of 
``` ruby
REGEXP_DETECT_RFC5424 = /^\<.*\>[1-9]\d{0,2}/
```
use this:
``` ruby
REGEXP_DETECT_RFC5424 = /^\<.[0-9]+\>[1-9]\d{0,2}/
```

**To Reproduce**

setup syslog parser plugin and run through it a message like:

        <15>Mar 12 21:58:53 localhost service: Connection established 1.2.3.4:1234<->4..2.1:4321

**Expected behavior**
Expected to be parsed as RFC3164 message:
```json
{ 
 "pri": 15, 
 "time": "Mar 12 21:58:53", 
 "host": "localhost",
 "ident": "service", 
 "message": "Connection established 1.2.3.4:1234<->4..2.1:4321"
}
```
**Actual behavior**

Parser error due to attempt to parse it as RFC5424 message

mlasevich:
Fixes #2877 

Regex to detect RFC5424 syslog format did not match  regex to parse it - leading to bugs

**Docs Changes**:
None

**Release Note**: 
Fix inappropriate parsing of some RFC3164 syslog messages with specific combination of characters (> followed by a number)

### DIFF
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -36,7 +36,7 @@ module Fluent
       REGEXP_RFC5424_WITH_PRI = Regexp.new(<<~'EOS'.chomp % REGEXP_RFC5424, Regexp::MULTILINE)
         \A<(?<pri>[0-9]{1,3})\>[1-9]\d{0,2} %s\z
       EOS
-      REGEXP_DETECT_RFC5424 = /^\<.*\>[1-9]\d{0,2}/
+      REGEXP_DETECT_RFC5424 = /^\<[0-9]{1,3}\>[1-9]\d{0,2}/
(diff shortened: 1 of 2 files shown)


ganmacs:
Thanks!

PR #2888, opened 2020-03-16
server: Add cert_verifier parameter for TLS transport

repeatedly:

No issue.

Support user defined verifier for certificate via `cert_verifier` parameter.
This parameter uses `SSLContext#verify_callback` internally.

- conf

```
<transport tls>
 # other parameters...
  client_cert_auth true
  cert_verifier /path/to/cert_checker.rb
</transport>
```

### cert_checker.rb example

Code must be return callbale object. It means returned object must have `#call` method with 2 arguments.

- Proc or lambda for simple case

```
Proc.new { |ok, ctx|
  # check code
}
```

- Use class for complicated case

```
class HooBarChecker
  def initialize
    # Setup data
  end

  def call(ok, ctx)
    # check code
  end
end

HooBarChecker.new
```

Here is actual command and conf: https://gist.github.com/repeatedly/6df9881d7ce62bcf321e2a07799d8909

**Docs Changes**:
Add `cert_verifier` to TLS parameters

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin_helper/cert_option.rb
+++ b/lib/fluent/plugin_helper/cert_option.rb
@@ -46,8 +46,13 @@ module Fluent
           ctx.extra_chain_cert = extra
         end
         if conf.cert_verifier
-          verifier = File.exist?(conf.cert_verifier) ? File.read(conf.cert_verifier) : conf.cert_verifier
-          ctx.verify_callback = self.instance_eval(verifier, File.basename(conf.cert_verifier))
+          sandbox = Class.new
+          ctx.verify_callback = if File.exist?(conf.cert_verifier)
+                                  verifier = File.read(conf.cert_verifier)
+                                  sandbox.instance_eval(verifier, File.basename(conf.cert_verifier))
+                                else
+                                  sandbox.instance_eval(conf.cert_verifier)
+                                end
         end
 
         Fluent::TLS.set_version_to_context(ctx, version, conf.min_version, conf.max_version)
--- a/lib/fluent/plugin_helper/cert_option.rb
+++ b/lib/fluent/plugin_helper/cert_option.rb
@@ -36,7 +36,7 @@ module Fluent
         end
 
         if conf.client_cert_auth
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin_helper/cert_option.rb
```diff
@@ -45,6 +45,11 @@ def cert_option_create_context(version, insecure, ciphers, conf)
         if extra && !extra.empty?
           ctx.extra_chain_cert = extra
         end
+        if conf.cert_verifier
+          verifier = File.exist?(conf.cert_verifier) ? File.read(conf.cert_verifier) : conf.cert_verifier
+          ctx.verify_callback = self.instance_eval(verifier)
```
ganmacs: Is it better to change the receiver of `instance_eval`?

If `File.exist?(conf.cert_verifier` returns  `true`, I think it's better to pass filename to the second argument of `instance_eval` so that users can debug easily.

https://ruby-doc.org/core-2.7.0/BasicObject.html#method-i-instance_eval

repeatedly: > Is it better to change the receiver of instance_eval?

Which object?

ganmacs: How about creating a new class like CerVerifierSandbox ?

lib/fluent/plugin_helper/cert_option.rb
```diff
@@ -47,7 +47,7 @@ def cert_option_create_context(version, insecure, ciphers, conf)
         end
         if conf.cert_verifier
           verifier = File.exist?(conf.cert_verifier) ? File.read(conf.cert_verifier) : conf.cert_verifier
-          ctx.verify_callback = self.instance_eval(verifier)
+          ctx.verify_callback = self.instance_eval(verifier, File.basename(conf.cert_verifier))
```
ganmacs: When  `File.exist?(conf.cert_verifier)` returns `false`,  `File.basename(conf.cert_verifier)` doesn't mean file path, right?

repeatedly:
Applied reviews.

PR #2890, opened 2020-03-16
Fixes #2889 by supporting CRLF and LF X.509 certificates

### LINKED ISSUE #2889: [TLS] Fluentd does not accept certificates containing CRLF instead of LF
**Describe the bug**
Enabling TLS transport using a valid X.509 certificate that contains CRLF instead of LF leads to an unexpected behavior. Fluentd is not able to parse the valid certificate and refuses to do TLS handshakes.

**To Reproduce**
Specify a forward input using TLS that points to a certificate containing CRLF instead of LF.

**Expected behavior**
Fluentd should warn if it could not parse and find a valid certificate in the given cert_path content. Also Fluentd should be able to handle certificates that contain CRLF instead of LF. 

**Your Environment**
- Fluentd or td-agent version: `fluentd 1.9.3`
- Operating system: `NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"`
- Kernel version: `4.14.171-136.231.amzn2.x86_64`

**Your Configuration**

```
<system>
  workers 1
  log_level debug
</system>

<source>
  @type forward
  @id input_forward
  port 24224
  <transport tls>
    cert_path /Users/pascalbudner/Certs/fluentd.dev.broken.cer
    private_key_path /Users/pascalbudner/Certs/fluentd.dev.broken.key
  </transport>
</source>
```

**Your Error Log**

```
[warn]: #0 [input_forward] unexpected error before accepting TLS connection by OpenSSL error_class=OpenSSL::SSL::SSLError error="SSL_accept returned=1 errno=0 state=SSLv3 read client hello C: no shared cipher"
```

**Additional context**

nothing to add here

pbudner:
Fixes #2889 

It makes Fluentd more robust when it parses X509 certificates by allowing CRLF besides currently only LF. Furthermore, it also adds a warn message when no certificate could be parsed during TLS initialization.

**Docs Changes**:
None

**Release Note**: 
Same as title


lib/fluent/plugin_helper/cert_option.rb
```diff
@@ -171,9 +171,12 @@ def cert_option_generate_server_pair_self_signed(generate_opts)
 
       def cert_option_certificates_from_file(path)
         data = File.read(path)
-        pattern = Regexp.compile('-+BEGIN CERTIFICATE-+\n(?:[^-]*\n)+-+END CERTIFICATE-+\n?', Regexp::MULTILINE)
+        pattern = Regexp.compile('-+BEGIN CERTIFICATE-+\r?\n(?:[^-]*\r?\n)+-+END CERTIFICATE-+\r?\n?', Regexp::MULTILINE)
         list = []
         data.scan(pattern){|match| list << OpenSSL::X509::Certificate.new(match) }
+        if list.length() == 0
```
ganmacs: `()` isn't necessary 
```suggestion
        if list.length == 0
```

pbudner:
I added a test that ensures CRLF certs can be parsed

repeatedly:
Thanks!

Issue #2871, opened 2020-03-10
Fluentd CPU usage increased by 50% from version 1.7 to 1.8.


qingling128:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**
When upgrading Fluentd 1.6 to the latest 1.9, our soak tests detected that the CPU usage increased by 50% given the same log throughputs. Further investigation revealed the CPU usage increase is triggered by some change between `1.7.4` and `1.8.1`.

**CPU usage increased by 50%**
![image](https://user-images.githubusercontent.com/5287526/76286651-14ef1880-6279-11ea-9577-a7a53bd9b284.png)
**Memory (RSS) is roughly the same** 
![image](https://user-images.githubusercontent.com/5287526/76286592-fbe66780-6278-11ea-8717-f0871650d9ba.png)
**Log throughputs decreased**
The agent can no longer catch up with the same log throughput of 1000 log entries per second:
![image](https://user-images.githubusercontent.com/5287526/76286743-423bc680-6279-11ea-83f5-81fb91487031.png)

**Expected behavior**
CPU usage does not increase dramastically

**Your Environment**
Fluentd runs inside a Debian 9 based container in GKE with fixed throughput of 1000 log entries per second.

**Your Configuration**

https://github.com/Stackdriver/kubernetes-configs/blob/b67c14df983107a5eacecf6bdc8371d284aa4d10/logging-agent.yaml#L131

qingling128:
Further experiments with various Fluentd versions (`1.7.4`, `1.8.0.rc1`, `1.8.0.rc2`, `1.8.0.rc3`, `1.8.0`, `1.8.1`) narrow down to the changes between `1.7.4` and `1.8.0.rc1`.

![image](https://user-images.githubusercontent.com/5287526/76288066-40bfcd80-627c-11ea-85f7-689ff4053149.png)

There are 64 commits in between these two versions: https://github.com/fluent/fluentd/compare/v1.7.4...v1.8.0.rc1.

Did anything look obvious that might contribute to the CPU usage?

ganmacs:
I'm not sure but  [this commit](https://github.com/fluent/fluentd/commit/fcd467c722c491919c7044c772e9e49ead3ef787) might affect usage of CPU. Could you please do the experiment on the previous commit(8535c9a)? 

[...]

PR #2903, opened 2020-03-23
Stop to output deprecated warning log about MessagePackFactory::Mixin multiple times for performance

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2871

Stop to output a deprecated warning multiple times.
fluent-plugin-google-cloud uses  Fluent::Engine.msgpack_factory and the method changed  deprecated in [this commit](https://github.com/fluent/fluentd/commit/7b166f08f1f1c5d7aa8f49709843bc0b332ee33e).
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/65b9d0e1afdec26ea14e01d7a731f05d499d3477/lib/fluent/plugin/out_google_cloud.rb#L1912

The deprecated method print [warning log]( https://github.com/fluent/fluentd/commit/fcd467c722c491919c7044c772e9e49ead3ef787#diff-2127cc829c2d543c1d9564680b1d4050R33) and it causes the performance issue.

**Docs Changes**:

no need

**Release Note**: 

same as the title.

**bechmark**

```
            original   346.805k i/100ms
                 now   243.599k i/100ms
                 new   345.928k i/100ms
            original     11.621M (± 5.3%) i/s -     57.916M in   4.999822s
                 now      4.617M (± 4.7%) i/s -     23.142M in   5.023242s
                 new     11.359M (± 4.6%) i/s -     56.732M in   5.005022s
```

```rb
require 'benchmark/ips'
require 'logger'
require 'msgpack'

$log = Logger.new(nil)

module MessagePackFactory
  @@engine_factory = nil

  def self.engine_factory(enable_time_support: false)
    @@engine_factory || factory(enable_time_support: enable_time_support)
  end

  def self.factory(enable_time_support: false)
    @@engine_factory = MessagePack::Factory.new
  end
end

module Mixin
  def msgpack_factory_original
    MessagePackFactory.engine_factory
  end

  def msgpack_factory
    if $log
      $log.warn('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.engine_factory')
    end

    MessagePackFactory.engine_factory
  end

  def msgpack_factory_cache
    unless @deprecated_log_done
      deprecated_log { 'Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.engine_factory' }
    end
    MessagePackFactory.engine_factory
  end

  def deprecated_log
    if $log
      $log.warn(yield)
      @deprecated_log_done = true
    end
  end
end

class A
  include Mixin
end

a = A.new
b = A.new
c = A.new

Benchmark.ips do |x|
  x.report "original" do
    a.msgpack_factory_original
  end

  x.report "now" do
    b.msgpack_factory
  end

  x.report "new" do
    c.msgpack_factory_cache
  end
end
```

### DIFF
--- a/lib/fluent/msgpack_factory.rb
+++ b/lib/fluent/msgpack_factory.rb
@@ -23,24 +23,31 @@ module Fluent
 
     module Mixin
       def msgpack_factory
-        if $log
-          $log.warn('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.engine_factory')
+        unless @deprecated_log_done
+          deprecated_log('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.engine_factory')
         end
         MessagePackFactory.engine_factory
       end
 
       def msgpack_packer(*args)
-        if $log
-          $log.warn('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.msgpack_packer')
+        unless @deprecated_log_done
+          deprecated_log('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.msgpack_packer')
         end
         MessagePackFactory.msgpack_packer(*args)
       end
 
       def msgpack_unpacker(*args)
-        if $log
-          $log.warn('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.msgpack_unpacker')
+        unless @deprecated_log_done
+          deprecated_log('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.msgpack_unpacker')
         end
         MessagePackFactory.msgpack_unpacker(*args)
+      end
+
+      def deprecated_log(str)
+        if $log
+          $log.warn(str)
+          @deprecated_log_done = true
+        end
       end
     end
 
--- a/test/test_msgpack_factory.rb
+++ b/test/test_msgpack_factory.rb
@@ -1,0 +1,18 @@
+require_relative 'helper'
+require 'fluent/msgpack_factory'
+
+class MessagePackFactoryTest < Test::Unit::TestCase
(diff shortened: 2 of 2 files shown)


lib/fluent/msgpack_factory.rb
```diff
@@ -23,25 +23,32 @@ module MessagePackFactory
 
     module Mixin
       def msgpack_factory
-        if $log
-          $log.warn('Deprecated method: this method is going to be deleted. Use Fluent::MessagePackFactory.engine_factory')
+        unless @deprecated_log_done
```
cosmo0920: This is just my curiosity. Feel free to ignore it.
Using single flag is reasonable for preventing duplicated warning logs?

test/test_msgpack_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class MessagePackFactoryTest < Test::Unit::TestCase
+  test 'call log.warn only once' do
+    klass = Class.new do
+      include Fluent::MessagePackFactory::Mixin
+    end
+
+    mp = klass.new
+
+    mock.proxy($log).warn(anything).once
```
cosmo0920: :+1:

lib/fluent/msgpack_factory.rb
ganmacs: I think printing at least one warning is enough becuase the point of this warning is "Do not use Fluent::MsgpackFactory::Mixin".

cosmo0920: Sounds reasonable.

repeatedly:
After verified by user, merge the patch :)

ganmacs:
https://github.com/fluent/fluentd/issues/2871#issuecomment-602882428 
It seems to be all right!

Issue #2735, opened 2019-12-12
Daemon doesn't work without supervision


jkohen:
**Describe the bug**
When fluentd runs with `--daemon --no-supervisor`, it ignores `--daemon`. The supervisor doesn't run, as expected.

**To Reproduce**
Run fluentd with `--daemon file.pid --no-supervisor`.

**Expected behavior**
Fluentd runs without the supervisor (single process, single worker) and goes to the background, writing its pid to `file.pid`.

**Your Environment**

- Fluentd or td-agent version: `fluentd --version` or `td-agent --version`
- Operating system: `cat /etc/os-release`
- Kernel version: `uname -r`

If you hit the problem with older fluentd version, try latest version first.

I used 1.6.3, but based on the code at https://github.com/fluent/fluentd/blob/master/lib/fluent/supervisor.rb, this shouldn't work at HEAD either.

**Additional context**

I filed this as a bug because nothing in the docs (https://docs.fluentd.org/deployment/command-line-option) says these options shouldn't work together, and fluentd accepts them. One could argue it's a feature request, because nothing in the implementation (https://github.com/fluent/fluentd/blob/master/lib/fluent/supervisor.rb) seems to indicate that this should work.

ganmacs:
I've confirmed it. will fix it later. Thank you for reporting!

ganmacs:
https://github.com/fluent/fluentd/pull/2912 should fix this patch.

PR #2912, opened 2020-03-25
Enable combination daemon and no supervisor

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2735

fix bugs when a user gives  `--daemon` and `--no-supervisor` as options.

**Docs Changes**:

no need

**Release Note**: 

no need

### DIFF
--- a/lib/fluent/daemonizer.rb
+++ b/lib/fluent/daemonizer.rb
@@ -1,0 +1,88 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


lib/fluent/daemonizer.rb
```diff
@@ -0,0 +1,88 @@
...
+      begin
+        Process.daemon(false, false)
+
+        File.write(pid_fullpath, Process.pid.to_s)
+
+        # install signal and set process name are performed by supervisor
+        install_at_exit_handlers(pid_fullpath)
+
+        yield
+      rescue NotImplementedError
```
repeatedly: When this exception happens?

ganmacs: It's for the environments that don't support fork(2) such as Windows.

repeatedly: I see.

cosmo0920: Can we use `Fluent.windows?` branch instead of `rescue NotImplementedError` guard?

ganmacs: I decided to use rescue instead of Fluent.windows? because using spawn here is not for windows but for the platforms that don’t support fork(2) and we now support many platforms. So I think rescue is proper in this context.

cosmo0920: I got it. Thanks!

Issue #2918, opened 2020-03-27
in_tail pos_file_compaction_interval corrupts position files


juliantaylor:
**Describe the bug**
when using the in_tail position file compaction via `pos_file_compaction_interval`  the position files become corrupted

**To Reproduce**
Reproduced on a kubernetes cluster running compaction every 24 seconds and several pods in crashloop backoff so there was stuff to compact.

**Expected behavior**
no corruption

**Your Environment**

    Fluentd or td-agent version: 1.9.2 td-agent 3.6.0
    Kernel version: uname -r

**Your Configuration**
standard in_tail configuration with `pos_file_compaction_interval 24`

**Your Error Log**
The corrupted position file looks like this, note the `ffffffffffffffff` part with the missing space

```
/var/log/containers/container-1c15005e5e6978f4652c9bcce679f416654fb6d9e304a0d5d3b476b3c4bfd734.log	0000000000031a96	00000000008f589b
ffffffffffffffff448/var/log/containers/container-3f06aac20b6a5b1eadee53f4a891fcfbc3f9c365fcf649c6a457b063bbb73671.log	0000000000016215	00000000008f58ca
/var/log/containers/container-319527df0aaf8a2c698c775340981126709f67e05e29566ab7689d72001a7b43.log	00000000000330ed	00000000008f5879
```

there are actually lots of null bytes added in the problematic spot

```
[...]

PR #2922, opened 2020-03-31
Update seek position of position file entry

ganmacs:
Fixes https://github.com/fluent/fluentd/issues/2918

Update seek position of position file entry after compaction. if some position files are compacted, others' seek position change.

**Docs Changes**:

no need

**Release Note**: 

same as the title.


ganmacs:
@repeatedly ?

PR #2923, opened 2020-04-01
Fix regression in the case with_priority and rfc5424

ganmacs:
Fixes https://github.com/treasure-data/omnibus-td-agent/issues/245

Convert string to integer of pri in syslog parser.

**Docs Changes**:

no need

**Release Note**: 

same as the title.

### DIFF
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -175,7 +175,7 @@ module Fluent
 
         if @with_priority
           if (m = RFC5424_PRI_REGEXP.match(text))
-            record['pri'] = m['pri']
+            record['pri'] = m['pri'].to_i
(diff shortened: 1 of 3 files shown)


repeatedly:
Could you add `pri` check here: https://github.com/fluent/fluentd/blob/297696b486a31b6571345fd03fe11ac6f0b3aefc/test/plugin/test_parser_syslog.rb#L453
rfc3164 case checks it: https://github.com/fluent/fluentd/blob/297696b486a31b6571345fd03fe11ac6f0b3aefc/test/plugin/test_parser_syslog.rb#L438

ganmacs:
https://github.com/fluent/fluentd/pull/2923/commits/167eb7c4ef82ef8336b8401b0c65d8663744b002 fixed

PR #2930, opened 2020-04-03
Allow cert chains in client auth

ganmacs:
ref: https://github.com/fluent/fluentd/issues/2699
follow up: https://github.com/fluent/fluentd/pull/2700

**Docs Changes**:

no need

**Release Note**:


aggarwalShivani:
Hi @repeatedly , @cosmo0920 

Will this fix be applicable to communication between fluentd and opensearch using [fluent-plugin-opensearch](https://github.com/fluent/fluent-plugin-opensearch) , when client-cert validation is enabled on the server (Opensearch) side?


The client (fluentd) certificate is a chained cert - like CN=fluentd, signed by CN=intermediate signed by CN=rootca. 
If I pass CN=fluentd and CN=intermediate as the client_cert, shouldn't fluentd send both the certs in the request and not just CN=fluentd?
Is there any additional config required to enable this extra_chain_cert?

PR #2937, opened 2020-04-08
log: Add ignore_repeated_log_interval parameter

repeatedly:

No issue

Specify interval to ignore repeated log/stacktrace messages.

Under high loaded environment, output destination sometimes becomes unstable and it causes lots of logs with same message. This parameter mitigates such situation.

**Docs Changes**:
Add `ignore_repeated_log_interval` to system config.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -509,7 +509,7 @@ module Fluent
 
       if @ignore_repeated_log_interval
         if ignore_repeated_log?(:last_repeated_log, time, message)
-          return nil
+          return nil, nil
         else
           Thread.current[:last_repeated_log] = CachedLog.new(message, time)
         end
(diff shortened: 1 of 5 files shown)


test/test_log.rb
```diff
@@ -367,6 +367,52 @@ def test_different_log_level
...
 
+  sub_test_case "ignore_repeated_log_interval" do
+    def test_same_message
+      message = "This is test"
+      logger = ServerEngine::DaemonLogger.new(@log_device, {log_level: ServerEngine::DaemonLogger::INFO})
+      log = Fluent::Log.new(logger, {ignore_repeated_log_interval: 5})
+
+      log.error message
+      10.times { |i|
+        Timecop.return
```
ganmacs, line 378: Does it need?

lib/fluent/log.rb
```diff
@@ -479,6 +507,14 @@ def event(level, args)
         end
       }
 
+      if @ignore_repeated_log_interval
+        if ignore_repeated_log?(:last_repeated_log, time, message)
+          return nil
```
ganmacs, line 512: nits
```suggestion
          return nil, nil
```

PR #2942, opened 2020-04-09
Fix a bug in in_syslog octet-counting mode

ganmacs:
Follow up https://github.com/fluent/fluentd/pull/2771

no need

**Docs Changes**:

no need

**Release Note**: 

same as title.


repeatedly:
No need to add test for `\n`?

ganmacs:
> No need to add test for \n?

I think so. For my understanding, this bug is off-by-one counting issue and is fixed in https://github.com/fluent/fluentd/pull/2942/commits/affd42ba34142de823aa677621b0f34c65559d63
So it's not related to `\n`.

Issue #2926, opened 2020-04-02
memory usage seems to be growing indefinitely with simple forward configuration


sp0f:
**Describe the bug**
Memory usage is growing indefinitely with very simple @type forward configuration. I'm shipping about 10-30G of logs daily. Mostly during evening hours. td-agent at the beginning is using about 100M of RES memory but after about 12h it's 500M and half of this time it's mostly idled because there are not many logs during the night. Memory is never freed. When I switch to any 'local' output like file or stdout memory usage is very stable. I've seen the same behavior with elasticsearch output so I guess it can be something connected with network outputs ... or just my stupidity :) There are no errors in the log file. I've tried to set RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0.9 but it didn't fix the problem. Memory usage is still growing. I have two workers in my configuration This problem only occurs on the one with a high amount of logs generated during day hours (worker 0). A worker is reading about 50 files at once. It may be relevant that I have a lot of `pattern not matched` - I'm in the middle of standardizing log format for all apps.

**To Reproduce**
Run with a high amount of not perfectly formate logs.

**Expected behavior**
Stable memory usage.

**Your Environment**

- Fluentd or td-agent version: `td-agent 1.9.2`
- Operating system: `NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"`
- Kernel version: `4.14.171-136.231.amzn2.x86_64`

**Your Configuration**

```
<system>
  workers 2
  log_level warn
</system>

<worker 0>
  <source>
    @type tail
    path "/somepath/*/*.log"
    read_from_head true
    pos_file /var/log/td-agent/mpservice-es-pos-file
    tag mpservice-raw.*
    enable_stat_watcher false
    <parse>
      @type multiline
      time_key time
      time_format %Y-%m-%d %H:%M:%S,%L
      timeout 0.5
      format_firstline /^\d{4}-\d{2}-\d{2}/
      format1 /^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}) \| (?<level>.*) \| (?<class>.*) \| (?<thread>.*) \| (?<message>.*)/

    </parse>
[...]

PR #2943, opened 2020-04-10
Creating timer of purging obsolete sockets if keepalive_timeout is nil

ganmacs:
ref https://github.com/fluent/fluentd/issues/2926

If connection sockets raise some errors, the cached socket is revoked.
Revoked cached socket(`@inactive_sockets`) must be deleted by
purge_obsolete_socks method so that memory usage doesn't increase.

https://github.com/fluent/fluentd/blob/0e901e7974232dd05f098ad74ea5be956aaba0e3/lib/fluent/plugin/out_forward/connection_manager.rb#L99
https://github.com/fluent/fluentd/blob/0e901e7974232dd05f098ad74ea5be956aaba0e3/lib/fluent/plugin/out_forward/socket_cache.rb#L63

This memory leak happens when keepalive and user_auth are on.
The memory leaking is easier to detect in TLS connection than in TCP connection since a lot of error can happen in [here](https://github.com/fluent/fluentd/blob/0e901e7974232dd05f098ad74ea5be956aaba0e3/lib/fluent/plugin/out_forward/connection_manager.rb#L99).

sender
```
<source>
  @type dummy
  size 1000
  rate 450
  tag test
</source>

<match **>
  @type forward

  keepalive true

  <server>
    host 127.0.0.1
    port 24284

    password testpassword
    username testuser
  </server>

  <buffer>
    @type file
    chunk_limit_size 1m
    flush_at_shutdown true
    flush_interval 1s
    path "#{Dir.pwd}/example/certs/buf"
    total_limit_size 128m
  </buffer>

  <security>
    self_hostname client
    shared_key sharedkeypassword
  </security>

  transport tls
  tls_verify_hostname false
  tls_cert_path "#{Dir.pwd}/example/certs/server.pem"
  tls_insecure_mode false
</match>
```

receiver
```
<source>
  @type forward
  port 24284

  <security>
    self_hostname server
    shared_key sharedkeypassword

    <user>
      password testpassword
      username testuser
    </user>

    user_auth true
  </security>

  <transport tls>
    cert_path "#{Dir.pwd}/example/certs/server.pem"
    insecure false
    private_key_path "#{Dir.pwd}/example/certs/server-key.pem"
    private_key_passphrase hoge
  </transport>
</source>

<match test>
  @type null
</match>
```

**Docs Changes**:

no need

**Release Note**: 

same as title

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -324,7 +324,7 @@ module Fluent::Plugin
         end
       end
 
-      if @keepalive && @keepalive_timeout
+      if @keepalive
         timer_execute(:out_forward_keep_alived_socket_watcher, @keep_alive_watcher_interval, &method(:on_purge_obsolete_socks))
       end
     end
--- a/test/plugin/test_out_forward.rb
+++ b/test/plugin/test_out_forward.rb
@@ -1200,6 +1200,15 @@ EOL
       end
     end
 
+    test 'create timer of purging obsolete sockets' do
+      output_conf = CONFIG + %[keepalive true]
+      d = create_driver(output_conf)
+
+      mock(d.instance).timer_execute(:out_forward_heartbeat_request, 1).once
+      mock(d.instance).timer_execute(:out_forward_keep_alived_socket_watcher, 5).once
+      d.instance_start
+    end
+
     sub_test_case 'with require_ack_response' do
       test 'Create connection per send_data' do
         target_input_driver = create_target_input_driver(conf: TARGET_CONFIG)


cosmo0920:
I also confirmed memory leaks do not occur with the following configuration:

```
% bundle exec fluent-ca-generate certs hoge 
```

```aconf
<system>
  workers 2
</system>
<worker 0>
  <source>
    @type dummy
    size 1000
    rate 450
    tag test
  </source>

  <match **>
    @type forward

    keepalive true

    <server>
      host 127.0.0.1
      port 24284

      password testpassword
      username testuser
    </server>

    <buffer>
      @type file
      chunk_limit_size 1m
      flush_at_shutdown true
      flush_interval 1s
      path "#{Dir.pwd}/certs/buf"
      total_limit_size 128m
    </buffer>

    <security>
      self_hostname client
      shared_key sharedkeypassword
    </security>

    transport tls
    tls_verify_hostname false
    tls_cert_path "#{Dir.pwd}/certs/ca_cert.pem"
    tls_insecure_mode false
  </match>
</worker>
<worker 1>
  <source>
    @type forward
    port 24284

    <security>
      self_hostname server
      shared_key sharedkeypassword

      <user>
        password testpassword
        username testuser
      </user>

      user_auth true
    </security>

    <transport tls>
      ca_cert_path "#{Dir.pwd}/certs/ca_cert.pem"
      insecure false
      ca_private_key_path "#{Dir.pwd}/certs/ca_key.pem"
      ca_private_key_passphrase hoge
      # client_cert_auth true
    </transport>
  </source>

  <match test>
    @type null
  </match>
</worker>
```

PR #2959, opened 2020-04-17
windows: Disable specifying linger_timeout again

cosmo0920:

None

Enabling SO_LINGER causes tcp port exhaustion on Windows.
This is because dynamic ports are only 16384 (from 49152 to 65535) and
expiring SO_LINGER enabled ports should wait 4 minutes where set by TcpTimeDelay.
Its default value is 4 minutes.
So, we should disable SO_LINGER on Windows to prevent flood of waiting ports.

**Docs Changes**:
No needed.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -382,7 +382,12 @@ module Fluent::Plugin
           cert_logical_store_name: @tls_cert_logical_store_name,
           cert_use_enterprise_store: @tls_cert_use_enterprise_store,
 
-          linger_timeout: @send_timeout,
+          # Enabling SO_LINGER causes tcp port exhaustion on Windows.
+          # This is because dynamic ports are only 16384 (from 49152 to 65535) and
+          # expiring SO_LINGER enabled ports should wait 4 minutes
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_forward.rb
```diff
@@ -382,7 +382,12 @@ def create_transfer_socket(host, port, hostname, &block)
           cert_logical_store_name: @tls_cert_logical_store_name,
           cert_use_enterprise_store: @tls_cert_use_enterprise_store,
 
-          linger_timeout: @send_timeout,
```
ganmacs, line 385: Does [L399](https://github.com/fluent/fluentd/pull/2959/files#diff-40c0e8c88efdc9a297a9fc63bc2bc69fR399) need the same change?

cosmo0920: I think L399 needn't to apply same change.

cosmo0920:
> Do we still need this change? 7d16531

I'm considering about it.
Enabling linger_timeout should be enabled with keepalive option.
But I do not have enough evidence which is enabling linger_timeout with keepalive is valid on Windows....

PR #2961, opened 2020-04-18
Avoid duplicates in path config

vimalk78:
Avoid duplicates in path config
 - removed duplicate in path config
 - added testcase

Fixes # None

**What this PR does / why we need it**: Avoid duplicates in path config

**Docs Changes**: Needed

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -136,7 +136,7 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "#{rc} are reserved words: #{@path_delimiter}"
       end
 
-      @paths = @path.split(@path_delimiter).map(&:strip)
+      @paths = @path.split(@path_delimiter).map(&:strip).uniq
       if @paths.empty?
         raise Fluent::ConfigError, "tail: 'path' parameter is required on tail input"
       end
@@ -296,7 +296,7 @@ module Fluent::Plugin
                end
         path.include?('*') ? Dir.glob(path) : path
       }.flatten.uniq
-      paths - excluded
+      paths.uniq - excluded
     end
 
     # in_tail with '*' path doesn't check rotation file equality at refresh phase.
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -102,6 +102,12 @@ class TailInputTest < Test::Unit::TestCase
       assert_equal ["tail.txt", "test2", "tmp,dev"], d.instance.paths
     end
 
+    test "multi paths with same path configured twice" do
+      c = config_element("ROOT", "", { "path" => "test1.txt,test2.txt,test1.txt", "tag" => "t1", "path_delimiter" => "," })
+      d = create_driver(c + PARSE_SINGLE_LINE_CONFIG, false)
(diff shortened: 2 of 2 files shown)


vimalk78:
how to fix the CI errors

lib/fluent/plugin/in_tail.rb
```diff
@@ -136,7 +136,7 @@ def configure(conf)
         raise Fluent::ConfigError, "#{rc} are reserved words: #{@path_delimiter}"
       end
 
-      @paths = @path.split(@path_delimiter).map(&:strip)
+      @paths = @path.split(@path_delimiter).map(&:strip).sort().uniq()
```
ganmacs, line 139: sort is unnecessary
```suggestion
      @paths = @path.split(@path_delimiter).map(&:strip).uniq
```

vimalk78:
> > how to fix the CI errors
> 
> CI errors seem to have nothing with this PR. so just ignore it.
> 
> 
> `@paths` is expanded with `*` or time. So I believe that `uniq` is executed at [here](https://github.com/fluent/fluentd/blob/eaf68654d4d80071a4a149daab5579f094f8ce37/lib/fluent/plugin/in_tail.rb#L308)

The `uniq` is executed on the `@exclude_path`

ganmacs:
Ah, you are right. then `path.uniq - excluded`
https://github.com/fluent/fluentd/blob/eaf68654d4d80071a4a149daab5579f094f8ce37/lib/fluent/plugin/in_tail.rb#L299

vimalk78:
perhaps testcase should assert multiple TailWatchers are not created for same path

vimalk78:
@ganmacs please check the changes.

```
  def create_driver(conf = SINGLE_LINE_CONFIG, use_common_conf = true)
    config = use_common_conf ? COMMON_CONFIG + conf : conf
    Fluent::Test::Driver::Input.new(Fluent::Plugin::TailInput).configure(config)
  end
```
while running the `test_in_tail.rb` , i found that input to `create_driver` already has `path` expanded.
If `path => /var/log/*.log`, the method will receive expanded value of path `/var/log/something/somethod.log`

So i could not fail the `expand_path` test i added. Still i hope the changes are useful.

lib/fluent/plugin/in_tail.rb
```diff
@@ -136,7 +136,7 @@ def configure(conf)
         raise Fluent::ConfigError, "#{rc} are reserved words: #{@path_delimiter}"
       end
 
-      @paths = @path.split(@path_delimiter).map(&:strip)
+      @paths = @path.split(@path_delimiter).map(&:strip).uniq()
```
ganmacs, line 139: unnecessary `()`
```suggestion
      @paths = @path.split(@path_delimiter).map(&:strip).uniq
```

vimalk78: updated

vimalk78:
how to fix the CI failure? need this to pass for PR merge?

ganmacs:
No, This PR has nothing to do with the failure. Just ignore it.

vimalk78:
Anything else i need to do get it merged?

repeatedly:
Looks good.

ganmacs:
Thank you 👍

PR #2962, opened 2020-04-18
Log typo

thejonanshow:
Thanks for the great project :)

No issue, just a typo

This changes a word. You need it because the other word is misspelled.

**Docs Changes**:
I don't think this log line needs any documentation.

**Release Note**: 
Vitally important log message update.


ganmacs:
Thanks!

PR #2967, opened 2020-04-21
removing git dependency

eduardoroliveira:
Fixes #
Removed the git dependency from  the project

We faced a problem installing fluentd from the source in a server that was isolated from internet and did not have git installed. Since the git is only listing the files, this change for a bash code would be enough and could remove that dependency

**Docs Changes**:

**Release Note**:


ganmacs:
-1

The behavior of `find`, `cut` and `grep` are different between OSs(macOS, Linux, Window)
It can cause unexpected behavior.

kenhys:
It seems that this PR is stalled so, I'll close it.

PR #2973, opened 2020-04-28
PoC: json_array switch to use json array data format for HTTP request body

### LINKED ISSUE #2909: Use out_http to send logs to in_http

I have 2 managed k8s clusters with individual fluentd services and want to forward some logs from cluster_1 to cluster_2. I can't use the `forward` protocol because the ingress does not allow raw TCP forwarding.

Instead I tried to use `out_http` in cluster_1 and `in_http` in cluster_2, since HTTP based ingress configurations are easy to do. But I couldn't find a working configuration, all failed with one of:
- `400 Bad Request\nReceived event is not json`
- `400 Bad Request\n'json' or 'msgpack' parameter is required`
- Server: `incoming event is invalid: path=/some.tag params={"json":"{\"hello\":\"world\",\"num\":18}\n{\"hello\":\"world\",\"num\":19}\n",...}`

There should be an easy configuration to allow using http to forward events between 2 fluentd instances.
I see 2 possible solutions:
- add an option to `out_http` to optionally wrap (json) events in an (json) array
- add an option to `in_http`/`parse_json` to be able to handle the json lines format used by `out_http`

Declare the `out_http -> in_http` case to be out of scope. In that case there should be some mention of it in the docs, to save the next person time. Two plugins which communicate over HTTP but can't talk to one another is very counter-intuitive in my opinion. 

**Additional context**
Configuration for reproducing the problem.
```
# sending side
<source>
  @type dummy
  dummy {"hello":"world"}
  auto_increment_key num
  tag kube.test.tag
</source>

<match **>
  @type http
  endpoint http://localhost:9881/$tag
  open_timeout 2
  #content_type application/json #results in 'invalid event' if uncommented
  <format>
    @type json
  </format>
  <buffer>
    flush_interval 2s
  </buffer>
</match>
```

```
# receiving side
<source>
  @type http
  port 9881
  bind 0.0.0.0
  <parse> # comment out for 'json parameter required'
    @type json
  </parse>
</source>
<match **>
  @type stdout
</match>
```

roman-geraskin:
Signed-off-by: Roman Geraskin <roman.n.geraskin@gmail.com>

Fixes #2909 

This PR implements the `json_array` option (bool, default false) for the out_http plugin. With this option, we can use the JSON array data format for the HTTP request body.

**Docs Changes**:
Docs update required.

**Release Note**: 
This is quick and easy feature implementation, sort of proof of concept. I will be grateful for any feedback.

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -37,6 +37,8 @@ module Fluent::Plugin
     config_param :proxy, :string, default: ENV['HTTP_PROXY'] || ENV['http_proxy']
     desc 'Content-Type for HTTP request'
     config_param :content_type, :string, default: nil
+    desc 'JSON array data format for HTTP request body'
+    config_param :json_array, :bool, default: false
     desc 'Additional headers for HTTP request'
     config_param :headers, :hash, default: nil
 
@@ -100,6 +102,13 @@ module Fluent::Plugin
       @proxy_uri = URI.parse(@proxy) if @proxy
       @formatter = formatter_create
       @content_type = setup_content_type unless @content_type
+
+      if @json_array
+        if @formatter_configs.first[:@type] != "json"
+            raise Fluent::ConfigError, "json_array option could be used with json formatter only"
+        end
+        define_singleton_method(:format, method(:format_json_array))
+      end
     end
 
     def multi_workers_ready?
@@ -112,6 +121,10 @@ module Fluent::Plugin
 
     def format(tag, time, record)
       @formatter.format(tag, time, record)
+    end
+
+    def format_json_array(tag, time, record)
+      @formatter.format(tag, time, record).strip() << ","
     end
 
     def write(chunk)
@@ -128,7 +141,7 @@ module Fluent::Plugin
     def setup_content_type
       case @formatter_configs.first[:@type]
       when 'json'
-        'application/x-ndjson'
+        @json_array ? 'application/json' : 'application/x-ndjson'
       when 'csv'
         'text/csv'
       when 'tsv', 'ltsv'
@@ -202,7 +215,7 @@ module Fluent::Plugin
         req.basic_auth(@auth.username, @auth.password)
       end
       set_headers(req)
-      req.body = chunk.read
+      req.body = @json_array ? "[#{chunk.read[0...-1]}]" : chunk.read
       req
     end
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_http.rb
```diff
@@ -202,7 +215,7 @@ def create_request(chunk, uri)
         req.basic_auth(@auth.username, @auth.password)
       end
       set_headers(req)
-      req.body = chunk.read
+      req.body = @json_array ? "[#{chunk.read[0...-1]}]" : chunk.read
```
repeatedly, line 218: Why use `[0...-1]`? `#read` returns String object.

roman-geraskin: I use `[0...-1]` to exclude the last character from the string:
- `"[{<data>},{<data>},]"` is not valid json string
- `"[{<data>},{<data>}]"` is valid json string afaik

Should I use another approach?

lib/fluent/plugin/out_http.rb
```diff
@@ -114,6 +123,10 @@ def format(tag, time, record)
       @formatter.format(tag, time, record)
     end
 
+    def format_json_array(tag, time, record)
+      @formatter.format(tag, time, record).strip() << ","
```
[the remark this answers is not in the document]
roman-geraskin: Yes, I use `strip` to ignore the `add_newline` setting value. I don't know will every endpoint parser be ok with such JSON payload:
```
[{<data>},
{<data>},
{<data>}]
```

So I try to produce a valid unformatted JSON. I could get rid of `strip` so plugin users should solve endpoint problems in a particular case themself. How do you think?

lib/fluent/plugin/out_http.rb
repeatedly: Ah, I see. If so, please use `chop!` method instead.
`[0...-1]` creates new object and it consumes bigger memory for large string.

repeatedly:
Please add test for `application/json` case in `test/plugin/test_out_http.rb`

lib/fluent/plugin/out_http.rb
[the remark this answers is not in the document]
repeatedly: I mean don't use `strip` for removing `\n`. setting `add_newline false` in `configure` is enough to remove `\n` from formatted result.

roman-geraskin:
Resolved all reviews, tests added. Please review. Thanks!

repeatedly:
Thanks!

roman-geraskin:
Thanks for reviewing!

PR #2990, opened 2020-05-12
in_forward: Set encoding us-ascii to source_host and source_address

okkez:
source_host and source_address should not be binary.

Signed-off-by: okkez <okkez000@gmail.com>

Fixes none

In the previous version, the values added by `source_hostname_key` and `souce_address_key` are encoded as `bin 8`. I expected `str N` or `fixstr`.

If `source_hostname` and `source_address` is encoded as `bin 8` (binary), we cannot convert records to parquet using `string` type defined by Avro schema. If we use `bytes` in Avro schema, we can convert records to parquet format but we cannot get human-readable dumped records from `*.parquet` file using parquet-tools.

In this version, encode `source_hostname` and `source_address` as `fixstr` or `str N` in msgpack.
We can use `string` type in Avro schema and we can convert records to parquet format. We can get human-readable dumped records from `*.parquet` file using parquet-tools.

**Docs Changes**: none

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_forward.rb
+++ b/lib/fluent/plugin/in_forward.rb
@@ -381,6 +381,12 @@ module Fluent::Plugin
       if @source_address_key && @source_hostname_key
         address = conn.remote_addr
         hostname = conn.remote_host
+        if address.ascii_only?
+          address.force_encoding(Encoding::US_ASCII)
+        end
+        if hostname.ascii_only?
+          hostname.force_encoding(Encoding::US_ASCII)
+        end
         es.each { |time, record|
           record[@source_address_key] = address
           record[@source_hostname_key] = hostname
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_forward.rb
```diff
@@ -379,21 +379,33 @@ def check_and_skip_invalid_event(tag, es, remote_host)
     def add_source_info(es, conn)
       new_es = Fluent::MultiEventStream.new
       if @source_address_key && @source_hostname_key
-        address = conn.remote_addr
-        hostname = conn.remote_host
+        address = conn.remote_addr.dup
+        hostname = conn.remote_host.dup
+        if address.ascii_only?
+          address.force_encoding(Encoding::US_ASCII)
```
cosmo0920, line 385: Just for my curiosity, it should be handled for other encoding such as cp932 or other encoding with other if clauses?
i.e. non ascii encoding cases will not be handled as binary with parquet format?

repeatedly:
When address and hostname are not only ascii?
In addition, current implementation creates lots of object for 1 stream and it is not memory friendly.
In-place encode change seems enough for me.

okkez:
> When address and hostname are not only ascii?

In my case, all hosts use only ascii .
If address and hostname are not only ascii, these are encoded as binary in the current implementation.
I think that automatic encoding detection may cause performance issues and there are a lot of encodings Ruby supported.
But we can add configuration to customize the parameter of `force_encoding`.

PR #2992, opened 2020-05-13
in_unix: Use v1 plugin API

repeatedly:

No issue

Use v1 plugin API in `in_unix` plugin.
In addition, add `tag` parameter to follow `in_forward`.

**Docs Changes**:
Add tag parameter

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/in_unix.rb
+++ b/lib/fluent/plugin/in_unix.rb
@@ -45,7 +45,6 @@ module Fluent::Plugin
 
     def configure(conf)
       super
-      #log.warn "'unix' input is obsoleted and will be removed. Use 'forward' instead."
     end
 
     def start
--- a/lib/fluent/plugin/in_unix.rb
+++ b/lib/fluent/plugin/in_unix.rb
@@ -66,6 +66,7 @@ module Fluent::Plugin
 
     def listen
       if File.exist?(@path)
+        log.warn "Found existing '#{@path}'. Remove this file for in_unix plugin"
         File.unlink(@path)
       end
       FileUtils.mkdir_p(File.dirname(@path))
--- a/lib/fluent/plugin/in_unix.rb
+++ b/lib/fluent/plugin/in_unix.rb
@@ -133,10 +133,10 @@ module Fluent::Plugin
     end
 
     def convert_time(time)
-      case
-      when time.nil? || (time == 0)
+      case time
+      when nil, 0
         Fluent::EventTime.now
-      when time === Fluent::EventTime
+      when Fluent::EventTime
         time
       else
         Fluent::EventTime.from_time(Time.at(time))
--- a/lib/fluent/plugin/in_unix.rb
+++ b/lib/fluent/plugin/in_unix.rb
@@ -14,49 +14,67 @@
 #    limitations under the License.
 #
 
-require 'fileutils'
-require 'socket'
+require 'fluent/env'
+require 'fluent/plugin/input'
+require 'fluent/msgpack_factory'
 
 require 'cool.io'
 require 'yajl'
+require 'fileutils'
+require 'socket'
 
-require 'fluent/input'
-require 'fluent/event'
+module Fluent::Plugin
+  # TODO: This plugin will be 3rd party plugin
+  class UnixInput < Input
+    Fluent::Plugin.register_input('unix', self)
 
-module Fluent
-  # obsolete
-  class StreamInput < Input
-    config_param :blocking_timeout, :time, default: 0.5
+    helpers :event_loop
+
+    def initialize
+      super
+
+      @lsock = nil
+    end
+
+    desc 'The path to your Unix Domain Socket.'
+    config_param :path, :string, default: Fluent::DEFAULT_SOCKET_PATH
+    desc 'The backlog of Unix Domain Socket.'
+    config_param :backlog, :integer, default: nil
+    desc "New tag instead of incoming tag"
+    config_param :tag, :string, default: nil
+
+    def configure(conf)
+      super
+      #log.warn "'unix' input is obsoleted and will be removed. Use 'forward' instead."
+    end
 
     def start
       super
 
-      @loop = Coolio::Loop.new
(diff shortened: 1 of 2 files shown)


ganmacs:
> Use v1 plugin API in in_unix plugin.
> In addition, add tag parameter to follow in_forward.

These two things might be in different PR. it's hard to review... (at least, please split them into different commits)

lib/fluent/plugin/in_unix.rb
```diff
@@ -14,50 +14,68 @@
...
+    desc 'The path to your Unix Domain Socket.'
+    config_param :path, :string, default: Fluent::DEFAULT_SOCKET_PATH
+    desc 'The backlog of Unix Domain Socket.'
+    config_param :backlog, :integer, default: nil
+    desc "New tag instead of incoming tag"
+    config_param :tag, :string, default: nil
+
+    def configure(conf)
+      super
+      #log.warn "'unix' input is obsoleted and will be removed. Use 'forward' instead."
```
ganmacs, line 48: is it necessary?

lib/fluent/plugin/in_unix.rb
```diff
@@ -14,50 +14,68 @@
...
+      end
 
       super
     end
 
-    #def listen
-    #end
+    def listen
+      if File.exist?(@path)
+        File.unlink(@path)
```
ganmacs, line 69: It might be better to print a warning log.

lib/fluent/plugin/in_unix.rb
```diff
@@ -80,22 +98,22 @@ def run
     # }
     def on_message(msg)
       # TODO format error
-      tag = msg[0].to_s
+      tag = @tag || (msg[0].to_s)
```
ganmacs, line 101: should it check if tag is `nil`?

lib/fluent/plugin/in_unix.rb
```diff
@@ -80,22 +98,22 @@ def run
...
+      when Array
         # Forward
-        es = MultiEventStream.new
+        es = Fluent::MultiEventStream.new
         entries.each {|e|
           record = e[1]
           next if record.nil?
-          time = e[0]
-          time = (now ||= EventTime.now) if time.to_i == 0
+          time = convert_time(e[0])
```
ganmacs, line 116: since in #convert_time check `time == 0`,  it's `e[0].to_i` ?

lib/fluent/plugin/in_unix.rb
```diff
@@ -105,39 +123,42 @@ def on_message(msg)
         record = msg[2]
         return if record.nil?
 
-        time = msg[1]
-        time = EventTime.now if time.to_i == 0
+        time = convert_time(msg[1])
```
ganmacs, line 126: `msg[1].to_i` ?

ganmacs:
Writing tests about tag parameter would be great.

lib/fluent/plugin/in_unix.rb
```diff
@@ -14,50 +14,68 @@
...
+    end
 
     def start
       super
 
-      @loop = Coolio::Loop.new
       @lsock = listen
-      @loop.attach(@lsock)
-      @thread = Thread.new(&method(:run))
+      event_loop_attach(@lsock)
```
cosmo0920, line 55: :+1:

lib/fluent/plugin/in_unix.rb
repeatedly: Maybe, msg type check is needed. Will add it.

lib/fluent/plugin/in_unix.rb
repeatedly: No. v1 supports nanosecond, so Float and EventTime are also supported.
`== 0` works with these types.

lib/fluent/plugin/in_unix.rb
repeatedly: ditto

lib/fluent/plugin/in_unix.rb
ganmacs: if `e[0]` is `nil`(I'm not sure this happens), should we do it?

lib/fluent/plugin/in_unix.rb
repeatedly: Can be removed. Will do

lib/fluent/plugin/in_unix.rb
repeatedly: Hmm... conflict path is rare case. Need warning?

ganmacs: I think so. if this file exists, it means other processes may use it.

repeatedly:
Applied reviews.

lib/fluent/plugin/in_unix.rb
```diff
@@ -105,39 +127,42 @@ def on_message(msg)
...
 
+    def convert_time(time)
+      case
+      when time.nil? || (time == 0)
+        Fluent::EventTime.now
+      when time === Fluent::EventTime
+        time
+      else
+        Fluent::EventTime.from_time(Time.at(time))
+      end
```
ganmacs, line 143: [nits]
```suggestion
  case time
  when nil, 0
    Fluent::EventTime.now
  when Fluent::EventTime
    time
  else
    Fluent::EventTime.from_time(Time.at(time))
  end
```

test/plugin/test_in_unix.rb
```diff
@@ -142,4 +142,20 @@ def test_message_json
...
+    @d = create_driver
+    @d.run(shutdown: false, timeout: 5) do
+      @d.instance.__send__(:on_message, data)
+    end
+
+    assert_equal 0, @d.events.size
+
+    logs = @d.instance.log.logs
+    assert_equal 1, logs.select { |line|
+      line =~ / \[warn\]: incoming data is broken: msg=#{data.inspect}/
```
ganmacs, line 178: [nits] match?

PR #3018, opened 2020-06-02
buffer: Fix calculation of timekey stats

jiping-s:

The `oldest_timekey` stat in buffer is broken.

The current approach of using `add_timekey` and `del_timekey` to track the count of chunks in each timekey doesn't work as intended - the count is never decreased to zero because of complications in chunk creation, as a result the `oldest_timekey` stat is almost never updated after initial flush(es).

A simple and not-so-timely alternative, to perform recalculation in every flush cycle.

Signed-off-by: Ji-Ping Shen <ji-ping.shen@relexsolutions.com>

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -197,18 +197,13 @@ module Fluent
         @stage, @queue = resume
         @stage.each_pair do |metadata, chunk|
           @stage_size += chunk.bytesize
-          if chunk.metadata && chunk.metadata.timekey
-            add_timekey(metadata.timekey)
-          end
         end
         @queue.each do |chunk|
           @queued_num[chunk.metadata] ||= 0
           @queued_num[chunk.metadata] += 1
           @queue_size += chunk.bytesize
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks! We will check the patch.
But buffer tests failed, so it seems to break backward compatibility.

lib/fluent/plugin/buffer.rb
```diff
@@ -456,9 +447,21 @@ def enqueue_unstaged_chunk(chunk)
         end
       end
 
+      def update_timekeys
+        synchronize do
```
jiping-s, line 451: The `synchronize` here is not some Ruby keyword as I imagined. Please take a look - maybe there is something I could do to reduce the lock(s) needed. The method does have to loop through everything in stage and queue, though the consistency of data it reads doesn't matter (e.g. it's okay for the same buffer to be found in `stage` and `queue`).

repeatedly:
Sorry. I confirmed this patch works before but forgot to merge...
Thanks for the patch!

PR #3015, opened 2020-06-02
syslog_parser: Add string parser for rfc5424

repeatedly:
Signed-off-by: Masahiro Nakagawa <repeatedly@gmail.com>

**Which issue(s) this PR fixes**: 
Fixes #2997
Follow up of https://github.com/fluent/fluentd/pull/2599

**What this PR does / why we need it**: 
Add non-regexp parser to fix issues and improve the performance
New parser is 2x faster than current regexp based parser.

```
# code
require 'benchmark/ips'
require 'fluent/env'
require 'fluent/engine'
require_relative 'lib/fluent/plugin/parser_syslog'


log1 = '<16>1 2017-02-06T13:14:15.003Z 192.168.0.1 fluentd - - - Hi, from Fluentd!'
log2 = '<16>1 2017-02-06T13:14:15.003Z 192.168.0.1 fluentd 11111 ID24224 [exampleSDID@20224 iut="3" eventSource="Application" eventID="11211"] [Hi] from Fluentd!'

parser = Fluent::Plugin::SyslogParser.new
parser.configure(Fluent::Config::Element.new('ROOT', '', {'with_priority' => true, 'message_format' => 'rfc5424', 'keep_time_key' => true}, []))

Benchmark.ips do |x|
  x.report "regexp" do
    parser.parse_rfc5424_regex(log1) { |t, r| }
  end
  x.report "string" do
    parser.parse_rfc5424(log1) { |t, r| }
  end
  x.report "regexp with extra" do
    parser.parse_rfc5424_regex(log2) { |t, r| }
  end
  x.report "string with extra" do
    parser.parse_rfc5424(log2) { |t, r| }
  end
end

# result
Warming up --------------------------------------
              regexp    15.374k i/100ms
              string    35.795k i/100ms
   regexp with extra    11.660k i/100ms
   string with extra    27.389k i/100ms
Calculating -------------------------------------
              regexp    154.042k (± 9.1%) i/s -    768.700k in   5.051709s
              string    377.800k (± 8.6%) i/s -      1.897M in   5.078498s
   regexp with extra    127.052k (± 4.5%) i/s -    641.300k in   5.057835s
   string with extra    300.726k (± 3.9%) i/s -      1.506M in   5.017230s
```

**Docs Changes**:
Mention it in `parser_type` parameter.

**Release Note**: 
Same as title.

PR #3019, opened 2020-06-03
Improve syslog parser

repeatedly:
None

Improve the syslog parser code

- bit faster with `match?` for `auto` mode
- make code clean and remove unnecessary condition

**Docs Changes**:
No change.

**Release Note**: 
Improve syslog parser and add new parser for rfc5424
Merge #3015 and this into 1 release note.


repeatedly:
> creating classes

Yeah, it is one approach.

PR #3037, opened 2020-06-10
Allow the Fluentd Windows Service to be "Automatic (Delayed start)".

igorpeshansky:
This will allow the fluentd Windows service to be registered as delayed auto-start (https://docs.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_delayed_auto_start_info).

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -185,6 +185,10 @@ if Fluent.windows?
     opts[:regwinsvcautostart] = s
   }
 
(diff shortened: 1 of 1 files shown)


repeatedly:
Could you follow DCO?

@cosmo0920 Code looks good. What do you think?

igorpeshansky:
Sorry about that. Done.

igorpeshansky:
@cosmo0920 Happy to fix the docs, but I would imagine it needs to be documented right next to the `--reg-winsvc-auto-start` flag, and I can't find any mention of that flag in the docs... Sent fluent/fluentd-docs-gitbook#186 — does that address your request?

repeatedly:
Merged.

PR #3044, opened 2020-06-18
Fix child process failure due to SIGPIPE if the command uses stdout

iwasakims:
If I use exec output plugin with `command` which writes data to stdout (i.e. /bin/echo foo), the command fails due to SIGPIPE(13).

```
2020-06-18 20:00:08 +0900 [debug]: #0 Executing command title=:out_exec_process spawn=[{}, "/bin/echo foo.bar.baz foo /tmp/fluent-plugin-out-exec-20200618-8149-1hyzffl"] mode=[] stderr=:connect
...
2020-06-18 20:00:08 +0900 [warn]: #0 command exits with error code prog="/bin/echo foo.bar.baz foo /tmp/fluent-plugin-out-exec-20200618-8149-1hyzffl" status=nil signal=13
```

Setting the stdout to /dev/null via spawn_opts rather than after spawning process should fix this.

### DIFF
--- a/lib/fluent/plugin_helper/child_process.rb
+++ b/lib/fluent/plugin_helper/child_process.rb
@@ -259,6 +259,9 @@ module Fluent
 
         if !mode.include?(:stderr) && !mode.include?(:read_with_stderr)
           spawn_opts[:err] = IO::NULL if stderr == :discard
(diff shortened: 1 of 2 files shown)


repeatedly:
Thanks!

iwasakims:
Thanks, @repeatedly.

PR #3065, opened 2020-07-06
Replace whitelist and dummy terminology

repeatedly:
None

Follow twitter's engineering team changes.

https://twitter.com/TwitterEng/status/1278733305190342656

**Docs Changes**:
Add deprecation note to in_dummy.

**Release Note**: 
Same as title


ganmacs:
Do we need to take care of these files?

https://github.com/fluent/fluentd/blob/6beca80f6467fd2e12ea25b21e0474d978007b08/example/in_dummy_blocks.conf
https://github.com/fluent/fluentd/blob/6beca80f6467fd2e12ea25b21e0474d978007b08/example/in_dummy_with_compression.conf

repeatedly:
Good catch. I will apply change to example conf.

PR #3071, opened 2020-07-13
MatchPattern: allow regex patterns

### LINKED ISSUE #3057: Allow not - match - pattern
I'm having several input plugins, for example tags aaa.* bbb.* ccc.* etc
I would like to apply record_modifier to everything but aaa.*

```
<filter /(?!aaa).*/ >
   @type record_modifier
   ....
```

== or ==

```
<filter !aaa.**>
   @type record_modifier
   ....
```

explicitly filter for bbb.** ccc.** etc

**Additional context**
(?!aaa) doesn't work at the moment because every non-word character is escaped in 
https://github.com/fluent/fluentd/blob/master/lib/fluent/match.rb#L113

I see 3 possible, easy to implement solutions:
1. allow regex with a / prefix, string must end with /
`<filter /some-regex/>
`
2. don't escape "(?!" and the corresponding ")"

3. allow pattern like
!aaa.**

and create a regex (?!aaa).* from it

ferbar:
Fixes #3057

Allow a (plain) regex as MatchPattern in <filter .... >
With a regex - assertion it allows a pattern that matches a tag not starting with a string. I think the regex MatchPattern can also replace the commented-out carachter classes.

**Docs Changes**:
Following lines should be added to https://docs.fluentd.org/configuration/config-file#how-match-patterns-work:
Starting with fluentd <version> a match pattern may be a regex by enclosing the regex with slashes.

### Regex match patterns
A regex match pattern looks like <filter /regex/>
This allows character classes or match patterns _not_ starting with a string.
```
# match tags starting with a
<filter a.** >
   ....
</filter
# match all tags not starting with a.
<filter /(?!a\.).*/ >
   ....
</filter>
```
For a complete list of supported regex expressions see: https://ruby-doc.org/core-2.7.1/Regexp.html

**Release Note**: 
Add regex match patterns

**Things I've considered**:
* The tag is checked with @regex.match(tag-name). I don't think that there's a way to break something with a bad regex.
* This patch wraps the regex with \A and \Z. I think this would be a more convenient behavior than without it

### DIFF
--- a/lib/fluent/match.rb
+++ b/lib/fluent/match.rb
@@ -33,9 +33,9 @@ module Fluent
 
   class GlobMatchPattern < MatchPattern
     def initialize(pat)
-      if pat[0,1] == "/"
-        if pat[pat.length - 1,1] == "/"
-          @regex = Regexp.new("\\A"+pat[1, pat.length-2]+"\\Z")
+      if pat.start_with?('/')
+        if pat.end_with?('/')
+          @regex = Regexp.new("\\A"+pat[1..-2]+"\\Z")
           return
         else
           raise Fluent::ConfigError,  "invalid match - regex"
--- a/lib/fluent/match.rb
+++ b/lib/fluent/match.rb
@@ -33,6 +33,15 @@ module Fluent
 
   class GlobMatchPattern < MatchPattern
     def initialize(pat)
+      if pat[0,1] == "/"
+        if pat[pat.length - 1,1] == "/"
+          @regex = Regexp.new("\\A"+pat[1, pat.length-2]+"\\Z")
+          return
+        else
+          raise Fluent::ConfigError,  "invalid match - regex"
+        end
+      end
+
       stack = []
       regex = ['']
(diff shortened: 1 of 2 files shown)


lib/fluent/match.rb
```diff
@@ -33,6 +33,15 @@ def match(str)
 
   class GlobMatchPattern < MatchPattern
     def initialize(pat)
+      if pat[0,1] == "/"
+        if pat[pat.length - 1,1] == "/"
```
repeatedly, line 37: `pat.start_with?('/')` and `pat.end_with?('/')` are better.

lib/fluent/match.rb
```diff
@@ -33,6 +33,15 @@ def match(str)
...
+          @regex = Regexp.new("\\A"+pat[1, pat.length-2]+"\\Z")
```
repeatedly, line 38: `pat[1..-2]` is better

repeatedly:
Thanks!

PR #3100, opened 2020-08-07
Upgrade win32 related gem dependency for Windows

cosmo0920:
Our td-agent 4 building system can detect dependent gem versions with
bundler.
So, we should update dependent gem versions to be "bundler friendly".

None. But it is needed for releasing td-agent 4.0.1.

Our new td-agent building system detects dependent gem versions.
But, currently, Fluentd locks win32 gems' version.

**Docs Changes**:
No need.
**Release Note**: 
Same as title.


repeatedly:
Thx!

PR #3101, opened 2020-08-10
out_http: adding support for intermediate certificates

tyarimi:
Signed-off-by: Tzach Yarimi <tyarimi@salesforce.com>

Fixes #

Add support for sending a client certificate chain for mutual TLS authentication in the `out_http` plugin.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -21,6 +21,14 @@ require 'fluent/tls'
 require 'fluent/plugin/output'
 require 'fluent/plugin_helper/socket'
 
+# patch Net::HTTP to support extra_chain_cert
+class Net::HTTP
+  SSL_IVNAMES << :@extra_chain_cert unless SSL_IVNAMES.include?(:@extra_chain_cert)
+  SSL_ATTRIBUTES << :extra_chain_cert unless SSL_ATTRIBUTES.include?(:extra_chain_cert)
+
+  attr_accessor :extra_chain_cert
+end
+
 module Fluent::Plugin
   class HTTPOutput < Output
     Fluent::Plugin.register_output('http', self)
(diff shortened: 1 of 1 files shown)


repeatedly:
@nurse This patch modifies Net::HTTP directly. Do you know this approach is safe or not?

tyarimi:
Hi folks. Any chance of merging this PR? We have been running it in production for some time now and it works fine.

nurse:
Of course it's not safe. It should use public APIs.

tyarimi:
@nurse `extra_chain_cert` was added in a later version of ruby. Is there any alternative?

nurse:
Well, it should add more comment for example referring https://github.com/ruby/ruby/commit/31af0dafba6d3769d2a39617c0dddedb97883712#diff-8c2ab8e0fb4f052e1d95ab6334e192c1
Also it should have a condition to avoid the monkey patch runs even if latest ruby.

tyarimi:
@nurse I added the comment, but I'm not sure how to implement the conditional patching (not a ruby expert). Can you please help with that?

lib/fluent/plugin/out_http.rb
```diff
@@ -171,7 +180,15 @@ def setup_http_option
         end
         if @tls_client_cert_path
           raise Fluent::ConfigError, "tls_client_cert_path is wrong: #{@tls_client_cert_path}" unless File.file?(@tls_client_cert_path)
-          opt[:cert] = OpenSSL::X509::Certificate.new(File.read(@tls_client_cert_path))
+
+          bundle = File.read(@tls_client_cert_path)
+          bundle_certs = bundle.scan(/-----BEGIN CERTIFICATE-----(?:.|\n)+?-----END CERTIFICATE-----/)
```
nobu, line 185: This regexp seems to match also the middle of a line, is this intentional?

lib/fluent/plugin/out_http.rb
```diff
@@ -21,6 +21,15 @@
...
 require 'fluent/plugin_helper/socket'
 
+# patch Net::HTTP to support extra_chain_cert which was added in Ruby feature #9758.
+# see: https://github.com/ruby/ruby/commit/31af0dafba6d3769d2a39617c0dddedb97883712
+class Net::HTTP
+  SSL_IVNAMES << :@extra_chain_cert unless SSL_IVNAMES.include?(:@extra_chain_cert)
+  SSL_ATTRIBUTES << :extra_chain_cert unless SSL_ATTRIBUTES.include?(:extra_chain_cert)
+
+  attr_accessor :extra_chain_cert
+end
```
nurse, line 31: With assuming SSL_ATTRIBUTES include :extra_chain_cert if SSL_IVNAMES include :@ extra_chain_cert, single condition can enclose the definitions.
```suggestion
unless SSL_IVNAMES.include?(:@extra_chain_cert)
  class Net::HTTP
    SSL_IVNAMES << :@extra_chain_cert 
    SSL_ATTRIBUTES << :extra_chain_cert
    attr_accessor :extra_chain_cert
  end
end
```

nurse (changes requested):

tyarimi:
Closing this PR in favor of #3146 (switching head branch)

PR #3119, opened 2020-09-01
Add ignore_same_log _interval parameter

repeatedly:
No issues

We noticed `ignore_repeated_log_interval` is not enough for some frequent log generation case.
For example, if plugin generates several log messages in one action, logs are not repeated:

```
# retry generates same error message with ignore_repeated_log_interval
def write(chunk)
  # process 1...
  log.error "message1"
  # process 2...
  log.error "message2"
end
```

`ignore_same_log_interval` resolves these cases. With this option, logger keeps to store all log messages for validation. The number of log message is limited, so increased memory usage is ignorable.

**Docs Changes**:
Add `ignore_same_log_interval` to system parameters.

**Release Note**: 
Same as title.

### DIFF
--- a/test/test_log.rb
+++ b/test/test_log.rb
@@ -411,6 +411,50 @@ class LogTest < Test::Unit::TestCase
     end
   end
 
+  sub_test_case "ignore_same_log_interval" do
+    teardown do
+      Thread.current[:last_same_log] = nil
+    end
+
+    def test_same_message
+      message = "This is test"
+      logger = ServerEngine::DaemonLogger.new(@log_device, {log_level: ServerEngine::DaemonLogger::INFO})
+      log = Fluent::Log.new(logger, {ignore_same_log_interval: 5})
+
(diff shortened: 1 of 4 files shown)


test/test_log.rb
```diff
@@ -411,6 +411,50 @@ def test_different_message
     end
   end
 
+  sub_test_case "ignore_same_log_interval" do
+    teardown do
+      Thread.current[:last_same_log] = nil
+    end
```
cosmo0920, line 417: Should we add `Timecop.return` in `#teardown`?

cosmo0920 (approved):
LGTM except for an added comment.

test/test_log.rb
repeatedly: `Timecop.return` is called in parent teardown.

cosmo0920: Oh! I got it.

Issue #3066, opened 2020-07-07
Unexpected error Broken pipe when trying to shutdown fluentd with multiple workers


flurreN:
**Describe the bug**
When trying to shutdown with ctrl+c in the td-agent command promt it crashes and wont shutdown properly.

**To Reproduce**
Run config and press ctrl+c or shut in the command promt

**Expected behavior**
Should shutdown properly withouth getting stuck with errors

**Your Environment**
Windows Server 2019
fluentd 1.10.2
Tried with both td-agent 4.0.0 and 3.7.1

**Your Configuration**

```
<system>
    workers                             2
</system>

<worker 0>
    <source>
        @type                           windows_eventlog2
        @id                             windows_eventlog2_0
        read_interval                   2
        tag                             serverwindows.general
        rate_limit                      200
        channels                        x
        parse_description               true
        render_as_xml                   true
        <storage>   
            @type                       local
            persistent                  true
            path                        x
        </storage>
    </source>

    <filter serverwindows.**>
        @type                           record_transformer
        enable_ruby                     true
        <record>
            time                        ${if record["TimeCreated"] then record["TimeCreated"] end}
            fluentd_host                ${hostname}
        </record>
    </filter>

    <match serverwindows.general>
        @type                           kafka2
        brokers                         x
        username                        x
        password                        x
        scram_mechanism                 x
        ssl_ca_cert                     x
        default_topic                   x

        <format>
            @type                       json
        </format>

        <buffer>
            @type                       file
            path                        x
            flush_mode                  interval
            flush_interval              2s
            flush_thread_count          8
            retry_wait                  5s
            chunk_limit_size            5m
            flush_at_shutdown           true
        </buffer>
    </match>

</worker>

<worker 1>
    <source>
        @type                           windows_eventlog2
        @id                             windows_eventlog2_1
        read_interval                   2
        tag                             serverwindows.security
        rate_limit                      200
        channels                        x
        parse_description               true
        render_as_xml                   true
        <storage>
            @type                       local
            persistent                  true
            path                        x
        </storage>
    </source>
	
    <filter serverwindows.**>
        @type                           record_transformer
        enable_ruby                     true
        <record>
            time                        ${if record["TimeCreated"] then record["TimeCreated"] end}
            fluentd_host                ${hostname}

        </record>
    </filter>

    <match serverwindows.security>
        @type                           kafka2
        brokers                         x
        username                        x
        password                        x
        scram_mechanism                 x
        ssl_ca_cert                     x
[...]

PR #3131, opened 2020-09-11
Add win32 events alternative to unix signals

ashie:
None

Since Windows doesn't have UNIX signals, features realized by them such as flushing buffers aren't implemented on Windows.
This PR adds triggers for them by using win32 events. Currently only supervisor process supports them.

In addition, this PR adds a new command `fluent-ctl`. It sends a simple command to a fluentd process.
The main purpose of this command is that providing a kill command alternative on Windows.
On UNIX, it just only send a UNIX signal to a specified process.

```
Usage: fluent-ctl COMMAND PID_OR_SIGNAME
    
Commands:
  shutdown
  restart
  flush
  reload
```

SIGNAME can be specified by fluentd's "-x" or "--signame" option.
Note that this command doesn't support sending events to windows service.
Probably we need to trigger these events via `sc` command for windows service.

This PR also fixes a bug that the graceful shutdown feature doesn't work correctly when Fluentd is launched as Windows service. On this case, worker processes were killed forcedly before receiving `GRACEFUL_STOP` command.

**Docs Changes**:
Add `fluent-ctl` to https://docs.fluentd.org/deployment/command-line-option

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -31,7 +31,7 @@ module Fluent
       include Windows::ServiceStructs
       include Windows::ServiceFunctions
 
-      WINDOWS_EVENT_MAP = {
+      COMMAND_MAP = {
         shutdown: "",
         restart: "HUP",
         flush: "USR1",
@@ -45,15 +45,13 @@ module Fluent
         flush: 129,
         reload: SERVICE_CONTROL_PARAMCHANGE,
       }
-      COMMAND_MAP = WINDOWS_EVENT_MAP
     else
-      UNIX_SIGNAL_MAP = {
+      COMMAND_MAP = {
         shutdown: :TERM,
         restart: :HUP,
         flush: :USR1,
         reload: :USR2,
       }
-      COMMAND_MAP = UNIX_SIGNAL_MAP
     end
 
     def initialize(argv = ARGV)
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -1,0 +1,132 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'optparse'
+require 'fluent/env'
+require 'fluent/version'
+require 'win32/event' if Fluent.windows?
+
+module Fluent
+  class Ctl
+    DEFAULT_OPTIONS = {}
+    UNIX_SIGNAL_MAP = {
+      shutdown: :TERM,
+      restart: :HUP,
+      flush: :USR1,
+      reload: :USR2,
+    }
+    WINDOWS_EVENT_MAP = {
+      shutdown: "",
+      restart: "HUP",
+      flush: "USR1",
+      reload: "USR2",
+    }
+    if Fluent.windows?
+      COMMAND_MAP = WINDOWS_EVENT_MAP
+    else
+      COMMAND_MAP = UNIX_SIGNAL_MAP
+    end
+
+    def initialize(argv = ARGV)
+      @argv = argv
+      @options = {}
+      @opt_parser = OptionParser.new
+      configure_option_parser
+      @options.merge!(DEFAULT_OPTIONS)
+      parse_options!
+    end
+
+    def help_text
+      text = "\n"
+      if Fluent.windows?
+        text << "Usage: #{$PROGRAM_NAME} COMMAND PID_OR_SIGNAME\n"
+      else
+        text << "Usage: #{$PROGRAM_NAME} COMMAND PID\n"
+      end
+      text << "\n"
+      text << "Commands: \n"
+      COMMAND_MAP.each do |key, value|
+        text << "  #{key}\n"
+      end
+      text
+    end
+
+    def usage(msg = nil)
+      puts help_text
+      if msg
+        puts
+        puts "Error: #{msg}"
+      end
+      exit 1
+    end
+
+    def call
+      if Fluent.windows?
+        call_windows_event(@command, @pid_or_signame)
+      else
+        call_signal(@command, @pid_or_signame)
+      end
+    end
+
+    private
+
+    def call_signal(command, pid)
+      signal = COMMAND_MAP[command.to_sym]
+      Process.kill(signal, pid.to_i)
+    end
+
+    def call_windows_event(command, pid_or_signame)
+      if pid_or_signame =~ /^[0-9]+$/
+        prefix = "fluentd_#{pid_or_signame}"
+      else
+        prefix = pid_or_signame
+      end
+      event_name = COMMAND_MAP[command.to_sym]
+      suffix = event_name.empty? ? "" : "_#{event_name}"
+
+      begin
+        event = Win32::Event.open("#{prefix}#{suffix}")
+        event.set
+        event.close
+      rescue Errno::ENOENT => e
+        puts "Error: Cannot find the fluentd process with the event name: \"#{prefix}\""
+      end
+    end
+
+    def configure_option_parser
+      @opt_parser.banner = help_text
+      @opt_parser.version = Fluent::VERSION
+    end
+
+    def parse_options!
+      @opt_parser.parse!(@argv)
+
+      @command = @argv[0]
+      @pid_or_signame = @argv[1]
+
+      usage("Command isn't specified!") if @command.nil? || @command.empty?
+      usage("Unknown command: #{@command}") unless COMMAND_MAP.has_key?(@command.to_sym)
+
+      if Fluent.windows?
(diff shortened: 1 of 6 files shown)


bin/fluentd-windows-cmd
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require 'win32/event'
+require 'optparse'
+
+$LOAD_PATH.unshift(File.join(__dir__, 'lib'))
+require 'fluent/command/windows_cmd'
+
+Fluent::WindowsCmd.new.call
```
cosmo0920, line 9: Should we raise an error when using non-Windows environment?

bin/fluentd-windows-cmd
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require 'win32/event'
```
cosmo0920, line 3: On non-Windows environment, `win32/event` shouldn't be installed as a dependency and loaded.
Should we use `Fluent.windows?` for guarding to require on non-Windows environment?

cosmo0920: Or, just raise LoadError here is enough?

bin/fluentd-windows-cmd
ashie: I renamed the command to fluentdctl and support also UNIX environment, it just only sends a UNIX signal to a specified process though.

bin/fluentd-windows-cmd
ashie: Ditto

ashie:
Matters to be discussed:

* I'm not sure the command name `fluentdctl` is appropriate or not.
  * `fluentd-ctl`? `fluentctl`? `fluent-ctl`?
* Although I've simplified the command names ASAP, I'm not sure they are appropriate or not.
  * e.g.) omit `buffers` of `flush` command, omit `graceful` of `reload` command
* UNIX version of `fluentdctl` may not be needed.
  * It just only send a UNIX signal to a specified process.

cosmo0920:
I'd confirmed that `fluentdctl` can work on both of Windows and Linux.

> * I'm not sure the command name `fluentdctl` is appropriate or not.
>    * `fluentd-ctl`? `fluentctl`? `fluent-ctl`?

For current command naming convention, I vote `fluentctl` or `fluent-ctl`.

ashie:
> For current command naming convention, I vote `fluentctl` or `fluent-ctl`.

I've renamed it to fluent-ctl.

ashie:
Now I've found that both the original implementation for shutdown and this PR doesn't support multiple workers #3066.

ashie:
Now I''ve found that fluent-ctl can't send events when fluentd is run as windows service, even if fluent-ctl is run as "SYSTEM" user.
So that I renamed the 3rd argument of the command from `PID_OR_SVCNAME` to `PID_OR_SIGNAME`, and removed the default parameter for it.

For Windows service, probably we need to trigger the events via `sc` command.

ashie:
> Now I've found that both the original implementation for shutdown and this PR don't support multiple workers: #3066

It's ServerEngine's fault, I'll fix it. This PR should follow the change.

ashie:
> Now I've found that fluent-ctl can't find events when fluentd is run as windows service, even if fluent-ctl is run as "SYSTEM" user.

Windows Service processes are separated as "Session 0" as of Windows Vista.
[PsExec](https://docs.microsoft.com/en-us/sysinternals/downloads/psexec) has a option to execute a command on "Session 0".
Now I've confirmed that I can trigger the events by the following command
```
\path\to\psexec.exe -s -i 0 \path\to\fluent-ctl restart fluentdwinsvc
```

ashie:
> This PR also fixes a bug that the graceful shutdown feature doesn't work correctly when Fluentd is launched as Windows service.

I've filed a new pull request to fix it: #3156

ashie:
The remaining issue of this PR is:

> SIGNAME can be specified by fluentd's "-x" or "--signame" option.
> Note that this command doesn't support sending events to windows service.
> Probably we need to trigger these events via sc command for windows service.

I'm now working on it.
`sc` command can send pre-defined control codes or user-defined control codes to a service:

https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-controlservice

* For `USER2` signal (gaceful-reload feature) on UNIX, probably mapping it to `SERVICE_CONTROL_PARAMCHANGE`  is suitable.
* For `USER1` (flush) and `HUP` (restart), there is no suitable pre-defined control code, so I'll map these features to user-defined control codes.
  * But [win32-sercvice](https://github.com/chef/win32-service/blob/master/lib/win32/daemon.rb) doesn't support receiving user-defined control codes. I'll send PR to it to enable receiving them.
  * Or we can map these features to rarely used control codes such as `SERVICE_CONTROL_NETBIND*` (deprecated control codes), but it's not so good design.

ashie:
>  * But [win32-sercvice](https://github.com/chef/win32-service/blob/master/lib/win32/daemon.rb) doesn't support receiving user-defined control codes. I'll send PR to it to enable receiving them.

https://github.com/chef/win32-service/pull/74

ashie:
And I've pushed the patch to send /receive a control code to/from fluend windows service.

shutdown:
```
# sc control 1 fluentdwinsvc
or
# sc control stop fluentdwinsvc
or
# fluent-ctl shutdown
```

reload:
```
# sc control paramchange fluentdwinsvc
or
# sc control  6 fluentdwinsvc
or
# fluent-ctl reload 
```

restart (need the patch for win32-service):
```
# sc control fluentdwinsvc 128
or
# fluent-ctl restart
```

flush (need the patch for win32-service):
```
# sc control fluentdwinsvc 129
or
# fluent-ctl flush
```

ashie:
I already completed this work but we need to wait next release of win32-service (probably 2.3.0).

lib/fluent/supervisor.rb
```diff
@@ -159,37 +165,101 @@ def stop_counter_server
...
-          stop(true)
         ensure
-          ev.close
+          events.each { |event| event.close }
         end
       end
     end
 
+    def stop_windows_event_thread
+      if Fluent.windows?
```
repeatedly, line 256: L73 is `stop_windows_event_thread if Fluent.windows?`.
Is this `if` for more safety?

lib/fluent/supervisor.rb
```diff
@@ -745,33 +831,45 @@ def install_main_process_signal_handlers
...
+      end
+    end
 
-      if Fluent.windows?
-        command_pipe = STDIN.dup
-        STDIN.reopen(File::NULL, "rb")
-        command_pipe.binmode
-        command_pipe.sync = true
+    def install_main_process_command_handlers
+      command_pipe = $stdin.dup
```
repeatedly, line 848: Why change from `STDIN` to `$stdin`?

ashie: For testing incoming commands at workers. See `test_main_process_command_handlers`.

lib/fluent/command/ctl.rb
```diff
@@ -0,0 +1,179 @@
...
+      }
+      WINSVC_CONTROL_CODE_MAP = {
+        shutdown: SERVICE_CONTROL_STOP,
+        # 128 - 255: user-defined control code
+        # See https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-controlservice
+        restart: 128,
+        flush: 129,
+        reload: SERVICE_CONTROL_PARAMCHANGE,
+      }
+      COMMAND_MAP = WINDOWS_EVENT_MAP
```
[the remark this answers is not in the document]
ashie: No particular reason. I'll remove them.

lib/fluent/supervisor.rb
ashie: > Is this if for more safety?

That's right.

repeatedly (approved):

test/test_supervisor.rb
```diff
@@ -190,6 +248,38 @@ def test_rpc_server
...
+    end
+    server.rpc_endpoint = sys_conf.rpc_endpoint
+
+    server.run_rpc_server
+
+    mock(server).restart(true) { nil }
+    Net::HTTP.get URI.parse('http://127.0.0.1:24447/api/plugins.flushBuffers')
+
+    server.stop_rpc_server
+  end
```
cosmo0920, line 281: `Net::HTTP.get URI.parse('http://127.0.0.1:24447/api/plugins.flushBuffers')`'s return value should be checked with some assertions?

cosmo0920:
win32-service 2.3.0 which includes user-defined control code handling has been released:
https://github.com/chef/win32-service/commit/bde5569a9512f2f7ae07a754ea29f4c7b7b1abd0

cosmo0920:
We should bump up dependent win32-service gem version to 2.3.0 before merging into master:
https://github.com/fluent/fluentd/blob/master/fluentd.gemspec#L36

cosmo0920 (approved):
I also confirmed that `fluent-ctl` can work for `shutdown/restart/flush/reload` on Windows Service.

BTW, I feel that fluent-ctl start is missing for Windows Service....
Starting Windows Service should be executed from sc/net command or Windows Management tool's Service?

ashie:
> Starting Windows Service should be executed from sc/net command or Windows Management tool's Service?

That's right.

repeatedly:
If there is no more comments, I will merge this patch.

cosmo0920:
> If there is no more comments, I will merge this patch.

:+1:

Bumping up dependent win32-service gem version to v2.3.0 should do later.
win32-service v2.3.0 is not released yet.

cosmo0920:
Let's :shipit:

cosmo0920:
:tada:

chook:
this is amazing, what about SIGCONT or forcing a thread dump @ashie

PR #3134, opened 2020-09-25
output: Replace ${chunk_id} before logging warning.

nairb774:
The updated implementation makes sure to replace instances of
${chunk_id} prior to looking for and warning on unreplaced keys in the
pattern. Prior to this, a warning line was printed for every chunk that
was flushed even though nothing was actually wrong (replacement happened
right after).

Signed-off-by: Brian Atkinson <brian@atkinson.mn>

None filed.

When using the s3 output plugin, with the `path` set to `"${tag}/%Y/%m/%d/%H/${chunk_id}_#{ENV['RUN_ID']}_#{worker_id}"` the fluentd process would print a log line like the following for every chunk flushed.
```
2020-09-25 00:32:49 +0000 [warn]: #1 [s3_output] chunk key placeholder 'chunk_id' not replaced. template:${tag}/%Y/%m/%d/%H/${chunk_id}_6ead9663854b0799849b47d91148d344f5fcb534329354a0dfad76ef6056_1.gz
```
This log line was being printed a little too early and is entirely misleading.

**Note to reviewer(s):** This is the first ruby I've worked with in about 10 years. Anything that doesn't feel normal about the code is likely due to this, and any improvements are appreciated.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -769,17 +769,19 @@ module Fluent
             end
           end
 
-          if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
-            log.warn "chunk key placeholder '#{$1}' not replaced. template:#{str}"
-          end
-
-          rvalue.sub(CHUNK_ID_PLACEHOLDER_PATTERN) {
+          rvalue = rvalue.sub(CHUNK_ID_PLACEHOLDER_PATTERN) {
             if chunk_passed
               dump_unique_id_hex(chunk.unique_id)
             else
               log.warn "${chunk_id} is not allowed in this plugin. Pass Chunk instead of metadata in extract_placeholders's 2nd argument"
             end
           }
+
+          if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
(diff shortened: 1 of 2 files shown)


repeatedly:
existing test fails. will check.

lib/fluent/plugin/output.rb
```diff
@@ -768,19 +767,13 @@ def extract_placeholders(str, chunk)
               end
             end
           end
+        end
 
-          if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN
-            log.warn "chunk key placeholder '#{$1}' not replaced. template:#{str}"
-          end
-
-          rvalue.sub(CHUNK_ID_PLACEHOLDER_PATTERN) {
```
repeatedly, line 776: This patch has a problem with `@chunk_key_time`.
How about moving this code before `if rvalue =~ CHUNK_KEY_PLACEHOLDER_PATTERN` check?

nairb774: Done.

repeatedly:
Thanks!

PR #3145, opened 2020-10-08
Add unixtime_millis to inject plugin helper

okkez:

It is useful to add timestamp as milliseconds since Unix epoch.
This is the short hand to add the filed using filter
record_transformer like followings:

```
<filter>
  @type record_transformer
  enable_ruby
  <record>
    time ${time.to_f.floor(3) * 1000}
  </record>
</filter>
```

This is useful to reduce data size with presto like system such as AWS Athena with S3.

**Docs Changes**: Need

**Release Note**: 

Add unixtime_millis to inject plugin helper

### DIFF
--- a/lib/fluent/plugin_helper/inject.rb
+++ b/lib/fluent/plugin_helper/inject.rb
@@ -76,7 +76,7 @@ module Fluent
           config_param :time_key, :string, default: nil
 
           # To avoid defining :time_type twice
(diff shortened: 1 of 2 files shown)


cosmo0920 (approved):
Looks reasonable.
Could you also add `unixtime_millis` parameter documentation in https://github.com/fluent/fluentd-docs-gitbook/tree/1.0?

repeatedly:
LGTM. Thanks!

PR #3152, opened 2020-10-22
Provide CRLF newline style choice on formatter plugins. Fix #3151

### LINKED ISSUE #3151: File output on Windows missing CRLF
**Describe the bug**
File outputs only have Unix line endings (LF), and some tools are having trouble reading files without CRLF on Windows.

**To Reproduce**
1. Read any input
2. Write to a file
3. File will always have a LF as a line ending

**Expected behavior**
at 3: use system-native line endings or at least make it configurable

**Your Environment**
In windows, I'm transferring logs and splitting them off between a couple of servers.

- td-agent fluentd 1.11.2
- Windows Server 2016

**Your Configuration**

```
<label @iis_text_logs>
  <match **>
    @type file
    path D:\Logs\IIS\W3SVC1\iis.%Y%m%dT%H%M%S%z
	append true
	<format>
	  @type single_value
	</format>
    <buffer>
      timekey 5s
      timekey_use_utc true
      timekey_wait 10s
	  path "D:\\Logs\\IIS\\W3SVC1\\buffer"
    </buffer>
   </match>
 </label>
```

**Your Error Log**
No errors.

**Additional context**
The reason for this curious configuration and the reason I'm hitting this is probably because I'm (ab)using fluentd to stream text logs from one server to another (as close to realtime as it will go - but buffer won't let me without writing billions of files...). The input is CRLF-terminated and the output needs to be CRLF-terminated as well for the follow-up tool that will then read the files. (and no, that tool cannot run on the source server, hence the setup)

cosmo0920:

Fixes #3151 

Currently, built-in formatter plugins use hard-coded LF newline.
We should provide CRLF style newline choice mainly for Windows users.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/247

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/formatter_hash.rb
+++ b/lib/fluent/plugin/formatter_hash.rb
@@ -22,10 +22,21 @@ module Fluent
       Plugin.register_formatter('hash', self)
 
       config_param :add_newline, :bool, default: true
+      config_param :newline, :enum, list: [:lf, :crlf], default: :lf
+
+      def configure(conf)
+        super
+        @newline = case newline
+                   when :lf
+                     "\n"
+                   when :crlf
+                     "\r\n"
+                   end
+      end
 
       def format(tag, time, record)
         line = record.to_s
-        line << "\n".freeze if @add_newline
+        line << @newline.freeze if @add_newline
         line
       end
     end
--- a/lib/fluent/plugin/formatter_hash.rb
+++ b/lib/fluent/plugin/formatter_hash.rb
@@ -19,20 +19,11 @@ require 'fluent/plugin/formatter'
 module Fluent
   module Plugin
     class HashFormatter < Formatter
+      include Fluent::Plugin::Newline::Mixin
+
       Plugin.register_formatter('hash', self)
 
       config_param :add_newline, :bool, default: true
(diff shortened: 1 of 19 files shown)


lib/fluent/plugin/formatter_hash.rb
```diff
@@ -22,10 +22,21 @@ class HashFormatter < Formatter
       Plugin.register_formatter('hash', self)
 
       config_param :add_newline, :bool, default: true
+      config_param :newline, :enum, list: [:lf, :crlf], default: :lf
```
repeatedly, line 25: How about using `Fluent.windows?` for changing default for windows?

In addition, `newline` code are duplicated in many places. How about providing mixin in Formatter?

cosmo0920: I'd applied the above suggestion as https://github.com/fluent/fluentd/pull/3152/commits/0943b1436a2882c5d0980c223b34e8f374661ecc and https://github.com/fluent/fluentd/pull/3152/commits/fdcfaf4ec0c5abd4a28c535e577541305b60f0e7.

repeatedly:
Looks good 👍

cosmo0920:
Let's shipping! :shipit:

PR #3153, opened 2020-10-27
Fix missing runtime bundler dependency

### LINKED ISSUE #1481: Bundler is not in dependencies
Fluentd's `--gemfile` option executes Fluentd itself via `bundler`, but Fluentd doesn't have any dependency about `bundler` gem. It makes that bare `fluentd --gemfile` command will fails always, without extra `fluent-gem install bundler` command.

(td-agent is configured to install bundler in its environment, though)

IMO it's better to add a runtime dependency for `bundler`, or update the error message to instruct to run `fluent-gem install bundler` here:
https://github.com/fluent/fluentd/blob/master/lib/fluent/command/bundler_injection.rb#L25

kenhys:
Fixes #1481

It fixes the following error with `fluentd --gemfile Gemfile`
when bundler is not installed.

  error: Gem::MissingSpecError

**Docs Changes**:

No need to change

**Release Note**: 

NOTE: since ruby 2.6 or later, bundler is bundled with ruby itself,
so this error has occurred with ruby 2.5 which will reach EOL
at 2021-03-31.

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.4'
 
+  gem.add_runtime_dependency("bundler")
   gem.add_runtime_dependency("msgpack", [">= 1.3.1", "< 2.0.0"])
   gem.add_runtime_dependency("yajl-ruby", ["~> 1.0"])
   gem.add_runtime_dependency("cool.io", [">= 1.4.5", "< 2.0.0"])


fluentd.gemspec
```diff
@@ -18,6 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.4'
 
+  gem.add_runtime_dependency("bundler", [">= 1.17.3", "< 2.2.0"])
```
repeatedly, line 21: Why specify exact number for version constraint?
regression or critical bugs?

kenhys: > regression or critical bugs?

This is because that I'm not confirmed with more older version.
The regression nor critical bugs has nothing to do with it.

kenhys: I've removed the explicit version constraint. If regression or critical bugs were found, it should be fixed.

kenhys:
Windows CI has failed but it has nothing to do with this PR itself.
So I've created another PR to fix it.
https://github.com/fluent/fluentd/pull/3160

PR #3155, opened 2020-11-02
Handle linux capability if available

### LINKED ISSUE #3053: in_tail ignores linux capabilities for opening/reading files
**Describe the bug**

In this case, we are enabling CAP_DAC_READ_SEARCH on the ruby binary in order to run as a non-root user but still read root owned log files. Calling file.readable?(path) results in "#{p} unreadable. It is excluded and would be examined next time." because file.readable? only considers the effective user/group id. 

**To Reproduce**

- switch to user root (or in Docker USER root)
- setcap cap_dac_read_search=+eip /usr/local/bin/ruby (or in Docker RUN setcap cap_dac_read_search=+eip /usr/local/bin/ruby)
- switch to user fluent (or in Docker USER fluent)
- set a tail configuration for root owned files, for example:
```    
    <source>
      @type tail
      pos_file /var/log/fluentd-containers.log.pos
      tag kubernetes.logs.*
      format json
      read_from_head true
      read_lines_limit 100
    </source>
```
Output:
> 2020-06-19 19:46:15 +0000 [warn]: #0 /var/log/containers/fluentd-logs-cr82n_genctl_fluentd-logs-816c787602a73f8e9b9c85b74bc340a783b1cd67de4b49735bd9e2d28085cc99.log unreadable. It is excluded and would be examined next time.

**Expected behavior**

Since we have CAP_DAC_READ_SEARCH, would expect to read the file normally.

**Your Environment**
fluentd:v1.9.1-debian-1.0 Docker image

**Your Configuration**

```
    <source>
      @type tail
      pos_file /var/log/fluentd-containers.log.pos
      tag kubernetes.logs.*
      format json
      read_from_head true
      read_lines_limit 100
    </source>
```

**Your Error Log**

```
2020-06-19 19:46:15 +0000 [warn]: #0 /var/log/containers/fluentd-logs-cr82n_genctl_fluentd-logs-816c787602a73f8e9b9c85b74bc340a783b1cd67de4b49735bd9e2d28085cc99.log unreadable. It is excluded and would be examined next time.
```

**Additional context**
Writing a custom in_tail plugin that avoids the file.readable?(p) check opens and reads the file normally.

cosmo0920:
Fixes #3053 

Currently, Fluentd core does not handle/refer Linux capability.
This class should handle/refer Linux capability.
Note that this class implementation is only working for Linux with
libcap-ng binding (capng_c) installed platform.
In other platform, this class only returns stubbed results.

in_tail should handle Linux capabilities.
In in_tail, File.readable? does not handles Linux capabilities.
So, we should handle them by libcap-ng binding wrapped class
which is Fluent::Capability.

On in_tail, the following capabilities should be handled:

* CAP_DAC_READ_SEARCH(:dac_read_search on capng_c)
* CAP_DAC_OVERRIDE(:dac_override on capng_c)

**Docs Changes**:
Needed. ~~I'll write and register Linux capability documentation for Fluentd later.~~
https://github.com/fluent/fluentd-docs-gitbook/pull/248

**Release Note**: 

Same as title.

**Additional Context**:

This PR should be reviewed with some Linux box. (macOS and Windows box is not installable capng_c that is Linux capability bindings for Ruby)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ branches:
 before_install: |
   gem update --system=3.1.2
   if [[ "x${USE_CAPNG}" == "xtrue" ]]; then
-    gem install capng_c --no-document
+    echo 'gem "capng_c"' >> Gemfile.local
   fi
 
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,44 +9,63 @@ matrix:
   include:
     - rvm: 2.4.9
       os: linux
+      env: USE_CAPNG=false
     - rvm: 2.4.9
       os: linux-ppc64le
+      env: USE_CAPNG=false
     - rvm: 2.5.7
       os: linux
+      env: USE_CAPNG=false
     - rvm: 2.5.7
       os: linux
       arch: s390x
       dist: xenial
+      env: USE_CAPNG=false
     - rvm: 2.6.5
       os: linux
+      env: USE_CAPNG=false
+    - rvm: 2.6.6
+      os: linux
+      env: USE_CAPNG=true
     - rvm: 2.7.0
       os: linux
+      env: USE_CAPNG=false
     - rvm: ruby-head
       os: linux
+      env: USE_CAPNG=false
     - rvm: ruby-head
       os: linux-ppc64le
+      env: USE_CAPNG=false
     - rvm: 2.4.6
       os: osx
       osx_image: xcode8.3 # OSX 10.12
+      env: USE_CAPNG=false
     - rvm: ruby-head
       os: osx
       osx_image: xcode8.3 # OSX 10.12
+      env: USE_CAPNG=false
   allow_failures:
     - rvm: 2.4.6
       os: osx
       osx_image: xcode8.3
+      env: USE_CAPNG=false
     - rvm: 2.5.7
       os: linux
       arch: s390x
       dist: xenial
+      env: USE_CAPNG=false
     - rvm: ruby-head
+      env: USE_CAPNG=false
 
 branches:
   only:
     - master
 
-before_install:
-  - gem update --system=3.1.2
+before_install: |
+  gem update --system=3.1.2
+  if [[ "x${USE_CAPNG}" == "xtrue" ]]; then
+    gem install capng_c --no-document
+  fi
 
 sudo: false
 dist: trusty # for TLSv1.2 support
@@ -55,3 +74,4 @@ addons:
   apt:
     packages:
       - libgmp3-dev
+      - libcap-ng-dev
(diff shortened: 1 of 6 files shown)


cosmo0920:
I'd confirmed that capability on Fluentd is working:
**https://travis-ci.org/github/fluent/fluentd/jobs/742347990**

lib/fluent/capability.rb
```diff
@@ -0,0 +1,71 @@
...
+  begin
+    require 'capng'
+  rescue LoadError
+  end
+end
+
+module Fluent
+  class Capability
+    def initialize(target = :current_process, pid_or_path = nil)
+      if defined?(CapNG)
```
repeatedly, line 29: Without `CapNG`, all methods don't do anything. So separate implementation is better.

```
module Fluent
  if defined?(CapNG)
    class Capability
      def apply(select_set)
        @capng.apply(select_set)
      end
    end
  else
    class Capability
      def apply(select_set)
      end
    end
  end
end
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -263,7 +270,9 @@ def expand_paths
           paths += Dir.glob(path).select { |p|
             begin
               is_file = !File.directory?(p)
-              if File.readable?(p) && is_file
+              if (File.readable?(p) ||
```
repeatedly, line 273: Maybe, one line is more readable. 3 line conditions are hard for me.

lib/fluent/capability.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3155/commits/6d247cdafd433b091e63c68b0946f4e361a1db30.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3155/commits/e4a833a5f7514ee7f2c6a36924b92e4456c0f8a2

cosmo0920:
I'd sent a separated PR to add Linux capability handling command.

.travis.yml
```diff
@@ -9,44 +9,63 @@ matrix:
...
 
 branches:
   only:
     - master
 
-before_install:
-  - gem update --system=3.1.2
+before_install: |
+  gem update --system=3.1.2
+  if [[ x"${USE_CAPNG}" == "xtrue" ]]; then
```
repeatedly, line 66: `x"true"`?

repeatedly (approved):

cosmo0920:
Let's :shipit:

PR #3160, opened 2020-11-09
windows: use protocol-http2 0.14.0 explicitly

kenhys:

Via #3153 PR, I've found Windows CI fails.

There is an install issue on Windows about http-protocol2.
It breaks Windows CI when executing bundle install.

ref. https://github.com/socketry/protocol-http2/issues/6

It seems that even though 0.14.0, it doesn't affect functionality because the only https://github.com/socketry/protocol-http2/commit/8769e9f8dfcb95c6e7bd015002a74a82bb14a83f is commited.

**Docs Changes**:

No need to change.

**Release Note**: 

No need to change.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,9 @@ install:
   - SET PATH=C:\Ruby%ruby_version%\bin;%PATH%
   - ruby --version
   - gem --version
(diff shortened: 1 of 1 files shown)


kenhys:
This PR focus on `bundle install` error which is caused by `protocol-httpd2` gem.
test failure should be fixed in another PR.

repeatedly:
Thx!

PR #3162, opened 2020-11-10
Add fluent-cap-ctl command to handle add/drop/get/clear capability operations

cosmo0920:
Please review #3155 first.

None. Follows up #3155

Linux capability operation should be handled in Fluentd management command.
This should be unified interface to add/drop/get/clear operations for Ruby executables which are used to execute Fluentd supervisor and worker(s).

**Docs Changes**:
Needed.  ~~I'll send a PR to describe this command.~~
https://github.com/fluent/fluentd-docs-gitbook/pull/249

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/command/cap_ctl.rb
+++ b/lib/fluent/command/cap_ctl.rb
@@ -24,7 +24,7 @@ module Fluent
     def prepare_option_parser
       @op = OptionParser.new
 
-      @op.on('--clear-cap', '--clear-capabilities', "Clear Fluentd Ruby capability") {|s|
+      @op.on('--clear', "Clear Fluentd Ruby capability") {|s|
         @opts[:clear_capabilities] = true
       }
 
--- a/lib/fluent/command/cap_ctl.rb
+++ b/lib/fluent/command/cap_ctl.rb
@@ -28,15 +28,15 @@ module Fluent
         @opts[:clear_capabilities] = true
       }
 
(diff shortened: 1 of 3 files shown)


repeatedly:
`fluent-cap-ctl --clear-cap` seems to be redundant. Need `-cap` suffix in the option?

cosmo0920:
> `fluent-cap-ctl --clear-cap` seems to be redundant. Need `-cap` suffix in the option?

Hmm..., right. I'll remove `-cap` suffix from `--clear-cap` option.

repeatedly:
How about other options? keep `-cap` is better? I'll leave it to you.

lib/fluent/command/cap_ctl.rb
```diff
@@ -0,0 +1,174 @@
...
+
+      @op.on('--clear', "Clear Fluentd Ruby capability") {|s|
+        @opts[:clear_capabilities] = true
+      }
+
+      @op.on('--add [CAPABILITITY1,CAPABILITY2, ...]', "Add capabilities into Fluentd Ruby") {|s|
+        @opts[:add_capabilities] = s
+      }
+
+      @op.on('--drop [CAPABILITITY1,CAPABILITY2, ...]', "Drop capabilities into Fluentd Ruby") {|s|
```
repeatedly, line 35: `from` instead of `into`?

repeatedly (approved):

cosmo0920:
I'll merge this PR after CIs are finishing.

PR #3174, opened 2020-11-20
github: add a guide to google groups

kenhys:

Fixes N/A

In the previous version, there are only two choices for
bug and feature request.

So Fluentd user tends to use "Bug Report" even though it is not a
bug report definitely.

This PR try to add third choice explicitly to guide to user forum.

Before:

  * Bug Report
  * Feature request

After:

  * Bug Report
  * Feature request
  * Ask a question

Additionally, the explanation about bug report is updated to emphasize
it is a "Bug Report".

**Docs Changes**:

N/A

**Release Note**: 

N/A


kenhys:
This image is not match to this PR because this preview image is borrowed from other project, but it help to understand what I mean.

image

repeatedly:
Thx!

PR #3182, opened 2020-11-25
Add follow_inodes parameter on in_tail

cosmo0920:
Revive of #1660  

From #1660:

> Add ability to track inodes inside fluentd in_tail plugin to enable combination of '*' in `path` configuration with rotation inside same directory and `read_from_head` set to true. In proposed solution we add `follow_inodes` flag (set to `false` by default to preserve current behavior). When `follow_inodes` is set to `true` and file rotation is detected - we do not reread file, instead we are checking if such inode is present inside pos file - if yes, we don't recreate TailWatcher. TailWatcher is closed after `rotate_wait` period and file marked as unwatched inside position file after `limit_recently_modified` time passed. Pos file is cleaned on startup as it was before, but also delete entries for non-existing files.

And we also think that this following inodes feature is useful for tracking k8s related logs.

**Docs Changes**:
Needed for `follow_inodes` parameter. I'll register a PR for it.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -245,7 +245,7 @@ module Fluent::Plugin
 
     def shutdown
       # during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.
-      stop_watchers(@tails.keys, immediate: true, remove_watcher: false)
+      stop_watchers(existence_path.values, immediate: true, remove_watcher: false)
       @pf_file.close if @pf_file
 
       super
@@ -408,39 +408,32 @@ module Fluent::Plugin
           log.warn "Skip #{path} because unexpected setup error happens: #{e}"
           next
         end
-        @tails[path] = tw
+        tuple = PathInodeTuple.new(path, Fluent::FileWrapper.stat(path).ino)
+        @tails[tuple] = tw
       }
     end
 
-    def stop_watchers(paths, immediate: false, unwatched: false, remove_watcher: true)
-      paths.each { |path|
+    def stop_watchers(paths_with_inodes, immediate: false, unwatched: false, remove_watcher: true)
+      paths_with_inodes.each { |path_with_inode|
         if remove_watcher
-          if @follow_inodes
-            tw = @tails.delete(path.ino)
-          else
-            tw = @tails.delete[path.path]
-          end
+          tw = @tails.delete(path_with_inode)
         else
-          if @follow_inodes
-            tw = @tails[path.ino]
-          else
-            tw = @tails[path.path]
-          end
+          tw = @tails[path_with_inode]
         end
         if tw
           tw.unwatched = unwatched
           if immediate
-            detach_watcher(tw, path.ino, false)
+            detach_watcher(tw, path_with_inode.ino, false)
           else
-            detach_watcher_after_rotate_wait(tw, path.ino)
+            detach_watcher_after_rotate_wait(tw, path_with_inode.ino)
           end
         end
       }
     end
 
     def close_watcher_handles
-      @tails.keys.each do |path|
-        tw = @tails.delete(path)
+      @tails.keys.each do |path_with_inode|
+        tw = @tails.delete(path_with_inode)
         if tw
           tw.close
         end
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -325,7 +325,7 @@ module Fluent::Plugin
 
     def existence_path
       hash = {}
-      @tails.keys.each {|path_ino|
+      @tails.each_key {|path_ino|
         if @follow_inodes
           hash[path_ino.ino] = path_ino
         else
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -312,21 +341,21 @@ def expand_paths
...
 
-      unwatched = existence_paths - target_paths
-      added = target_paths - existence_paths
+      unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
+      added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
 
-      stop_watchers(unwatched, immediate: false, unwatched: true) unless unwatched.empty?
-      start_watchers(added) unless added.empty?
+      stop_watchers(unwatched_hash.values, immediate: false, unwatched: true) unless unwatched_hash.empty?
+      start_watchers(added_hash.values) unless added_hash.empty?
```
repeatedly, line 353: Calling `values` is needed? Passes `Hash` and call `each_value` has a problem? Latter approach doesn't create temporal object.

lib/fluent/plugin/in_tail.rb
```diff
@@ -303,7 +308,31 @@ def expand_paths
...
+        else
+          hash[tuple.path] = tuple
+        end
+      }
+      hash
+    end
+
+    def existence_path
+      hash = {}
+      @tails.keys.each {|path_ino|
```
repeatedly, line 328: How about `each_key` ?

lib/fluent/plugin/in_tail/position_file.rb
```diff
@@ -224,5 +263,7 @@ def read_inode
         @inode
       end
     end
+
+    PathInodeTuple = Struct.new(:path, :ino)
```
repeatedly, line 267: What is the differences from `PathWithInode`?

cosmo0920: They are same behavior except for member names.
`PathWithInode` is not used anywhere. It should be removed the PR.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3182/commits/e294104fa28b1ed6dca2c7eda7a105834493a1b5.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3182/commits/d8019edd6c3625bd7e0fda7f5109becaa4909ac1.

lib/fluent/plugin/in_tail/position_file.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3182/commits/c1f41da80dbf9ff6289bb01031a956a3b7dc1cec.

lib/fluent/plugin/in_tail.rb
```diff
@@ -608,10 +657,12 @@ def on_timer
     end
 
     class TailWatcher
-      def initialize(path, pe, log, read_from_head, update_watcher, line_buffer_timer_flusher, io_handler_build)
+      def initialize(path, ino, pe, log, read_from_head, follow_inodes, update_watcher, line_buffer_timer_flusher, io_handler_build)
```
repeatedly, line 660: `path` and `ino` can be merged into `PathInodeTuple`?
It also reduces method argument in other places, e.g `unwatch`, `setup_watcher` and others.

Of course, separate PR is ok for me.

lib/fluent/plugin/in_tail/position_file.rb
```diff
@@ -224,5 +254,7 @@ def read_inode
...
```
repeatedly, line 258: I'm re-consindering this struct name. This name depends on members.
If we add more member in the future, this name will be invalid.
So `TargetInfo` or other name is better for the future.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3182/commits/72b9d4c6c7e0505e51da2f1f5024056bf71ef6c7.

lib/fluent/plugin/in_tail.rb
cosmo0920: Oh, that's reasonable. Will do.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3182/commits/08ee2da9c3ac895b747dbabd2ef2a0cada1bcc25.

cosmo0920: Ah, wait. Merging path and inode arguments makes slight a bit unstable for testing...
Perhaps we should do in the separated PR.

repeatedly (approved):
LGTM. 

There are several refactoring points.

- Some code still use `path_with_inode` name
- Some methods still take two `path`/`ino` arguments. It is not needed because we can use TargetInfo or TailWatcher instance for it, TailInput methods, TailWatcher methods, PositionFile methods and more. If we need to keep this two argument version, e.g. for path and different inode check, adding new TargetInfo version is better for avoiding `func(target_info.path, target_info.ino)` like redundant code.

New PR is better for large refactoring.

cosmo0920:
OK, let's merge this PR.
Refactoring should do in the another PR.

PR #3185, opened 2020-11-30
Add log throttling per file (revised)

cosmo0920:
None

From #2702:

> Running in a big cluster with high volume of log, it would be nice to throttle the log shipping to avoid network saturation and make it easier to calculate the max throughput per node for example in a Kubernetes cluster.

> Tail plugin is watching files and every second reading from the last pointer to the end of the file.
This change allow to stop reading the file after X number of logs lines read and update the pointer in the pos file as usual.

> This [commit](https://github.com/fluent/fluentd/pull/2702/commits/2ce05b1aa46737409e95e9d9725386fd1e385804) adds log throttling per bytes for each files, should work only when `watch_timer` is enabled and the `stat_watcher` (inotify) is disabled. 

> In order to have this feature for any watch configuration (timer or inotify..), I've updated to add a sleep when you have been reaching the bytes read limit, the sleep would block process and affect other file ingestion, I've added a basic thread array to have a multi-threading ingestion. 
However I've noticed you are relying on `cool.io`, and I was wondering if I should use this library instead.

> Would you be interested in this feature?

> Some [discussions](https://github.com/rewiko/fluentd/pull/1) before submitting the PR.

**Docs Changes**:

https://github.com/fluent/fluentd-docs-gitbook/pull/259
~~I'll register a PR to describe `read_bytes_limit_per_second` parameter.~~

**Release Note**: 

Same as title.


cosmo0920:
I'd confirmed that the following issue also exists in this PR for now:

>  In the [commit](https://github.com/fluent/fluentd/pull/2702/commits/2ce05b1aa46737409e95e9d9725386fd1e385804), I've been implementing the log throttling per file but this implementation work nicely only with you use the timer because you it will stop reading and if it did not get notified again then some logs might not be read. 
Maybe I could reuse that concept and more breaking I can push the function with a sleep and a notify which will be consume by a thread pool. In that case only the file throttled will create thread.

Disabling watch timer and enabling stat watcher circumstance, Fluentd reads left logs again only if they gets notified again.

cosmo0920:
I confirmed that:

* Enable watcher timer is needed. 
  * Because w/o timer watcher, only StatWatcher#on_change is caught and kicks to ship logs. log throttling feature is needed to kick to ship with periodic, not ought to be filesystem events based.
* Need to create thread pool mechanism to prevent stop the world with Kernel#sleep
  * Default thread pool size is 1 (single threaded)

cosmo0920:
@rewiko Could you try this patch out? I'd implemented thread pool feature for preventing "stop the world" due to `Kernel#sleep`.

rewiko:
@cosmo0920 Sure let me try.

rewiko:
## Test setup 

```
git clone https://github.com/fluent/fluentd.git 
git checkout log-throttling-per-file-revised 
sudo gem install bundler  
bundle install
bundle exec rake build
sudo gem install pkg/fluentd-1.12.0.rc2.gem 
```

```
fluentd -c ./fluentd.conf -v
```

with `fluentd.conf`:

```
<source>
  @type tail
  path logs.log
  tag all
  refresh_interval 1
  enable_stat_watcher false
  read_from_head true
  read_bytes_limit_per_second 8192
  max_thread_pool_size 2
  <parse>
  @type none
  </parse>
</source>

<match all>
  @type stdout
</match>
``` 

`logs.log` contains multiples lines of "lorum ipsum". 

I think everything looks good except `number_bytes_read += @lines.last ? @lines.last.size : 0`.
`@lines.size` does not give the number of bytes, not sure exactly what does it count. 
Also, it seems that will add 0 if that's not the last line, but in fact, we read 8192 bytes. 
I think that should be `number_bytes_read += 8192`, does not matter if it reached the end of the file at any read it try to read 8192 bytes. 
By using `number_bytes_read += 8192`, the throttling works as expected.

lib/fluent/plugin/in_tail.rb
```diff
@@ -892,16 +967,35 @@ def opened?
...
               read_more = false
 
               if !io.nil? && @lines.empty?
                 begin
                   while true
                     @fifo << io.readpartial(8192, @iobuf)
                     @fifo.read_lines(@lines)
-                    if @lines.size >= @read_lines_limit
+
+                    number_bytes_read += @lines.last ? @lines.last.size : 0
```
rewiko, line 980: I think it should be `number_bytes_read += 8192`. see comment https://github.com/fluent/fluentd/pull/3185#issuecomment-754545966

cosmo0920: Thanks! I’ll rewrite to use 8192 here.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/7a542c7405eae248a0d01189969efa238b5f4bd7.

cosmo0920:
Thanks for testing. Yeah, you are right.
I'd used `8192` for number_bytes_read instead of `number_bytes_read += @lines.last ? @lines.last.size : 0`.

test/plugin/test_in_tail.rb
```diff
@@ -285,6 +312,69 @@ def test_emit_with_read_lines_limit(data)
       assert num_events <= d.emit_count
     end
 
+    sub_test_case "log throttoling per file" do
```
rewiko, line 315: `throttoling` to `throttling`

rewiko:
Found a typo, but otherwise looks good to me. thanks for the help!

lib/fluent/plugin/in_tail.rb
```diff
@@ -892,16 +967,35 @@ def opened?
...
               read_more = false
 
               if !io.nil? && @lines.empty?
                 begin
                   while true
                     @fifo << io.readpartial(8192, @iobuf)
                     @fifo.read_lines(@lines)
-                    if @lines.size >= @read_lines_limit
+
+                    number_bytes_read += 8192
```
rewiko, line 980: `8192` could become a variable that is also used on line 977 

but not a big deal

test/plugin/test_in_tail.rb
cosmo0920: Oh! I’ll fix it. 😖

lib/fluent/plugin/in_tail.rb
cosmo0920: I think that they should be a constant instead of `8192` or variables.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/8d26d11073baa140e93e0a39765e732a0e78f11f.

rewiko:
@cosmo0920 what would be the next step ?

cosmo0920:
@repeatedly Could you kindly take a look? I'm not sure what we should do next....

repeatedly:
Okay, I will check it.

cosmo0920:
I'll rebase off master.

cosmo0920:
@ashie Could you take a look on this PR?

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1000,35 @@ def opened?
         def handle_notify
           with_io do |io|
             begin
+              number_bytes_read = 0
+              start_reading = Fluent::EventTime.now
```
ashie, line 1004: Probably it should be `Fluent::Clock.now` instead of `Fluent::EventTime.now`.
Otherwise it may cause unexpected high load when system time is changed.

ashie (changes requested):
I've add a comment but not yet complete to review.

lib/fluent/plugin/in_tail.rb
```diff
@@ -669,6 +710,51 @@ def on_timer
...
+        def stop
+          return unless @running
+          @running = false
+
+          until @queue.num_waiting == @threads.size
+            sleep 0.01
+          end
+          begin
+            Timeout.timeout(1) do
+              @threads.join
```
ashie, line 740: Do you intent `@threads.each {|th| th.join}`?
`@threads.join` is `Array`'s join so this code doesn't make sense (will return a joined string).

ashie (changes requested):

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/a03e43cad542af3d0a556e0f0dd0cd5ee2440d59.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/2a994ab074cee483723e3489dffc2816ec753ce5.

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1000,35 @@ def opened?
...
+                    number_bytes_read += BYTES_TO_READ
+                    limit_bytes_per_second_reached = (number_bytes_read >= @read_bytes_limit_per_second && @read_bytes_limit_per_second > 0)
+                    @log.debug("reading file: #{@path}")
+                    if @lines.size >= @read_lines_limit || limit_bytes_per_second_reached
                       # not to use too much memory in case the file is very large
                       read_more = true
+
+                      if limit_bytes_per_second_reached
+                        # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
+                        time_spent_reading = Fluent::EventTime.now - start_reading
```
ashie, line 1022: `Fluent::Clock.now`

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1000,35 @@ def opened?
...
+                        @log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
+                        if time_spent_reading < 1
+                          needed_sleeping_time = 1 - time_spent_reading
+                          @log.debug("sleep: #{needed_sleeping_time}")
+                          sleep(needed_sleeping_time)
+                        end
+                        start_reading = Fluent::EventTime.now
```
ashie, line 1029: Ditto

ashie (changes requested):

lib/fluent/plugin/in_tail.rb
cosmo0920: Oops! Fixed in https://github.com/fluent/fluentd/pull/3185/commits/dba436545387b426583d9404fa1fddb9659708b6.

lib/fluent/plugin/in_tail.rb
```diff
@@ -440,6 +477,9 @@ def stop_watchers(targets_info, immediate: false, unwatched: false, remove_watch
           end
         end
       }
+      if @thread_pool
+        @thread_pool.stop
```
ashie, line 481: Please consider adding `@thread_pool = nil` to clarify that `@thread_pool`never reused.

lib/fluent/plugin/in_tail.rb
```diff
@@ -669,6 +710,51 @@ def on_timer
...
+      class Pool
+        def initialize(max_size, &session)
+          @max_size = max_size
+          @queue = Queue.new
+          @threads = []
+          @running = false
+          session.call(self)
+          @running = true
+        ensure
+          stop
```
ashie, line 723: It will not clean up because `@running` is `false` while running `session.call(self)`, and `stop` doesn't nothing when `@running` is `false`.

ashie (changes requested):

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1000,35 @@ def opened?
...
+                      if limit_bytes_per_second_reached
+                        # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
+                        time_spent_reading = Fluent::Clock.now - start_reading
+                        @log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
+                        if time_spent_reading < 1
+                          needed_sleeping_time = 1 - time_spent_reading
+                          @log.debug("sleep: #{needed_sleeping_time}")
+                          sleep(needed_sleeping_time)
+                        end
+                        start_reading = Fluent::Clock.now
```
ashie, line 1029: Is it needed? It's reset at the head of this block.
If it's needed to reset here, I seems that `number_bytes_read` is also reset to 0.

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1000,35 @@ def opened?
...
+                    if @lines.size >= @read_lines_limit || limit_bytes_per_second_reached
                       # not to use too much memory in case the file is very large
                       read_more = true
+
+                      if limit_bytes_per_second_reached
+                        # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
+                        time_spent_reading = Fluent::Clock.now - start_reading
+                        @log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
+                        if time_spent_reading < 1
+                          needed_sleeping_time = 1 - time_spent_reading
```
ashie, line 1025: I think it's too deep nest, please consider splitting the method, or simplify the code.

test/plugin/test_in_tail.rb
```diff
@@ -323,6 +350,69 @@ def test_emit_with_read_lines_limit(data)
       assert num_events <= d.emit_count
     end
 
+    sub_test_case "log throttling per file" do
+      teardown do
+        FileUtils.rm_f("#{TMP_DIR}/tail.txt")
```
ashie, line 355: Should use `cleanup_file`

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/d15eae9304651c4f888355496ca08b0c9db24f60.

lib/fluent/plugin/in_tail.rb
```diff
@@ -669,6 +710,51 @@ def on_timer
...
+            @threads.each {|th| th.kill }
+          end
+        end
+
+        protected def create_thread
+          Thread.start(@queue) do |q|
+            loop do
+              task = q.pop
+              task.call
+            end
```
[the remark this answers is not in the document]
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/d15eae9304651c4f888355496ca08b0c9db24f60.

test/plugin/test_in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/5fdabd3f6238d0f5524dcf966d94c84583f8f36b.

lib/fluent/plugin/in_tail.rb
cosmo0920: I tried to handle more properly to terminate `@queue`.
But I not sure this patch https://github.com/fluent/fluentd/pull/3185/commits/e20d0f30e6844a10f05363bb81f5f10ba6d7c0f6 is appropriate for our use case.

lib/fluent/plugin/in_tail.rb
cosmo0920: Oh. You are right.
Fixed in https://github.com/fluent/fluentd/pull/3185/commits/758f03c8fcd00ffcea0703e5baab58c930218204.

lib/fluent/plugin/in_tail.rb
```diff
@@ -911,16 +1011,34 @@ def opened?
...
 
               if !io.nil? && @lines.empty?
                 begin
                   while true
-                    @fifo << io.readpartial(8192, @iobuf)
+                    @fifo << io.readpartial(BYTES_TO_READ, @iobuf)
                     @fifo.read_lines(@lines)
-                    if @lines.size >= @read_lines_limit
+
+                    number_bytes_read += BYTES_TO_READ
```
ashie, line 1024: It doesn't count actual read bytes correctly.
`BYTES_TO_READ` isn't actual read bytes, it maximum bytes to read.

test/plugin/test_in_tail.rb
```diff
@@ -323,6 +350,69 @@ def test_emit_with_read_lines_limit(data)
...
+            end
+          }
+        end
+
+        assert do
+          count_thread_pool_object >= 1
+        end
+
+        events = d.events
+        assert_true(events.length <= num_events)
```
ashie, line 410: I think this test doesn't check this feature correctly, this assertion is too lough.
In addition, this feature seems cause hard time out while shutting down the plugin. The test shows following error:

```
E, [2021-04-05T18:47:53.212805 #298497] ERROR -- : unexpected error while after_shutdown, RuntimeError:event loop does NOT exit until hard timeout.
E, [2021-04-05T18:47:53.212914 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/plugin_helper/event_loop.rb:130:in `after_shutdown'
E, [2021-04-05T18:47:53.212929 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:156:in `block in instance_shutdown'
E, [2021-04-05T18:47:53.212940 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:144:in `block in instance_shutdown'
E, [2021-04-05T18:47:53.212950 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:156:in `instance_shutdown'
...
```

ashie (changes requested):

test/plugin/test_in_tail.rb
ashie: > I think this test doesn't check this feature correctly, this assertion is too rough.

For example `num_events` of first data is `21` but it doesn't make sense because only 21 lines are inserted to the test file.

lib/fluent/plugin/in_tail.rb
cosmo0920: Previously, I'd tried to count actual bytes.

But the previous approach causes unintended behavior:

> I think that should be number_bytes_read += 8192, does not matter if it reached the end of the file at any read it try to read 8192 bytes.
> By using number_bytes_read += 8192, the throttling works as expected.

ref: https://github.com/fluent/fluentd/pull/3185#issuecomment-754545966

ashie: > I think everything looks good except number_bytes_read += @lines.last ? @lines.last.size : 0.

It should be like this:
```diff
diff --git a/lib/fluent/plugin/in_tail.rb b/lib/fluent/plugin/in_tail.rb
index 2ca25494..ac8794ac 100644
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -1018,10 +1018,11 @@ module Fluent::Plugin
               if !io.nil? && @lines.empty?
                 begin
                   while true
-                    @fifo << io.readpartial(BYTES_TO_READ, @iobuf)
+                    data = io.readpartial(BYTES_TO_READ, @iobuf)
+                    number_bytes_read += data.bytesize
+                    @fifo << data
                     @fifo.read_lines(@lines)
 
-                    number_bytes_read += BYTES_TO_READ
                     limit_bytes_per_second_reached = (number_bytes_read >= @read_bytes_limit_per_second && @read_bytes_limit_per_second > 0)
                     @log.debug("reading file: #{@path}")
                     if @lines.size >= @read_lines_limit || limit_bytes_per_second_reached
```
In the previous code you only count the last line, but incoming data may be like this
```
test
test
...
test
```

cosmo0920: Oh! I'd overlooked for `#bytesize`. :stuck_out_tongue:

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/92aadec7038204cc54cf241a21eb9e8fc0b5c573.

lib/fluent/plugin/in_tail.rb
```diff
@@ -669,6 +710,51 @@ def on_timer
...
+        def stop
+          return unless @running
+          @running = false
+
+          until @queue.num_waiting == @threads.size
+            sleep 0.01
+          end
+          begin
+            Timeout.timeout(1) do
+              @threads.each{|th| th.join }
```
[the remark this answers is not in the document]
cosmo0920: I'd tried to more proper queue termination in: https://github.com/fluent/fluentd/pull/3185/commits/382df6cfcdba3341b95aa5efd3050e0a398199ca.
But with this patch, I got still `terminating event_loop forcedly` errors.

test/plugin/test_in_tail.rb
cosmo0920: > > I think this test doesn't check this feature correctly, this assertion is too rough.
> 
> For example `num_events` of first data is `21` but it doesn't make sense because only 21 lines are inserted to the test file.

I made testcases to be more robust:
https://github.com/fluent/fluentd/pull/3185/commits/f0228afa164343480336aba299dd3c956c10d0c9

lib/fluent/plugin/in_tail.rb
cosmo0920: I decided to use `@queue` instance status to check whether thread pool is running or not:
https://github.com/fluent/fluentd/pull/3185/commits/41aa0cb0b193f9da0d477d2d57b36bd22b39ae5f

lib/fluent/plugin/in_tail.rb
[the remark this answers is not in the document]
cosmo0920: On testing, in_tail plugin instance is always blocked for termination by `shutdown: true` (default driver#run keyword arguments).
This is normal behavior on `in_tail` during tailing and emitting events phase.
`in_tail` does not support immediate stop on tailing files left.
 
In https://github.com/fluent/fluentd/pull/3185/commits/2d67e44608408cea32f7abfdfe2f2447e3915243, I specified `shutdown: false` to prevent hard timeout messages generating.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/520b335c38f58435a94edfeb0f2637d9fc84b08a.

cosmo0920:
I fixed the issues which was pointed out.
Could you take a look again, @ashie, @kenhys?

test/plugin/test_in_tail.rb
ashie: > In addition, this feature seems cause hard time out while shutting down the plugin. The test shows following error:
> 
> ```
> E, [2021-04-05T18:47:53.212805 #298497] ERROR -- : unexpected error while after_shutdown, RuntimeError:event loop does NOT exit until hard timeout.
> E, [2021-04-05T18:47:53.212914 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/plugin_helper/event_loop.rb:130:in `after_shutdown'
> E, [2021-04-05T18:47:53.212929 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:156:in `block in instance_shutdown'
> E, [2021-04-05T18:47:53.212940 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:144:in `block in instance_shutdown'
> E, [2021-04-05T18:47:53.212950 #298497] ERROR -- : 	/home/aho/Projects/Fluentd/fluentd/lib/fluent/test/driver/base.rb:156:in `instance_shutdown'
> ...
> ```

I'm discussing with @cosmo0920 that this issue doesn't seems this PR's problem, it may be a same issue with #3323.

ashie:
Looks good, but I'm still further checking...
BTW I'll release 1.12.3 before 1.13 to fix #3327.

cosmo0920:
> Looks good, but I'm still further checking...
> BTW I'll release 1.12.3 before 1.13 to fix #3327.

Yeah, that's sure. I'm waiting further reviewing comments.

lib/fluent/plugin/in_tail.rb
```diff
@@ -107,6 +109,8 @@ def initialize
     config_param :path_timezone, :string, default: nil
     desc 'Follow inodes instead of following file names. Guarantees more stable delivery and allows to use * in path pattern with rotating files'
     config_param :follow_inodes, :bool, default: false
+    desc 'Specify max size of thread pool'
+    config_param :max_thread_pool_size, :integer, default: 1
```
ashie, line 113: We should describe that this parameter is for the log throttling feature.

lib/fluent/plugin/in_tail.rb
```diff
@@ -178,6 +182,25 @@ def configure(conf)
...
+        end
+        if !@enable_watch_timer
+          raise Fluent::ConfigError, "Need to enable watch timer when using log throttling feature"
+        end
+        if @read_bytes_limit_per_second < 8192
+          log.warn "Should specify greater equal than 8192. Use 8192 for read_bytes_limit_per_second"
+          @read_bytes_limit_per_second = 8192
+        end
+        if @max_thread_pool_size < 2
+          raise Fluent::ConfigError, "Specify 2 or more on max_thread_pool_size"
```
ashie, line 200: Although the default value of `@max_thread_pool_size` is `1`, it should be `2` or more.
Otherwise users are always required to add `@max_thread_pool_size` config.

lib/fluent/plugin/in_tail.rb
```diff
@@ -392,37 +415,51 @@ def setup_watcher(target_info, pe)
...
+        detach_watcher(tw, target_info.ino, false)
+      end
+    end
+
+    def start_watchers(targets_info)
+      if @enable_read_bytes_limit_thread_pool
+        @thread_pool = TailThread::Pool.new(@max_thread_pool_size) do |pool|
+          targets_info.each_value {|target_info|
+            pool.run {
+              construct_watcher(target_info)
```
ashie, line 454: Probably only start-up task is multi-threaded, trailing all timer or inotify handlers are run on the same (main) thread because they are invoked by the main event loop.

lib/fluent/plugin/in_tail.rb
```diff
@@ -909,19 +1004,40 @@ def opened?
...
         private
 
+        def sleep_for_log_ingestion(start_reading)
+          # sleep to stop reading files when we reach the read bytes per second limit, to throttle the log ingestion
+          time_spent_reading = Fluent::Clock.now - start_reading
+          @log.debug("time_spent_reading: #{time_spent_reading} #{ @watcher.path}")
+          if time_spent_reading < 1
+            needed_sleeping_time = 1 - time_spent_reading
+            @log.debug("sleep: #{needed_sleeping_time}")
+            sleep(needed_sleeping_time)
```
ashie, line 1014: Why it's needed to sleep? I think just abort reading the file and wait next timer event is more elegant.

ashie (changes requested):
Now I'm doubting the base design of multi-threading model of this patch.
Probably only start-up task is multi-threaded, trailing all timer or inotify handlers are run on the same (main) thread because they are invoked by the main event loop.

https://github.com/fluent/fluentd/blob/af2ea9bb3af594bf0a1fbb18eaf701c26ebcf291/lib/fluent/plugin/in_tail.rb#L401-L403

lib/fluent/plugin/in_tail.rb
ashie: Although this feature requires `enable_watch_timer`:

https://github.com/fluent/fluentd/blob/af2ea9bb3af594bf0a1fbb18eaf701c26ebcf291/lib/fluent/plugin/in_tail.rb#L192-L194

I think it's not needed if we use `sleep` instead of waiting next timer.

lib/fluent/plugin/in_tail.rb
cosmo0920: Hmm..., that's right.
Using multi-threaded only for start-up tasks is not enough for preventing to block from stop the world with `#sleep`.
But, I guess that we needn't to use queue which is constructed by thread pool.

lib/fluent/plugin/in_tail.rb
cosmo0920: After removing thread pool feature, we need to remove this needless parameter.

lib/fluent/plugin/in_tail.rb
cosmo0920: After removing `max_thread_pool_size` parameter, this validation is not needed and ought to be removed.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/d3a3023fc8e3d6b06fc37c7f432fc91b5694471f.

lib/fluent/plugin/in_tail.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3185/commits/478b487383e1dec1af86cb0def2000b515eb6577.

lib/fluent/plugin/in_tail.rb
cosmo0920: Removed in https://github.com/fluent/fluentd/pull/3185/commits/478b487383e1dec1af86cb0def2000b515eb6577.

lib/fluent/plugin/in_tail.rb
```diff
@@ -909,19 +928,41 @@ def opened?
...
+                    limit_bytes_per_second_reached = (number_bytes_read >= @read_bytes_limit_per_second && @read_bytes_limit_per_second > 0)
+                    @log.debug("reading file: #{@path}")
+                    if @lines.size >= @read_lines_limit || limit_bytes_per_second_reached
                       # not to use too much memory in case the file is very large
                       read_more = true
+
+                      read_more = false if limit_bytes_per_second_reached && need_to_get_out_for_log_ingestion_loop?(start_reading)
+
                       break
                     end
```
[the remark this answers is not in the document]
cosmo0920: I got it.
Fixed in https://github.com/fluent/fluentd/pull/3185/commits/147ffbfad2c1a24dce219528b99c5d933a98aa01.

lib/fluent/plugin/in_tail.rb
```diff
@@ -909,21 +928,49 @@ def opened?
 
         private
 
+        def limit_bytes_per_second_reached?(start_reading, number_bytes_read)
+          if number_bytes_read >= @read_bytes_limit_per_second && @read_bytes_limit_per_second > 0
```
ashie, line 932: Can return earlier.

ashie:
Thanks for revising the patch.
It seems much better than before.
Interestingly it's similar with #2478.
I'll do final check (probably no more comments though).
Although it still might have remaining issue like mentioned in https://github.com/fluent/fluentd/issues/2478#issuecomment-512086345, I think they should be addressed in another pull request.

cosmo0920:
> Although it still might have remaining issue like mentioned in [#2478 (comment)](https://github.com/fluent/fluentd/issues/2478#issuecomment-512086345), I think they should be addressed in another pull request.

Sure. This pull request too long to add more commits/features. Let's move out the another PR.

ashie:
Sorry, I missed one more problem.
If a log file is growing frequently (less than 1sec) and inotify is enabled, probably it doesn't limit reading pace expectedly.
It can be another PR though.

ashie (approved):
LGTM
Remaining issues should be addressed in other PR.

cosmo0920:
> > Oh. That is one reason that I used `#sleep` to prevent log ingestion. I have no idea how to prevent log tailing when inotify and timer watcher are enabled confounded
> 
> I think storing last read time & bytes into member variables will resolve it.

Thanks for the hints,
I'll send an another PR to implement this.

rewiko:
@ashie @cosmo0920  thank you 😃

PR #3189, opened 2020-11-30
Add ignore_if_prev_successes to copy plugin

tiagogoddard-hotmart:
None

This PR makes a change in the Copy Output plugin so you can inform the fluentd to ignore said copy if the previous output succeeded.

This enables having multiple nested outputs as backup of each other by doing:

```
    <store ignore_error>
      @type test
      name c0
    </store>
    <store ignore_if_prev_success ignore_error>
      @type test
      name c1
    </store>
    <store ignore_if_prev_success>
      @type test
      name c2
    </store>
```

**Docs Changes**:

Adding the defined ignore_if_prev_successes attribute to the <store> tag makes the copy plugin ignore said store if the previous store succeed in pushing the output.

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,3 @@
+---
+BUNDLE_RETRY: "3"
+BUNDLE_JOBS: "3"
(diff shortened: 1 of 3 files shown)


TiagoGoddard:
Added another PR, I did a lot of wrong on this PR and branches trying to do the DCO

PR #3197, opened 2020-12-10
feat: added support for query params in in_http plugin

### LINKED ISSUE #3188: Add support for query params in in_http input plugin

I am using fluentd to collect CSP violation report, I would like to add query params, like the website version to help us quickly spot issues related to a release.

Add a property like `add_query_params`, that when set will parse the query params and add them to the record.
 
Passing it using a path and parsing the path with regex

**Additional context**

I'm willing to open a PR if the solution is acceptable

omerlh:
Fix #3188

Added `add_query_params` config option to in_http, when set it will add all query params to the record

**Docs Changes**:
I think so

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -80,6 +80,8 @@ module Fluent::Plugin
     config_param :use_204_response, :bool, default: false
     desc 'Dump error log or not'
     config_param :dump_error_log, :bool, default: true
+    desc 'Add QUERY_ prefix query params to record'
+    config_param :add_query_params, :bool, default: false
 
     config_section :parse do
       config_set_default :@type, 'in_http'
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_http.rb
```diff
@@ -533,7 +543,14 @@ def on_message_complete
         end
         path_info = uri.path
 
+        query_params = WEBrick::HTTPUtils.parse_query(uri.query)
+
+        query_params.each_pair {|k,v|
+          params["QUERY_#{k.gsub('-','_').upcase}"] = v
+        }
```
repeatedly, line 550: Adding `if @add_query_params` is better.
Pass `@add_query_params` to `Handler`.

omerlh: Sure, added!

omerlh:
I am not sure why the tests are failing, but it does not look related to my changes - unless I am missing something?

repeatedly:
Thanks!

> it does not look related to my changes

Yes. travis-ci recently has random failure problem with openssl. It seems travis-ci specific issue.
We are now migrating travis-ci into github actions for CI.

PR #3213, opened 2020-12-23
in_tcp: show listening address and port for consistency with in_udp

kenhys:

Not bound to specific issue

in_udp plugin shows log message when starting server, but this
behavior is not true for in_tcp plugin.
For consistency it is better to show startup message.

  [info]: #0 listening tcp socket bind="..." port=...

This change is introduced that there is a case that inconsistency log
leads confusion when in_tcp and in_udp is used at the same time.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/in_tcp.rb
+++ b/lib/fluent/plugin/in_tcp.rb
@@ -98,6 +98,7 @@ module Fluent::Plugin
     def start
       super
 
+      log.info "listening tcp socket", bind: @bind, port: @port
       del_size = @delimiter.length
       if @_extract_enabled && @_extract_tag_key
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_tcp.rb
```diff
@@ -98,6 +98,7 @@ def multi_workers_ready?
     def start
       super
 
+      log.info "listening tcp socket", bind: @bind, port: @port
```
cosmo0920, line 101: :+1:

NOTE: in_udp plugin:
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_udp.rb#L71

cosmo0920 (approved):

repeatedly:
Thanks!

PR #3220, opened 2020-12-31
Support unixtime_micros and unixtime_nanos to inject time

abicky:

This PR adds new time types "unixtime_micros" and "unixtime_nanos" to inject time for the same reason as https://github.com/fluent/fluentd/pull/3145.

If we collect docker container logs using the fluentd logging driver, we need to sort the logs by the time to merge logs from various containers. Although we can now specify unixtime_millis, logs with line feed sometimes have the same time in milliseconds. As a result, the logs are sorted in the wrong order. This PR will resolve the problem by introducing new types of unixtime with higher precision.

**Docs Changes**:
Needed. I'll create a pull request to change [configuration/inject-section.md#time-parameters](https://github.com/fluent/fluentd-docs-gitbook/blob/71e49c869fb4e5416a528f6d77b35f8a67b17bac/configuration/inject-section.md#time-parameters) if this PR is merged.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin_helper/inject.rb
+++ b/lib/fluent/plugin_helper/inject.rb
@@ -132,7 +132,7 @@ module Fluent
           if @_inject_time_key
             @_inject_time_formatter = case @inject_config.time_type
                                       when :float then ->(time){ time.to_r.truncate(+6).to_f } # microsecond floating point value
-                                      when :unixtime_millis then ->(time) { time.to_f.floor(3) * 1000 }
+                                      when :unixtime_millis then ->(time) { time.respond_to?(:nsec) ? time.to_i * 1_000 + time.nsec / 1_000_000 : (time * 1_000).floor }
                                       when :unixtime then ->(time){ time.to_i }
                                       else
                                         localtime = @inject_config.localtime && !@inject_config.utc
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin_helper/inject.rb
```diff
@@ -132,7 +132,9 @@ def configure(conf)
           if @_inject_time_key
             @_inject_time_formatter = case @inject_config.time_type
                                       when :float then ->(time){ time.to_r.truncate(+6).to_f } # microsecond floating point value
-                                      when :unixtime_millis then ->(time) { time.to_f.floor(3) * 1000 }
+                                      when :unixtime_millis then ->(time) { time.respond_to?(:nsec) ? time.to_i * 1_000 + time.nsec / 1_000_000 : (time * 1_000).floor }
```
abicky, line 135: By this change, this proc will return a different value in some cases as below, but I think the new value is more appropriate:

```
irb(main):001:0> time = Fluent::EventTime.new(1466464211, 320999980)
=> 2016-06-21 08:10:11.320999980 +0900
irb(main):002:0> time.to_f.floor(3) * 1000
=> 1466464211321.0
irb(main):003:0> time.to_i * 1_000 + time.nsec / 1_000_000
=> 1466464211320
```

repeatedly:
Looks good. This will be included in v1.12.0.

PR #3240, opened 2021-02-02
fluent-plugin-config-format: use markdown table

kenhys:
You can see list of parameters at a glance with --compact option

Before:

```
  * name (type) (required?) description

  * Available values: ...

  * Default value: ...

  * Alias:

  * Deprecated:

  * Obsoleted:
```

After:

```
  ### Configuration

  |parameter|type|description|default|
  |---|---|---|---|
  |...|...|...|...|
```

### DIFF
--- a/lib/fluent/command/plugin_config_formatter.rb
+++ b/lib/fluent/command/plugin_config_formatter.rb
@@ -162,9 +162,20 @@ class FluentPluginConfigFormatter
     else
       sections, params = base_section.partition {|_name, value| value[:section] }
     end
+    if @table and not params.empty?
+      dumped << "### Configuration\n\n"
+      dumped << "|parameter|type|description|default|\n"
+      dumped << "|---|---|---|---|\n"
+    end
     params.each do |name, config|
       next if name == :section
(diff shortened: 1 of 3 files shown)


cosmo0920 (changes requested):
It seems that this patch contains breaking changes.
We should add a option `--table` or something.

lib/fluent/command/plugin_config_formatter.rb
```diff
@@ -162,6 +162,11 @@ def dump_section_markdown(base_section, level = 0)
     else
       sections, params = base_section.partition {|_name, value| value[:section] }
     end
+    if @compact and not params.empty?
+      dumped << "### Configuration\n\n"
+      dumped << "|parameter|type|description|default|\n"
+      dumped << "|---|---|---|---|\n"
+    end
```
cosmo0920, line 169: I want to add a separated instance variable to handle markdown table....

kenhys: It is reasonable, I'll fix.

cosmo0920 (approved):
Looks good. Thanks for your work! :muscle:

PR #3241, opened 2021-02-02
out_http: Add headers_from_placeholders. fix #3237

### LINKED ISSUE #3237: Support dynamic placeholder for out_http headers parameter

Today I can't use dynamic headers for out_http plugin.  

We need the ability for headers to be dynamic in nature-based upon values inside the record.  Today they can only be strings, but the goal would be to do something like `header X-Custom-Host:${$.kubernetes.host},X-Custom-Datacenter:${$.datacenter}`

I have tried all configuration options I could think with no luck.  After exhausting all options, I finally reached out to the community and got a response from @repeatedly saying:

> Currently, out_http's headers parameter doesn't support dynamic placeholder.
> For it, we need to add headers_from_record like parameter with buffer placeholder.

**Additional context**

The question was posted on the Fluentd Google Group with the subject, "out_http headers not dynamic".

repeatedly:

Fixes #3237

Add `headers_from_placeholders` parameter to set placeholder based header.

```
headers_from_placeholders {"x-foo-bar":"${$.foo.bar}","x-tag":"app-${tag}"}
<buffer tag,$.foo.bar>
</buffer>
```

**Docs Changes**:
Add `headers_from_placeholders` to `out_http` article.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -51,6 +51,8 @@ module Fluent::Plugin
     config_param :json_array, :bool, default: false
     desc 'Additional headers for HTTP request'
     config_param :headers, :hash, default: nil
+    desc 'Additional placeholder based headers for HTTP request'
+    config_param :headers_from_placeholders, :hash, default: nil
 
     desc 'The connection open timeout in seconds'
     config_param :open_timeout, :integer, default: nil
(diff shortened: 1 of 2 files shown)


cosmo0920 (approved):
Looks good. :+1:

PR #3250, opened 2021-02-15
Add an option to disable shared socket forcedly

ashie:
none

Fluentd always creates ServerEngine::SocketManager's shared socket even if it's not required. The shared socket isn't needed if there is no plugin that uses it such as in_forward, in_http and in_syslog.
Although it can be disabled if Fluentd doesn't use multiple workers, this patch adds `disable_shared_socket` instead of checking number of workers to keep backward compatibility for third-party plugins (`PluginHelper::Server.create_server_connection` uses shared socket by default).

Opening a needless socket may cause conflict with other applications especially on Windows because `ServerEngine::SocketManager` uses TCP socket on Windows.

**Docs Changes**:

* Add `disable_shared_socket` to https://docs.fluentd.org/deployment/system-config#parameters.
* Add `--disable-shared-socket` to https://docs.fluentd.org/deployment/command-line-option

**Release Note**: 
Same with the title.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -163,6 +163,10 @@ op.on('--conf-encoding ENCODING', "specify configuration file encoding") { |s|
   opts[:conf_encoding] = s
 }
 
+op.on('--disable-shared-socket', "Don't open shared socket for multiple workers") { |b|
+  opts[:disable_shared_socket] = b
+}
+
 if Fluent.windows?
   require 'windows/library'
   include Windows::Library
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -45,6 +45,7 @@ module Fluent
   module ServerModule
     def before_run
       @fluentd_conf = config[:fluentd_conf]
+      @rpc_endpoint = nil
       @rpc_server = nil
       @counter = nil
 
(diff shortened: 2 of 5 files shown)


ashie:
Hmm, a test failed:

```
Failure: test_disable_shared_socket(SupervisorTest): <"/tmp/SERVERENGINE_SOCKETMANAGER_2021-02-15T08:49:14Z_8557"> was expected to be nil.
/home/travis/build/fluent/fluentd/test/test_supervisor.rb:539:in `test_disable_shared_socket'
     536:     end
     537:     begin
     538:       server.before_run
  => 539:       assert_nil(ENV['SERVERENGINE_SOCKETMANAGER_PATH'])
     540:     ensure
     541:       server.after_run
     542:     end
```

ashie:
> Hmm, a test failed while it's not reproduced on my local environment:

I've add a commit to fix it but still still broken on windows:
```
C:/projects/fluentd/test/test_supervisor.rb:544:in `test_disable_shared_socket'
     541:       server.before_run
     542:       assert_nil(ENV['SERVERENGINE_SOCKETMANAGER_PATH'])
     543:     ensure
  => 544:       server.after_run
     545:       ENV.delete('SERVERENGINE_SOCKETMANAGER_PATH')
     546:     end
     547:   end
```

ashie:
> I've add a commit to fix it but still still broken on windows:

I've fixed for Windows too.

Other failed tests doesn't seem concerned with this PR. They are still failed even if I revert this PR.
I'll investigate them in another issue or PR.

repeatedly:
Thanks!

ashie:
> **Docs Changes**:
> 
>     * Add `disable_shared_socket` to https://docs.fluentd.org/deployment/system-config#parameters.
>     * Add `--disable-shared-socket` to https://docs.fluentd.org/deployment/command-line-option

I'll send PR for it.

PR #3095, opened 2020-08-04
buffer: Enable optimization for metadata comparison in all platform

cosmo0920:
Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>

With this patch, Fluentd on armv7l Linux environment can handle optimized `#hash` method in Metadata.

```
fluent@497f8dcf9876:/fluentd$ uname -a
Linux 497f8dcf9876 4.9.140-tegra #1 SMP PREEMPT Mon Dec 9 22:52:02 PST 2019 aarch64 GNU/Linux
```

<details>

From outside armhf container:

```
$ for i in {1..100}; do echo '{"status":"OK"}' | bundle exec fluent-cat test; done
```

```
fluent@497f8dcf9876:/$ echo '<source>
  @type forward
  port 24224
</source>

<match test.**>
  @type file
  path log/%Y%m%d.${tag}
  append true
  <buffer tag,time>
    @type file
    path log/
    timekey 1d
  </buffer>
</match>
' > /fluentd/etc/issue3088.conf
fluent@497f8dcf9876:/$ cd fluentd
fluent@497f8dcf9876:/fluentd$ fluentd -c /fluentd/etc/issue3088.conf -vvv
2020-08-04 08:32:01 +0000 [info]: fluent/log.rb:327:info: parsing config file is succeeded path="/fluentd/etc/issue3088.conf"
2020-08-04 08:32:01 +0000 [info]: fluent/log.rb:327:info: gem 'fluentd' version '1.11.1'
2020-08-04 08:32:01 +0000 [trace]: fluent/log.rb:284:trace: registered output plugin 'file'
2020-08-04 08:32:02 +0000 [trace]: fluent/log.rb:284:trace: registered buffer plugin 'file'
2020-08-04 08:32:02 +0000 [trace]: fluent/log.rb:284:trace: registered formatter plugin 'out_file'
2020-08-04 08:32:02 +0000 [trace]: fluent/log.rb:284:trace: registered input plugin 'forward'
2020-08-04 08:32:02 +0000 [debug]: fluent/log.rb:306:debug: No fluent logger for internal event
2020-08-04 08:32:02 +0000 [info]: fluent/log.rb:327:info: using configuration file: <ROOT>
  <source>
    @type forward
    port 24224
  </source>
  <match test.**>
    @type file
    path "log/%Y%m%d.${tag}"
    append true
    <buffer tag,time>
      @type "file"
      path "log/"
      timekey 1d
    </buffer>
  </match>
</ROOT>
[...]

PR #3252, opened 2021-02-17
Support multiple kind of timestamp format

### LINKED ISSUE #3248: Fluentd json parser to support multiple timestamp format for same set of logs

Fluentd json parser only supports a timestamp format in filter rule. I am using fluentd to parse Kubernetes docker logs and they have different timestamp formats. Some of them are in unix time format and others are in rfc3339 format. In order to support this use case, I have to explicitly define different filters and specify different time format for those json logs

fluentd json plug that accepts multiple timestamp formats and use them to parse json logs. there are probably better ways to do it as that doesn't sound very efficient.

I use multiple filters to specify different timestamp formats.
```
<source>
  @type http
  port 9880
  bind 0.0.0.0
</source>

<filter vault.**>
  @type parser
  <parse>
    @type json
    json_parser json
    time_type string
    time_format "%Y-%m-%dT%H:%M:%S"
  </parse>
  key_name log
  hash_value_field json_log
  replace_invalid_sequence true
  remove_key_name_field true
  reserve_data true
</filter>

<filter **>
  @type parser
  <parse>
    @type json
    json_parser json
  </parse>
  key_name log
  hash_value_field json_log
  replace_invalid_sequence true
  emit_invalid_record_to_error true
  remove_key_name_field true
  reserve_data true
</filter>

<match **>
  @type stdout
</match>
```

**Additional context**

kenhys:
Use Case 1: Most of timestamp format is unixtime, but some entry use %iso8601

```
time_type mixed
time_format_fallbacks unixtime, %iso8601
```

Use Case 2: Most of timestamp format is %iso8601, but some entry use unixtime

```
time_type mixed
time_format_fallbacks %iso8601, unixtime
```

Fixes #3248

Even though time format is changed from time to time, there is no
need to workaround (define multiple filter for each format)

**Docs Changes**:

It should be documented on https://docs.fluentd.org/configuration/parse-section#time-parameters

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -132,13 +132,14 @@ module Fluent
   end
 
   module TimeMixin
-    TIME_TYPES = ['string', 'unixtime', 'float']
+    TIME_TYPES = ['string', 'unixtime', 'float', 'mixed']
 
     TIME_PARAMETERS = [
       [:time_format, :string, {default: nil}],
       [:localtime, :bool, {default: true}],  # UTC if :localtime is false and :timezone is nil
       [:utc,       :bool, {default: false}], # to turn :localtime false
       [:timezone, :string, {default: nil}],
+      [:time_format_fallbacks, :array, {default: []}], # try time_format, then try fallbacks
     ]
     TIME_FULL_PARAMETERS = [
       # To avoid to define :time_type twice (in plugin_helper/inject)
@@ -170,6 +171,12 @@ module Fluent
           raise Fluent::ConfigError, "both of utc and localtime are specified, use only one of them"
         end
 
+        if conf.has_key?('time_type') and @time_type == :mixed
+          if @time_format.nil? and @time_format_fallbacks.empty?
+            raise Fluent::ConfigError, "time_type is :mixed but time_format and time_format_fallbacks is empty."
+          end
+        end
+
         Fluent::Timezone.validate!(@timezone) if @timezone
       end
     end
@@ -180,6 +187,7 @@ module Fluent
       end
 
       def time_parser_create(type: @time_type, format: @time_format, timezone: @timezone, force_localtime: false)
+        return MixedTimeParser.new(type, format, @localtime, timezone, @utc, force_localtime, @time_format_fallbacks) if type == :mixed
         return NumericTimeParser.new(type) if type != :string
         return TimeParser.new(format, true, nil) if force_localtime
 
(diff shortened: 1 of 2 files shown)


kenhys:
This is PoC PR, so  I make it draft PR.

cosmo0920:
I think that we should do a micro benchmarking like as: https://github.com/fluent/fluentd/pull/3095#issuecomment-668607116

cosmo0920:
I executed a micro benchmark for this implementation.

```ruby
require 'benchmark/ips'
require 'fluent/time'
require 'fluent/config'
require 'fluent/configurable'

class DummyForTimeParser
  include Fluent::Configurable
  include Fluent::TimeMixin::Parser
end

def config_element(name = 'test', argument = '', params = {}, elements = [])
  Fluent::Config::Element.new(name, argument, params, elements)
end

a = Fluent::TimeParser.new('%d/%b/%Y:%H:%M:%S %z')
i = DummyForTimeParser.new
i.configure(config_element('parse', '',
                           {'time_type' => 'mixed',
                            'time_format' => '%iso8601',
                            'time_format_fallbacks' => ['unixtime']}))
b = i.time_parser_create
i2 = DummyForTimeParser.new
i2.configure(config_element('parse', '', {'time_type' => 'mixed',
                                          'time_format_fallbacks' => ['unixtime', '%iso8601']}))
c = i2.time_parser_create
time = Time.now
timesec = "#{time.sec}".freeze

Benchmark.ips do |x|
  x.report("Normal time parser") do
    a.parse('28/Feb/2013:12:00:00 +0900'.freeze)
  end

  x.report("primary format is %iso8601, secondary is unixtime") do
    b.parse('2021-01-01T12:00:00+0900'.freeze)
  end

  x.report("fallback to unixtime") do
    c.parse(timesec)
  end
end
```

And I got the result:

```log
  Normal time parser   354.439k i/100ms
primary format is %iso8601, secondary is unixtime
                        89.717k i/100ms
fallback to unixtime    41.537k i/100ms
  Normal time parser      3.919M (±38.1%) i/s -     16.304M in   5.084468s
primary format is %iso8601, secondary is unixtime
                          1.970M (±31.0%) i/s -      7.447M in   5.013618s
fallback to unixtime      1.675M (±37.2%) i/s -      6.521M in   5.016029s
```

kenhys:
As we expected, this feature affects performance.

Pros.

* We can handle mixed timestamp format (good if we can't modify collector behavior)

Cons.

* Need to pay performance penalty cost when fallback occurs.

Then, I doubt that the current fallback notation is reasonable or not.

Anyway I'll change this PR status if CI succeeds.

cosmo0920 (approved):
I think this is reasonable change. Thanks!

lib/fluent/time.rb
```diff
@@ -170,6 +171,12 @@ def configure(conf)
           raise Fluent::ConfigError, "both of utc and localtime are specified, use only one of them"
         end
 
+        if conf.has_key?('time_type') and @time_type == :mixed
+          if @time_format.nil? and @time_format_fallbacks.empty?
+            raise Fluent::ConfigError, "Missing time_type is :mixed but time_format and time_format_fallbacks is empty."
```
ashie, line 176: Is `Missing` in the message needed?

lib/fluent/time.rb
```diff
@@ -452,4 +460,56 @@ def format_float(time)
...
+          else
+            localtime = localtime && (timezone.nil? && !utc)
+            @parsers << TimeParser.new(fallback, localtime, timezone)
+          end
+        end
+      end
+    end
+
+    def parse(value)
+      parsed = nil
```
ashie, line 494: Is the variable `parsed` needed?

test/test_time_parser.rb
```diff
@@ -228,4 +228,69 @@ class DummyForTimeParser
...
+    test 'primary format is %iso8601, secondary is unixtime' do
+      i = DummyForTimeParser.new
+      i.configure(config_element('parse', '',
+                                 {'time_type' => 'mixed',
+                                  'time_format' => '%iso8601',
+                                  'time_format_fallbacks' => ['unixtime']}))
+      parser = i.time_parser_create
+      time = event_time('2021-01-01T12:00:00+0900')
+      assert_equal_event_time(time, parser.parse("#{time.sec}"))
+    end
```
ashie, line 276: It seems that there is no test for no fallback case so that whether `time_format` is correctly used or not isn't confirmed by this test set.

ashie (changes requested):

test/test_time_parser.rb
kenhys: I'll add more test case to cover.

kenhys: I've added more test cases.

lib/fluent/time.rb
kenhys: Removed

lib/fluent/time.rb
kenhys: Removed

ashie (approved):

Issue #3224, opened 2021-01-08
Log forwarding stopped after "No such file or directory" error in in_tail


yvespp:
**Describe the bug**

Fluentd stops forwarding logs after error: 
```
Unexpected error raised. Stopping the timer. title=:in_tail_refresh_watchers error_class=Errno::ENOENT error="No such file or directory @ rb_file_s_stat - 
```
This error only startted happening when I updated to fluentd 1.12.0 and set `follow_inodes true` in the tail plugin.
Sometime there is also this warn:
```
#0 Unparsable line in pos_file:
```

**To Reproduce**
Happens randomly in our environment, happens more often on servers with more log volume.

**Expected behavior**
No error should happen. If an error happens the log should still be forwarded.

**Your Environment**

- Fluentd or td-agent version: fluentd 1.12.0
- Operating system: we use the docker image fluent/fluentd:v1.12.0-1.0 on Ubuntu 20.04.1 LTS
- Kernel version: 5.4.0-54-generic
- containerd: v1.4.3
- Kubernetes: v1.19.5

**Your Configuration**
```
<source>
  @type tail
  path /var/log/pods/*/*/*.log
  pos_file /fluentd/log/fluent_output.log.pos
  pos_file_compaction_interval 1h
  follow_inodes true
  read_from_head true
  refresh_interval 2
  tag except.kubernetes.*
  path_key tailed_path
  <parse>
    @type regexp
    expression "/^(?<time>.+) (stdout|stderr) [FP] (?<log>.+)?$/."
    time_format %Y-%m-%dT%H:%M:%S.%NZ
  </parse>
</source>

# drop empty log lines
<filter except.kubernetes.**.log>
  @type grep
  <exclude>
    key log
    pattern /^$/
  </exclude>
</filter>

# get kubernetes pod infos from log path
<filter except.kubernetes.**.log>
  @type parser
  key_name tailed_path
  reserve_time true
  reserve_data true
  remove_key_name_field true
  hash_value_field kubernetes
  <parse>
    @type regexp
    expression /^\/var\/log\/pods\/(?<namespace_name>.+)_(?<pod_name>.+)_.*\/(?<container_name>.*)\/.*\.log$/
  </parse>
</filter>

<filter except.kubernetes.**.log>
  @type record_transformer
  enable_ruby true
  <record>
    # Add cluster name
    # Line breaks may be trimmed when collecting from files. Add them back so that
    # multi line logs are still in multi line after combined by detect_exceptions.
[...]

PR #3275, opened 2021-03-02
in_tail: expect ENOENT during stat (try to fix #3274 and #3224)

### LINKED ISSUE #3274: 1.12.0: tail plugin crashes worker upon ENOENT (cannot stat() file)
**Describe the bug**

The `in_tail` plugin might discover a new log file, and then crash the worker upon the `stat()` system call due to ENOENT:

```
[info]: gem 'fluentd' version '1.12.0'

...

[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/root_agent.rb:200:in `block in start'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:234:in `start'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:353:in `refresh_watchers'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:390:in `start_watchers'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:390:in `each_value'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:409:in `block in start_watchers'
[error]: #0 /usr/local/bundle/gems/fluentd-1.12.0/lib/fluent/plugin/in_tail.rb:409:in `stat'
[error]: #0 unexpected error error_class=Errno::ENOENT error="No such file or directory @ rb_file_s_stat - /var/log/containers/foo-55fc795779-l66hb_foo_foo-ee7d580da4f6fac071b1e0fa8533fcada239a88404ed2fcde16480fbcd4a39fb.log"
```

I think this is a classical race condition. As this is part of the inotify-based watch system, of course there's the chance that between being delivered the signal that a new file is there and the code accessing it the file might already have disappeared again.

**To Reproduce**
Run in an environment that has log files popping up and quickly disappear thereafter (I guess, no clear repro on our side).

**Expected behavior**

At the system call boundary between application and OS the individual system calls like `stat()` / `open()` should always be wrapped with local error handling, expecting the system call to fail.

This can then be retried in graceful degradation fashion, w/o affecting the other operations performed by the worker process.

An appropriate action would be to log a warning/error showing how the system call failed.

**Your Environment**

Fluentd 1.12.0 on Linux

Not retried with 1.12.1 -- same problem seems to be there, based on commits and changelog.

**Your Configuration**

```...```

**Your Error Log**

see above

**Additional context**

- This could be related to https://github.com/fluent/fluentd/issues/3224.
- This might be a regression introduced via https://github.com/fluent/fluentd/pull/3155 / https://github.com/fluent/fluentd/pull/2992

The `stat()` call that crashes the worker is https://github.com/fluent/fluentd/blame/65a9edf4e05cf64c7ed0de56e12b3c9a0774e0d6/lib/fluent/plugin/in_tail.rb#L409

Looks like even before the last refactor this stat() wasn't protected: 
https://github.com/fluent/fluentd/pull/3196/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802R409

The worker subsequently restarts. As this affects the super early worker startup phase and results in an immediate crash, potentially a crash loop, I am not sure if this might result in data loss. https://github.com/fluent/fluentd/issues/3224 has more detail about that.

jgehrcke:

This is a minimal attempt to fix #3274. While I was here I was also trying to address #3224.

**Docs Changes**:

**Release Note**: 

I'd appreciate quick review first, happy to add changelog/release note entry later.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -313,11 +313,17 @@ module Fluent::Plugin
       (paths - excluded).select { |path|
         FileTest.exist?(path)
       }.each { |path|
-        target_info = TargetInfo.new(path, Fluent::FileWrapper.stat(path).ino)
-        if @follow_inodes
-          hash[target_info.ino] = target_info
-        else
-          hash[target_info.path] = target_info
+        # Even we just checked for existence, there is a race condition here as
+        # of which stat() might fail with ENOENT. See #3224.
+        begin
+          target_info = TargetInfo.new(path, Fluent::FileWrapper.stat(path).ino)
+          if @follow_inodes
+            hash[target_info.ino] = target_info
+          else
+            hash[target_info.path] = target_info
+          end
+        rescue Errno::ENOENT
+          $log.warn "expand_paths: stat() for #{target_info.path} failed with ENOENT. Skip file."
         end
       }
       hash
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -418,7 +418,8 @@ module Fluent::Plugin
           @tails[target_info] = tw
         rescue Errno::ENOENT
           $log.warn "stat() for #{target_info.path} failed with ENOENT. Drop tail watcher for now."
-          # explicitly detach and close watcher `tw`, or is it being garbage-collected?
+          # explicitly detach and unwatch watcher `tw`.
+          stop_watchers(target_info, immediate: true, unwatched: true)
         end
       }
     end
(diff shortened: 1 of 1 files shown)


cosmo0920:
Thanks for the patch.
I'd added an additional commit to detach and unwatch for broken target info entries.

ashie:
I've add 419ef5041a8dc59224f98783ab0e2c33a436e68b to fix an error on AppVeyor.

cosmo0920 (approved):

ashie (approved):

ashie:
Thanks!

PR #3281, opened 2021-03-05
Tweak GitHub Actions and Remove AppVeyor

ashie:
none

We need to tune CI because recently it always fail.
This PR is an intermediate result of tuning it:
* CI on Linux will be moved from TravisCI to GitHub Actions
  * .travis.yml isn't removed yet because macOS isn't migrated yet.
  * ppc64le and s390x will be completely removed.
* Windows CI has been moved to GitHub Actions, AppVeyor will be removed
  * Dozens of in_tail tests are still failed, but it's same with AppVeyor. I'm now investigating it.
* Disable some failed tests on Ruby 3.0
  * We need to fix RR to re-enable them
* We'll continue to use Drone for Arm architecture.

**Docs Changes**:
none

**Release Note**: 
none


.github/workflows/linux-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby-version: ['3.0', '2.7', '2.6', '2.5']
+        os: [ubuntu-latest]
+        experimental: [false]
+        include:
+          - ruby-version: head
+            os: ubuntu-latest
+            experimental: true
```
cosmo0920, line 22: :+1:
I didn't know how to setup ruby-head and allow-failure like behavior.

cosmo0920 (approved):
LGTM.
Yeah, key words arguments failures on Ruby 3.0 should be fixed later.

PR #3283, opened 2021-03-05
test in_tail: use safer method to cleanup

kenhys:

N/A

This PR aims to fix in_tail test cases on Windows.

On Windows, when the file or directory is removed and created
frequently, there is a case that creating file or directory will fail.
This situation is caused by pending file or directory deletion which
is mentioned on win32 API document.

ref. https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#files

As a workaround, execute rename and remove method.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -6,6 +6,7 @@ require 'fluent/system_config'
 require 'net/http'
 require 'flexmock/test_unit'
 require 'timecop'
+require 'securerandom'
 
 class TailInputTest < Test::Unit::TestCase
   include FlexMock::TestCase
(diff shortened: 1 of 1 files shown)


test/plugin/test_in_tail.rb
```diff
@@ -22,28 +24,42 @@ def teardown
...
   def cleanup_directory(path)
+    unless Dir.exist?(path)
+      FileUtils.mkdir_p(path)
+      return
+    end
+
+    dir = File.join(Dir.tmpdir, SecureRandom.hex(10))
     begin
-      FileUtils.rm_f(path, secure: true)
+      files = Dir.glob("*", base: path)
```
ashie, line 34: Could you add a comment as similar as the commit message?

cosmo0920:
@kenhys How long does it take to be ready to review?

ashie (changes requested):

ashie:
Finally we got success status on Windows! I think it's worth to merge soon.
If you are considering add more fixes, you can open new pull request.

kenhys:
> Finally we got success status on Windows! I think it's worth to merge soon.

No. The root of failure is not resolved...

ashie:
> No. The root of failure is not resolved...

Hmm, I see.
We shouldn't merge it to make it clear.
Is the root cause that file deletion is pending?
We should resolve the root cause of such pending, shouldn't we?

test/plugin/test_in_tail.rb
```diff
@@ -22,28 +24,42 @@ def teardown
...
+      if files.empty?
+        FileUtils.mv(path, dir)
+      else
+        files.each do |name|
+          FileUtils.rm_rf(File.join(path, name))
+        end
```
kenhys, line 40: `path` should be renamed to `dir` here,
But it raises access denied exception.

kenhys:
1. TMP_DIR is empty, safer cleanup works as expected,
2. TMP_DIR is not empty, safer cleanup logic does not work as expected.

1. is ok, as GitHub CI indicates as well
2. is not OK. this is the blocker.

kenhys:
I've changed to use `if Fluent.windows?` block to reduce in_tail testing time overhead.

kenhys:
> I've changed to use if Fluent.windows? block to reduce in_tail testing time overhead.

Oops, It introduces unexpected CI degration, I'll fix it.

test/plugin/test_in_tail.rb
kenhys: Fixed.

kenhys:
waiting CI result.

kenhys:
Ruby 3.0(windows)

4 failure is related to test_server.rb

Ruby 2.5 (windows)

```
2021-03-09T03:24:32.4985691Z Error: test: a node supporting responses after stop(ForwardOutputTest): Fluent::Test::Driver::TestTimedOut: Test case timed out with hard limit.
2021-03-09T03:24:32.4989241Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:201:in `rescue in run_actual'
2021-03-09T03:24:32.4991540Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:196:in `run_actual'
2021-03-09T03:24:32.4993666Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:95:in `run'
2021-03-09T03:24:32.4995739Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base_owner.rb:130:in `run'
2021-03-09T03:24:32.5000793Z /home/runner/work/fluentd/fluentd/test/plugin/test_out_forward.rb:651:in `block in <class:ForwardOutputTest>'
```

This error should be fixed in another PR.

test/plugin/test_in_tail.rb
kenhys: Changed to use cleanup_file

test/plugin/test_in_tail.rb
```diff
@@ -22,28 +23,64 @@ def teardown
...
+          cleanup_file(File.join(path, name))
+        rescue
+          # expect test driver block release already owned file handle.
+        end
+      end
+    else
+      begin
+        FileUtils.rm_f(path, secure:true)
+      rescue ArgumentError
+        FileUtils.rm_f(path)
```
ashie, line 43: Could you revive the comment `# For Ruby 2.6 or before.` which exists in the original code?

ashie (changes requested):

test/plugin/test_in_tail.rb
kenhys: Fixed.

kenhys:
I've fixed and rebased against master.

ashie (changes requested):
Please follow DCO https://github.com/fluent/fluentd/pull/3283/checks?check_run_id=2063272446

cosmo0920 (approved):
LGTM.
I'd confirmed that this PR makes to be more stable CI on Windows.

kenhys:
As https://github.com/fluent/fluentd/pull/3289 was merged, then rebased again.

cosmo0920:
Ruby 2.5, 2.6, and 2.7 jobs on Windows get green! :muscle:

ashie (approved):

Issue #3130, opened 2020-09-10
 failed to flush the buffer. retry_time=0 next_retry_seconds=2020-09-10 09:32:24 +0000 chunk="5aef23b4b29e2cd5ef6cec983d594292" error_class=Errno::ECONNRESET error="Connection reset by peer"


upengs:
![image](https://user-images.githubusercontent.com/8267975/92712155-deab6680-f38b-11ea-9d3e-f3f677124072.png)

PR #3287, opened 2021-03-08
Revise feat ignore_if_prev_successes to copy plugin

cosmo0920:
Revised #3190

In this revised version, I added a testcase fix for https://github.com/fluent/fluentd/pull/3190#discussion_r538052270.

According to #3130:

> This PR makes a change in the Copy Output plugin so you can inform the fluentd to ignore said copy if the previous output succeeded.

> This enables having multiple nested outputs as backup of each other by doing:

> ```aconf
>    <store ignore_error>
>      @type test
>      name c0
>    </store>
>    <store ignore_if_prev_success ignore_error>
>      @type test
>      name c1
>    </store>
>    <store ignore_if_prev_success>
>      @type test
>      name c2
>    </store>
> ```

**Docs Changes**:

> Adding the defined ignore_if_prev_successes attribute to the tag makes the copy plugin ignore said store if the previous store succeed in pushing the output.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/out_copy.rb
+++ b/lib/fluent/plugin/out_copy.rb
@@ -46,6 +46,9 @@ module Fluent::Plugin
         @ignore_errors << (store.arg.include?('ignore_error'))
         @ignore_if_prev_successes << (store.arg.include?('ignore_if_prev_success'))
       }
+      if @ignore_errors.uniq.size == 1 && @ignore_errors.include?(true) && @ignore_if_prev_successes.include?(false)
+        log.warn "ignore_errors are specified in all <store>, but ignore_if_prev_success is not specified. Is this intended?"
+      end
     end
 
     def multi_workers_ready?
--- a/lib/fluent/plugin/out_copy.rb
+++ b/lib/fluent/plugin/out_copy.rb
@@ -27,11 +27,12 @@ module Fluent::Plugin
     desc 'Pass different record to each `store` plugin by specified method'
     config_param :copy_mode, :enum, list: [:no_copy, :shallow, :deep, :marshal], default: :no_copy
 
-    attr_reader :ignore_errors
+    attr_reader :ignore_errors, :ignore_if_prev_successes
 
     def initialize
       super
       @ignore_errors = []
+      @ignore_if_prev_successes = []
     end
 
     def configure(conf)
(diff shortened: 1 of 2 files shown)


test/plugin/test_out_copy.rb
```diff
@@ -217,5 +220,47 @@ def test_ignore_error
...
+
+    # override to raise an error
+    d.instance.outputs[0].define_singleton_method(:process) do |tag, es|
+      raise ArgumentError, 'Failed'
+    end
+
+    # check ingore_if_prev_success functionality:
+    # 1. output 2 is succeeded.
+    # 2. output 3 is not cailed.
+    flexstub(d.instance.outputs[1]) do |output|
```
kenhys, line 250: Typo 'cailed'

lib/fluent/plugin/out_copy.rb
```diff
@@ -55,10 +57,15 @@ def process(tag, es)
         }
         es = m
       end
-
+      success = Array.new(outputs.size)
       outputs.each_with_index do |output, i|
         begin
-          output.emit_events(tag, @copy_proc ? @copy_proc.call(es) : es)
+          if i > 0 && success[i - 1] && @ignore_if_prev_successes[i]
```
kenhys, line 63: If ignore_if_prev_success is specified in first `<store>` section, 
it may be better to raise an error.

kenhys: Then, `ignore_error` is specified in all `store` without `ignore_if_prev_success` it may be better to warn.

kenhys (changes requested):
There are some requests to make it user-friendly.

test/plugin/test_out_copy.rb
cosmo0920: :scream:

lib/fluent/plugin/out_copy.rb
cosmo0920: > If ignore_if_prev_success is specified in first `<store>` section,
> it may be better to raise an error.

This should be fixed at https://github.com/fluent/fluentd/pull/3287/commits/2f5a6e1450b3e3dbbb6d3e856e76dfb4224118b5.

cosmo0920: > Then, `ignore_error` is specified in all `store` without `ignore_if_prev_success` it may be better to warn.

This also should be fixed at https://github.com/fluent/fluentd/pull/3287/commits/bdbaef288c73868d414becfbbb47bf663265af07.

kenhys (approved):
LGTM

PR #3288, opened 2021-03-08
formatter_ltsv: suppress delimiters in output

ashie:
Revise #1666

formatter_ltsv creates malformed LTSV files when values include tabs.
This PR prevent it by replacing tabs with spaces.

**Docs Changes**:
`replacement` should be added to https://docs.fluentd.org/parser/ltsv

**Release Note**: 
Same with the title


cosmo0920 (approved):

ashie:
Thanks @MikePlayle!

PR #3291, opened 2021-03-12
Use more secure methods

cosmo0920:
None

DeepSource complains that fluentd uses insecure methods:

* `Kernel.open`: https://deepsource.io/gh/cosmo0920/fluentd/issue/RB-SC1004/
* `JSON.load`: https://deepsource.io/gh/cosmo0920/fluentd/issue/RB-SC1002/occurrences

**Docs Changes**:
No need.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -186,7 +186,7 @@ module Fluent
       return nil if val.nil?
 
       param = if val.is_a?(String)
-                val.start_with?('{') ? JSON.load(val) : Hash[val.strip.split(/\s*,\s*/).map{|v| v.split(':', 2)}]
+                val.start_with?('{') ? JSON.parse(val) : Hash[val.strip.split(/\s*,\s*/).map{|v| v.split(':', 2)}]
               else
                 val
               end
@@ -213,7 +213,7 @@ module Fluent
       return nil if val.nil?
 
       param = if val.is_a?(String)
-                val.start_with?('[') ? JSON.load(val) : val.strip.split(/\s*,\s*/)
+                val.start_with?('[') ? JSON.parse(val) : val.strip.split(/\s*,\s*/)
               else
                 val
               end
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -87,7 +87,7 @@ module Fluent
           if File.exist?(@path)
             raise Fluent::ConfigError, "Plugin storage path '#{@path}' is not readable/writable" unless File.readable?(@path) && File.writable?(@path)
             begin
-              data = open(@path, 'r:utf-8') { |io| io.read }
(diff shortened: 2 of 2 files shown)


lib/fluent/config/types.rb
```diff
@@ -186,7 +186,7 @@ def self.hash_value(val, opts = {}, name = nil)
       return nil if val.nil?
 
       param = if val.is_a?(String)
-                val.start_with?('{') ? JSON.load(val) : Hash[val.strip.split(/\s*,\s*/).map{|v| v.split(':', 2)}]
+                val.start_with?('{') ? JSON.parse(val) : Hash[val.strip.split(/\s*,\s*/).map{|v| v.split(':', 2)}]
```
ashie, line 189: Memo:
`JSON.load` wraps `JSON.parse` and it modifies some default options of `JSON::Parser`:
https://github.com/flori/json/blob/7b452b290502a5cca8fc6403f31275c83e0e3d48/lib/json/common.rb#L569
https://github.com/flori/json/blob/7b452b290502a5cca8fc6403f31275c83e0e3d48/lib/json/common.rb#L422
https://docs.ruby-lang.org/ja/latest/class/JSON=3a=3aParser.html
So that the behavior is also changed a bit but I believe it's no problem on this case.

ashie (approved):

kenhys (commented):
Could you add a reference to why it should be fixed in the commit message?
Then LGTM.

lib/fluent/config/types.rb
cosmo0920: Thanks for the clarification. Yeah, in this case, using `JSON.parse` instead of `JSON.load` does not cause issues.

kenhys:
I doubt whether https://deepsource.io/gh/cosmo0920/fluentd/issue/RB-SC1002 is consistent or not.

How about quoting the following:

* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Security/Open

```
`Kernel#open` and `URI.open` enable not only file access but also process invocation by prefixing a pipe symbol (e.g., `open(“| ls”)`). So, it may lead to a serious security risk by using variable input to the argument of `Kernel#open` and `URI.open`. It would be better to use `File.open`, `IO.popen` or `URI.parse#open` explicitly.
```


* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Security/JSONLoad

```
Autocorrect is disabled by default because it's potentially dangerous. If using a stream, like `JSON.load(open('file'))`, it will need to call `#read` manually, like `JSON.parse(open('file').read)`. If reading single values (rather than proper JSON objects), like `JSON.load('false')`, it will need to pass the `quirks_mode: true` option, like `JSON.parse('false', quirks_mode: true)`. Other similar issues may apply.
```

kenhys (approved):
LGTM

Issue #3292, opened 2021-03-15
in_tail may not send rotated logs when mv is used to rotate


aYukiSekiguchi:
**Describe the bug**
After upgrading fluentd from 0.14.24 to 1.12.1, tail plugin + my script which rotates the watched log using `mv` sometimes won't watch the rotated log file and send the first part of rotated logs.

**To Reproduce**

Watch `/tmp/script.log`
```
<source>
  @type tail
  path /tmp/script.log
  tag my-tag
  pos_file /var/log/td-agent/script.pos
  <parse>
    @type json
    time_key time
  </parse>
</source>
```

`mv` `/tmp/script.log` to `/tmp/script.log.20210314` and write to `/tmp/script.log`

**Expected behavior**
The content in `/tmp/script.log` should be sent.

**Your Environment**

- Fluentd or td-agent version: td-agent 4.1.0 fluentd 1.12.1 (e3effa337593618cbd7f0f4ef071766df1ec69a0)
- Operating system: Amazon Linux 2
- Kernel version: 4.14.192-147.314.amzn2.x86_64

**Your Configuration**

```
<source>
  @type tail
  path /tmp/script.log
  tag my-tag
  pos_file /var/log/td-agent/script.pos
  <parse>
    @type json
    time_key time
  </parse>
</source>
```

**Your Error Log**

none (No rotate detection log)

**Additional context**

I'm trying to upgrade fluentd from 0.14.24 to 1.12.1 and faced two problems.

My script is called by cron every day. At the start of the script, it `mv`s log to rotate, do something great and dump logs to a new file.
```ruby
FILE_NAME = '/tmp/script.log'

def rotate_file_if_exist
  if File.exists?(FILE_NAME)
    new_file_name = "#{FILE_NAME}.#{Time.now.strftime('%Y%m%d')}"
    FileUtils.mv(FILE_NAME, new_file_name)
  end
end
```

fluentd is watching the file name. My config is:
```
<source>
  @type tail
  path /tmp/script.log
  tag my-tag
  pos_file /var/log/td-agent/script.pos
  <parse>
    @type json
    time_key time
  </parse>
</source>
```

Here are problems I faced:
1. TailWatcher sometimes won't restart
1. TailWatcher may not send the first part of rotated logs

I debugged them using `mv` and `cp` command.

# TailWatcher won't restart

When I run `mv`, the file of the filename becomes none. Then, `TailWatcher#@rotate_handler` becomes `nil`. This stops to watch the inode.

[...]

PR #3294, opened 2021-03-15
in_tail: Handle to send rotated logs when mv is used for rotating

cosmo0920:

Related to #3292

If a file is mv-ed and a new file is created during calling `#refresh_watchers`s, and `#refresh_watchers`
won't run `#start_watchers` and `#stop_watchers` for the path because `#target_paths_hash` always contains the path.

**Docs Changes**:
No need.

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -780,16 +780,22 @@ module Fluent::Plugin
           end
 
           if watcher_needs_update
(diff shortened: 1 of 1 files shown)


cosmo0920:
I'm trying to write testcase for this edge case, but writing testcase for this case is unstable test....

ashie (approved):
We'll try to write a test for it later.

PR #3305, opened 2021-03-31
Follow depended gem version by fluent-plugin-generate

kenhys:

Fixes N/A

In the previous versions, generated .gemspec doesn't follow
depended gem version which is used in fluentd.gemspec at all.

Instead, it is changed to retrieve depended gem version from
Gemfile.lock to sync with it.

Before:

```
  spec.add_development_dependency "bundler", "~> 1.14"
  spec.add_development_dependency "rake", "~> 12.0"
  spec.add_development_dependency "test-unit", "~> 3.0"
```

It is required to update manually.

After:

```
  spec.add_development_dependency "bundler", "~> 2.2.15"
  spec.add_development_dependency "rake", "~> 13.0"
  spec.add_development_dependency "test-unit", "~> 3.3"
```

Above depended version is replaced by <%= %> in template.
No need to update manually.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/command/plugin_generator.rb
+++ b/lib/fluent/command/plugin_generator.rb
@@ -151,6 +151,36 @@ BANNER
     underscore_name
   end
 
+  def gem_file_path
+    File.expand_path(File.join(File.dirname(__FILE__),
+                               "../../../",
+                               "Gemfile"))
+  end
+
+  def lock_file_path
+    File.expand_path(File.join(File.dirname(__FILE__),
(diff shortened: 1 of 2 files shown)


cosmo0920 (approved):

cosmo0920:
off topic: fluent-plugin-generate command does not support `storage` and `service_discovery` plugin types.
We should also support them.

ashie (approved):

Issue #3263, opened 2021-02-19
Some unit tests are broken on Ruby 3.0


kenhys:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**

Drone CI is continuously broken for two months.
See https://cloud.drone.io/fluent/fluentd

**To Reproduce**

Follow steps in .drone.yml

See https://github.com/fluent/fluentd/blob/master/.drone.yml#L26-L35

**Expected behavior**

There are 3 test case failures.

```
===============================================================================
Failure: test: it does not create accessor methods for parameters(Fluent::Config::TestConfigurable:::skip_accessor option)
/drone/src/test/config/test_configurable.rb:1456:in `block (2 levels) in <class:TestConfigurable>'
     1453:         @example = ConfigurableSpec::ExampleWithSkipAccessor.new
     1454:         @example.configure(config_element('ROOT'))
     1455:         assert_equal 'example7', @example.instance_variable_get(:@name)
  => 1456:         assert_raise NoMethodError.new("undefined method `name' for #{@example}") do
     1457:           @example.name
     1458:         end
     1459:       end
```

```
===============================================================================
Failure: test: when out_forward has @id(ForwardOutputTest):
  Exception raised:
  ArgumentError(<wrong number of arguments (given 2, expected 1)>)
  /drone/src/lib/fluent/plugin.rb:120:in `new_sd'
  /drone/src/vendor/bundle/ruby/3.0.0/gems/rr-1.2.1/lib/rr/method_dispatches/method_dispatch.rb:30:in `call_original_method'
  /drone/src/vendor/bundle/ruby/3.0.0/gems/rr-1.2.1/lib/rr/method_dispatches/method_dispatch.rb:41:in `call_implementation'
  /drone/src/vendor/bundle/ruby/3.0.0/gems/rr-1.2.1/lib/rr/method_dispatches/method_dispatch.rb:16:in `call'
  /drone/src/vendor/bundle/ruby/3.0.0/gems/rr-1.2.1/lib/rr/injections/double_injection.rb:183:in `dispatch_method'
  /drone/src/vendor/bundle/ruby/3.0.0/gems/rr-1.2.1/lib/rr/injections/double_injection.rb:40:in `dispatch_method'
[...]

PR #3306, opened 2021-03-31
test: omit test case because exception is not raised with Ruby 3.0

kenhys:

Related: #3263

There is a case that expected exception is not raised with Ruby 3.0.

Because of that behavior, exception is assumed test case always fails.

It should be fixed in Fiddle so, when fixed version is
released, we should enable this test case again for Ruby 3.0.

ref. https://github.com/ruby/fiddle/issues/72

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/test/plugin_helper/test_server.rb
+++ b/test/plugin_helper/test_server.rb
@@ -313,6 +313,7 @@ class ServerPluginHelperTest < Test::Unit::TestCase
       # 'server_create_connection unix' => [:server_create, :unix, {}],
     )
     test 'cannot create 2 or more servers using same bind address and port if shared option is false' do |(m, proto, kwargs)|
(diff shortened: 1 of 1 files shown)


ashie:
I'd like to leave the test as is, because it's not only the test's issue but also occur on actual deployment.
This issue should be clarified by this test.
Instead of omitting the test, I think marking Windows on Ruby 3.0 as "experimental" [like macOS](https://github.com/fluent/fluentd/blob/master/.github/workflows/macos-test.yaml#L18) is better.

kenhys:
> Instead of omitting the test, I think marking the workflow on Windows + Ruby 3.0 as "experimental" (like macOS) is better.

It is better to do so, I'll close this PR.

PR #3307, opened 2021-03-31
test: mark Ruby 3.0 as experimental on windows

kenhys:

Related #3263

There are some failed unit test with Ruby 3.0 on windows,
Instead of omitting those test cases, it may be better
to mark "experimental" to clarify the issues.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/.github/workflows/windows-test.yaml
+++ b/.github/workflows/windows-test.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-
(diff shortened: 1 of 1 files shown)


ashie (changes requested):
`continu-on-error` is also needed: https://github.com/fluent/fluentd/blob/d19b7061ab003b9c469976d6137b39d461a65d4b/.github/workflows/macos-test.yaml#L18

kenhys:
Thanks,

ashie (approved):

PR #3308, opened 2021-03-31
plugin_helper: Allow TLS to use keep-alive socket option

jiping-s:

Allow incoming TLS connections to use keep-alive socket option when a listener is created, such as the`send_keepalive_packet` option in `forward` input. It's currently blocked by `socket_option_validate!`

We've been using this on several busy production servers for 2 months and haven't experienced any problem.

`send_keepalive_packet` is critical in some network setup, for instance in ours where AWS sometimes turned connections from private network into zombies - they linger and keep accumulating until fluentd is restarted, while according to clients they should have been closed normally days or weeks earlier. (I suspect there might be some bug in async socket closing but  couldn't confirm or investigate)

Signed-off-by: Ji-Ping Shen <ji-ping.shen@relexsolutions.com>

### DIFF
--- a/lib/fluent/plugin_helper/socket_option.rb
+++ b/lib/fluent/plugin_helper/socket_option.rb
@@ -38,8 +38,8 @@ module Fluent
           end
         end
         if send_keepalive_packet
-          if protocol != :tcp
-            raise ArgumentError, "BUG: send_keepalive_packet is available for tcp"
+          if protocol != :tcp && protocol != :tls
+            raise ArgumentError, "BUG: send_keepalive_packet is available for tcp/tls"
           end
         end
       end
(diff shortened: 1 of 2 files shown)


cosmo0920 (approved):
It sounds reasonable change for me. :punch:

ashie:
I'm checking the commit which introduces this option: a79c6d2f7246a9fa277c7fdfd5aa8657ed8f1557
It seems that some tests are needed to update.
I'll do it.

kenhys (approved):
> It seems that some tests are needed to update.

:+1: 

The concept is reasonable,
if the test cases have been updated && passed.

ashie (approved):

ashie:
Thanks!

PR #3316, opened 2021-04-06
test: add tiny wait not to lose globbed entries

kenhys:

Related #3263

In busy loop with wait_flush, there may be a case that Dir.glob() lose
existing entries under target directory in some test cases.

As a result Timeout::Error: execution expired will be raised.

These failures in test/plugin/test_output_as_buffered_backup
are observed with Ruby 3.0 on Windows.

**Docs Changes**:

N/A

**Release Note**: 

N/A


kenhys:
```
Timeout::Error: execution expired
<internal:dir>:135:in `glob'
./test/plugin/test_output_as_buffered_backup.rb:175:in `block in wait_flush'
./test/plugin/test_output_as_buffered_backup.rb:112:in `block in waiting'
```

It seems that the above error doesn't occur.

ashie (approved):

PR #3325, opened 2021-04-12
WindowsFile: Fix incorrect error message

ashie:
none

Fluent::WindowsFile passes win32 error codes to SystemCallError so that it doesn't show
correct message for them. For example when ERROR_SHARING_VIOLATION(32)
of win32 occurs, it shows Errno::EPIPE(32).

**Docs Changes**:
none

**Release Note**: 
same with the title

### DIFF
--- a/lib/fluent/plugin/file_wrapper.rb
+++ b/lib/fluent/plugin/file_wrapper.rb
@@ -107,7 +107,7 @@ module Fluent
       if @file_handle == INVALID_HANDLE_VALUE
         err = GetLastError.call
         if err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND || err == ERROR_ACCESS_DENIED
-          raise SystemCallError.new(2)
+          raise Errno::ENOENT
         end
         raise Win32Error.new(err)
       end
--- a/lib/fluent/plugin/file_wrapper.rb
+++ b/lib/fluent/plugin/file_wrapper.rb
@@ -46,6 +46,34 @@ module Fluent
     end
   end
 
+  class Win32Error < StandardError
+    require 'windows/error'
+    include Windows::Error
+
+    attr_reader :errcode
+
+    def initialize(errcode)
+      @errcode = errcode
+    end
+
+    def format_english_message(errcode)
+      buf = 0.chr * 260
+      flags = FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/file_wrapper.rb
```diff
@@ -46,6 +46,33 @@ def s.ino; @ino; end
...
+    attr_reader :errcode
+
+    def initialize(errcode)
+      @errcode = errcode
+    end
+
+    def format_english_message(errcode)
+      buf = 0.chr * 260
+      flags = FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY
+      english_lang_id = 1033 # The result of MAKELANDID(LANG_ENGLISH, SUBLANG_ENGLISH_US)
```
kenhys, line 62: Should it be `MAKELANGID` ?

lib/fluent/plugin/file_wrapper.rb
```diff
@@ -46,6 +46,33 @@ def s.ino; @ino; end
...
+    include Windows::Error
+
+    attr_reader :errcode
+
+    def initialize(errcode)
+      @errcode = errcode
+    end
+
+    def format_english_message(errcode)
+      buf = 0.chr * 260
```
kenhys, line 60: Can we use `MAX_PATH` constant here?

test/plugin/test_file_wrapper.rb
```diff
@@ -0,0 +1,77 @@
...
+      ensure
+        file1.close if file1
+      end
+    end
+
+    test 'Errno::ENOENT raised' do
+      path = "#{TMP_DIR}/nofile.txt"
+      file = nil
+      assert_raise(SystemCallError.new(2)) do
+        file = Fluent::WindowsFile.new(path)
```
kenhys, line 56: Can we avoid using the numeric value e.g. 2 ? (meaningful constant is better)

kenhys (commented):
We need cosmetic changes only.
Except that, it looks good.

lib/fluent/plugin/file_wrapper.rb
ashie: Can't use because it's a macro in C.

lib/fluent/plugin/file_wrapper.rb
ashie: Fixed.

test/plugin/test_file_wrapper.rb
ashie: Fixed.

kenhys (approved):
LGTM

PR #3329, opened 2021-04-13
WindowsFile: Show the path when Win32Error is occured

ashie:
none

In #3325 we added Win32Error to show Windows specific errors correctly,
but it's not enough for debugging because it doesn't show which file
causes the error.

**Docs Changes**:
none

**Release Note**: 
none, same with #3325


cosmo0920 (approved):
Looks attractive for me.

kenhys (approved):
LGTM.

ashie:
Hmm, it's still not enough. `to_s` is missing so that it shows only "Fluent::Win32Error".

PR #3331, opened 2021-04-14
Win32Error: Add to_s and inspect methods

ashie:
none

Additional fix for #3325 and #3329.
Since the previous code doesn't have to_s method, it shows only` Fluent::Win32Error`, doesn't show the detail.

**Docs Changes**:
none

**Release Note**: 
Same with #3325

### DIFF
--- a/lib/fluent/plugin/file_wrapper.rb
+++ b/lib/fluent/plugin/file_wrapper.rb
@@ -65,10 +65,15 @@ module Fluent
       buf.force_encoding(Encoding.default_external).strip
     end
 
-    def message
-      msg = "code: #{@errcode}, #{format_english_message(@errcode)}"
-      msg << ": #{@msg}" if @msg
+    def to_s
+      msg = super
+      msg << ": code: #{@errcode}, #{format_english_message(@errcode)}"
+      msg << " - #{@msg}" if @msg
       msg
+    end
+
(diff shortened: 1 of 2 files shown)


kenhys (approved):
I'm not sure #to_s and #message  should be implemented separately.
Anyway this change LGTM.

ashie:
> I'm not sure #to_s and #message should be implemented separately.

Ah sorry, I already removed #message locally but forgot to push it.

cosmo0920 (approved):
Looks reasonable for me. :+1:

ashie:
Thanks.

BTW I found a weird behavior of td-agent 4.x on Windows by this patch.
Sometimes `WindowsFile` returns a weird return code on calling `CreateFile` against a non-existent file.
Sometimes `ERROR_FILE_NOT_FOUND (158)` or sometimes `ERROR_SHARING_VIOLATION (32)`.
Probably Ruby's internal code resets the error code.
We may have to add a method to get correct error code to win32-api, like [`Fiddle::win32_last_error`](https://rubyapi.org/3.0/o/fiddle#method-c-win32_last_error) or [`FFI::LastError.winapi_error`](https://www.rubydoc.info/github/ffi/ffi/FFI/LastError#winapi_error-class_method)

PR #3352, opened 2021-04-26
Enable system.log.rotate_... system configuration

kenhys:

Fixes N/A

In the previous versions, log rotation options were supported as
--log-rotate-age or --log-rotate-size via command line options.

On Windows, as fluentd is launched as a windows service, it is
required to configure again via --reg-winsvc-fluentdopt or edit
fluentdopt registry key for customization.

This approach is not convenient for Windows users, so it may be better
to support more comprehensive solution - customize via configuration
file.

This commit introduces such a configurable parameter in .conf

```
  <system>
   <log>
     rotate_age 5
     rotate_size 1048576
   </log>
  </system>
```

**Docs Changes**:

TODO: https://docs.fluentd.org/deployment/logging#by-config-file should be updated.

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -518,6 +518,8 @@ module Fluent
         dl_opts = {}
         # subtract 1 to match serverengine daemon logger side logging severity.
         dl_opts[:log_level] = @level - 1
+        dl_opts[:log_rotate_age] = @log_rotate_age if @log_rotate_age
+        dl_opts[:log_rotate_size] = @log_rotate_size if @log_rotate_size
         logger = ServerEngine::DaemonLogger.new(@logdev, dl_opts)
         $log = Fluent::Log.new(logger, @opts)
         $log.enable_color(false) if @path
@@ -606,6 +608,19 @@ module Fluent
 
       @cl_opt = opt
       @conf = nil
+      # parse configuration immediately to initialize logger in early stage
+      if @config_path and File.exist?(@config_path)
+        @conf = Fluent::Config.build(config_path: @config_path,
+                                     encoding: @conf_encoding ? @conf_encoding : 'utf-8',
+                                     additional_config: @inline_config ? @inline_config : nil,
+                                     use_v1_config: !!@use_v1_config)
+        @system_config = build_system_config(@conf)
+        if @system_config.log
+          @log_rotate_age ||= @system_config.log.rotate_age
+          @log_rotate_size ||= @system_config.log.rotate_size
+        end
+        @conf = nil
+      end
 
       log_opts = {suppress_repeated_stacktrace: opt[:suppress_repeated_stacktrace], ignore_repeated_log_interval: opt[:ignore_repeated_log_interval],
                   ignore_same_log_interval: opt[:ignore_same_log_interval]}
--- a/lib/fluent/system_config.rb
+++ b/lib/fluent/system_config.rb
@@ -55,6 +55,20 @@ module Fluent
     config_section :log, required: false, init: true, multi: false do
       config_param :format, :enum, list: [:text, :json], default: :text
       config_param :time_format, :string, default: '%Y-%m-%d %H:%M:%S %z'
+      config_param :rotate_age, default: nil do |v|
+        if %w(daily weekly monthly).include?(v)
+          v.to_sym
+        else
+          begin
+            Integer(v)
+          rescue ArgumentError => e
+            raise Fluent::ConfigError, e.message
+          else
+            v.to_i
+          end
+        end
+      end
+      config_param :rotate_size, :size, default: nil
     end
 
     config_section :counter_server, multi: false do
--- a/lib/fluent/system_config.rb
+++ b/lib/fluent/system_config.rb
@@ -56,7 +56,7 @@ module Fluent
       config_param :format, :enum, list: [:text, :json], default: :text
       config_param :time_format, :string, default: '%Y-%m-%d %H:%M:%S %z'
       config_param :rotate_age, default: nil do |v|
-        if %w(daily weekly monthly).include?(v)
+        if Fluent::Log::LOG_ROTATE_AGE.include?(v)
           v.to_sym
         else
           begin
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -160,7 +160,8 @@ class ConfigTest < Test::Unit::TestCase
     prepare_config
     opts = {
       :config_path => "#{TMP_DIR}/config_test_1.conf",
-      :inline_config => "<source>\n  type http\n  port 2222\n </source>"
+      :inline_config => "<source>\n  type http\n  port 2222\n </source>",
+      :use_v1_config => false
     }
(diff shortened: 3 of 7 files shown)


kenhys:
Is it not appropriate approach to parse in Fluent::Supervisor#initialize?
Above approach is aimed to pass `log_rotate_age`, and `log_rotate_size` to `LoggerInitializer.new`.
But `Fluent::SystemConfig` returns `@log Fluent::Config::Section` which fills rotate_age: null, rotate_size: null.

kenhys:
> @log Fluent::Config::Section which fills rotate_age: null, rotate_size: null.

because of it, Fluent::DaemonLogger is not initialized expectedly.

kenhys:
Hmm, it affects test_config test case.

```
.F                                                                    
============================================================================================================================================
Failure: test_include(ConfigTest)                                                                                                           
test/test_config.rb:77:in `test_include'                                                                                                    
     74:     assert_equal 'root_config', c['k1']
     75:     assert_equal 'relative_path_include', c['k2']                                                                                  
     76:     assert_equal 'relative_include_in_included_file', c['k3']                                                                      
  => 77:     assert_equal 'absolute_path_include', c['k4']                                                                                  
     78:     assert_equal 'uri_include', c['k5']                                                                                            
     79:     assert_equal 'wildcard_include_1', c['k6']
     80:     assert_equal 'wildcard_include_2', c['k7']
<"absolute_path_include"> expected but was
<nil>

diff:
? "absolute_path_include"
? ni                     
````

**Comment by kenhys** [comment]
In lib/fluent/supervisor.rb, the following code doesn't assign `<log>` section parameters appropriately.
(but format will be assigned :thinking: )

```
@conf = Fluent::Config.build(config_path: @config_path).elements.find { |e| e.name == 'system' }
@system_config = Fluent::SystemConfig.new(@conf)
```

**Comment by kenhys** [comment]
Even though master, test_include test_inline may fail.

**Comment by kenhys** [comment]
test_inline fails because it tries to handle the following .conf.

```
 k1 root_config
      include dir/config_test_2.conf  #
      include test/tmp/config/config_test_4.conf                                                                                            
      include file://test/tmp/config/config_test_5.conf                                                                                     
      <include config.d/*.conf />
                                                                                                                                            
<source>                                                              
  type http                                                                                                                                 
  port 2222                                                                                                                                 
 </source>
```

File: lib/fluent/supervisor.rb [hunk 1]
```diff
@@ -600,12 +602,35 @@ def initialize(opt)
...
       @signame = opt[:signame]
 
       @cl_opt = opt
       @conf = nil
+      # parse configuration immediately to initialize logger in early stage
+      if @config_path and File.exist?(@config_path)
+        @conf = Fluent::Config.build(config_path: @config_path,
+                                     encoding: @conf_encoding ? @conf_encoding : 'utf-8',
+                                     additional_config: @inline_config ? @inline_config : nil,
+                                     use_v1_config: !!@use_v1_config).elements.find { |e| e.name == 'system' }
```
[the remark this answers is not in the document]
kenhys: Yes, the following `Fluent::SystemConfog.new(@conf)` expects `system` section. thus without it, conf file can not be parsed expectedly.

ref. https://github.com/fluent/fluentd/blob/master/test/config/test_system_config.rb#L59-L62

lib/fluent/system_config.rb
```diff
@@ -55,6 +55,8 @@ class SystemConfig
     config_section :log, required: false, init: true, multi: false do
       config_param :format, :enum, list: [:text, :json], default: :text
       config_param :time_format, :string, default: '%Y-%m-%d %H:%M:%S %z'
+      config_param :rotate_age, :string, default: nil
```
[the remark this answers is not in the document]
kenhys: I've misunderstood that compound multiple types can't be allowed.
It should be enum here for consistency.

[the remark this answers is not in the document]
cosmo0920: Validating on latter do block is one approach.
ref: https://github.com/fluent/fluentd/blob/master/lib/fluent/config/configure_proxy.rb#L317

```ruby
      config_param :rotate_age, :string, default: nil do |v|
        # some `v` value validation.
      end
```

lib/fluent/system_config.rb
```diff
@@ -55,6 +55,8 @@ class SystemConfig
...
+      config_param :rotate_size, :integer, default: nil
```
cosmo0920, line 59: How about use `size` type instead of `integer`?
`size` type can handle `k`, `m`, `g` SI prefix.

lib/fluent/system_config.rb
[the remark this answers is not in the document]
kenhys: Using validation block may be a good approach, thanks!

lib/fluent/system_config.rb
kenhys: Using `size` seems reasonable, thanks!

lib/fluent/supervisor.rb
```diff
@@ -600,12 +602,35 @@ def initialize(opt)
...
+        @system_config = Fluent::SystemConfig.new(@conf)
+        if @system_config.log
+          if @system_config.log.rotate_age
+            if %w(daily meekly monthly).include?(@system_config.log.rotate_age)
```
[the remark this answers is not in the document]
kenhys: Changed to simplify the code.

test/test_config.rb
```diff
@@ -160,7 +160,8 @@ def test_inline
     prepare_config
     opts = {
       :config_path => "#{TMP_DIR}/config_test_1.conf",
-      :inline_config => "<source>\n  type http\n  port 2222\n </source>"
+      :inline_config => "<source>\n  type http\n  port 2222\n </source>",
+      :use_v1_config => false
```
ashie, line 164: Sorry, I can't understand why adding log rotation setting affects to this even I read the commit message.
Could you describe more detail?

kenhys: I've added more explanation in the commit message.

lib/fluent/system_config.rb
```diff
@@ -55,6 +55,20 @@ class SystemConfig
     config_section :log, required: false, init: true, multi: false do
       config_param :format, :enum, list: [:text, :json], default: :text
       config_param :time_format, :string, default: '%Y-%m-%d %H:%M:%S %z'
+      config_param :rotate_age, default: nil do |v|
+        if %w(daily weekly monthly).include?(v)
```
ashie, line 59: It seems better that this list is placed in lib/fluent/log.rb and `ROTATE_AGE` in lib/fluent/command/fluentd.rb refers same one.

ashie (changes requested):

lib/fluent/system_config.rb
kenhys: Ah, It should be,  Thanks!

kenhys: Fixed

ashie (approved):
LGTM

PR #3358, opened 2021-04-28
Add ability to get the root-router

ferbar:
This patch adds a way in event_emitter_router(label_name) to get the root event router. To use an empty string "" for the default router is just a proposal, maybe `@DEFAULT` would fit better.
Why I needed it:
I did not see another way to get an event it is once assigned to a label back the default route. The concat filter can send timedout events to a special label. With the _route_ plugin the tag can be prefixed and with `@label ""` the event can be sent to the default filters and output.

**Docs Changes**:
Will be updated if there is any interest.

**Release Note**: 
Enable to get root router by `Fluent::PluginHelper::EventEmitter::event_emitter_router("@ROOT")`

### DIFF
--- a/lib/fluent/plugin_helper/event_emitter.rb
+++ b/lib/fluent/plugin_helper/event_emitter.rb
@@ -47,7 +47,7 @@ module Fluent
 
       def event_emitter_router(label_name)
         if label_name
-          if label_name == ""
+          if label_name == "@ROOT"
             Engine.root_agent.event_router
           else
             Engine.root_agent.find_label(label_name).event_router
--- a/lib/fluent/plugin_helper/event_emitter.rb
+++ b/lib/fluent/plugin_helper/event_emitter.rb
@@ -47,7 +47,11 @@ module Fluent
 
       def event_emitter_router(label_name)
         if label_name
-          Engine.root_agent.find_label(label_name).event_router
+          if label_name == ""
+            Engine.root_agent.event_router
+          else
+            Engine.root_agent.find_label(label_name).event_router
+          end
         elsif self.respond_to?(:as_secondary) && self.as_secondary
           if @primary_instance.has_router?
             @_event_emitter_lazy_init = true
--- a/test/plugin_helper/test_event_emitter.rb
+++ b/test/plugin_helper/test_event_emitter.rb
@@ -56,8 +56,8 @@ class EventEmitterTest < Test::Unit::TestCase
   end
 
   test 'should have event_emitter_router' do
-    d0 = Dummy.new
-    assert d0.respond_to?(:event_emitter_router)
+    d = Dummy.new
+    assert d.respond_to?(:event_emitter_router)
   end
 
   test 'get router' do
--- a/test/plugin_helper/test_event_emitter.rb
+++ b/test/plugin_helper/test_event_emitter.rb
@@ -1,6 +1,7 @@
 require_relative '../helper'
 require 'fluent/plugin_helper/event_emitter'
 require 'fluent/plugin/base'
+require 'flexmock/test_unit'
 
 class EventEmitterTest < Test::Unit::TestCase
   setup do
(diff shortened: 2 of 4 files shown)


lib/fluent/plugin_helper/event_emitter.rb
```diff
@@ -47,7 +47,11 @@ def event_emitter_used_actually?
 
       def event_emitter_router(label_name)
         if label_name
-          Engine.root_agent.find_label(label_name).event_router
+          if label_name == ""
```
cosmo0920, line 50: > This patch adds a way in event_emitter_router(label_name) to get the root event router. To use an empty string "" for the default router is just a proposal, maybe @DEFAULT would fit better.

I guess `@DEFAULT` is better.
I don't want to use empty string: `""` as default router.
I don't like to implicit meaning literals.

cosmo0920 (changes requested):

lib/fluent/plugin_helper/event_emitter.rb
ashie: I'm not sure `@DEFAULT` is an appropriate name.
`@ROOT` might be more suitable because:

> get the root event router.

ashie (changes requested):
Please add tests for it.

ashie:
I'm not sure adding it and the label name are reasonable or not.
@repeatedly Do you have any opinions?

lib/fluent/plugin_helper/event_emitter.rb
cosmo0920: label_name cannot be empty string(`""`).
Due to: https://github.com/fluent/fluentd/blob/master/lib/fluent/root_agent.rb#L133

repeatedly:
> candidates of the label (Empty or @DEFAULT or @ROOT) name are reasonable or not.

We use `ROOT` in some places, so `@ROOT` is better than `@DEFAULT` and others.

https://github.com/fluent/fluentd/blob/da162e15bc95b1263f92eb0622c714f8b9d303b9/lib/fluent/config/v1_parser.rb#L44

ashie:
@ferbar Could you replace with `@ROOT` and add unit test?
If you hard to add unit test, please let me know.
We'll add it.

ferbar:
@ashie I changed '' to '@ROOT', will test it tomorrow. I'm sorry I don't get it with the unit-tests. I found test/plugin_helper/test_event_emitter.rb, but I didn't find out how to do the

`assert(event_emitter_router("@ROOT" == Engine.root_agent.event_router)
`

I will check what
`<label @ROOT>
`
will do. That should probably be prevented...

Thanks!

ashie:
I've added tests for `event_emitter_router`.

But ... I'm now doubting to use the label name `@ROOT` as a reserved word in this context.
I am sorry that I repeat the discussion...

Because it's too simple, users might be already using it in their configurations. It means it breaks backward compatibility.
Although the prefix `@` is known as reserved word in Fluentd's configuration file, I think it's not applicable for label names because example configurations in [the document](https://docs.fluentd.org/configuration/routing-examples#with-label-input-greater-than-filter-greater-than-output) uses `@` prefix as label names.  Most users use `@` prefix as label names  since the document exemplify it, it's not always necessary though.

When we add a new reserved word, we should ensure to select a word which is used rarely.
e.g.)
* `@ROOT_EVENT_ROUTER`
* `@__ROOT__`
* ...

How do you think about it? @cosmo0920 @kenhys @repeatedly

kenhys:
* `@__ROOT__` 

It may be better. 
And another idea, how about  `@__ROOT__@` or `@ROOT@` (with prefix `@` and suffix `@` )

ferbar:
I think this is the right place to check if someone is using the _root_event_router_ in a label tag:
https://github.com/fluent/fluentd/blob/master/lib/fluent/root_agent.rb#L133

ashie:
BTW fluentd has a built-in label [`@ERROR`](https://docs.fluentd.org/configuration/config-file#error-label).
To keep consistency of naming-convention, `@ROOT` might be better than names like `@__ROOT__`.

test/plugin_helper/test_event_emitter.rb
```diff
@@ -48,4 +49,32 @@ class Dummy < Fluent::Plugin::TestBase
...
     d1.terminate
   end
+
+  test 'should not have event_emitter_router' do
+    d0 = Dummy0.new
+    assert !d0.respond_to?(:event_emitter_router)
+  end
+
+  test 'should have event_emitter_router' do
+    d0 = Dummy.new
```
cosmo0920, line 59: This line should be confusable on first glance.
We should use to store `d` for Dummy class' instance.

cosmo0920 (changes requested):

test/plugin_helper/test_event_emitter.rb
ashie: Ah, it should be so.

ferbar:
Added an error message for 

```
<label @ROOT>
  <match debug.**>
    @type stdout
    @id stdout_output
  </match>
</label>
```


```
2021-07-29 15:08:16 +0200 [info]: parsing config file is succeeded path="fluent.conf"
2021-07-29 15:08:16 +0200 [error]: config error file="fluent.conf" error_class=Fluent::ConfigError error="Invalid label name (@ROOT)"
```

cosmo0920 (approved):

ashie (approved):
> Added an error message for
> 
> ```
> <label @ROOT>
>   <match debug.**>
>     @type stdout
>     @id stdout_output
>   </match>
> </label>
> ```
> 
> ```
> 2021-07-29 15:08:16 +0200 [info]: parsing config file is succeeded path="fluent.conf"
> 2021-07-29 15:08:16 +0200 [error]: config error file="fluent.conf" error_class=Fluent::ConfigError error="Invalid label name (@ROOT)"
> ```

Thanks, now it can notify users about incompatibility of config file.
Let's go with this.

ferbar:
@ashie, @cosmo0920:

Should we add some documentation on these pages:
https://docs.fluentd.org/configuration/config-file#error-label

https://docs.fluentd.org/configuration/routing-examples

What about removing the @ in `@SYSTEM` `@METRICS` ?

PR #3360, opened 2021-05-06
Navigate to community forum instead of mailing list

kenhys:

Fixes N/A

It is better to use a community forum instead of
the traditional mailing list in most of the cases.

Recommends using https://discuss.fluentd.org/.

**Docs Changes**:

N/A

**Release Note**: 

N/A


ashie:
Could you add it to https://github.com/fluent/fluentd#more-information too?

kenhys:
I've fixed more-information section, too.

ashie (approved):

PR #3364, opened 2021-05-11
Handle both of stat and timer watchers are enabled and steady growing file case

cosmo0920:

Follows up #3185 

In this case, timer watcher tells file contents every 1 seconds and file
contents changes is also notified from stats watcher.

In such circumstances, if some tailing files are steadily growing,
previous implementation does not prevent log ingestion for written
contents which is notified by stat watcher.

This commit also prevents log ingestion in such cases.

**Docs Changes**:

No needed.

**Release Note**: 

Same as title.

**Additional context**

I can implement this edge case patch, but how can we write down into test cases...?

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -908,6 +908,8 @@ module Fluent::Plugin
           @io = nil
           @notify_mutex = Mutex.new
           @log = log
+          @last_read_time = 0
+          @number_bytes_read = 0
 
           @log.info "following tail of #{@path}"
         end
@@ -948,10 +950,17 @@ module Fluent::Plugin
           false
         end
 
+        def refresh_bytes_read_counter
+          if Fluent::Clock.now - @last_read_time > 1
+            @number_bytes_read = 0
+          end
+
+          yield
+        end
+
         def handle_notify
           with_io do |io|
             begin
-              number_bytes_read = 0
               start_reading = Fluent::Clock.now
               read_more = false
 
@@ -959,7 +968,9 @@ module Fluent::Plugin
                 begin
                   while true
                     data = io.readpartial(BYTES_TO_READ, @iobuf)
-                    number_bytes_read += data.bytesize
+                    refresh_bytes_read_counter do
+                      @number_bytes_read += data.bytesize
+                    end
                     @fifo << data
                     @fifo.read_lines(@lines)
 
@@ -969,11 +980,12 @@ module Fluent::Plugin
                       read_more = true
                       break
                     end
-                    if limit_bytes_per_second_reached?(start_reading, number_bytes_read)
+                    if limit_bytes_per_second_reached?(start_reading, @number_bytes_read)
                       # Just get out from tailing loop.
                       read_more = false
                       break
                     end
+                    @last_read_time = Fluent::Clock.now
                   end
                 rescue EOFError
                 end
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -908,6 +908,7 @@ module Fluent::Plugin
           @io = nil
           @notify_mutex = Mutex.new
           @log = log
+          @start_reading = 0
           @last_read_time = 0
           @number_bytes_read = 0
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -969,11 +980,12 @@ def handle_notify
                       read_more = true
                       break
                     end
-                    if limit_bytes_per_second_reached?(start_reading, number_bytes_read)
+                    if limit_bytes_per_second_reached?(start_reading, @number_bytes_read)
                       # Just get out from tailing loop.
                       read_more = false
                       break
                     end
+                    @last_read_time = Fluent::Clock.now
```
ashie, line 988: Probably storing `start_reading` and reset also it in `refresh_bytes_read_counter` is more appropriate.
By this, `limit_bytes_per_second_reached?` doesn't need arguments any more.

lib/fluent/plugin/in_tail.rb
```diff
@@ -948,18 +950,27 @@ def limit_bytes_per_second_reached?(start_reading, number_bytes_read)
...
+        def refresh_bytes_read_counter
+          if Fluent::Clock.now - @last_read_time > 1
+            @number_bytes_read = 0
+          end
+
+          yield
+        end
+
         def handle_notify
           with_io do |io|
```
ashie, line 962: Should return immediately when we are already encountering the read limit (until passes 1 sec).

lib/fluent/plugin/in_tail.rb
cosmo0920: That's reasonable.
I decided to make `start_reading` as instance variable to store starting to read time.

lib/fluent/plugin/in_tail.rb
cosmo0920: With previous implementation, your suggestion shouldn't be handled here.
But, current patch https://github.com/fluent/fluentd/pull/3364/commits/34fcda1d69f9692473a8ad0ac9b2c8cbffc96357 should handle here due to stored time with `@start_reading`.

ashie: Hmm, hard to tell my intention by words.
I've added a commit to simplify it.

cosmo0920:
> An idea is that adding a hook to touch the test log file to `limit_bytes_per_second_reached?` by
> [rr's Proxy](https://www.rubydoc.info/gems/rr/RR/DoubleDefinitions/Strategies/Implementation/Proxy).
> By this, it will make sure to update the file and emit inotify event on appropriate timing.
> 
> Please see the following example:
> [1c1ef8e#diff-b8d2266d24d2fe9459a77854ada3159ca968433b87c23b5b716ba3c132b57714R1931](https://github.com/fluent/fluentd/commit/1c1ef8ecf3cba3fa7d5da6b901ba3baff2f439ef#diff-b8d2266d24d2fe9459a77854ada3159ca968433b87c23b5b716ba3c132b57714R1931)
> Probably other test code can be almost same with existing one in #3185.

I'd considered this and I decided to add two testing scenarios.
One is already throttled scenario:
https://github.com/fluent/fluentd/pull/3364/commits/bc11e6c477744185974e2a38ebbdfc7ef2fff0f1
The other is glitched writing timing scenario:
https://github.com/fluent/fluentd/pull/3364/commits/bee91259b98398e6ae00c00b52a71bb08851e4db

Both scenarios  return same number of events per emits.

ashie:
I'm checking the tests, I feel some curious points.
In addition, I found that the test in #3185 is still inappropriate.
I'm now try to fix them.

ashie:
> I'm now trying to fix them.

I've added a commit but probably it will fail on mac & windows.

test/plugin/test_in_tail.rb
```diff
@@ -345,53 +345,109 @@ def test_emit_with_read_lines_limit(data)
...
+            config = CONFIG_READ_FROM_HEAD + config_element("", "", { "read_lines_limit" => limit, "read_bytes_limit_per_second" => limit_bytes }) + PARSE_SINGLE_LINE_CONFIG
+          end
+          d = create_driver(config)
+          msg = 'test' * 2000 # in_tail reads 8192 bytes at once.
+
+          mock.proxy(d.instance).io_handler(anything, anything) do |io_handler|
+            require 'fluent/config/types'
+            limit_bytes_value = Fluent::Config.size_value(limit_bytes)
+            io_handler.instance_variable_set(:@number_bytes_read, limit_bytes_value)
+            mock.proxy(io_handler).handle_notify.at_least(5)
```
cosmo0920, line 424: In Windows and macOS, `handle_inotify` is called 2 times instead of 5.
But I'm not sure why they don't call it 5 times.... :confounded:

ashie (approved):

cosmo0920:
Thanks!

test/plugin/test_in_tail.rb
ashie: handle_inotify will be called on every file change at https://github.com/fluent/fluentd/blob/e2e51d1182188d8d498e096dfab1a031dafacb75/test/plugin/test_in_tail.rb#L442-L447

ashie: Why I choose `at_least(5)` instead of `at_least(9)` is that considering instability of CI.

PR #3367, opened 2021-05-11
filter plugin add track_key to track the forward info, and  add record_size_key to record source data size

jiejieling:
Which issue(s) this PR fixes:
None

What this PR does / why we need it:
In our use scenario, we need to monitor and track the data flow link status, and do some cleaning according to the size of the data, so we have added two fields of original data size and data flow node information to solve this need

Docs Changes:
None

Release Note:
Same as title

### DIFF
--- a/lib/fluent/plugin/filter_record_transformer.rb
+++ b/lib/fluent/plugin/filter_record_transformer.rb
@@ -40,7 +40,11 @@ module Fluent::Plugin
     desc 'When set to true, the full Ruby syntax is enabled in the ${...} expression.'
     config_param :enable_ruby, :bool, default: false
     desc 'Use original value type.'
-    config_param :auto_typecast, :bool, default: true
+    config_param :auto_typecast, :bool, default: false # false for lower version compatibility
+    desc 'Add forward tracking info.'
+    config_param :track_key, :string, default: nil
+    desc 'Add record length info.'
+    config_param :record_size_key, :string, default: nil
 
     def configure(conf)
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/filter_record_transformer.rb
```diff
@@ -103,6 +107,21 @@ def filter_stream(tag, es)
...
+            else
+              new_record[@track_key] = "#{@hostname},#{Time.new.to_i}"
+            end
+          end
+
+          # add record length information
+          if @record_size_key
+            new_record[@record_size_key] = new_record.to_s.size
+          end
+
```
cosmo0920, line 124: First, could you write test for your implementation?

---

>               new_record[@track_key] = "#{new_record[@track_key]},#{@hostname},#{Time.new.to_i}"

`Time.new.to_i` decays nanosecond precision. We should use `Time.now.to_f` here.

>               new_record[@track_key] = "#{@hostname},#{Time.new.to_i}"

ditto.

cosmo0920 (changes requested):
I feel that your suggestion seems to be a bit different target for this record_transporter.

I suggest that your implementation might have to extract a new plugin which name is like `filter_tracking_information`.

What do you think?

lib/fluent/plugin/filter_record_transformer.rb
jiejieling: Ok, I got it,   think you.

PR #3373, opened 2021-05-13
in_http: Add support for HTTP GET requests

### LINKED ISSUE #3372: http_in plugin respond with 200 OK to GET /
I am trying to run Fluentd by deploying a docker container to an Azure App Service. App Service is designed for running HTTP APIs and websites and expects an application to respond with HTTP 200 to GET / before it will route any requests to it.

I would like to have http_in respond with 200 OK to a GET / request.

Running the http_in plugin behind nginx within the same container andn having nginx handle GET /. This isn't preferred for a few reasons though.

**Additional context**
N/A

Is this reasonable? I am happy to work on it & open a PR.

fujimotos:
Fixes #3372

Azure App uses GET requests to check if the HTTP server is working all right.
Since Fluentd responds with "400 Bad Requests" to HTTP GET, it does not work well.

Teach Fluentd to respond nicely with "200 OK" to HTTP GET.

**Docs Changes**:

Not needed

**Release Note**: 

* Starting this version,  `in_http` responds with `200 OK` to GET requests (instead of `404 Bad request`)

### DIFF
--- a/test/plugin/test_in_http.rb
+++ b/test/plugin/test_in_http.rb
@@ -770,6 +770,15 @@ class HttpInputTest < Test::Unit::TestCase
     end
   end
 
+  def test_get_request
+    d = create_driver(CONFIG)
+
+    d.run do
+      res = get("/cors.test", {}, {})
+      assert_equal "200", res.code
+    end
+  end
+
   def test_cors_preflight
     d = create_driver(CONFIG + 'cors_allow_origins ["*"]')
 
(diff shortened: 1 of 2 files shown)


test/plugin/test_in_http.rb
```diff
@@ -985,6 +994,12 @@ def test_if_content_type_is_initialized_properly
     assert_equal $test_in_http_connection_object_ids[0], $test_in_http_connection_object_ids[1]
   end
 
+  def get(path, params, header = {})
+    http = Net::HTTP.new("127.0.0.1", PORT)
+    req = Net::HTTP::Get.new(path, header)
```
ashie, line 999: I think always returning `200` on all paths might not appropriate, it would be better that it's configurable.
On the other hand, it's not worth to introduce such complexity until we find actual inconvenience.

ashie (approved):

Issue #3375, opened 2021-05-14
NameError: uninitialized constant RubyVM for non-CRuby


eregon:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

**Describe the bug**
Starting fluentd on any other implementation than CRuby (e.g. TruffleRuby) causes:
```
$ bundle exec fluentd -c receiver.conf
bundler: failed to load command: fluentd (/home/eregon/.rubies/truffleruby-dev/bin/fluentd)
NameError: uninitialized constant RubyVM
  /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/fluentd-1.12.3/bin/fluentd:4:in `const_missing'
  /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/fluentd-1.12.3/bin/fluentd:4:in `<top (required)>'
  <internal:core> core/kernel.rb:400:in `load'
  <internal:core> core/kernel.rb:400:in `load'
  /home/eregon/.rubies/truffleruby-dev/bin/fluentd:23:in `<top (required)>'
  <internal:core> core/kernel.rb:400:in `load'
  <internal:core> core/kernel.rb:400:in `load'
```
due to
https://github.com/fluent/fluentd/blob/5bd334613fa0561298cf9b7acbdd65043ad9e473/bin/fluentd#L4
https://github.com/fluent/fluentd/pull/1764

**To Reproduce**
I used https://github.com/fluent/fluentd-benchmark/tree/master/one_forward but any fluent command would do

**Expected behavior**
It should work.

The `rescue nil` is not enough, for instance this also fails on CRuby:
```
$ ruby -ve 'RubyVM2::InstructionSequence.compile_option = {trace_instruction: false} rescue nil'
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux]
Traceback (most recent call last):
-e:1:in `<main>': uninitialized constant RubyVM2 (NameError)
```

I think the issue is the `rescue nil` only applies to `{trace_instruction: false}` and not the whole assignment.

**Your Environment**

- Fluentd or td-agent version: `fluentd --version` or `td-agent --version`: 1.12.3

cosmo0920:
I'm not familiar with TruffleRuby. But, other C-Ruby implementation can be handle to install C extension such as cool.io, msgpack-ruby and other gems?
[...]

PR #3376, opened 2021-05-17
bin: fluentd: Specify trace_instruction as false if available

cosmo0920:

Related to #3375 

In recent CRuby, RubyVM::InstructionSequence is always defined.
But other implementation such as truffleruby does not provide such
class.

But even if this commit, fluentd does not work correctly with truffleruby. 
Especially, Fluentd sometimes does not terminate correctly with TERM signal.

**Docs Changes**:

No needed.

**Release Note**: 

Same as title.

**Additional Note**:

`TruffleRuby` can be installed with `rbenv install truffleruby-21.0.0`. We can use it to investigate issue which is caused by truffleruby.

### DIFF
--- a/bin/fluentd
+++ b/bin/fluentd
@@ -1,6 +1,11 @@
 #!/usr/bin/env ruby
 # -*- coding: utf-8 -*-
 
+# In recent CRuby, RubyVM::InstructionSequence is always defined.
+# This optimization aims to reduce memory (avg 0.5 megabytes) at startup.
+# ref: http://atdot.net/~ko1/activities/2017_fukuoka_rubykaigi_02.pdf
+# But other implementation such as truffleruby does not provide such
+# class.
 if defined?(RubyVM::InstructionSequence)
   RubyVM::InstructionSequence.compile_option = {trace_instruction: false}
 end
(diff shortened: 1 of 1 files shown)


bin/fluentd
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env ruby
 # -*- coding: utf-8 -*-
 
-RubyVM::InstructionSequence.compile_option = {trace_instruction: false} rescue nil
+if defined?(RubyVM::InstructionSequence)
+  RubyVM::InstructionSequence.compile_option = {trace_instruction: false}
```
kenhys, line 5: Could you add an explanation of which platform is available for this kind of optimization?

kenhys: If an explanation is added, it looks ok for me.

cosmo0920: I added in https://github.com/fluent/fluentd/pull/3376/commits/2404104e305e68fb568d0f52f4e1745130bb922a.

kenhys (approved):
LGTM

cosmo0920:
@ashie Any comments?

ashie (approved):

cosmo0920:
Thanks!

PR #3378, opened 2021-05-18
in_tail: Safely skip files used by another process on Windows

ashie:
None.
Reported at https://discuss.fluentd.org/t/windiws-server-fluentd-stops-working/61

On Windows, a win32 error ERROR_SHARING_VIOLATION is occured on
`Fluent::FileWrapper.stat(path)` when a file is temporarily used by
another process. Such files should be skipped safely until the file is
released as same as ENOENT case.

This patch rescues also other EACCES cases although they rarely occur.

**Docs Changes**:
None.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/file_wrapper.rb
+++ b/lib/fluent/plugin/file_wrapper.rb
@@ -52,6 +52,8 @@ module Fluent
 
     attr_reader :errcode, :msg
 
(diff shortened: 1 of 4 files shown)


cosmo0920 (approved):
Looks reasonable for me. :+1:

Issue #3333, opened 2021-04-16
in_tail plugin crashes on multiple gracefulReloads of fluentd


alex-vmw:
**Describe the bug**
We were performing tests that were issuing multiple /api/config.gracefulReload calls to fluentd. After issuing 2+ gracefulReloads one right after another (without letting fluentd fully go through the reload) we can see the below error in the logs:
```
2021-04-16 01:34:19 +0000 [error]: #0 [in_tail_container_logs] closed stream
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail/position_file.rb:212:in `pos='
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail/position_file.rb:212:in `block in update_pos'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail/position_file.rb:211:in `synchronize'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail/position_file.rb:211:in `update_pos'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail.rb:933:in `block in handle_notify'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail.rb:964:in `with_io'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail.rb:912:in `handle_notify'
  2021-04-16 01:34:19 +0000 [error]: #0 /usr/local/lib/ruby/gems/2.7.0/gems/fluentd-1.12.2/lib/fluent/plugin/in_tail.rb:895:in `block in on_notify'
[...]

Issue #1504, opened 2017-03-15
timer: Add immediate option to timer_execute


repeatedly:
Enhancement request for input plugins.
By default, `PluginHelper::Timer#timer_execute` launches timer immediately.
This is no problem on almost cases but some plugins, e.g. in_tail, takes long time in `start` by executing code.
To avoid this problem, delaying timer execution mitigates this situation.

See also: https://github.com/fluent/fluentd/pull/1502

fujimotos:
FWIW, I also want to use this option in https://github.com/prometheus-community/windows_exporter.

 * We use `timer_execute` to periodically collect system metrics (default interval is 60s).
 * I feel the current user experience is sub-optimal, because our user needs
   to wait 1 minute to confirm that the plugin is actually working.

I was considering to migrate to manual timer implementation because of this issue,
but it is preferable if `timer_execute` supports `immediate` option.

PR #3380, opened 2021-05-18
in_tail: Shutdown immediately & safely even if reading huge files

### LINKED ISSUE #3323: Tail input plugin doesn't respect SIGTERM  signal

**Describe the bug**
We use tail input plugin to read logs. And Fluentd is deployed as daemonset in Kubernetes. During deployment, Kubernetes will first send a SIGTERM signal to Fluentd container to ask it to gracefully shut down. But tail input plugin keeps reading data until the container eventually killed. 

**To Reproduce**
Run Fluentd in Kubernetes and trigger a deployment. 

**Expected behavior**
When receive SIGTERM, tail should stop reading, keep record of current position and flush out memory buffers. 

**Your Environment**

- Fluentd or td-agent version: `td-agent 4.1.0 fluentd 1.12.1 (e3effa337593618cbd7f0f4ef071766df1ec69a0)`
- Operating system: `Ubuntu 18.04.5 LTS (Bionic Beaver)`
- Kernel version: `5.4.0-1039-azure`

If you hit the problem with older fluentd version, try latest version first.

**Your Configuration**

```
<source>
  @type tail
  pos_file /var/log/fluentd-containers.log.pos
  time_key @timestamp
  tag kubernetes.*
  format json
  read_from_head true
</source>
```

**Your Error Log**

```
N/A
```

**Additional context**

N/A

ashie:
Fixes #3323 #3345 #3333 (maybe)

in_tail takes very long time to shutdown when it's reading huge files.
This patch ensure to shutdown ASAP just after receiving a shutdown
command.
    
* Constructed tail watchers should be registered ASAP to be recognized
  by the shutdown process.
* Break `TailWatcher#on_notify` ASAP
  Wait 5 secdonds before breaking it to ensure reading remaining lines.
    
It takes effect only when `skip_refresh_on_startup` is enabled, because
reading files is occurred on main thread without it. It should be
enabled by default in the future.

**Docs Changes**:
none

**Release Note**: 
Same as title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -56,6 +56,7 @@ module Fluent::Plugin
       @pf_file = nil
       @pf = nil
       @ignore_list = []
+      @shutdown_start_time = nil
     end
 
     desc 'The paths to read. Multiple paths can be specified, separated by comma.'
@@ -256,6 +257,7 @@ module Fluent::Plugin
     end
 
     def shutdown
+      @shutdown_start_time = Fluent::Clock.now
       # during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.
       stop_watchers(existence_path, immediate: true, remove_watcher: false)
       @pf_file.close if @pf_file
@@ -385,8 +387,6 @@ module Fluent::Plugin
         tw.register_watcher(tt)
       end
 
-      tw.on_notify
-
       tw.watchers.each do |watcher|
         event_loop_attach(watcher)
       end
@@ -398,7 +398,7 @@ module Fluent::Plugin
           event_loop_detach(watcher)
         end
 
-        tw.detach
+        tw.detach(@shutdown_start_time)
         tw.close
       end
       raise e
@@ -427,6 +427,7 @@ module Fluent::Plugin
       begin
         target_info = TargetInfo.new(target_info.path, Fluent::FileWrapper.stat(target_info.path).ino)
         @tails[target_info] = tw
+        tw.on_notify
       rescue Errno::ENOENT, Errno::EACCES => e
         $log.warn "stat() for #{target_info.path} failed with #{e.class.name}. Drop tail watcher for now."
         # explicitly detach and unwatch watcher `tw`.
@@ -438,6 +439,7 @@ module Fluent::Plugin
     def start_watchers(targets_info)
       targets_info.each_value {|target_info|
         construct_watcher(target_info)
+        break if before_shutdown?
       }
     end
 
@@ -490,9 +492,11 @@ module Fluent::Plugin
 
         if new_position_entry.read_inode == 0
           @tails[new_target_info] = setup_watcher(new_target_info, new_position_entry)
+          @tails[new_target_info].on_notify
         end
       else
         @tails[new_target_info] = setup_watcher(new_target_info, pe)
+        @tails[new_target_info].on_notify
       end
       detach_watcher_after_rotate_wait(rotated_tw, pe.read_inode) if rotated_tw
     end
@@ -505,7 +509,7 @@ module Fluent::Plugin
       tw.watchers.each do |watcher|
         event_loop_detach(watcher)
       end
-      tw.detach
+      tw.detach(@shutdown_start_time)
 
       tw.close if close_io
 
@@ -717,8 +721,11 @@ module Fluent::Plugin
         @watchers << watcher
       end
 
-      def detach
-        @io_handler.on_notify if @io_handler
+      def detach(shutdown_start_time = nil)
+        if @io_handler
+          @io_handler.ready_to_shutdown(shutdown_start_time)
+          @io_handler.on_notify
+        end
         @line_buffer_timer_flusher&.close(self)
       end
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -965,16 +976,18 @@ def handle_notify
...
-                    end
                     if limit_bytes_per_second_reached?
                       # Just get out from tailing loop.
                       read_more = false
                       break
                     end
+                    if @lines.size >= @read_lines_limit
+                      # not to use too much memory in case the file is very large
+                      @ready_to_shutdown_mutex.synchronize {
+                        read_more = !@ready_to_shutdown
```
ashie, line 987: I think it should wait a few seconds instead of aborting immediately.

lib/fluent/plugin/in_tail.rb
```diff
@@ -718,6 +720,7 @@ def register_watcher(watcher)
       end
 
       def detach
+        @io_handler.ready_to_shutdown if @io_handler
```
ashie, line 723: `detach` is called not only when shutdown but also on refreshing watchers.
But I think we don't need consider the later case because it will occur only when idling case (not occur while running `on_notify`).

lib/fluent/plugin/in_tail.rb
```diff
@@ -490,9 +490,11 @@ def update_watcher(target_info, pe)
 
         if new_position_entry.read_inode == 0
           @tails[new_target_info] = setup_watcher(new_target_info, new_position_entry)
+          @tails[new_target_info].on_notify
```
ashie, line 493: Constructed watchers should be registered ASAP, before running `on_notify`.
Otherwise shutting down process can't recognize the watcher so that it won't be killed.
This is the why the error I mentioned at https://github.com/fluent/fluentd/issues/3323#issuecomment-830634918 is occurred.

lib/fluent/plugin/in_tail.rb
ashie: Done.

ashie:
This patch might fixes also #3333, but a bit more change may be needed to address it.
I'll confirm it before merging it.

ashie:
> It takes effect only when `skip_refresh_on_startup` is enabled, because reading files is occurred on main thread without it.

Because it's executed at:

https://github.com/fluent/fluentd/blob/b6ba1c27db9a1f1b736b20c1c5018ed6cd88a01e/lib/fluent/plugin/in_tail.rb#L246

it blocks:

https://github.com/fluent/fluentd/blob/b6ba1c27db9a1f1b736b20c1c5018ed6cd88a01e/lib/fluent/engine.rb#L144-L148

so that trailing main loop isn't executed yet on this case.

I'm now considering how to stop on this case.
Even if in this case, `TERM` signal handler is worked and `@engine_stopped` of `Fluent::Engine` is set.
So we can stop `handle_notify` by checking it.
But I think it's not smart design, make `skip_refresh_on_startup` default is the appropriate way I think.
Probably it's harmless but not certain.

fujimotos:
@ashie I'm testing this PR on my machine. WFM.

fujimotos (approved):

fujimotos:
I can confirm this PR fixes #3323: Sending `killall -TERM fluentd` 
stops `in_tail` after 5 seconds (when this PR is applied):

```console
$ bundle exec fluentd -c fluentd.conf
...
2021-05-24 23:44:17 +0900 [info]: #0 shutting down fluentd worker worker=0
2021-05-24 23:44:17 +0900 [info]: #0 shutting down input plugin type=:tail plugin_id="object:2adaf83ea9ec"                                                              
2021-05-24 23:44:22 +0900 [info]: #0 shutting down output plugin type=:null plugin_id="object:2adaf837b128"
```

The configuration file I used for the testing was:

```xml
<source>
  @type tail
  tag debug.log
  path data.txt
  read_from_head true
  skip_refresh_on_startup true
  refresh_interval 1s
  <parse>
    @type json
  </parse>
</source>

<match debug.**>
  @type null
</match>
```

and the data file was:

```console
$ ls -hl data.txt
-rw-r--r-- 1 jailbird jailbird 181M May 24 23:29 data.txt
$ head -n3 data.txt
{"hello": 0}
{"hello": 1}
{"hello": 2}
```

So I believe this PR is mergeable.

fujimotos:
> It takes effect only when skip_refresh_on_startup is enabled, because reading files is occurred on main thread without it.

I expect that this won't be so much a big issue, because:

 * The "main thread" blocking occurs just once: during the initialization phase of `in_tail`.
 * After we pass that phase, `SIGTERM`  works, regardless of `skip_refresh_on_startup` enabled or not.
 * For this reason, this issue occurs mostly when users try to shutdown Fluentd immediately after boot.
   * ... which I believe a rare thing to do.

> I'm now considering how to stop on this case.
> ...
> Probably it's harmless but not certain...

I think the most expedient fix is _"Don't read the file in main thread."._

Probably we can just create a new thread and kick `refresh_watcher` there:

```ruby
def start:
    ...
    Thread.new do 
        refresh_watchers unless @skip_refresh_on_startup
    end     
end
```

... or maybe we should add a new option `immediate` to `timer_execute()`,
to let it start executing the callback without delays.

But either way, I think we can fix that in another PR.

ashie:
Thanks for your review!

ashie:
> ... or maybe we should add a new option `immediate` to `timer_execute()`,
> to let it start executing the callback without delays.

refs: #1502 #1504

PR #3379, opened 2021-05-18
in_tail: read_bytes_limit_per_second should precede read_lines_limit

ashie:
**Which issue(s) this PR fixes**: 
Follow up #3185

**What this PR does / why we need it**: 
in_tail: read_bytes_limit_per_second should precede read_lines_limit
    
Otherwise it doesn't limit reading pace on reading many short lines.

**Docs Changes**:
none

**Release Note**: 
Same with #3185

PR #3390, opened 2021-05-25
in_tail: Ensure to reach EOF after rotate even if throttling is enabled

### LINKED ISSUE #2478: in_tail take long time to found rotated files

ashie:
Fixes #2478

Although log throttling feature is landed at #3185 (and #3379), it has still remaining issue mentioned at https://github.com/fluent/fluentd/issues/2478#issuecomment-512086345
When the throttling feature is enabled and rotation is occurred, remaining lines will be discarded.
(When it's not enabled, in_tail make sure to reach EOF with blocking other tasks.)
This patch ensure to reach EOF even if throttling is enabled.

**Docs Changes**:
none

**Release Note**: 
~~Same with  #3185 (if this PR is landed on v1.13)~~
Same with the title

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -522,8 +522,25 @@ module Fluent::Plugin
     def detach_watcher_after_rotate_wait(tw, ino)
       # Call event_loop_attach/event_loop_detach is high-cost for short-live object.
       # If this has a problem with large number of files, use @_event_loop directly instead of timer_execute.
-      timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
+      if @open_on_every_update
+        # Detach now because it's already closed, waiting it doesn't make sense.
         detach_watcher(tw, ino)
+      elsif @read_bytes_limit_per_second < 0
+        # throttling isn't enabled, just wait @rotate_wait
+        timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
+          detach_watcher(tw, ino)
+        end
+      else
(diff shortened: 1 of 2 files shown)


ashie:
I've added tests for it, it seems work as expected.

kenhys (approved):
LGTM

kenhys:
@cosmo0920 could you check, please?

cosmo0920 (approved):
LGTM! :+1:

ashie:
Thanks!

cosmo0920:
:tada:

Issue #3365, opened 2021-05-11
in_tail: Fluentd v1.12.3 fails to handle file rotations on Windows


fujimotos:
**Describe the bug**

After a few file rotations, `in_tail` stops tracking log files.

**To Reproduce**

1.  Run Fluentd with the attached configuration.

2. Rotate the file twice using cmd.exe

   ```cmd
   % cd /opt/td-agent/
   % echo aiueo >> test.log
   % del test.log
   % echo aiueo >> test.log
   % del test.log
   ```

3. Now Fluentd just gets stuck, keeping `test.log` in `DeletePending` status forever.

**Expected behavior**

File rotation should work properly.

**Your Environment**

- Windows Server 2016
- td-agent v4.1.1
- Fluentd v1.12.3

**Your Configuration**

```xml
<source>
@type tail
 path /opt/td-agent/test.log
 tag test.log
 <parse>
   @type none
 </parse>
</source>

<match test**>
  @type stdout
</match>
```

**Additional context**

* This issue is evidenlty occurring since #3196.

ashie:
@cosmo0920 Could you take look this?
Feel free to reassign me if you can't.

cosmo0920:
Sure. I'll tackle on this in the next week. :eyes:

cosmo0920:
I'd confirmed that the current master (b6ba1c27db9a1f1b736b20c1c5018ed6cd88a01e) can be stucked with the attached configuration.
And I found that taking over 5 seconds to adding lines into rotated file can be reproduce this issue:
```cmd
% del test.log
del Adding a line after long time (over 5 seconds)
% echo aiueo >> test.log
```

cosmo0920:
I'd confirmed that this issue is also happened on Linux. Not for Windows specific.

AmeerAssi:
thanks @cosmo0920.
@fujimotos mentioned that the root cause of this issue: https://github.com/fluent/fluentd/issues/3340 might be the issue here.

we are using Fluentd v.1.11.2, do you know if its reproducing with this version?

cosmo0920:
> we are using Fluentd v.1.11.2, do you know if its reproducing with this version?

This version is not reproducible this issue. This issue is only for Fluentd v1.12.x.

cosmo0920:
I've sent a PR to fix this issue: https://github.com/fluent/fluentd/pull/3391

And I have never reproduced this issue with https://github.com/fluent/fluentd/pull/3391.

[...]

PR #3391, opened 2021-05-25
in_tail: position_file: Define path based equalities on TargetInfo

cosmo0920:

Related to https://github.com/fluent/fluentd/issues/3365

In the previous Fluentd v1.11.x implementation uses path based tailing
keys.
We wouldn't rewrite tailing mechanism entirely.
And also we're implicitly assuming path based equality for TargetInfo in tailing logic.
We should handle TargetInfo with path based equality instead of path and
inode based equality.

**Docs Changes**:

No needed.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/in_tail/position_file.rb
+++ b/lib/fluent/plugin/in_tail/position_file.rb
@@ -248,6 +248,20 @@ module Fluent::Plugin
       end
     end
 
-    TargetInfo = Struct.new(:path, :ino)
+    TargetInfo = Struct.new(:path, :ino) do
+      def ==(other)
+        return false unless other.is_a?(TargetInfo)
+        self.path == other.path
+      end
+
+      def hash
+        self.path.hash
+      end
+
+      def eql?(other)
+        return false unless other.is_a?(TargetInfo)
+        self.path == other.path
+      end
+    end
   end
 end
--- a/test/plugin/in_tail/test_position_file.rb
+++ b/test/plugin/in_tail/test_position_file.rb
@@ -282,4 +282,29 @@ class IntailPositionFileTest < Test::Unit::TestCase
       assert_equal 2, f.read_inode
     end
   end
+
+  sub_test_case "TargetInfo" do
+    def test_equal
+      t1 = Fluent::Plugin::TailInput::TargetInfo.new("test", 1234)
+      t2 = Fluent::Plugin::TailInput::TargetInfo.new("test", 1235)
+
+      assert_equal t1, t2
+    end
+
+    def test_eql?
+      t1 = Fluent::Plugin::TailInput::TargetInfo.new("test", 1234)
+      t2 = Fluent::Plugin::TailInput::TargetInfo.new("test", 5321)
+
+      assert do
+        t1.eql? t2
+      end
+    end
+
+    def test_hash
+      t1 = Fluent::Plugin::TailInput::TargetInfo.new("test", 1234)
+      t2 = Fluent::Plugin::TailInput::TargetInfo.new("test", 7321)
+
+      assert_equal t1.hash, t2.hash
+    end
+  end
 end
--- a/test/plugin/in_tail/test_position_file.rb
+++ b/test/plugin/in_tail/test_position_file.rb
@@ -291,6 +291,13 @@ class IntailPositionFileTest < Test::Unit::TestCase
       assert_equal t1, t2
     end
 
+    def test_not_equal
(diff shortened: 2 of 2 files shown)


ashie:
Typo?
```diff
-In the previous Fluentd v1.11.x implementation uses patch based tailing keys.
+In the previous Fluentd v1.11.x implementation uses path based tailing keys.
```

cosmo0920:
> Typo?
> 
> ```diff
> -In the previous Fluentd v1.11.x implementation uses patch based tailing keys.
> +In the previous Fluentd v1.11.x implementation uses path based tailing keys.
> ```

Yep. This is typo. I'd fixed it.

ashie:
Hmm, we should release v1.12.4 with this patch before releasing v1.13?

ashie:
I'll create v1.12 branch.

cosmo0920:
> Hmm, we should release v1.12.4 with this patch before releasing v1.13?

Yes, please.

ashie:
I'll release v1.12.4 in a few days.

lib/fluent/plugin/in_tail/position_file.rb
```diff
@@ -248,6 +248,20 @@ def read_inode
...
 
-    TargetInfo = Struct.new(:path, :ino)
+    TargetInfo = Struct.new(:path, :ino) do
+      def ==(other)
+        return false unless other.is_a?(TargetInfo)
+        self.path == other.path
+      end
+
+      def hash
+        self.path.hash
```
ashie, line 258: I'm not sure follow_inodes feature but does it work well also on `@follow_inodes`?
I think that keys of `@tails` should be inode on `@follow_inodes`.

ashie: Or paths are updated by `update_watcher` on rotation so paths are never duplicate?

cosmo0920: I'll take a look on this.

ashie: I'm checking it too.
Probably above my concern is wrong, it should be path maybe.
Otherwise `update_watcher` can't discard old `TailWather` from `@tails`.

ashie: Memo:

Instead of overriding them, I thought using `target_info.path` as a key is more straightforward and understandable.
`@tails[target_info.path]`

But it's also wrong.
`TargetInfo` is referred by `each_keys` so keys have to `TargetInfo`.

ashie: Hmm, probably it's O.K. as a first aid but I think more refactoring is needed to make easy to understand in the future.

test/plugin/in_tail/test_position_file.rb
```diff
@@ -282,4 +282,29 @@ def build_files(file)
...
+
+    def test_eql?
+      t1 = Fluent::Plugin::TailInput::TargetInfo.new("test", 1234)
+      t2 = Fluent::Plugin::TailInput::TargetInfo.new("test", 5321)
+
+      assert do
+        t1.eql? t2
+      end
+    end
+
```
kenhys, line 302: To emphasize the difference of eql? and equal? behavior, how about adding one more test case using `assert_not_same t1, t2`

lib/fluent/plugin/in_tail/position_file.rb
cosmo0920: > Instead of overriding them, I thought using `target_info.path` as a key is more straightforward and understandable: `@tails[target_info.path]`
> 
> But it's also wrong.
> `TargetInfo` is referred by `each_keys` so keys have to be `TargetInfo`.

Yes! I'm also considering to use it to be made easier to understand, but overriding equality on TargetInfo is one of the simplest way to dispose discarded old TailWatcher. (This strategy dosen't request more refactoring.)

test/plugin/in_tail/test_position_file.rb
cosmo0920: This should be done in https://github.com/fluent/fluentd/pull/3391/commits/ab36738b86e908cd9a5867a704008bd687b9a1eb and https://github.com/fluent/fluentd/pull/3391/commits/2f2214b5f0840990227ad2fcab709c1e9092f657.

ashie:
I've created a pull request for v1.12 and merged already: #3393
I'll merge v1.12 branch to master after releasing v1.12.4, then close this PR.
Thank you for investigating it!

PR #3393, opened 2021-05-26
in_tail: position_file: Define path based equalities on TargetInfo (for v1.12)

ashie:
#3365
Same change with #3391. This PR is for v1.12

(from #3365)
In the previous Fluentd v1.11.x implementation uses path based tailing keys.
We wouldn't rewrite tailing mechanism entirely.
And also we're implicitly assuming path based equality for TargetInfo in tailing logic.
We should handle TargetInfo with path based equality instead of path and
inode based equality.

**Docs Changes**:
none

**Release Note**: 
in_tail: Fix a bug that in_tail fails to handle file rotations

### DIFF
--- a/lib/fluent/plugin/in_tail/position_file.rb
+++ b/lib/fluent/plugin/in_tail/position_file.rb
@@ -248,6 +248,20 @@ module Fluent::Plugin
       end
     end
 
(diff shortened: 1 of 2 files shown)


cosmo0920:
:raised_hands:

PR #3402, opened 2021-05-30
test: Omit test_EACCES_error_after_setup_watcher if current user is root

ashie:
none

Suppress a test failure on Drone CI:
https://cloud.drone.io/fluent/fluentd/1444/1/2

**Docs Changes**:
none

**Release Note**: 
none

### DIFF
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2088,6 +2088,7 @@ class TailInputTest < Test::Unit::TestCase
   end
 
   def test_EACCES_error_after_setup_watcher
+    omit "Cannot test with root user" if Process::UID.eid == 0
     path = "#{TMP_DIR}/noaccess/tail.txt"
     begin
       FileUtils.mkdir_p("#{TMP_DIR}/noaccess")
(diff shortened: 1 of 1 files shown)


test/plugin/test_in_tail.rb
```diff
@@ -2110,8 +2111,10 @@ def test_EACCES_error_after_setup_watcher
       assert($log.out.logs.any?{|log| log.include?("stat() for #{path} failed with Errno::EACCES. Drop tail watcher for now.\n") })
     end
   ensure
-    FileUtils.chmod(0755, "#{TMP_DIR}/noaccess")
-    FileUtils.rm_rf("#{TMP_DIR}/noaccess")
+    if File.exist?("#{TMP_DIR}/noaccess")
+      FileUtils.chmod(0755, "#{TMP_DIR}/noaccess")
+      FileUtils.rm_rf("#{TMP_DIR}/noaccess")
+    end
   end unless Fluent.windows?
```
kenhys, line 2118: I'm not sure, but is this `if ... end` needed?
It seems that `omit` is enough.

kenhys: If it is needed, could you explain in the commit message, too?

ashie: It's needed because It's in ensure block. It's always called.
I think writing it in commit message is too verbose.

kenhys: Sorry, I've overlooked `ensure` block behavior, so no need to too verbose explanation.

kenhys (approved):
LGTM.

ashie:
Thanks!

PR #3408, opened 2021-06-03
test_in_tail: Remove meaningless test data for throttling tests

ashie:
none

Although throttling tests try many data, about half of them are
meaningless. Since these tests take 1-2 seconds per datum, they
should be removed to reduce whole test time.

**Docs Changes**:
none

**Release Note**: 
none


ashie:
In addition, removing them is more likely to avoid instability of CI environment.

cosmo0920 (approved):
Exactly. It is reasonable for me. :grin:

ashie:
Thanks!

Issue #1293, opened 2016-10-25
Buffered input plugins


tagomoris:
Input plugins may emit events one-by-one (for example, `in_exec`), but it degrades total performance of Fluentd event routing.
If we can support buffer plugins in input plugins, it can improve performance with users' configurations for such cases.

It may be implemented as:
* an extended API (method) for existing input plugin class (if plugin implements it, input plugin can work with buffer)
* a plugin helper (or a new method of event emitter plugin helper)

Output plugin supports both of buffered and non-buffered way. So Input plugin SHOULD support both too.

reversi-fun:
+1;  
"input plugin can work with buffer","new method of event emitter plugin helper"

It might be useful,To count throughput.

https://github.com/fluent/fluentd/issues/1424

PR #3411, opened 2021-06-04
Measure emitting records counts and their sizes for input and output metrics and used it in in_monitor_agent

cosmo0920:
None

Currently, Fluentd metrics do not have emit size for calculating flow rate (per sec or  per hour).
Fluentd Input plugin and Output should have `emit_records` and `emit_size` metrics.

But, currently, I disabled input metrics by default for performance concerns.

`enable_input_metrics` in system config is `false` by default. (By default, metrics of input plugin are not updated.)
In the future, we could adopt this measurement implementation on input plugin as [buffered input plugin](https://github.com/fluent/fluentd/issues/1293).

**Docs Changes**:
Needed.

**Release Note**: 

Same as title.


lib/fluent/system_config.rb
```diff
@@ -46,6 +47,7 @@ class SystemConfig
     config_param :strict_config_value, :bool, default: nil
     config_param :enable_msgpack_time_support, :bool, default: nil
     config_param :disable_shared_socket, :bool, default: nil
+    config_param :without_input_metrics, :bool, default: true
```
kenhys, line 50: Similar to other parameters, `enable_input_metrics` may be better.
(Even though `without_source` exists)

lib/fluent/plugin/input.rb
```diff
@@ -29,6 +29,27 @@ class Input < Base
...
+          'emit_records' => @emit_records,
+          'emit_size' => @emit_size,
+        }
+
+        { 'input' => stats }
+      end
+
+      def metric_callback(es)
+        @counter_mutex.synchronize { @emit_records += 1 }
+        @counter_mutex.synchronize { @emit_size += es.to_msgpack_stream.bytesize }
```
kenhys, line 50: Maybe I'm missing, does it need to be separated `synchronize` ?

test/plugin/test_in_monitor_agent.rb
```diff
@@ -350,7 +356,9 @@ def get(uri, header = {})
         "plugin_category" => "input",
         "plugin_id"       => "test_in",
         "retry_count"     => nil,
-        "type"            => "test_in"
+        "type"            => "test_in",
+        "emit_records"    => 0,
+        "emit_size"       => 0,
       }
       expected_test_in_response.merge!("config" => {"@id" => "test_in", "@type" => "test_in"}) if with_config
```
kenhys, line 363: Maybe extra test cases are needed to confirm `emit_records` and `emit_size` is incremented correctly.

lib/fluent/plugin/input.rb
cosmo0920: I guess it should be better to minimize synchronize regions.

test/plugin/test_in_monitor_agent.rb
cosmo0920: Sure, will do.

lib/fluent/system_config.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3411/commits/334b0490a5f73ce4540835f1d030d512f45c1ebd.

lib/fluent/root_agent.rb
```diff
@@ -55,9 +55,11 @@ def initialize(log:, system_config: SystemConfig.new)
       @suppress_emit_error_log_interval = 0
       @next_emit_error_log_time = nil
       @without_source = false
+      @enable_input_metrics = false
 
       suppress_interval(system_config.emit_error_log_interval) unless system_config.emit_error_log_interval.nil?
       @without_source = system_config.without_source unless system_config.without_source.nil?
+      @enable_input_metrics = system_config.enable_input_metrics unless system_config.enable_input_metrics.nil?
```
ashie, line 62: How about `!!system_config.enable_input_metrics.nil?` to simplify it?

cosmo0920: I think the suggested changes does not have equal meaning:

```irb
irb(main):001:0> value = nil
=> nil
irb(main):002:0> !!value.nil?
=> true
irb(main):003:0> value = false
=> false
irb(main):004:0> !!value.nil?
=> false
irb(main):005:0> value = true
=> true
irb(main):006:0> !!value.nil?
=> false
```

This is not intended.

ashie: Oops, sorry I intended `!!system_config.enable_input_metrics`.

cosmo0920: Hehe, no worries. I'll use it.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3411/commits/b00397af5962cf25bb3c3ba245d6799bae63b472.

test/plugin/test_in_monitor_agent.rb
cosmo0920: I'm investigating how to test `emit_records` and `emit_size` with current test driver.
But I found that, with current input test driver, it is hard to test against input metrics because event router is replaced with the following rules:
https://github.com/fluent/fluentd/blob/master/lib/fluent/test/driver/test_event_router.rb
https://github.com/fluent/fluentd/blob/master/lib/fluent/test/driver/base_owner.rb#L49

lib/fluent/plugin/input.rb
repeatedly: By separating `synchronize`, `emit_records` and `emit_size` may have inconsistent value.

cosmo0920: Oh, I see....

test/plugin/test_in_monitor_agent.rb
cosmo0920: I'm trying to write assertion for input plugin statistics on: https://github.com/fluent/fluentd/pull/3411/commits/ad64150d186d576bfa4a06e4797bd8086ecbe37e

cosmo0920: But I'm not sure why root agent does not ignite metrics callbacks....

lib/fluent/plugin/output.rb
```diff
@@ -800,7 +801,10 @@ def emit_sync(tag, es)
         @counter_mutex.synchronize{ @emit_count += 1 }
         begin
           process(tag, es)
-          @counter_mutex.synchronize{ @emit_records += es.size }
+          @counter_mutex.synchronize do
+            @emit_records += es.size
+            @emit_size += es.to_msgpack_stream.bytesize
```
ashie, line 806: I'm being afraid performance regression of output plugins due to this.
I think it 's too high cost just only for getting stream size even though users might not need this information in most cases.
This patch doesn't provide a way to disable it for output plugins (it can be disabled for input plugins though).

I've measured the performance regression with the following patch and benchmark code:

```diff
diff --git a/lib/fluent/plugin/output.rb b/lib/fluent/plugin/output.rb
index 7d0e3500..e3bd6026 100644
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -171,9 +171,11 @@ module Fluent
       attr_accessor :output_enqueue_thread_waiting, :dequeued_chunks, :dequeued_chunks_mutex
       # output_enqueue_thread_waiting: for test of output.rb itself
       attr_accessor :retry_for_error_chunk # if true, error flush will be retried even if under_plugin_development is true
+      attr_accessor :use_size_metric
 
       def initialize
         super
+        @use_size_metric = true
         @counter_mutex = Mutex.new
         @buffering = false
         @delayed_commit = false
@@ -803,7 +805,7 @@ module Fluent
           process(tag, es)
           @counter_mutex.synchronize do
             @emit_records += es.size
-            @emit_size += es.to_msgpack_stream.bytesize
+            @emit_size += es.to_msgpack_stream.bytesize if @use_size_metric
           end
         rescue
           @counter_mutex.synchronize{ @num_errors += 1 }
@@ -972,7 +974,7 @@ module Fluent
         end
         @counter_mutex.synchronize do
           @emit_records += records
-          @emit_size += es.to_msgpack_stream.bytesize
+          @emit_size += es.to_msgpack_stream.bytesize if @use_size_metric
         end
         true
       end
@@ -992,7 +994,7 @@ module Fluent
         end
         @counter_mutex.synchronize do
           @emit_records += records
-          @emit_size += es.to_msgpack_stream.bytesize
+          @emit_size += es.to_msgpack_stream.bytesize if @use_size_metric
         end
         true
       end
@@ -1020,7 +1022,7 @@ module Fluent
         end
         @counter_mutex.synchronize do
           @emit_records += records
-          @emit_size += es.to_msgpack_stream.bytesize
+          @emit_size += es.to_msgpack_stream.bytesize if @use_size_metric
         end
         true
       end
```

```ruby
#!/usr/bin/env ruby

require 'benchmark/ips'

require_relative 'test/helper'
require 'fluent/event'

class DummyOutput < Fluent::Plugin::Output
  def process(tag, es)
    nil
  end
end

def create_output
  o = DummyOutput.new
  o.configure(config_element())
  o
end

o1 = create_output
o1.use_size_metric = false

o2 = create_output

t = event_time()
es = Fluent::ArrayEventStream.new([ [t, {"key" => "value1"}], [t, {"key" => "value2"}] ])

Benchmark.ips do |x|
  x.report("before") do
    o1.emit_events('tag', es)
  end

  x.report("after") do
    o2.emit_events('tag', es)
  end
end
```

Result:
```
              before   215.850k i/100ms
               after    18.082k i/100ms
              before      2.124M (± 0.5%) i/s -     10.792M in   5.082330s
               after    181.924k (± 0.7%) i/s -    922.182k in   5.069309s
```

It's about 10 times slower than before.
Of course, I think it's too malicious comparison because actual performance is determined by other various aspects and probably it's not bottleneck. Probably it's not so behind in actual cases.

But the performance regression actually exists.
Do you have any concerns about it?

cosmo0920: > This patch doesn't provide a way to disable it for output plugins (it can be disabled for input plugins though).

Oh, it's not good for us. I think that providing disabling to measure way on output plugin or system config is reasonable for our use case.

I'll implement disable event size measurement system config parameter.

cosmo0920: BTW, I don't have accurate benchmark yet, but this regression is more sensitive for high frequency loaded environment such as processing over 5000 ~ 6000 events per second case.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3411/commits/9b13ab283f9e1f3a7b457a0eaa613534f72ec8a0, https://github.com/fluent/fluentd/pull/3411/commits/e400ee6eb460d8f6153f6a207d446ba4c92d0a8c, and https://github.com/fluent/fluentd/pull/3411/commits/c73a83dd8555cbe61ea2604164cf0d3375338665.

I'd confirmed that record size metrics does not update when specifying `enable_size_metrics` as `false` with the following configuration:

```aconf
<system>
  enable_input_metrics true
  enable_size_metrics true
</system>
<source>
  @type monitor_agent
  bind 0.0.0.0
  port 24220
  tag monitor.agents
  emit_interval 5
</source>
<source>
  @type sample
  tag sample.raw
  rate 10
</source>

<match /(monitor|sample).*/>
  @type null
</match>
```

cosmo0920: Increasing what should be needed to specify enabling size of each record and number of records is no problem for our use case.

PR #3412, opened 2021-06-07
CI: Drop Ruby 2.5

ashie:
none

Ruby 2.5 is already EOL:
https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/

**Docs Changes**:
none

**Release Note**: 
none


cosmo0920 (approved):
Looks reasonable. Let's keep to be simpler CI tasks.

PR #3421, opened 2021-06-14
Remove duplicated license text

kenhys:

Fixes N/A

The license text is duplicated in service_discovery.rb

This is introduced by https://github.com/fluent/fluentd/commit/ce870c4a2f8af84d925c15a5c12b79e66e7e6546.
But I guess that it is not intended.

**Docs Changes**:

N/A

**Release Note**: 

N/A


ashie:
Good catch!

PR #3440, opened 2021-07-02
Implement metrics plugin mechanism

cosmo0920:

None.

It would nice to have configurable and flexible Fluentd's metrics.
Currently, Fluentd's metrics should be tightly coupled on its implementation.
But, in our usecase, Fluentd should load different plugin such as [fluent-plugin-metrics-cmetrics](https://github.com/calyptia/fluent-plugin-metrics-cmetrics) which handles [cmetrics](https://github.com/calyptia/cmetrics) context. 
Of course, [cmetrics Ruby bindings](https://github.com/calyptia/cmetrics-ruby) should be used in Fluentd metrics plugin.  

In the real metrics plugin usecase with metrics_local is: [cosmo0920/fluentd@implement-metrics-plugin branch](https://github.com/fluent/fluentd/compare/f73c601...cosmo0920:implement-metrics-plugin)

**Docs Changes**:
Needed.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/metrics.rb
+++ b/lib/fluent/plugin/metrics.rb
@@ -1,0 +1,119 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'socket'
+
+require 'fluent/plugin/base'
+
+require 'fluent/log'
+require 'fluent/unique_id'
+require 'fluent/plugin_id'
+
+module Fluent
+  module Plugin
+    class Metrics < Base
+      include PluginId
+      include PluginLoggerMixin
+      include UniqueId::Mixin
+
+      DEFAULT_TYPE = 'local'
+
+      configured_in :metrics
+
+      config_param :default_labels, :hash, default: {agent: "Fluentd", hostname: "#{Socket.gethostname}"}
+      config_param :labels, :hash, default: {}
+
+      attr_reader :use_gauge_metric
+      attr_reader :has_methods_for_gauge, :has_methods_for_counter
+
+      def initialize
+        super
+
+        @has_methods_for_counter = false
+        @has_methods_for_gauge = false
+        @use_gauge_metric = false
+      end
+
+      def configure(conf)
+        super
+
+        if use_gauge_metric
+          @has_methods_for_gauge = has_methods_for_gauge?
+        else
+          @has_methods_for_counter = has_methods_for_counter?
+        end
+      end
+
+      # Some metrics should be counted by gauge.
+      # ref: https://prometheus.io/docs/concepts/metric_types/#gauge
+      def use_gauge_metric=(use_gauge_metric=false)
+        @use_gauge_metric = use_gauge_metric
+      end
+
+      def create(namespace:, subsystem:,name:,help_text:,labels: {})
+        # This API is for cmetrics type.
+      end
+
+      def get(key)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def inc(key)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def dec(key)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def add(key, value)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def sub(key, value)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def set(key, value)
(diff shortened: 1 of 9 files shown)


lib/fluent/plugin/metrics.rb
```diff
@@ -0,0 +1,119 @@
...
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def inc(key)
+        raise NotImplementedError, "Implement this method in child class"
+      end
+
+      def dec(key)
+        raise NotImplementedError, "Implement this method in child class"
+      end
```
kenhys, line 80: mention which class must be implemented ("gauge" or "counter" ) seems better.

test/plugin/test_metrics.rb
```diff
@@ -0,0 +1,294 @@
...
+    end
+  end
+
+  sub_test_case 'BasicCounterMetric' do
+    setup do
+      @m = BasicCounterMetrics.new
+      @m.configure(config_element('metrics', '', {'@id' => '1'}))
+    end
+
+    test 'all operations work well' do
```
kenhys, line 195: all "counter" operations?

test/plugin/test_metrics.rb
```diff
@@ -0,0 +1,294 @@
...
+    end
+  end
+
+  sub_test_case 'AliasedCounterMetric' do
+    setup do
+      @m = AliasedCounterMetrics.new
+      @m.configure(config_element('metrics', '', {}))
+    end
+
+    test 'all operations work well' do
```
kenhys, line 222: maybe better mention ":set" because it is important in this test case.

test/plugin/test_metrics.rb
```diff
@@ -0,0 +1,294 @@
...
+
+  sub_test_case 'AliasedGaugeMetric' do
+    setup do
+      @m = AliasedGaugeMetrics.new
+      @m.use_gauge_metric = true
+      @m.configure(config_element('metrics', '', {}))
+    end
+
+    test 'all operations work well' do
+      assert_false @m.has_methods_for_counter
```
kenhys, line 277: maybe better to mention dec set,sub

lib/fluent/plugin/metrics.rb
cosmo0920: https://github.com/fluent/fluentd/pull/3440/files#diff-6826edd8ce4eb8c37724e029bb0fa852f74c76cd11a2a3e2e4ece1ea4761cee4R96-R105 and https://github.com/fluent/fluentd/pull/3440/files#diff-6826edd8ce4eb8c37724e029bb0fa852f74c76cd11a2a3e2e4ece1ea4761cee4R107-R116 mentions what is needed for counter or gauge? Are they not enough for you?

kenhys: For me, it is enough. I just think that it may be kind for metrics plugin developer/users.
Instead of "something method is not implemented", developer/users find "gauge metrics plugin must implement xxx but not it is not implemented".

Hmm, it may be redundant... :thinking:

cosmo0920: Yeah, It may be redundant. Because most of developers (except I was crazy someday 😛  Haha.) will check with try-and-error on developing plugins.

test/plugin/test_metrics.rb
cosmo0920: Like rspec context, "all operation work well" testcase should be interpret as `BasicCounterMetric` should `all operations work well`.
With your suggestion, it seems that it is redundant for "counter".
What do you think?

test/plugin/test_metrics.rb
cosmo0920: `#has_methods_for_counter` checks for `set` method definition and necessary methods.
Do you want to write down more descriptive comments or assertion, right?

test/plugin/test_metrics.rb
cosmo0920: #has_methods_for_gauge checks for set method definition and necessary methods.
Do you want to write down more descriptive comments or assertion, right?

kenhys: It is important whether the aliased method works as expected, so it may be better to focus on it.
another idea I mean:

```
test 'all common operations except aliased methods' do
...
end

test 'all aliased gauge operations' do
  assert_false @m.has_methods_for_counter
  ...
end
```

But it may be ok as is in the current way.

ashie:
I'm checking this and https://github.com/fluent/fluentd/compare/91b12e8...cosmo0920:implement-metrics-plugin
I'm not yet check the detail but I have no objection for the design for now.
It would be better that we can get benchmark data to check performance regression.

cosmo0920:
> I'm checking this and [91b12e8...cosmo0920:implement-metrics-plugin](https://github.com/fluent/fluentd/compare/91b12e8...cosmo0920:implement-metrics-plugin)
> I'm not yet check the detail but I have no objection for the design for now.
> It would be better that we can get benchmark data to check performance regression.

Thanks for the checking. Yeah, we should do it.
Sadly, we don't have any performance regression benchmark result yet.

cosmo0920:
Also, I'm planning to create prometheus node exporter plugin like as fluent-bit does:
https://github.com/fluent/fluent-bit/pull/3447
This exporter will use [cmetrics metrics plugin](https://github.com/calyptia/fluent-plugin-metrics-cmetrics).

ashie (approved):

PR #3451, opened 2021-07-08
Avoid to crash on outputting log at the early stage config parser

ashie:
Fixes https://github.com/fluent/fluentd-kubernetes-daemonset/issues/583

By #3352, an early stage config parser is introduced to enable setting
log rotation by system config. While this stage, the global logger
doesn't exist yet, so that it causes crash when the config parser try to
output log. This commit avoid the crash by checking $log. Same logs will
be output at the later stage so that it's not needed in the early stage.

See also: https://github.com/fluent/fluentd-kubernetes-daemonset/issues/583

**Docs Changes**:
None

**Release Note**: 
Same with the title

### DIFF
--- a/lib/fluent/config.rb
+++ b/lib/fluent/config.rb
@@ -62,7 +62,7 @@ module Fluent
         Parser.parse(str, fname, basepath)
       when :ruby
         require 'fluent/config/dsl'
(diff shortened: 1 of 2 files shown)


kenhys (approved):
Oh, I've overlooked such a situation.

LGTM.

ashie:
I've fount one more case that `$log` is used while parsing a config file, it's rarely used though.

Issue #3433, opened 2021-06-23
in_tail pos file compaction doesn't seem to work as expected


alex-vmw:
**Describe the bug**
in_tail pos file compaction doesn't seem to work as expected. The expectation is that compaction will remove entries for deleted log files from the pos file, but we are not seeing that happen. We see in the logs that compaction was triggered, but when we look into pos file the entries for deleted log files still exist.

For example, we see that compaction took place via this log:
```
2021-06-23 21:55:07 +0000 [info]: #0 [in_tail_container_logs] Clean up the pos file
```
However, when we look at the pos file on the node, it contains entries for logs files that we know no longer exist. For example, we see these lines in the pos file:
```
/var/log/containers/alex-busybox_default_alex-busybox-0584af93443bad99c0b64bf291190609f5d22557c7be937f36c9af2dde9ce54b.log      000000000001e6c3        0000000018776252
/var/log/containers/alex-busybox_default_alex-busybox-d8180a67753ee1ab99c49d028a1f58676c211f233b74934930aecfeef54e7d0d.log      000000000000134f        000000001012930f
/var/log/containers/alex-busybox_default_alex-busybox-a01184a2cb8064d7c776488188c1d1bbd5bf57c509ea3024da9dbee7fabc7c38.log      00000000000046b0        0000000018605f76
```
But these logs do NOT exist on the node anymore as corresponding pods have long been deleted:
```
# ls -l /var/log/containers/alex-busybox_default_alex-busybox-0584af93443bad99c0b64bf291190609f5d22557c7be937f36c9af2dde9ce54b.log
ls: cannot access '/var/log/containers/alex-busybox_default_alex-busybox-0584af93443bad99c0b64bf291190609f5d22557c7be937f36c9af2dde9ce54b.log': No such file or directory
# ls -l var/log/containers/alex-busybox_default_alex-busybox-d8180a67753ee1ab99c49d028a1f58676c211f233b74934930aecfeef54e7d0d.log
ls: cannot access 'var/log/containers/alex-busybox_default_alex-busybox-d8180a67753ee1ab99c49d028a1f58676c211f233b74934930aecfeef54e7d0d.log': No such file or directory
# ls -l /var/log/containers/alex-busybox_default_alex-busybox-a01184a2cb8064d7c776488188c1d1bbd5bf57c509ea3024da9dbee7fabc7c38.log
[...]

PR #3459, opened 2021-07-15
Fix some pos_file bugs

ashie:
None

I found these bugs while addressing #3433, but fix for #3433 isn't included in this pull request yet.

This pull request fixes following two bugs:

* When both `follow_inodes` and `pos_file_compaction_interval` are specified, file entries that didn't exist on start up will be deleted unexpectedly when compaction is triggered.
* `String#bytesize` should be used instead of `String#size` for path length.

**Docs Changes**:
None

**Release Note**: 
TBD


kenhys (approved):
LGTM.

cosmo0920 (approved):
:+1:

PR #3460, opened 2021-07-16
WindowsFile: Add testing for DeletePending

fujimotos:

Ammend #3457

I think I found a way to add testing for the changeset f61c08ca.

 1. Create a test file with FILE_SHARE_DELETE mode.
 2. Call `DeleteFile()` on that file.
 3. Call `WindowsFile.stat()` on that file.
 4. Check if ENOENT is raised after step 3.

Confirmed to work fine on GitHub Action:

    WindowsFile exceptions:
     test: ERROR_SHARING_VIOLATION raised:        .: (0.001220)
     test: Errno::ENOENT raised:                  .: (0.000633)
     test: Errno::ENOENT raised on DeletePending: .: (0.000717)
     test: nothing raised:                        .: (0.000835)

**Docs Changes**:

Not required.

**Release Note**: 

Not required.

### DIFF
--- a/test/plugin/test_file_wrapper.rb
+++ b/test/plugin/test_file_wrapper.rb
@@ -96,6 +96,17 @@ class FileWrapperTest < Test::Unit::TestCase
       end
     end
 
+    test 'Errno::ENOENT raised on DeletePending' do
+      path = "#{TMP_DIR}/deletepending.txt"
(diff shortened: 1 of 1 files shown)


ashie:
I confirmed that `delete_pending` returns true on this case :+1:

PR #3473, opened 2021-07-28
metrics: metrics_local: Remove key argument to optimize performance

cosmo0920:

Folows up #3440 

This is because our use cases doesn't need key for distinguish plugins
and agents.
One of the use cases to specify plugin_id calling has a possibility to become the bottleneck for performance.
The reason is Kernel#instance_variable_defined? method is relatively
heavy operation.

**Docs Changes**:

No needed.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/metrics.rb
+++ b/lib/fluent/plugin/metrics.rb
@@ -67,27 +67,27 @@ module Fluent
         # This API is for cmetrics type.
       end
 
-      def get(key)
+      def get
         raise NotImplementedError, "Implement this method in child class"
       end
 
-      def inc(key)
+      def inc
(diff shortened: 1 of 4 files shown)


kenhys:
I was just curious about it, how much performance is improved by this PR?

kenhys (approved):

cosmo0920:
> I was just curious about it, how much performance is improved by this PR?

For synchronous output plugin case, 2x faster than the previous implementation.
For buffered/asynchronous buffered case, 10% faster at the maximum which depends on batch operation frequency, I think.

ashie (approved):

Issue #903, opened 2016-04-18
in_tail: throws and exception on logrotation


giner:
**Problem**
If td-agent is not running as root and in_tail plugin is in use then it throws and exception on log rotation (if create option is in use) from time to time.

```
2016-04-15 13:00:32 +0000 [error]: Permission denied - /var/log/nginx/nginx.log
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:364:in `initialize'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:364:in `open'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:364:in `on_notify'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:151:in `on_notify'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:258:in `call'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:258:in `on_change'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/cool.io-1.1.1/lib/cool.io/loop.rb:96:in `run_once'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/cool.io-1.1.1/lib/cool.io/loop.rb:96:in `run'
  2016-04-15 13:00:32 +0000 [error]: /usr/lib/fluent/ruby/lib/ruby/gems/1.9.1/gems/fluentd-0.10.38/lib/fluent/plugin/in_tail.rb:81:in `run'
```

[...]

PR #3477, opened 2021-07-30
in_tail: Fix a regression of retrying unaccessible files

ashie:
None.
Follow up of #3378

When a fluentd proccess doesn't have permision to read a file, it should
retry to open it again. But the current implementation doesn't do it due
to a bug introduced at #3378. This pull request intend to rescue EACCESS
on stat(2) but it also rescue it on open(2) wrongly.

**Docs Changes**:
None

**Release Note**: 
Same with the title.


ashie:
Tests for it should be added.

ashie:
In addition, it's better to decrease frequency of retrying: #903

ashie:
Hmm, it's not regression, it retries on every tail watcher timer (1 sec). 
It's more desired behaviour than before as mentioned in: #903

The only thing we should do is adding warning message for it. 
I'll do it in anther pull request.

PR #3478, opened 2021-07-30
in_tail: Revive a warning message of retrying unaccecible file

ashie:
None.
Follow up of #3378

When a fluentd procces doesn't have permission to read a file, it should
retries to open it with a warning message. But the current
implementation suppresses the message due to #3378.
This patch revive the warning message.

**Docs Changes**:
None.

**Release Note**: 
Same with the title.


cosmo0920 (approved):
Looks good.

PR #3479, opened 2021-07-30
input: filter: output: bare_output: multi_output: Use metrics plugin mechanism on plugin base classes

cosmo0920:
None

This is the last PR series for implementing metrics plugin mechanism.
With these changes, our cmetrics metrics plugin can work with Fluentd and Fluentd can handle native metrics to send prometheus via fluent-plugin-metrics-cmetrics which binds [cmetrics library](https://github.com/calyptia/cmetrics).

Performance degradation will be prevented with https://github.com/fluent/fluentd/commit/97486d1170f75814c2e92779cb9ca22664379c6e and https://github.com/fluent/fluentd/commit/e4995895574ca48d1f75b324b50dde8c0e1b7ede. Locking with Monitor is heavy operation when it needn't to use due to disable size metrics on each plugin base classes. 

**Docs Changes**:
Needed.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/compat/output.rb
+++ b/lib/fluent/compat/output.rb
@@ -310,7 +310,7 @@ module Fluent
       # original implementation of v0.12 BufferedOutput
       def emit(tag, es, chain, key="")
         # this method will not be used except for the case that plugin calls super
-        @emit_count += 1
+        @emit_count_metrics.inc
         data = format_stream(tag, es)
         if @buffer.emit(key, data, chain)
           submit_flush
@@ -337,14 +337,14 @@ module Fluent
       # because v0.12 BufferedOutput may overrides #format_stream, but original #handle_stream_simple method doesn't consider about it
       def handle_stream_simple(tag, es, enqueue: false)
         if @overrides_emit
-          current_emit_count = @emit_count
+          current_emit_count = @emit_count_metrics.get
           size = es.size
           key = data = nil
           begin
             emit(tag, es, NULL_OUTPUT_CHAIN)
             key, data = self.last_emit_via_buffer
           ensure
-            @emit_count = current_emit_count
+            @emit_count_metrics.set(current_emit_count)
             self.last_emit_via_buffer = nil
           end
           # on-the-fly key assignment can be done, and it's not configurable if Plugin#emit does it dynamically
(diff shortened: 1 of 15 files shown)


cosmo0920:
Hmm, all CI tasks are failed. We should check them.

ashie:
> Hmm, all CI tasks are failed. We should check them.

I fixed it.

lib/fluent/compat/output.rb
```diff
@@ -337,22 +337,29 @@ def format_stream(tag, es)
...
             self.last_emit_via_buffer = nil
           end
           # on-the-fly key assignment can be done, and it's not configurable if Plugin#emit does it dynamically
           meta = @buffer.metadata(variables: (key && !key.empty? ? {key: key} : nil))
           write_guard do
             @buffer.write({meta => data}, format: ->(_data){ _data }, size: ->(){ size }, enqueue: enqueue)
           end
-          @counter_mutex.synchronize{ @emit_records += size }
+          if @enable_size_metrics
+            @counter_mutex.synchronize do
```
ashie, line 356: I think `@counter_mutex` isn't needed any more because each `Metrics` object have a mutex lock.

Generally speaking, wrapping into same lock is desired as mentioned in https://github.com/fluent/fluentd/pull/3411/files/97d1692c05538de71b285fbaa374ff38ec443aa3#r647202429.
But keeping consistency of both value and using 2 locks aren't so meaningful in this case.
I think keeping the code simple is more important in this case.

```ruby
@emit_records_metrics.add(es.size)
@emit_size_metrics.add(es.to_msgpack_stream.bytesize) if @enable_size_metrics
```

ashie (changes requested):

lib/fluent/compat/output.rb
cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3479/commits/aa79a4cd16bdab3190dd8fe8f5e2dc92a69c72ea.

ashie (approved):

cosmo0920:
I found that we should do additional work for prometheus plugin. I'll send another PR/

cosmo0920:
I'd confirmed `that,` in Fluentd side, there is no additional work is needed but prometheus plugin should follow this change.

PR #3481, opened 2021-08-03
in_http: Enable to add Access-Control-Allow-Credentials header

### LINKED ISSUE #3422: http plugin set Access-Control-Allow-Credentials header

I'm trying to use the [http](https://docs.fluentd.org/input/http) input plugin to collect logs from a client web app.  I want to do my best to send any remaining logs from the client when the window is closed, so am attempting to use the [sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) api to do so:

```javascript
  let headers = {
    type: 'application/json'
  };
  let blob = new Blob([JSON.stringify(logs)], headers);
  window.navigator.sendBeacon(logUrl, blob)
```
unfortunately, it sets `credentials: "include"`  (which is not customizable) on the request, and I get an error on the console:
`Access to resource at 'https://app.westus2.stage.stage.hoy.lu/logs/login.app' from origin 'http://localhost:3002' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'.`

I believe all that I need is a way to have fluentd set the `Access-Control-Allow-Credentials` header to 'true' on the response and everything will work as expected.

ashie:
Fixes #3422

To support [Beacon API](https://www.w3.org/TR/beacon/), setting [Access-Control-Allow-Credentials](https://fetch.spec.whatwg.org/#http-access-control-allow-credentials) header is required because it always set [`credentials: include`](https://fetch.spec.whatwg.org/#concept-request-credentials-mode): https://www.w3.org/TR/beacon/#sec-processing-model

**Docs Changes**:
Add `cors_allow_credentials` to https://docs.fluentd.org/input/http

**Release Note**: 
Same with the title

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -74,6 +74,8 @@ module Fluent::Plugin
     config_param :blocking_timeout, :time, default: 0.5
     desc 'Set a allow list of domains that can do CORS (Cross-Origin Resource Sharing)'
     config_param :cors_allow_origins, :array, default: nil
+    desc 'Tells browsers whether to expose the response to frontend when the credentials mode is "include".'
+    config_param :cors_allow_credentials, :bool, default: nil
     desc 'Respond with empty gif image of 1x1 pixel.'
     config_param :respond_with_empty_img, :bool, default: false
     desc 'Respond status code with 204.'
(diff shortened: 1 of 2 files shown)


ashie:
Failed tests on Windows aren't related with this pull request, they are caused by #3490

cosmo0920 (approved):
Looks reasonable for me.

kenhys (approved):
LGTM

ashie:
Thanks!

PR #3484, opened 2021-08-05
buffer: Use metrics plugin mechanism on a plugin base class

cosmo0920:
None

We should add more metrics plugin supported plugin base classes.
With these changes, our cmetrics metrics plugin also can work with Fluentd and Fluentd can handle native metrics to send prometheus format encodable metrics via fluent-plugin-metrics-cmetrics which binds [cmetrics library](https://github.com/calyptia/cmetrics).

Almost metrics plugin support will be handled in `#statistics` methods. So, it won't be caused significant performance degradation.

**Docs Changes**:
Needed.

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin_helper/metrics.rb
+++ b/lib/fluent/plugin_helper/metrics.rb
@@ -42,7 +42,11 @@ module Fluent
         @plugin_type_or_id = if self.plugin_id_configured?
                                self.plugin_id
                              else
-                               "#{conf["@type"] || conf["type"]}.#{self.plugin_id}"
+                               if type = (conf["@type"] || conf["type"])
+                                 "#{type}.#{self.plugin_id}"
+                               else
+                                 "#{self.class.to_s.split("::").last.downcase}.#{self.plugin_id}"
+                               end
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin_helper/metrics.rb
```diff
@@ -42,7 +42,11 @@ def configure(conf)
         @plugin_type_or_id = if self.plugin_id_configured?
                                self.plugin_id
                              else
-                               "#{conf["@type"] || conf["type"]}.#{self.plugin_id}"
+                               if type = (conf["@type"] || conf["type"])
+                                 "#{type}.#{self.plugin_id}"
+                               else
+                                 "#{self.class.to_s.split("::").last.downcase}.#{self.plugin_id}"
```
cosmo0920, line 48: This is for `out_file`'s default buffer plugin settings:
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_file.rb#L61-L65

ashie (approved):

PR #3489, opened 2021-08-18
in_tail: Simplify TargetInfo related code

ashie:
None

Since TargetInfo related code is too confusing, we introduced some fatal bugs in the past.
We should refactor it.

* Using `TargetInfo` as hash table key is confusing, only path is used in actual.
* Simlify error handling.

**Docs Changes**:
None

**Release Note**: 
Same with the title


ashie:
I'm planning to merge this after #3535

cosmo0920 (approved):
Looks attractive! Thanks for the simplification of TargetInfo implementation.

ashie:
Thanks for your review :+1:

PR #3491, opened 2021-08-24
in_http: Fix default value of cors_allow_credentials

ashie:
Follow up for #3481

There is no reason I chose the default value `nil` for `cors_allow_credentials` at #3481.
It should be `false`.

**Docs Changes**:
none

**Release Note**: 
none


kenhys (approved):
I've overlooked it.

LGTM.

PR #3503, opened 2021-09-12
out_forward: Include node statictics information correctly

cosmo0920:

None.

In monitor_agent, it assumes output plugin metrics are put in under `output` branch.
With current implementation, they will be ignored on retrieving with https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_monitor_agent.rb#L338

**Docs Changes**:
No needed.

**Release Note**: 
Same as title.


ashie (approved):

ashie:
Good catch!

Issue #3508, opened 2021-09-15
ignore_if_prev_successes not work


hoanpv3993:
### Describe the bug

Hi all,
I want use output copy with ignore_if_prev_successes but it not work

### To Reproduce

- I config 2 plugin output : forward and file
- Then I stop service of 2 agg node to set output forward fail

### Expected behavior

- when plugin forward fail, i can't see log file is created by plugin 2

### Your Environment

```markdown
- Fluentd version: 1.13.3
- TD Agent version: 4.2.0
- Operating system: Ubuntu 20.04LTS
- Kernel version: 5.8.0-1038-gcp
```

### Your Configuration
```
<source>
  @type tail
  path /var/log/nginx/test.demo.com.access.log
  pos_file /var/log/td-agent/nginx-access.log.pos
  refresh_interval 2
  tag apache.access
  <parse>
    @type json
  </parse>
</source>

<filter **.**>
  @type prometheus
  <metric>
    name fluentd_input_status_num_records_total
    type counter
    desc The total number of incoming records
    <labels>
      tag ${tag}
      hostname ${hostname}
    </labels>
  </metric>
</filter>

<match apache.access>
  @type copy
  <store ignore_error>
    @type forward
    send_timeout 10s
    recover_wait 3s
    hard_timeout 10s
    <buffer tag>
      @type file
      path /var/log/td-agent/buffer/
      flush_mode interval
      flush_interval 2s
      retry_max_times 3
    </buffer>

    <server>
      name agg-01
      host agg-01
      port 24224
      weight 50
    </server>
    <server>
      name agg-02
      host agg-02
      port 24224
      weight 50
    </server>
  </store>

  <store ignore_if_prev_success ignore_error>
    @type file
    path /var/log/td-agent/backup-store-%Y%m%d
    append true
    <buffer>
      @type memory
      flush_mode interval
      flush_interval 2s
    </buffer>
  </store>

  <store>
    @type prometheus
    <metric>
      name fluentd_output_status_num_records_total
      type counter
      desc The total number of outgoing records
      <labels>
        tag ${tag}
        hostname ${hostname}
      </labels>
    </metric>
  </store>
</match>
```
### Your Error Log

```shell
[...]

PR #3515, opened 2021-09-24
out_copy: Suppress a wrong warning for ignore_if_prev_success

ashie:
None.
I found the bug while checking #3508

out_copy outputs a following warning even if a user set `ignore_if_prev_success`:

```
2021-09-24 14:10:54 +0900 [warn]: #0 ignore_errors are specified in all <store>, but ignore_if_prev_success is not specified. Is this intended?
```

This commit fixes it.

**Docs Changes**:
None

**Release Note**: 
Same with the title.


cosmo0920 (approved):
:+1:

Issue #3521, opened 2021-09-30
`FileUtils.mkdir_p` function call inside `backup_chunk` does not copy file permissions (e.g. sticky bit) into middle directories in the path. 


Simon-Janos:
### Is your feature request related to a problem? Please describe.

The issue there is that when fluentd fails, it creates the bad chunks in deep /tmp/../../ directory structure via 
 `FileUtils.mkdir_p` function call inside `backup_chunk` which seems to be just a system call to the underlying linux “mkdir -p -m /tmp/..” command.
-m, --mode=MODE set file mode (as in chmod), not a=rwx – umask
-p, --parents no error if existing, make parent directories as needed

However when used together -m is only applies at the lowest/deepest directory of the specified directory structure not copying the sticky bit of /tmp dricetory to the middle directories.
This is how Linux works, but this is a CIS vulnerability which is pointed out by vulnerability scan.

### Describe the solution you'd like

In order to resolve this vulnerability, fluentd have to be changed to set -m parameters to all of the members of the directory tree somehow.
https://gitlab.cncf.ci/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L1251: 
```ruby
FileUtils.mkdir_p(backup_dir, mode: system_config.dir_permission || 0755) unless Dir.exist?(backup_dir)
```
Solution could be to change mkdir_p to some custom function which iterates through the specified directory struct applying -m (mode).

### Describe alternatives you've considered

I do not see.

kenhys:
Thanks,
@Simon-Janos 
Do you know whether this issue is assigned to CVE-*?

kenhys:
To be clear, if `dir_permission` was specified (like the above example), is it not vulnerable?

Simon-Janos:
> Thanks, @Simon-Janos Do you know whether this issue is assigned to CVE-*?

Not to my knowledge.

Simon-Janos:
> To be clear, if `dir_permission` was specified (like the above example), is it not vulnerable?

Yes the problem is that you specify a  `dir_permission` (e.g. with sticky bit inside) but it does not apply to the whole directory structure of the specified path only to the deepest ones (and sometimes but it is to be tested not even to the deepest one), This is how Linux works, but it is not a desired behaviour.

Simon-Janos:
Any update?

PR #3522, opened 2021-10-01
Use defined constant for consistency

kenhys:

Fixes #

The default value of dir or file permission is defined in fluent/env.rb.
So it should be used everywhere.

I've found this issue when checking #3521.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/storage_local.rb
+++ b/lib/fluent/plugin/storage_local.rb
@@ -14,6 +14,7 @@
 #    limitations under the License.
 #
 
+require 'fluent/env'
 require 'fluent/plugin'
 require 'fluent/plugin/storage'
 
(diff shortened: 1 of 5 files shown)


ashie (approved):
LGTM

storage_local.rb should also use them?

https://github.com/fluent/fluentd/blob/da65272e498b4d7c32f53ace003bb659115b63f7/lib/fluent/plugin/storage_local.rb#L28-L37

kenhys:
It should be, I'll fix it too.

kenhys:
Fixed.

cosmo0920 (approved):

lib/fluent/plugin/storage_local.rb
```diff
@@ -29,10 +30,10 @@ class LocalStorage < Storage
       DEFAULT_FILE_MODE = 0644
```
ashie, line 30: `DEFAULT_FILE_MODE` and `DEFAULT_DIR_MODE` aren't used anymore, right?

ashie (changes requested):

lib/fluent/plugin/storage_local.rb
kenhys: Yes, I've removed it.

kenhys:
Removed needless constant definition.

ashie (approved):

PR #3523, opened 2021-10-01
Ensure to specify dir permission

### LINKED ISSUE #3521: `FileUtils.mkdir_p` function call inside `backup_chunk` does not copy file permissions (e.g. sticky bit) into middle directories in the path.

kenhys:

Fixes #3521

FileUtils.mkdir_p is not used with mode: so even though
system_config.dir_permission is specified, it was not used at all.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/command/ca_generate.rb
+++ b/lib/fluent/command/ca_generate.rb
@@ -2,6 +2,7 @@ require 'openssl'
 require 'optparse'
 require 'fileutils'
 require 'fluent/version'
+require 'fluent/env'
 
 module Fluent
   class CaGenerate
@@ -40,7 +41,7 @@ HELP
         exit 1
       end
 
-      FileUtils.mkdir_p(ca_dir)
+      FileUtils.mkdir_p(ca_dir, mode: Fluent::DEFAULT_DIR_PERMISSION)
 
       cert, key = Fluent::CaGenerate.generate_ca_pair(@options)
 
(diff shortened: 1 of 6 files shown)


kenhys:
It seems that `mock.proxy(File).chmod(0o777, TMP_DIR).never` raise exception.
It is cased by `FileUtils.mkdir_p(, mode: xxx)`, it is okay if `FileUtils.mkdir_p(path)` or FileUtils.mkdir_p(, mode: nil)` :thinking:

kenhys:
Test case also should be changed, done.

PR #3526, opened 2021-10-06
Detect wrong plugin naming which contains extra kind prefix

kenhys:

Fixes N/A

Bit related issue: https://github.com/fluent-plugins-nursery/fluent-package-builder/issues/334

As a plugin naming scheme, register_xxx should not be pre-fixed
with kind.

For examples, input plugin in_xxx.rb, @type xxx is specified in <source>
section. output plugin out_xxx.rb, @type xxx is specified in
<match> section.

There is a well known failure case, plugin doesn't detected
because of not following above plugin naming rule.

In this PR, it checks whether prefixed plugin name is specified, then
it raise an error with more informative message.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/registry.rb
+++ b/lib/fluent/registry.rb
@@ -15,6 +15,7 @@
 #
 
 require 'fluent/config/error'
+require 'fluent/validator/plugin_name'
 
 module Fluent
   class Registry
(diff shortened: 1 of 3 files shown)


kenhys:
Even though launch with --gemfile, it cause an error with "  error: You seem to use `bundle exec` already."
It seems that BUNDLE_BIN_PATH checked and exit there.

ashie:
I close this pull request that has been pending for a long time and has no solution for now, and not so serious even if it's not resolved.

PR #3533, opened 2021-10-13
Issue Template: Render configuration

ashie:
None

Most users doesn't quote `Your Configuration` field manually when they report a bug.
So it should be quoted automatically.

**Docs Changes**:
None

**Release Note**: 
None


ashie:
Available Languages:
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

PR #3535, opened 2021-10-20
Add log throttling in files based on group rules

Pranjal-Gupta2:
None

Fluentd running in a big cluster with a high volume of logs has to deal with a lot of applications logging at different rates. High log generation rates for low priority applications can lead to log loss for high priority applications. We need to have a data flow control in the log collector so that we can limit the log collection rate of applications based on the priority of application workloads in the cluster. 

Tail Plugin is watching a file and reads logs till it reaches the end of the file. Earlier accepted PR #3185 introduced throttling at the source level which applies a limit to the number of bytes collected by the in_tail plugin every second. However, it uniformly applies the same limit to every log file.

This PR enables us to form different groups based on configuration rules provided by the user. Within each group, one can set the rate limits for log collection. The set rate limit gets uniformly applied to each member (log file) of the group. If the group line limit of the log is reached within the `rate_period` time interval, Fluentd stops reading logs from that group. Once, `rate_period` (time) is over, Fluentd resets its group counters and restarts the reading process. 

I strongly believe group level rate limiting can really help one to control low priority applications vs high priority applications log flow. Currently based on the namespace and app name (application name) extracted from file names, one can form groups for rate limiting.

**Docs Changes**:
[fluent/fluentd-docs-gitbook#376](https://github.com/fluent/fluentd-docs-gitbook/pull/376)

**Release Note**: 
Same as Title


pmoogi-redhat:
@ashie can you further guide on this PR merging  ? please note both in_tail and in_tail_with_throttle can co-exists and one can use both of them as set of <source>.  we didn't want to disturb the base line of in_tail hence we created a new plugin with this group based rate limiting functionality.  let us know what step we need to take next.

ashie:
My additional impressions:

* An equivalent feature might be able to realized by multiple `<source>` and adding regexp based `path` parameter to in_tail. Could you describe the strength of your patch compared with other ideas?
  * I think one of the strength of your idea is fallback mechanism, it makes easy to configure unmatched paths.
* Again, implementing as an additional plugin isn't good idea, please consider integrating it into in_tail if you want to merge it into fluentd.
  * If you can integrate it cleanly, we might not need waiting refactoring.
* The config format seems too rely on k8s's concept. It would be nice if you can generalize it.
* We don't want maintaining Copy & Pasted codes. For example `detach_watcher_after_rotate_wait` and `TailWatcher::IOHandler#rate_limit_handle_notify` are almost same with parent's one. When parent one is modified, we have to modify both codes. Please consider integrating them.
* Although in_tail_with_throttle might change the behaviour of in_tail, in_tail's tests aren't executed against in_tail_with_throttle.

Pranjal-Gupta2:
> My additional impressions:
> 
> * Equivalent feature might be able to realized by multiple `<source>` and adding regexp based `path` parameter to in_tail. Could you describe the advantage of your patch compared with other ideas?
>   
>   * I think one of the strength of your idea is fallback mechanism, it makes easy to configure unmatched paths.

@ashie Our plugin has the following advantages : 

1. Using multiple threads -> context switching and cpu usage

Adding regexp based path parameter to in_tail will add multiple source plugins which will create additional threads for reading from files. Hence cpu cycles will be wasted in context switching.

2. Paths can be overlapped when using multiple source plugins

in_tail_with_throttle allows you to have generic and context-specifc rules for log rate limit. It will be difficult to manage with in_tail's path parameter based grouping and can lead to files falling In more than one group.

For example, consider the following requirement: 
	1. Rate limit all files with a particular namespace (Generic rule).
	2. Give priority to specific app name in that namespace with a different limit (Context Specific Rule)

	Configuration for such requirement will be like: 
	<group_rule>
		...
		<rule>
			## rule1
			namespace ABC
			limit 1000 ## Generic rules for all files in a namespace
		</rule>	
		<rule>
			## rule2
			namespace ABC
			appname DEF  ## Specific rules with adjustable limit
			limit 1000
		</rule>	
		...
	</group_rule>

in_tail_with_throttle will add a file with DEF app name and ABC namespace to rule 2 (Higher precedence over rule1). Using in_tail's path parameter will add the file in two groups and hence repeated log lines will be collected.

3. Our plugin will take care of unexpected files.

In cases where we forget to add rules for some files, in_tail_with_throttle will add them to the default group and still collect logs. in_tail plugin with regex based path parameter grouping will fail in these situations.

4. Config will get pretty cluttered.

Adding multiple source configs for in_tail's path based grouping will make your configuration file big in size. It will be difficult to manage and update configurations in case of very large number of groups.

Pranjal-Gupta2:
> My additional impressions:
> 
> * The config format seems too rely on k8s's concept. It would be nice if you can generalize it.

@ashie can you please elaborate on what this generalisation means? Should we replace our namespace and app-name fields with something else? 

For example,
```
<rule> 
    ## level1 instead of namespace
    level1 ABC

    ## level2 instead of appname
    level2 DEF

    limit 1000
</rule>
```

Pranjal-Gupta2:
@ashie Could you please comment?

agup006:
@cosmo0920 would you also be able to take a look?

lib/fluent/plugin/in_tail.rb
```diff
@@ -548,7 +636,7 @@ def detach_watcher_after_rotate_wait(tw, ino)
       if @open_on_every_update
         # Detach now because it's already closed, waiting it doesn't make sense.
         detach_watcher(tw, ino)
-      elsif @read_bytes_limit_per_second < 0
+      elsif @read_bytes_limit_per_second < 0 || (!tw.group_watcher.nil? && tw.group_watcher.limit <= 0)
```
cosmo0920, line 639: Could you extract this condition as a method for readability and testability?

lib/fluent/plugin/in_tail.rb
```diff
@@ -775,6 +936,11 @@ def initialize(target_info, pe, log, read_from_head, follow_inodes, update_watch
       attr_reader :line_buffer_timer_flusher
       attr_accessor :unwatched  # This is used for removing position entry from PositionFile
       attr_reader :watchers
+      attr_accessor :group_watcher
+
+      def group_watcher=(group_watcher)
```
cosmo0920, line 941: `attr_accessor :group_watcher` creates `group_watcher=` method.
What is the expected behavior for this method?

lib/fluent/plugin/in_tail.rb
```diff
@@ -1054,6 +1224,7 @@ def should_shutdown_now?
 
         def handle_notify
           return if limit_bytes_per_second_reached?
+          return if !group_watcher.nil? && group_watcher.limit_lines_reached?(@path)
```
cosmo0920, line 1227: We can use `&.` operator as nil guard:

```ruby
          return if group_watcher&.limit_lines_reached?(@path)
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -753,6 +841,79 @@ def on_timer
...
+          false
+        end
+      end
+
+      def to_s
+        super + " current_paths: #{@current_paths} rate_period: #{@rate_period} limit: #{@limit}"
+      end
+    end
+
+
```
cosmo0920, line 916: Please remove a redundant newline.

lib/fluent/plugin/in_tail.rb
```diff
@@ -1063,12 +1234,19 @@ def handle_notify
...
+                    group_watcher.update_reading_time(@path) unless group_watcher.nil?
                     data = io.readpartial(BYTES_TO_READ, @iobuf)
                     @eof = false
                     @number_bytes_read += data.bytesize
                     @fifo << data
+                    group_watcher.update_lines_read(@path, -@lines.size) unless group_watcher.nil?
                     @fifo.read_lines(@lines)
+                    group_watcher.update_lines_read(@path, @lines.size) unless group_watcher.nil?
 
+                    if !group_watcher.nil? && group_watcher.limit_lines_reached?(@path) || should_shutdown_now?
```
cosmo0920, line 1246: Can we extract this condition as a method for simplicity and readability?

cosmo0920 (changes requested):
I reviewed and added a few concerns. Could you take a look about them?

lib/fluent/plugin/in_tail.rb
```diff
@@ -120,7 +126,25 @@ def initialize
...
+                   :regexp, 
+                   default: /var\/log\/containers\/(?<appname>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\/[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container>.+)-(?<docker_id>[a-z0-9]{64})\.log$/
+      desc 'Period of time in which the group_line_limit is applied'
+      config_param :rate_period, :time, default: 5
+
+      config_section :rule, multi: true, required: true do
+        desc 'Namespace key'
+        config_param :namespace, :array, value_type: :string, default: [DEFAULT_NAMESPACE]
+        desc 'App name key'
+        config_param :appname, :array, value_type: :string, default: [DEFAULT_APPNAME]
```
ashie, line 141: I'm considering how to generalize this part.
As I mentioned before, I think this part has following 2 issues:

* Depends on k8s's concept
* config_param `appname` and `namespace` are hardcoded while `pattern` can be modified by user

The former one is acceptable for me since it's only a default value, it's customizable (and this default value is appropriate since k8s is most major application recent days).
But the later one is a bit uncomfortable for me.

One idea I'm considering is extracting named caputures from `pattern` (by `pattern.named_captures`) and generate `config_param` dynamically. But it might too complicate.

Do you have any other idea to resolve it? @Pranjal-Gupta2 @cosmo0920

ashie (commented):
Thanks for revising the patch, it's much better than before!
I'm sorry for my late response, I was thinking how to resolve remaining issues.
I commented about it above, please check it.

lib/fluent/plugin/in_tail.rb
```diff
@@ -120,7 +126,25 @@ def initialize
       config_argument :usage, :string, default: 'in_tail_parser'
     end
 
-    attr_reader :paths
+    config_section :group, param_name: :group, required: false, multi: false do
+      desc 'Regex for extracting group\'s metadata'
+      config_param :pattern, 
+                   :regexp, 
+                   default: /var\/log\/containers\/(?<appname>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\/[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container>.+)-(?<docker_id>[a-z0-9]{64})\.log$/
```
ashie, line 133: Doesn't it need `^` and '/' prefix for this regexp?

lib/fluent/plugin/in_tail.rb
ashie: Or using hash & array param might be enough for this case

Pranjal-Gupta2: Thanks for your suggestions @ashie! 
I was also thinking in the direction of generating `config_param` dynamically. 

```ruby
      config_section :rule, multi: true, required: true do 
        Regexp.compile(conf.elements('group')[0]['pattern']).named_captures.keys.each do |key|
          config_param key.to_sym, :array, value_type: :string, default: [DEFAULT_APPNAME]
        end
        desc 'Maximum number of log lines allowed per group over a period of rate_period'
        config_param :limit, :integer, default: DEFAULT_LIMIT
      end
```

I added this snippet in the `configure` method just before calling `super` ([in_tail.rb#L163](https://github.com/fluent/fluentd/blob/514126ca4329c81739925aa6bd140fabe87faaf2/lib/fluent/plugin/in_tail.rb#L163)). However, I am getting this error: 
```bash
/.rvm/gems/ruby-3.0.0/gems/fluentd-1.14.4/lib/fluent/plugin/in_tail.rb:160:in `configure': undefined method `config_section' for #<Fluent::Plugin::TailInput:000000000010b8> (NoMethodError)
Did you mean?  configured_section_create
```

If we can call `config_section` and `config_param` in the `configure` method then I believe we could generate dynamic parameters for our plugin. What do you think? @ashie @cosmo0920

ashie: Now I think generating `config_param` dynamically is too complicate.
How about using `:hash` type like this?:

```ruby
config_param :match, :hash, default: {namespace: [DEFAULT_NAMESPACE], appname: [DEFAULT_APPNAME]}
```

usage:
```xml
<rule>
  match {
    "namespace": ["aaa", "bbb"],
    "appname": ["ccc", "ddd"]
  }
  limit 20
</rule>
```

It would be better that single string is also acceptable:
```xml
<rule>
  match {
    "namespace": "namespace1",
    "appname": "app1"
  }
 limit 20
</rule>
```

ashie: Now I think that generating `config_params` is too complicate and probably too hard or impossible for this case.
I'm considering another idea, using `:hash` type of `config_param` like this:

```ruby
    config_param :match, :hash, default: {namespace: [DEFAULT_NAMESPACE], appname: [DEFAULT_APPNAME}, value_type: :xxx
```

Although current implementation can't accept an array as a value of hash, I think there are 2 ways to resolve it.

1. Do not use array, express multiple items by regexp
2. Add a new `value_type` like `:string_array`

Example usage for 1.:
```xml
<rule>
  match {
         "namespace": "/spac1|space2|scape3/", 
         "appname": "/app1|app2|app3/"
       }
  ])
  limit 20
</rule>
```

Example patch & usage for 2.:

```diff
diff --git a/lib/fluent/config/types.rb b/lib/fluent/config/types.rb
index 98e7d0d9..b6836b8a 100644
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -191,6 +191,10 @@ module Fluent
         when :time then Config.time_value(value, opts, name)
         when :regexp then Config.regexp_value(value, opts, name)
         when :symbol then Config.symbol_value(value, opts, name)
+        when :string_array
+          child_opts = opts.dup
+          child_opts[:value_type] = :string
+          Config.array_value(value, child_opts, name)
         else
           raise "unknown type in REFORMAT: #{type}"
         end
```

Example usage for 2.:
```xml
<rule>
  match {
         "namespace": ["spac1", "space2", "scape3"], 
         "appname": ["app1, "app2", "app3"]
       }
  ])
  limit 20
</rule>
```

Pranjal-Gupta2: I think we can use a hash with `regexp` key values for defining rules. It will be easier to assign a group to a file using regular expressions.

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: Thanks! I have added these prefixes in the pattern.

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: @ashie I have renamed `appname` to `podname`. So, now the default values of group are k8s specific but can be overridden in conf for other use cases.

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: Thanks! I have changed it to:
```ruby
if group_watcher&.limit_lines_reached?(@path) || should_shutdown_now?
```

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: Thanks! I have modified it.

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: Thanks for pointing this out! I have updated the code.

lib/fluent/plugin/in_tail.rb
Pranjal-Gupta2: I have changed it to: 
```ruby
elsif @read_bytes_limit_per_second < 0 || tw.group_watcher&.limit <= 0
```

cosmo0920 (commented):
Implementation idea is good. But, I found a nitpick issue for a constant name.

test/plugin/test_in_tail.rb
```diff
@@ -143,13 +143,84 @@ def create_target_info(path)
                       "format_firstline" => "/^[s]/"
                     })
     ])
+  
+  PATTERN = "/#{TMP_DIR}\/(?<podname>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\/[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container>.+)-(?<docker_id>[a-z0-9]{6})\.log$/"
```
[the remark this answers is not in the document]
Pranjal-Gupta2: Yeah, we can simply use `GROUP_PATTERN` as well. What do you think?

lib/fluent/plugin/in_tail.rb
```diff
@@ -120,7 +128,24 @@ def initialize
...
+      config_param :pattern, 
+                   :regexp,
+                   default: /^\/var\/log\/containers\/(?<podname>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\/[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?<namespace>[^_]+)_(?<container>.+)-(?<docker_id>[a-z0-9]{64})\.log$/
+
+      desc 'Period of time in which the group_line_limit is applied'
+      config_param :rate_period, :time, default: 5
+
+      config_section :rule, param_name: :rule, required: true, multi: true do
+        desc 'Key-value pairs for grouping'
+        config_param :match, :hash, value_type: :regexp, default: {namespace: [DEFAULT_KEY], appname: [DEFAULT_KEY]}
```
ashie, line 142: It seems the default value should be `{namespace: [DEFAULT_KEY], podname: [DEFAULT_KEY]}`.
```diff
- {namespace: [DEFAULT_KEY], appname: [DEFAULT_KEY]}
+ {namespace: [DEFAULT_KEY], podname: [DEFAULT_KEY]}
```

Pranjal-Gupta2: Ah, thanks for pointing this out! I will fix this typo.

ashie:
The big picture seems good, I'll check further more details, please wait for a few days...

ashie:
A test always fails:
```
2022-02-07T07:27:29.7369178Z Error: test_shorter_than_rotate_wait(TailInputTest::singleline::log throttling per file::EOF with reads_bytes_per_second): Fluent::Test::Driver::TestTimedOut: Test case timed out with hard limit.
2022-02-07T07:27:29.7398448Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:201:in `rescue in run_actual'
2022-02-07T07:27:29.7398936Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:196:in `run_actual'
2022-02-07T07:27:29.7399358Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base.rb:95:in `run'
2022-02-07T07:27:29.7399753Z /home/runner/work/fluentd/fluentd/lib/fluent/test/driver/base_owner.rb:130:in `run'
2022-02-07T07:27:29.7400214Z /home/runner/work/fluentd/fluentd/test/plugin/test_in_tail.rb:714:in `test_shorter_than_rotate_wait'
2022-02-07T07:27:29.7400553Z      711:             tw
2022-02-07T07:27:29.7400872Z      712:           end.twice
2022-02-07T07:27:29.7401240Z      713: 
2022-02-07T07:27:29.7401498Z   => 714:           d.run(timeout: 10) do
2022-02-07T07:27:29.7403168Z      715:             until detached do
2022-02-07T07:27:29.7403680Z      716:               if d.events.size > 0 && !rotated
2022-02-07T07:27:29.7405460Z      717:                 cleanup_file("#{TMP_DIR}/tail.txt")
2022-02-07T07:27:29.7406123Z ===============================================================================
2022-02-07T07:29:08.6385152Z ...............................................................................
2022-02-07T07:29:08.6500370Z ...............................................................................
2022-02-07T07:29:08.6725292Z ...............................................................................
```

Pranjal-Gupta2:
> The big picture seems good, I'll check further more details, please wait for a few days...

@ashie @cosmo0920 I have updated the PR and all tests are passing. Please review the code.

ashie:
Thanks for update, and sorry for my delay.
I'll check it.

cosmo0920 (approved):
Looks reasonable for me.

ashie (commented):
I'll add more commits directly to your branch to polish up the code.
Please feel free to comment on them if you have any problems.

lib/fluent/plugin/in_tail/group_watch.rb
```diff
@@ -0,0 +1,204 @@
...
+
+      def add_path_to_group_watcher(path)
+        return nil if @group.nil?
+        group_watcher = find_group_from_metadata(path)
+        group_watcher.add(path) unless group_watcher.include?(path)
+        group_watcher
+      end
+
+      def remove_path_from_group_watcher(path)
+        return if @group.nil?
```
cosmo0920, line 89: Should we return `nil` on `remove_path_from_group_watcher` as well?
This is just nitpick and not important.

cosmo0920 (approved):
Looks good with a nitpick comment.

ashie:
Hmm, it takes too long time to run test, more than 1 minutes longer than before:

master branch:
```console
$ time bundle exec rake test TEST=test/plugin/test_in_tail.rb
...
real	2m33.910s
user	0m3.410s
sys	0m0.322s
```

this branch:
```console
$ time bundle exec rake test TEST=test/plugin/test_in_tail.rb
...
real	3m36.044s
user	0m3.807s
sys	0m0.335s
```

We should shorten it if it's possible since it's affect to debug efficiency.

ashie:
> Hmm, it takes too long time to run tests, more than 1 minutes longer than before:

Trying to improve it by efa7d2dd081eab33b2aad343b6be1393e3e936de
It seems fine on my local environment but not sure on CI yet.

ashie:
> > Hmm, it takes too long time to run tests, more than 1 minutes longer than before:
> 
> Trying to improve it by [efa7d2d](https://github.com/fluent/fluentd/commit/efa7d2dd081eab33b2aad343b6be1393e3e936de) It seems fine on my local environment but not sure on CI yet.

It's also fine on CI.

ashie:
I'll merge this after checking group_watch.rb once more.

ashie:
> > Hmm, it takes too long time to run tests, more than 1 minutes longer than before:
> 
> Trying to improve it by [efa7d2d](https://github.com/fluent/fluentd/commit/efa7d2dd081eab33b2aad343b6be1393e3e936de) It seems fine on my local environment but not sure on CI yet.

```console
$ time bundle exec rake test TEST=test/plugin/test_in_tail.rb
...
real	2m45.993s
user	0m3.642s
sys	0m0.266s
```

ashie (approved):

ashie:
Merged.
I'm sorry for taking long time to merge this.
And thank you for contributing a nice feature :smiley:

PranjalGupta2199:
Thanks @ashie @cosmo0920 @agup006 for your help and support! I am glad that it got merged! 🎉

yuzs2:
looks this is a very useful feature, it would be even greater if you could provide more detailed user-guide for this. I'm new to fluentd, and the current example looks too simple for me: https://docs.fluentd.org/input/tail#less-than-group-greater-than-section, thank you

yuzs2:
Hi folk, I tried to add the <group> section in my fluent.conf, but it was giving parse error:
/usr/local/rvm/rubies/ruby-2.7.6/lib/ruby/gems/2.7.0/gems/fluentd-1.15.0/lib/fluent/config/basic_parser.rb:92:in `parse_error!': got incomplete JSON hash configuration at kubernetes.conf line 230,10 (Fluent::ConfigParseError)

below is what my fluent.conf be like, removing <group> section it will be working fine:
<source>
  @type tail
  @id in_tail_container_logs
  pos_file /var/log/{{.ID}}-fluentd-containers.log.pos
  pos_file_compaction_interval 1h
  tag kubernetes.*
  read_from_head true
  read_bytes_limit_per_second 8192
  <parse>
    @type multiline
    # cri-o
    format1 /^(?<partials>([^\n]+ (stdout|stderr) P [^\n]+\n)*)/
    format2 /(?<time>[^\n]+) (?<stream>stdout|stderr) F (?<log>[^\n]*)/
    # docker
    format3 /|^(?<json>{.*})/
    time_format %Y-%m-%dT%H:%M:%S.%N%:z
  </parse>
  <group>
    rate_period 600

    <rule>
      match {
          namespace: [/gulel-core/],
          podname: [/redis-master-0/]
      }
      limit 2
    </rule>
  </group>
</source>

could you please help? thanks @ashie @Pranjal-Gupta2 @cosmo0920 @agup006 @pmoogi-redhat

ashie:
`match` value should be like this:

```diff
  match {
-      namespace: [/gulel-core/],
-      podname: [/redis-master-0/]
+      "namespace": "/gulel-core/",
+      "podname": "/redis-master-0/"
  }
```

The usage is describe in the [document](https://docs.fluentd.org/input/tail#less-than-group-greater-than-section) but it's also wrong, sorry for that. I'll fix it.

ashie:
> `match` value should be like this:
> 
> ```diff
>   match {
> -      namespace: [/gulel-core/],
> -      podname: [/redis-master-0/]
> +      "namespace": "/gulel-core/",
> +      "podname": "/redis-master-0/"
>   }
> ```
> 
> The usage is describe in the [document](https://docs.fluentd.org/input/tail#less-than-group-greater-than-section) but it's also wrong, sorry for that. I'll fix it.

https://github.com/fluent/fluentd-docs-gitbook/pull/426

lib/fluent/plugin/in_tail/group_watch.rb
```diff
@@ -0,0 +1,204 @@
...
+      def limit_time_period_reached?(path)
+        time_spent_reading(path) < @rate_period
+      end
+
+      def limit_lines_reached?(path)
+        return true unless include?(path)
+        return true if @limit == 0
+
+        return false if @limit < 0
+        return false if @current_paths[path].number_lines_read < @limit / size
```
daipom, line 188: Thanks for adding this great feature!
I have a question about this code line.

The `@limit` controls the total number of lines collected for a group, right? (https://docs.fluentd.org/input/tail#less-than-group-greater-than-section)

That judgment seems to be done by this code line, but this line judges it by the average number of lines collected for a group, not the total number.

Either the code or the documentation is wrong?

ashie: Hmm, that's right. It should be total.

daipom: Thanks! For now, I made an issue for this: 

* #4184

Pranjal-Gupta2: @daipom @ashie 
>This PR enables us to form different groups based on configuration rules provided by the user. Within each group, one can set the rate limits for log collection. The set rate limit gets uniformly applied to each member (log file) of the group. 

The PR description states that the limit set in the configuration distributes the rate uniformly to all the members in a group. That's why the condition checks for average rate limit for a particular group member. I get why this is confusing and we may either change the documentation with an example or change the code itself. 

The reason we are using average limit (for a group member) is because it puts a hard upper limit on the rate of logs emitted by a group. For example, a two groups with same limit but one having 1000 containers and the other having 10 containers, both will emit the same number of log lines. 
However, if you apply the same limit to all the group members (without having a hard upper limit control) then there's no limit on the amount of data that is sent via fluentd. Considering the previous example, the rate of data sent from one group will be 100x larger.

daipom: @Pranjal-Gupta2 Thanks!

> Considering the previous example, the rate of data sent from one group will be 100x larger.

If we apply the limit to each file independently, it will be 100 times larger, right?

I think the preferred specification is to calculate the total number in the group in #4184.
In this way, it will be the same upper amount independent of the number of files.

Pranjal-Gupta2: >If we apply the limit to each file independently, it will be 100 times larger, right? 

Yes, correct! 

> I think the preferred specification is to calculate the total number in the group in https://github.com/fluent/fluentd/issues/4184.
> In this way, it will be the same upper amount independent of the number of files.

Can you please elaborate on this?

Pranjal-Gupta2: > I think the preferred specification is to calculate the total number in the group in https://github.com/fluent/fluentd/issues/4184.
> In this way, it will be the same upper amount independent of the number of files.

Are you saying that we calculate the limit per file once and then use it for rate limiting for each group member, right? Taking average ensures that if a file is being added/removed dynamically, then rate limit adjusts itself.

daipom: I'm thinking of having just one counter per group and comparing the total as a group with the limit.

Code image (This is just an image of a modification):

```diff
--- a/lib/fluent/plugin/in_tail/group_watch.rb
+++ b/lib/fluent/plugin/in_tail/group_watch.rb
@@ -138,17 +138,18 @@ module Fluent::Plugin
       )
 
       def initialize(rate_period = 60, limit = -1)
-        @current_paths = {}
+        @current_paths = []
         @rate_period = rate_period
         @limit = limit
+        @counter = FileCounter.new(0, nil)
       end
 
       def add(path)
-        @current_paths[path] = FileCounter.new(0, nil)
+        @current_paths << path
       end
 
       def include?(path)
-        @current_paths.key?(path)
+        @current_paths.include?(path)
       end
 
       def size
@@ -159,25 +160,25 @@ module Fluent::Plugin
         @current_paths.delete(path)
       end
 
-      def update_reading_time(path)
-        @current_paths[path].start_reading_time ||= Fluent::Clock.now
+      def update_reading_time
+        @counter.start_reading_time ||= Fluent::Clock.now
       end
 
-      def update_lines_read(path, value)
-        @current_paths[path].number_lines_read += value
+      def update_lines_read(value)
+        @counter.number_lines_read += value
       end
 
-      def reset_counter(path)
-        @current_paths[path].start_reading_time = nil
-        @current_paths[path].number_lines_read = 0
+      def reset_counter
+        @counter.start_reading_time = nil
+        @counter.number_lines_read = 0
       end
 
-      def time_spent_reading(path)
-        Fluent::Clock.now - @current_paths[path].start_reading_time
+      def time_spent_reading
+        Fluent::Clock.now - @counter.start_reading_time
       end
 
-      def limit_time_period_reached?(path)
-        time_spent_reading(path) < @rate_period
+      def limit_time_period_reached?
+        time_spent_reading < @rate_period
       end
 
       def limit_lines_reached?(path)
@@ -185,13 +186,13 @@ module Fluent::Plugin
         return true if @limit == 0
 
         return false if @limit < 0
-        return false if @current_paths[path].number_lines_read < @limit / size
+        return false if @counter.number_lines_read < @limit
 
         # update_reading_time(path)
-        if limit_time_period_reached?(path) # Exceeds limit
+        if limit_time_period_reached? # Exceeds limit
           true
         else # Does not exceed limit
-          reset_counter(path)
+          reset_counter
           false
         end
       end
```

Pranjal-Gupta2: Does this implementation ensure that the fluentd **collects equally** from all the files in a group?

daipom: Oh! I see!
Now I can see the essence of this.
We can prevent unbalanced collections by judging the limit with the average number.
As you say, simply judging by the total number may result in unbalanced collections.

We need to think more carefully about this feature's specifications.
Thanks!

Pranjal-Gupta2: Yup! We can write a better articulation of this point in the documentation. 
@ashie Can I take this up?

daipom: I will think some more about what to do with #4184.
If it looks good to fix the document, I will fix it.

daipom: > Yup! We can write a better articulation of this point in the documentation. @ashie Can I take this up?

Oh, sorry. I missed this comment.
At this point, fixing the document would be a good idea!

daipom: We welcome the document modifications of course!

Pranjal-Gupta2: Thanks @daipom @ashie! 
Will start working on this.

PR #3541, opened 2021-10-31
in_tail: Fix StatWatcher uses wrong path

majimenez-stratio:
No new logs are read when `enable_stat_watcher true` and `enable_watch_timer false`. Stat watchers for new files are created with the wrong path, so they aren't notified of any events.

Use path from target_info when creating StatWatcher

**Docs Changes**:
None

**Release Note**:
Fix in_tail not reading new lines with stat watcher enabled and watch timer disabled

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -370,7 +370,7 @@ module Fluent::Plugin
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
 
-      log.debug { "tailing paths: target = #{target_paths.join(",")} | existing = #{existence_paths.join(",")}" }
+      log.debug { "tailing paths: target = #{target_paths_hash.keys.join(",")} | existing = #{existence_paths_hash.keys.join(",")}" }
 
       unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
       added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
@@ -389,7 +389,7 @@ module Fluent::Plugin
       end
 
       if @enable_stat_watcher
-        tt = StatWatcher.new(path, log) { tw.on_notify }
+        tt = StatWatcher.new(target_info.path, log) { tw.on_notify }
         tw.register_watcher(tt)
       end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -369,8 +369,10 @@ module Fluent::Plugin
     def refresh_watchers
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
-
-      log.debug { "tailing paths: target = #{target_paths_hash.keys.join(",")} | existing = #{existence_paths_hash.keys.join(",")}" }
+      
+      target_paths_str = target_paths_hash.collect { |key, target_info| target_info.path }.join(",")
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -389,7 +389,7 @@ def setup_watcher(target_info, pe)
       end
 
       if @enable_stat_watcher
-        tt = StatWatcher.new(path, log) { tw.on_notify }
+        tt = StatWatcher.new(target_info.path, log) { tw.on_notify }
```
ashie, line 392: Oh my god....

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,7 +370,7 @@ def refresh_watchers
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
 
-      log.debug { "tailing paths: target = #{target_paths.join(",")} | existing = #{existence_paths.join(",")}" }
+      log.debug { "tailing paths: target = #{target_paths_hash.keys.join(",")} | existing = #{existence_paths_hash.keys.join(",")}" }
```
ashie, line 373: The keys of `target_paths_hash` and `existence_paths_hash` are inode when `follow_inode = true`.
So that paths should be collected by `xxx_hash.collect { |key, target_info| target_info.path }`.

ashie (changes requested):

lib/fluent/plugin/in_tail.rb
ashie: I got it why tests aren't failed even for such broken code.
When there is no wildcard in the path, `path` is valid because it's defined as config_param:
```
    config_param :path, :string
```
and `path` and `target_info.path` are same.

On the other hand, when a wildcard is used in the `path` config, the expanded path (`target_info.path`) and the `path` param aren't matched. 
e.g.)

* `path` param: tail*.txt
* `target_info.path`: tail1.txt

In addition, there is no test for the later case.

ashie:
I've added a test for it.

lib/fluent/plugin/in_tail.rb
majimenez-stratio: Done and tested ok with `follow_inodes = true`

lib/fluent/plugin/in_tail.rb
```diff
@@ -369,8 +369,10 @@ def existence_path
     def refresh_watchers
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
-
-      log.debug { "tailing paths: target = #{target_paths.join(",")} | existing = #{existence_paths.join(",")}" }
+      
+      target_paths_str = target_paths_hash.collect { |key, target_info| target_info.path }.join(",")
+      existence_paths_str = existence_paths_hash.collect { |key, target_info| target_info.path }.join(",")
```
ashie, line 374: The debug string should be built only when debug mode is enabled. I'll add a commit to fix it.

ashie: Done.

ashie (approved):

ashie:
I'll merge it after waiting CI.

ashie:
Merged. Thanks a lot!

majimenez-stratio:
Glad to help!

lib/fluent/plugin/in_tail.rb
toyaser: :clap: so impressed with this find!

PR #3542, opened 2021-10-31
in_tail: Always read from head for new/rotated files after startup

majimenez-stratio:
Signed-off-by: majimenez-stratio <majimenez@stratio.com>

None

I noticed that read_from_head affect files read during startup and new files equally. I see a few drawbacks with the current situation if `read_from_head = false` and path includes '*', namely:
- New files detected on refresh are not read from the beggining and initial lines are lost.
- Same as above for rotated files

As a result, I think we can improve the way read_from_head is applied, so that it only affects new files detected during startup. This PR changes the behavior so that:
- During startup, files are read from the position stored in the position file, if available. If not, they are read as specified in `read_from_head`.
- After startup, if a new file is detected (or rotated), it is **always** read from the beginning.
- If `skip_refresh_on_startup = true`, new files are **always** read from the beggining. I'm not sure about this side effect, but I think `skip_refresh_on_startup` only makes sense in combination with `read_from_head = true`, so I think we're fine here.

I tried to mimic the way fluent-bit tail input works as it makes more sense to me. However, I'm quite new to Ruby, fluentd and logging in general, so I'm open to discussion if you feel this change is not appropiate or should be approached from another angle.

**Docs Changes**:
Pending

**Release Note**:
Always read from head for new/rotated files after startup

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -250,7 +250,7 @@ module Fluent::Plugin
         end
       end
 
-      refresh_watchers unless @skip_refresh_on_startup
+      refresh_watchers(true) unless @skip_refresh_on_startup
       timer_execute(:in_tail_refresh_watchers, @refresh_interval, &method(:refresh_watchers))
     end
 
@@ -366,7 +366,7 @@ module Fluent::Plugin
     # It will cause log duplication after updated watch files.
     # In such case, you should separate log directory and specify two paths in path parameter.
     # e.g. path /path/to/dir/*,/path/to/rotated_logs/target_file
-    def refresh_watchers
+    def refresh_watchers(startup = false)
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
 
@@ -376,12 +376,13 @@ module Fluent::Plugin
       added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
 
       stop_watchers(unwatched_hash, immediate: false, unwatched: true) unless unwatched_hash.empty?
-      start_watchers(added_hash) unless added_hash.empty?
+      start_watchers(added_hash, startup) unless added_hash.empty?
     end
 
-    def setup_watcher(target_info, pe)
+    def setup_watcher(target_info, pe, startup = false)
       line_buffer_timer_flusher = @multiline_mode ? TailWatcher::LineBufferTimerFlusher.new(log, @multiline_flush_interval, &method(:flush_buffer)) : nil
-      tw = TailWatcher.new(target_info, pe, log, @read_from_head, @follow_inodes, method(:update_watcher), line_buffer_timer_flusher, method(:io_handler), @metrics)
+      read_from_head = startup ? @read_from_head : true
+      tw = TailWatcher.new(target_info, pe, log, read_from_head, @follow_inodes, method(:update_watcher), line_buffer_timer_flusher, method(:io_handler), @metrics)
 
       if @enable_watch_timer
         tt = TimerTrigger.new(1, log) { tw.on_notify }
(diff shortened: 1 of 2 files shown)


ashie (changes requested):
Good catch!
Your patch seems almost reasonable, thanks!

> * If skip_refresh_on_startup = true, new files are always read from the beggining. I'm not sure about this side effect, but I think skip_refresh_on_startup only makes sense in combination with read_from_head = true, so I think we're fine here.

We should care this case too to keep consistency.
Otherwise people who reading the code later will be confused.
How about using an instance variable to check the first `setup_watchers` is already called or not.
By this, additional argument aren't needed anymore.

majimenez-stratio:
Sounds good, I’ll change the code to address this case

majimenez-stratio:
I followed your suggestions and refactored the code so that startup is tracked with an instance variable instead of parameters. It looks much nicer now, thanks for the tip.

Also, with the latest changes, the behavior when `skip_refresh_on_startup = true` seems more consistent, as it is basically the same as when it's false: first refresh works as specified in `read_from_head`, but will always read from the beginning for new/rotated files.

Additionally, this is the code that resets `@startup` instance var:
https://github.com/fluent/fluentd/blob/98feddc12014a120a789593593e6a502aa41920f/lib/fluent/plugin/in_tail.rb#L381
I'm not sure it is the best/safer way to do it or the most idiomatic, I'll take your guidance here in case there's a better way.

lib/fluent/plugin/in_tail.rb
```diff
@@ -371,17 +372,19 @@ def refresh_watchers
       existence_paths_hash = existence_path
 
       log.debug { "tailing paths: target = #{target_paths.join(",")} | existing = #{existence_paths.join(",")}" }
-
+
```
ashie, line 375: Please remove needless spaces.

ashie (changes requested):
Could you resolve conflict?

ashie:
> Additionally, this is the code that resets `@startup` instance var:
> 
> https://github.com/fluent/fluentd/blob/98feddc12014a120a789593593e6a502aa41920f/lib/fluent/plugin/in_tail.rb#L381
> 
> I'm not sure it is the best/safer way to do it or the most idiomatic, I'll take your guidance here in case there's a better way.

LGTM

lib/fluent/plugin/in_tail.rb
```diff
@@ -371,17 +372,19 @@ def refresh_watchers
       existence_paths_hash = existence_path
 
       log.debug { "tailing paths: target = #{target_paths.join(",")} | existing = #{existence_paths.join(",")}" }
-
+      
       unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
       added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
-
+
```
kenhys, line 378: Same as "Please remove needless spaces."

kenhys:
@majimenez-stratio could you resolve the conflict again, please?

lib/fluent/plugin/in_tail.rb
majimenez-stratio: Done!

lib/fluent/plugin/in_tail.rb
majimenez-stratio: Sorry, I forgot this

ashie:
Various memo:

* We need to update the document about `read_from_head`
  * https://docs.fluentd.org/input/tail#read_from_head
  * https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/input/tail.md#read_from_head
* I had a concern about incompatibility of the behaviour. But now I think that adding a notice to above document is enough, the revised behaviour seems reasonable for most users.
* I considered to add a test for rotation case but probably it's not needed
  * Rotation is usually processed by `TailWatcher#on_rorate`, not by `refresh_watcher`. So that read position is correctly determined by it.
  * In actual, I tried to add a test for it but both previous code & new code are passed the test.

ashie:
Thanks for your contribution!

majimenez-stratio:
My pleasure!

PR #3544, opened 2021-11-01
Relax http_parse.rb version

ashie:
Fixes #

http_parser.rb 0.8.0 is ready for Ractor and compatible with previous versions.
Although current Fluentd is far from supporting Ractor, we want to support it in the future.

**Docs Changes**:
None

**Release Note**: 
Accept http_parse.rb 0.8.0.


cosmo0920 (approved):
Seems reasonable for me.
Yeah, we have lots of tasks for making "Ractor safe".

mcgrewjeramie2936 (commented):
#

Issue #1849, opened 2018-02-07
Large message overflow logic incorrect


amread:
Check [CONTRIBUTING guideline](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md) first and here is the list to help us investigate the problem.

- fluentd or td-agent version. v1.0.2 (also seems to be a problem on master right now)

I have a use case that allows for very log messages. While load testing using 2m buffers, I noticed that if I try to quickly jam these messages through back-to-back, the system will complain:

```
a 524596bytes record is larger than buffer chunk limit size
``` 

Looking at the code, I see the test for a single record overflowing the current chunk in `write_step_by_step`:

```ruby
                  if split.size == 1 && original_bytesize == 0
                    big_record_size = format ? format.call(split).bytesize : split.first.bytesize
                    raise BufferChunkOverflowError, "a #{big_record_size}bytes record is larger than buffer chunk limit size"
                  end
```

Since `original_bytesize` is set outside of the inner loop, it does not detect if the current split is the first chunk. The fix to this is straightforward enough, but the retry logic isn't clever enough to do something different the next time.

Ideally, we would know ahead of time that the message would overflow the buffer and just not write it. I imagine the code does not take the approach for the sake of efficiency. It would be nice if there was something like the `commit` and `rollback` functionality for use while constructing the chunk, perhaps `mark()` and `rewindToMark()` functions, or maybe even restructure `append` and `concat` to take blocks and allow for rewinding on failure.

I took a stab at some minimal change, but found that I don't have a full appreciation for the nuance of this function's existing behavior, given the various assertions that failed in my attempts to fix this bug. It does seem that the underlying assumption of the current code is that records are small in comparison to the buffer so that the `chunk_full_threshold` will always be crossed while filling a chunk. I can probably work around this issue by either increasing my buffer size, or changing the `chunk_full_threshold`. 

Here is a test that shows the problem adding one message that fills half the buffer, and then another that goes one byte over (added to 'with configuration for test with lower limits` sub_test_case):
```ruby
    test '#write does not raise BufferOverflowError when chunk is not empty' do
      m = @p.metadata(timekey: Time.parse('2016-04-11 16:40:00 +0000').to_i)

      assert_nothing_raised Fluent::Plugin::Buffer::BufferOverflowError do
        @p.write({m => ["a" * 512, "b" * 513 ]})
      end
    end
```

[...]

PR #3553, opened 2021-11-05
Fix a bug that BufferChunkOverflowError cause a whole data loss

kenhys:

Fixes #

In the previous version, if BufferChunkOverflowError occurs,
whole data is not sent because of exception.
It is not appropriate because the problem exists only a partial split
data.

In this commit, detect BufferChunkOverflowError and postpone to raise
it as an exception.
So a valid partial split is processed as intended.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -354,7 +354,7 @@ module Fluent
                 unstaged_chunks[metadata] ||= []
                 unstaged_chunks[metadata] << chunk
               end
-              if error and not error.empty?
+              if error && !error.empty?
                 buffer_chunk_overflow_errors << error
               end
             end
@@ -773,11 +773,18 @@ module Fluent
                 split = splits[writing_splits_index]
                 formatted_split = format ? format.call(split) : split.first
                 if split.size == 1 && original_bytesize == 0
-                  big_record_size = formatted_split.bytesize
-                  if chunk.bytesize + big_record_size > @chunk_limit_size
-                    errors << "a #{big_record_size} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size"
-                    writing_splits_index += 1
-                    next
+                  if format == nil && @compress != :text
+                    # The actual size of chunk is not determined until after chunk.append.
+                    # so, keep already processed 'split' content here.
+                    # (allow performance regression a bit)
+                    chunk.commit
+                  else
+                    big_record_size = formatted_split.bytesize
+                    if chunk.bytesize + big_record_size > @chunk_limit_size
+                      errors << "a #{big_record_size} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size"
+                      writing_splits_index += 1
+                      next
+                    end
                   end
                 end
 
@@ -788,7 +795,18 @@ module Fluent
                 end
 
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
+                  adding_bytes = chunk.instance_eval { @adding_bytes } || "N/A" # 3rd party might not have 'adding_bytes'
                   chunk.rollback
+
+                  if split.size == 1 && original_bytesize == 0
+                    # It is obviously case that BufferChunkOverflowError should be raised here,
+                    # but if it raises here, already processed 'split' or
+                    # the proceeding 'split' will be lost completely.
+                    # so it is a last resort to delay raising such a exception
+                    errors << "a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size"
+                    writing_splits_index += 1
+                    next
+                  end
 
                   if chunk_size_full?(chunk) || split.size == 1
                     enqueue_chunk_before_retry = true
@@ -811,6 +829,7 @@ module Fluent
             end
 
             block.call(chunk, chunk.bytesize - original_bytesize, errors)
+            errors = []
           end
         end
       rescue ShouldRetry
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -332,12 +332,14 @@ module Fluent
         unstaged_chunks = {} # metadata => [chunk, chunk, ...]
         chunks_to_enqueue = []
         staged_bytesizes_by_chunk = {}
+        # track internal BufferChunkOverflowError in write_step_by_step
+        buffer_chunk_overflow_errors = []
 
         begin
           # sort metadata to get lock of chunks in same order with other threads
           metadata_and_data.keys.sort.each do |metadata|
             data = metadata_and_data[metadata]
-            write_once(metadata, data, format: format, size: size) do |chunk, adding_bytesize|
+            write_once(metadata, data, format: format, size: size) do |chunk, adding_bytesize, error|
               chunk.mon_enter # add lock to prevent to be committed/rollbacked from other threads
               operated_chunks << chunk
               if chunk.staged?
@@ -351,6 +353,9 @@ module Fluent
               elsif chunk.unstaged?
                 unstaged_chunks[metadata] ||= []
                 unstaged_chunks[metadata] << chunk
+              end
+              if error and not error.empty?
+                buffer_chunk_overflow_errors << error
               end
             end
           end
@@ -443,6 +448,10 @@ module Fluent
               chunk.purge rescue nil # to prevent leakage of unstaged chunks
             end
             chunk.mon_exit rescue nil # this may raise ThreadError for chunks already committed
+          end
+          unless buffer_chunk_overflow_errors.empty?
+            # Notify delayed BufferChunkOverflowError here
+            raise BufferChunkOverflowError, buffer_chunk_overflow_errors.join(", ")
           end
         end
       end
@@ -716,6 +725,7 @@ module Fluent
 
       def write_step_by_step(metadata, data, format, splits_count, &block)
         splits = []
+        errors = []
         if splits_count > data.size
           splits_count = data.size
         end
@@ -761,19 +771,24 @@ module Fluent
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                formatted_split = format ? format.call(split) : split.first
+                if split.size == 1 && original_bytesize == 0
+                  big_record_size = formatted_split.bytesize
+                  if chunk.bytesize + big_record_size > @chunk_limit_size
+                    errors << "a #{big_record_size} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size"
+                    writing_splits_index += 1
+                    next
+                  end
+                end
+
                 if format
-                  chunk.concat(format.call(split), split.size)
+                  chunk.concat(formatted_split, split.size)
                 else
                   chunk.append(split, compress: @compress)
                 end
 
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
                   chunk.rollback
-
-                  if split.size == 1 && original_bytesize == 0
-                    big_record_size = format ? format.call(split).bytesize : split.first.bytesize
-                    raise BufferChunkOverflowError, "a #{big_record_size}bytes record is larger than buffer chunk limit size"
-                  end
 
                   if chunk_size_full?(chunk) || split.size == 1
                     enqueue_chunk_before_retry = true
@@ -795,7 +810,7 @@ module Fluent
               raise
             end
 
-            block.call(chunk, chunk.bytesize - original_bytesize)
+            block.call(chunk, chunk.bytesize - original_bytesize, errors)
           end
         end
       rescue ShouldRetry
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -1247,6 +1247,7 @@ class BufferTest < Test::Unit::TestCase
   sub_test_case 'when compress is gzip' do
     setup do
       @p = create_buffer({'compress' => 'gzip'})
+      @dm0 = create_metadata(Time.parse('2016-04-11 16:00:00 +0000').to_i, nil, nil)
     end
 
     test '#compress returns :gzip' do
@@ -1256,6 +1257,30 @@ class BufferTest < Test::Unit::TestCase
     test 'create decompressable chunk' do
       chunk = @p.generate_chunk(create_metadata)
       assert chunk.singleton_class.ancestors.include?(Fluent::Plugin::Buffer::Chunk::Decompressable)
+    end
+
+    test '#write compressed data which exceeds chunk_limit_size, it raises BufferChunkOverflowError' do
+      @p = create_buffer({'compress' => 'gzip', 'chunk_limit_size' => 70})
+      timestamp = event_time('2016-04-11 16:00:02 +0000')
+      es = Fluent::ArrayEventStream.new([[timestamp, {"message" => "012345"}], # overflow
+                                         [timestamp, {"message" => "aaa"}],
+                                         [timestamp, {"message" => "bbb"}]])
+      assert_equal [], @p.queue.map(&:metadata)
+      assert_equal 70, @p.chunk_limit_size
+
+      # calculate the actual boundary value. it varies on machine
+      c = @p.generate_chunk(create_metadata)
+      c.append(Fluent::ArrayEventStream.new([[timestamp, {"message" => "012345"}]]), compress: :gzip)
+      overflow_bytes = c.bytesize
+
+      messages = "a #{overflow_bytes} bytes record (nth: 0) is larger than buffer chunk limit size"
+      assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError.new(messages) do
+        # test format == nil && compress == :gzip
+        @p.write({@dm0 => es})
+      end
+      # message a and b occupies each chunks in full, so both of messages are queued (no staged chunk)
+      assert_equal([2, [@dm0, @dm0], [1, 1], nil],
+                   [@p.queue.size, @p.queue.map(&:metadata), @p.queue.map(&:size), @p.stage[@dm0]])
     end
   end
 
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -18,7 +18,7 @@ module FluentPluginBufferTest
   end
   class DummyMemoryChunkError < StandardError; end
   class DummyMemoryChunk < Fluent::Plugin::Buffer::MemoryChunk
-    attr_reader :append_count, :rollbacked, :closed, :purged
+    attr_reader :append_count, :rollbacked, :closed, :purged, :chunk
     attr_accessor :failing
     def initialize(metadata, compress: :text)
       super
@@ -943,6 +943,52 @@ class BufferTest < Test::Unit::TestCase
       assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError do
         @p.write({@dm0 => es}, format: @format)
       end
+    end
+
+    data(
+      first_chunk: Fluent::ArrayEventStream.new([[event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}],
+                                                 [event_time('2016-04-11 16:00:02 +0000'), {"message" => "a"}],
+                                                 [event_time('2016-04-11 16:00:02 +0000'), {"message" => "b"}]]),
+      intermediate_chunk: Fluent::ArrayEventStream.new([[event_time('2016-04-11 16:00:02 +0000'), {"message" => "a"}],
+                                                        [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}],
+                                                        [event_time('2016-04-11 16:00:02 +0000'), {"message" => "b"}]]),
+      last_chunk: Fluent::ArrayEventStream.new([[event_time('2016-04-11 16:00:02 +0000'), {"message" => "a"}],
+                                                [event_time('2016-04-11 16:00:02 +0000'), {"message" => "b"}],
+                                                [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}]]),
+      multiple_chunks: Fluent::ArrayEventStream.new([[event_time('2016-04-11 16:00:02 +0000'), {"message" => "a"}],
+                                                     [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}],
+                                                     [event_time('2016-04-11 16:00:02 +0000'), {"message" => "b"}],
+                                                     [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}]])
+    )
+    test '#write exceeds chunk_limit_size, raise BufferChunkOverflowError, but not lost whole messages' do |(es)|
+      assert_equal [@dm0], @p.stage.keys
+      assert_equal [], @p.queue.map(&:metadata)
+
+      assert_equal 1_280_000, @p.chunk_limit_size
+
+      nth = []
+      es.entries.each_with_index do |entry, index|
+        if entry.last["message"].size == @p.chunk_limit_size
+          nth << index
+        end
+      end
+      messages = []
+      nth.each do |n|
+        messages << "a 1280025 bytes record (nth: #{n}) is larger than buffer chunk limit size"
+      end
+
+      assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError.new(messages.join(", ")) do
+        @p.write({@dm0 => es}, format: @format)
+      end
+      # message a and b are concatenated and staged
+      staged_messages = Fluent::MessagePackFactory.msgpack_unpacker.feed_each(@p.stage[@dm0].chunk).collect do |record|
+        record.last
+      end
+      assert_equal([2, [{"message" => "a"}, {"message" => "b"}]],
+                   [@p.stage[@dm0].size, staged_messages])
+      # only es0 message is queued
+      assert_equal [@dm0], @p.queue.map(&:metadata)
+      assert_equal [5000], @p.queue.map(&:size)
     end
   end
 


kenhys:
TODO: need to fix error.

kenhys:
> TODO: need to fix error.

Resolved.

kenhys:
micro bench:

It has slightly performance drawbacks.

Before:

      no BufferChunkOverflowError
                              116.903k (± 1.4%) i/s -      1.172M in  10.030026s
      raise BufferChunkOverflowError in every 100 calls
                               40.906k (± 4.1%) i/s -    411.060k in  10.065067s

      Comparison:
      no BufferChunkOverflowError:   116903.2 i/s
      raise BufferChunkOverflowError in every 100 calls:    40905.8 i/s - 2.86x  (± 0.00) slower

After:

      no BufferChunkOverflowError
                              116.276k (± 1.1%) i/s -      1.163M in  10.007341s
      raise BufferChunkOverflowError in every 100 calls
                               40.179k (± 4.9%) i/s -    401.472k in  10.016561s

      Comparison:
      no BufferChunkOverflowError:   116276.1 i/s
      raise BufferChunkOverflowError in every 100 calls:    40179.0 i/s - 2.89x  (± 0.00) slower

```
+    test '#write BufferChunkOverflowError micro benchmark' do
+      require 'benchmark/ips'
+      es = Fluent::ArrayEventStream.new([ [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x"}] ])
+      large = Fluent::ArrayEventStream.new([ [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}] ])
+      Benchmark.ips do |x|
+        x.config(:time => 10, :warmup => 2)
+        x.report("no BufferChunkOverflowError") do
+          @p.write({@dm0 => es}, format: @format)
+        end
+        x.report("raise BufferChunkOverflowError in every 100 calls") do |n|
+          n.times do |i|
+            if i % 100 == 0
+              assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError do
+                @p.write({@dm0 => large}, format: @format)
+              end
+            else
+              @p.write({@dm0 => es}, format: @format)
+            end
+          end
+        end
+        x.compare!
+      end
+    end
```

lib/fluent/plugin/buffer.rb
```diff
@@ -444,6 +451,12 @@ def write(metadata_and_data, format: nil, size: nil, enqueue: false)
             end
             chunk.mon_exit rescue nil # this may raise ThreadError for chunks already committed
           end
+          if @delay_buffer_chunk_overflow_error
+            # Notify delayed chunk overflow here
+            message = @delay_buffer_chunk_overflow_error
+            @delay_buffer_chunk_overflow_error = nil
```
cosmo0920, line 457: This buffer should be handled by multiple output threads.
I believe this implementation covered for most cases but we should block other threads with `synchronize` blocks.

lib/fluent/plugin/buffer.rb
```diff
@@ -772,7 +791,13 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
 
                   if split.size == 1 && original_bytesize == 0
                     big_record_size = format ? format.call(split).bytesize : split.first.bytesize
-                    raise BufferChunkOverflowError, "a #{big_record_size}bytes record is larger than buffer chunk limit size"
+                    writing_splits_index += 1
+                    # It is obviously case that BufferChunkOverflowError should be raised here,
+                    # but if it raises here, already processed 'split' or
+                    # the proceeding 'split' will be lost completely.
+                    # so it is a last resort to delay raising such a exception
+                    @delay_buffer_chunk_overflow_error = "a #{big_record_size}bytes record is larger than buffer chunk limit size"
```
cosmo0920, line 799: Substitution into instance variable should be blocked with `synchronize` block from other output threads.

cosmo0920 (changes requested):
I believe almost this implementation is very awesome, but we should consider that buffer should be handled by multiple output threads. 
Thus, we should use `synchronize` blocks to block processing by other output threads while updating values with substitutions.

lib/fluent/plugin/buffer.rb
kenhys: Thanks, fixed.

lib/fluent/plugin/buffer.rb
kenhys: Fixed, too.

kenhys:
Fixed and squashed commit.

cosmo0920 (approved):
Looks reasonable for me.
We should wait for @ashie's review.

lib/fluent/plugin/buffer.rb
```diff
@@ -444,6 +451,12 @@ def write(metadata_and_data, format: nil, size: nil, enqueue: false)
             end
             chunk.mon_exit rescue nil # this may raise ThreadError for chunks already committed
           end
+          if @delay_buffer_chunk_overflow_error
+            # Notify delayed chunk overflow here
+            message = synchronize { @delay_buffer_chunk_overflow_error }
```
ashie, line 456: Should wrap the whole of the transaction of `@delay_buffer_chunk_overflow_error` instead of each lines.

lib/fluent/plugin/buffer.rb
ashie: It doesn't seem fixed.

ashie, line 799: I'm not sure whether it's possible or not but it might be overwritten by other following errors.
How about using an array to store all errors?

ashie (changes requested):
Please consider not to use an instance variable `@delay_buffer_chunk_overflow_error`.
I'm not sure but I think the feature can be realized without bothering synchronization.
For example, store errors into a local variable in `write_step_by_step` and pass it to `write_once` or `write` via 3rd argument of given blocks.

lib/fluent/plugin/buffer.rb
kenhys: Looks nice.

kenhys:
I've considered two-approaches to fix again.
* propagate error via #write_once block
* propagate error by raising #write_step_by_step

The first one is confirmed that it shows better performance.

ashie:
It seems good, but I don't yet fully understand it.
Please wait for a while.

test/plugin/test_buffer.rb
```diff
@@ -944,6 +944,52 @@ def create_chunk_es(metadata, es)
...
+      messages = []
+      nth.each do |n|
+        messages << "a 1280025 bytes record (nth: #{n}) is larger than buffer chunk limit size"
+      end
+
+      assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError.new(messages.join(", ")) do
+        @p.write({@dm0 => es}, format: @format)
+      end
+      # message a and b are concatenated and staged
+      messages = Fluent::MessagePackFactory.msgpack_unpacker.feed_each(@p.stage[@dm0].chunk).collect do |record|
```
ashie, line 984: Please rename it with `staged_messages` to clarify the thing written in the comment.

lib/fluent/plugin/buffer.rb
```diff
@@ -761,6 +771,12 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                if split.size == 1 && original_bytesize == 0 && (chunk.bytesize + split.first.to_msgpack.bytesize) > @chunk_limit_size
+                  # When it will be predictable that BufferChunkOverflowError occur,
+                  # commit current chunk to keep it, otherwise valid 'split' data can be lost by
+                  # the following chunk.rollback
+                  chunk.commit
+                end
```
ashie, line 779: This is the point I'm concerned.

* Doesn't it have performance regression?
* It seems that it cares only the case on msgpack.
  * It might not exceed when `compress gzip` case

ashie: >       * It might not exceed when `compress gzip` case

In my understanding it's harmless because it doesn't exceed the limit as a result.

ashie:
@kenhys Could you take a look #1849 too?
Although it's a problem that is different from the one we are going to solve (whole size of a chunk vs 1 message's size), but a very near issue.

test/plugin/test_buffer.rb
kenhys: It is reasonable, thanks.

kenhys:
> @kenhys Could you take a look #1849 too?

okay, I'll check it.

lib/fluent/plugin/buffer.rb
kenhys: > Doesn't it have performance regression?

Yes, noted in a commit message, it has performance regression.
I'll reconsider another approach again.

kenhys:
With extending memory chunk approach (PoC), it seems that it overcome the performance regression:

```
 Before:

      no BufferChunkOverflowError
                              11.488k i/100ms
      raise BufferChunkOverflowError in every 100 calls
                               4.110k i/100ms
      no BufferChunkOverflowError
                              113.353k (± 1.0%) i/s -      1.137M in  10.034300s
      raise BufferChunkOverflowError in every 100 calls
                               40.599k (± 3.5%) i/s -    406.890k in  10.034076s

      Comparison:
      no BufferChunkOverflowError:   113353.4 i/s
      raise BufferChunkOverflowError in every 100 calls:    40599.5 i/s - 2.79x  (± 0.00) slower

    After:

      no BufferChunkOverflowError
                              11.399k i/100ms
      raise BufferChunkOverflowError in every 100 calls
                               4.301k i/100ms
      no BufferChunkOverflowError
                              113.855k (± 1.1%) i/s -      1.140M in  10.013090s
      raise BufferChunkOverflowError in every 100 calls
                               43.683k (± 4.0%) i/s -    438.702k in  10.059251s

      Comparison:
      no BufferChunkOverflowError:   113854.6 i/s
      raise BufferChunkOverflowError in every 100 calls:    43683.3 i/s - 2.61x  (± 0.00) slower

```

kenhys:
TODO: also do the same thing for file chunk.

ashie:
Looks better than before.
But I'm wondering it's really needed to modify chunk's code.
I think the size can be determined before calling `chunk.append` or `chunk.concat`.

https://github.com/fluent/fluentd/blob/2722b799e2aba3f82c279deb5d554fb44c077f4d/lib/fluent/plugin/buffer.rb#L774-L778

https://github.com/fluent/fluentd/blob/2722b799e2aba3f82c279deb5d554fb44c077f4d/lib/fluent/plugin/buffer/chunk.rb#L87-L99

kenhys:
tried another approach without modifying chunk code: just calculate size in advance.
It also has no performance drawbacks.

kenhys:
As for #1849, all 3 records exceeds the chunk_size_limit.

https://github.com/fluent/fluentd/issues/1849#issuecomment-968781444

```
2021-11-16 16:45:55 +0900 [warn]: #0 emit transaction failed: error_class=Fluent::Plugin::Buffer::BufferChunkOverflowError error="a 67 bytes record (nth: 0) is larger than buffer chunk limit size, a 67 bytes record (nth: 1) is larger than buffer chunk limit size, a 67 bytes record (nth: 2) is larger than buffer chunk limit size" location="/home/kenhys/.rvm/gems/ruby-2.7.4@fluentd-test/gems/fluentd-1.14.2/lib/fluent/plugin/buffer.rb:454:in `write'" tag="tail.log"
```

ashie:
> As for #1849, all 3 records exceeds the chunk_size_limit.

Hmm, it's same problem with this?

kenhys:
> Hmm, it's same problem with this?

I think so.

* https://github.com/fluent/fluentd/issues/1849#issue-295004393
  * indicates possibility to clean up overflow logic
* https://github.com/fluent/fluentd/issues/1849#issuecomment-968781444
  * point out data loss (same as #3553)

It can be improved overflow logic, butI'm not sure yet.

ashie:
DCO check is failed: https://github.com/fluent/fluentd/pull/3553/checks?check_run_id=4221605400

lib/fluent/plugin/buffer.rb
```diff
@@ -761,6 +771,15 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                if split.size == 1 && original_bytesize == 0
+                  big_record_size = format ? format.call(split).bytesize : split.first.bytesize
```
ashie, line 775: The result of `format.call(split)` should be stored to pass it to following `chunk.concat`.
Otherwise `format.call(split)` will be called twice on non-error case, it's wasteful.

kenhys: Got it, I've fixed it.

test/plugin/test_buffer.rb
```diff
@@ -944,6 +944,52 @@ def create_chunk_es(metadata, es)
...
+
+      assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError.new(messages.join(", ")) do
+        @p.write({@dm0 => es}, format: @format)
+      end
+      # message a and b are concatenated and staged
+      staged_messages = Fluent::MessagePackFactory.msgpack_unpacker.feed_each(@p.stage[@dm0].chunk).collect do |record|
+        record.last
+      end
+      assert_equal([2, staged_messages],
+                   [@p.stage[@dm0].size, [{"message" => "a"}, {"message" => "b"}]])
```
ashie, line 988: The expected data and the actual data seem inverted.

`staged_messages` is the actual data and ` [{"message" => "a"}, {"message" => "b"}]` is the expected data, aren't they?

kenhys: Ah, yes, it's inverted. I've fixed.

ashie:
> > As for #1849, all 3 records exceeds the chunk_size_limit.
> 
> Hmm, it's same problem with this?

No, it's a different problem.
For exmaple, in the case of https://github.com/fluent/fluentd/issues/1849#issuecomment-968781444, `BufferChunkOverflow` is occurred on writing 2nd record although all records don't exceed `chunk_size_limit`. They should be separated into different chunks.

We should work on it in another pull request.

lib/fluent/plugin/buffer.rb
```diff
@@ -761,20 +771,25 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                formatted_split = format ? format.call(split) : split.first
+                if split.size == 1 && original_bytesize == 0
+                  big_record_size = formatted_split.bytesize
+                  if chunk.bytesize + big_record_size > @chunk_limit_size
+                    errors << "a #{big_record_size} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size"
+                    writing_splits_index += 1
+                    next
+                  end
```
ashie, line 781: It's OK for the case that `format` is given.
But the case on `format == nil` and `@compress == :gzip`, `BufferChunkOverflowError` might not occurred in actual.
In such case, doing `chunk.commit` here and delaying to store errors  until `if chunk_size_over?(chunk)` block is better like you did in the previous patch,

ashie: > But the case on `format == nil` and `@compress == :gzip`, `BufferChunkOverflowError` might not occurred in actual. 

https://github.com/fluent/fluentd/blob/46dad90aac2d463bf3d5b0d1be372ac97c0d0724/lib/fluent/plugin/buffer/chunk.rb#L187-L199

kenhys: okay, I'll fix  it.

kenhys: Fixed.

kenhys:
A bit strange because it pass on my local environment.

```
% bundle exec ruby test/plugin/test_buffer.rb 
Loaded suite test/plugin/test_buffer
Started
..............................................................
Finished in 0.579545554 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------
62 tests, 376 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------
106.98 tests/s, 648.78 assertions/s
```

kenhys:
Even though `bundle exec rake test` on my local, it passes. (not depending the other test cases)

kenhys:
```
      c = @p.generate_chunk(create_metadata)
      c.append(Fluent::ArrayEventStream.new([timestamp, {"message" => "0123"}]), compress: :gzip)
      p c.bytesize
```

it shows 49 when invoked from test case. it is expected 71

kenhys:
> it shows 49 when invoked from test case. it is expected 71

digging internal flow, message "0123", "aaa", "bbb" are occupy 71, 70,70 bytes (chunk.bytesize)

kenhys:
chunk.bytesize 49:

```
#<Fluent::CompressedMessagePackEventStream:0x0000560ddcbb0a68 @data="1460390402{\"message\"=>\"0123\"}", @size=29, @unpacked_times=[49, 52, 54, 48, 51, 57, 48, 52, 48, 50, 123, 34, 109, 101, 115, 115, 97, 103, 101, 34, 61, 62, 34, 48, 49, 50, 51, 34, 125], @unpacked_records=[nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil], @decompressed_data="1460390402{\"message\"=>\"0123\"}", @compressed_data="\x1F\x8B\b\x00!\xDD\x95a\x00\x0334130\xB64010\xAAV\xCAM-.NLOU\xB2\xB5S2042V\xAA\x05\x00D\xC0\xA2C\x1D\x00\x00\x00">
```

chunk.bytesize: 71
```
#<Fluent::CompressedMessagePackEventStream:0x0000560ddcbad2f0 @data="[2016-04-12 01:00:02.000000000 +0900, {\"message\"=>\"0123\"}]", @size=58, @unpacked_times=[91, 50, 48, 49, 54, 45, 48, 52, 45, 49, 50, 32, 48, 49, 58, 48, 48, 58, 48, 50, 46, 48, 48, 48, 48, 48, 48, 48, 48, 48, 32, 43, 48, 57, 48, 48, 44, 32, 123, 34, 109, 101, 115, 115, 97, 103, 101, 34, 61, 62, 34, 48, 49, 50, 51, 34, 125, 93], @unpacked_records=[nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil], @decompressed_data="[2016-04-12 01:00:02.000000000 +0900, {\"message\"=>\"0123\"}]", @compressed_data="\x1F\x8B\b\x00!\xDD\x95a\x00\x03\x8B6204\xD350\xD154R00\xB420\xB020\xD23\x80\x01\x05m\x03K\x03\x03\x1D\x85j\xA5\xDC\xD4\xE2\xE2\xC4\xF4T%[;%\x03C#c\xA5\xDAX\x00\x17\x9D\xF9\x82:\x00\x00\x00">
```

kenhys:
This glitch caused by the following:

*  `c.append(Fluent::ArrayEventStream.new([timestamp, {"message" => "0123"}]), compress: :gzip)` 49bytes
*   `c.append(Fluent::ArrayEventStream.new([[timestamp, {"message" => "0123"}]]), compress: :gzip)` 71bytes

kenhys:
check the boundary value in CI.

kenhys:
Got it.

```
Failure: test: #write compressed data which exceeds chunk_limit_size, it raises BufferChunkOverflowError(BufferTest::when compress is gzip)
/home/runner/work/fluentd/fluentd/test/plugin/test_buffer.rb:1274:in `block (2 levels) in <class:BufferTest>'
     1271:       # ensure the boundary value
     1272:       c = @p.generate_chunk(create_metadata)
     1273:       c.append(Fluent::ArrayEventStream.new([[timestamp, {"message" => "0123"}]]), compress: :gzip)
  => 1274:       assert_equal 71, c.bytesize
     1275: 
     1276:       messages = "a 71 bytes record (nth: 0) is larger than buffer chunk limit size"
     1277:       assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError.new(messages) do
<71> expected but was
<69>
```

kenhys:
Added workaround a147de5, it seem to work well.

kenhys:
NOTE:

https://github.com/fluent/fluentd/pull/3553#issuecomment-972552443 msgpack version is same 1.4.2.
69 bytes: libmsgpackc2 3.0.1-3
71 bytes: libmsgpackc2 3.3.0-4
but, even though running test on focal, it does not match to 69bytes.

lib/fluent/plugin/buffer.rb
```diff
@@ -761,18 +771,41 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                formatted_split = format ? format.call(split) : split.first
+                if split.size == 1 && original_bytesize == 0
+                  if format == nil && @compress
```
ashie, line 776: `@compress` is `:text` or `:gzip`

https://github.com/fluent/fluentd/blob/6d386249f3225ffa5794d5a45afe19c9adbbabb5/lib/fluent/plugin/buffer.rb#L67

Only `:gzip` is undetermined here.

kenhys: Thanks, I'll check it.

kenhys: Added check for `:gzip`

lib/fluent/plugin/buffer.rb
```diff
@@ -761,18 +771,41 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                formatted_split = format ? format.call(split) : split.first
+                if split.size == 1 && original_bytesize == 0
+                  if format == nil && @compress == :gzip
```
ashie, line 776: `@compress != :text` is more robust against future expanding.

kenhys: Thanks, It is reasonable!

kenhys: Fixed.

kenhys:
@ashie what is the blocker still in this PR?

cosmo0920 (approved):
Looks good to me.
With the micro benchmark on my MacBook Pro M1Pro environment:

```console
$ sysctl machdep.cpu.brand_string
machdep.cpu.brand_string: Apple M1 Pro
$ sysctl -a machdep.cpu 
machdep.cpu.brand_string: Apple M1 Pro
machdep.cpu.core_count: 10
machdep.cpu.cores_per_package: 10
machdep.cpu.logical_per_package: 10
machdep.cpu.thread_count: 10
```

```diff
diff --git a/test/plugin/test_buffer.rb b/test/plugin/test_buffer.rb
index acf48b7d..2a494fb4 100644
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -990,6 +990,30 @@ class BufferTest < Test::Unit::TestCase
       assert_equal [@dm0], @p.queue.map(&:metadata)
       assert_equal [5000], @p.queue.map(&:size)
     end
+
+    test '#write BufferChunkOverflowError micro benchmark' do
+      require 'benchmark/ips'
+      es = Fluent::ArrayEventStream.new([ [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x"}] ])
+      large = Fluent::ArrayEventStream.new([ [event_time('2016-04-11 16:00:02 +0000'), {"message" => "x" * 1_280_000}] ])
+      Benchmark.ips do |x|
+        x.config(:time => 10, :warmup => 2)
+        x.report("no BufferChunkOverflowError") do
+          @p.write({@dm0 => es}, format: @format)
+        end
+        x.report("raise BufferChunkOverflowError in every 100 calls") do |n|
+          n.times do |i|
+            if i % 100 == 0
+              assert_raise Fluent::Plugin::Buffer::BufferChunkOverflowError do
+                @p.write({@dm0 => large}, format: @format)
+              end
+            else
+              @p.write({@dm0 => es}, format: @format)
+            end
+          end
+        end
+        x.compare!
+      end
+    end
   end
 
   sub_test_case 'custom format with configuration for test with lower chunk limit size' do
```

```console
no BufferChunkOverflowError
                        13.682k i/100ms
raise BufferChunkOverflowError in every 100 calls
                         2.938k i/100ms
no BufferChunkOverflowError
                        113.712k (±15.0%) i/s -      1.122M in  10.077790s
raise BufferChunkOverflowError in every 100 calls
                         49.707k (±23.5%) i/s -    470.080k in  10.034925s

Comparison:
no BufferChunkOverflowError:   113711.8 i/s
raise BufferChunkOverflowError in every 100 calls:    49706.8 i/s - 2.29x  (± 0.00) slower
```

No severe performance degradation is observed.

kenhys:
Changed not to extend chunk API.

lib/fluent/plugin/buffer.rb
```diff
@@ -761,18 +771,41 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
+
                 if format
-                  chunk.concat(format.call(split), split.size)
+                  chunk.concat(formatted_split, split.size)
                 else
                   chunk.append(split, compress: @compress)
                 end
 
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
+                  adding_bytes = chunk.instance_eval { @adding_bytes }
```
ashie, line 798: 3rd party plugins might not have `@adding_bytes`, we should care such case too.
e.g.)
`adding_bytes = chunk.instance_eval { @adding_bytes } || "?"`

kenhys: Fixed, thanks!

lib/fluent/plugin/buffer.rb
```diff
@@ -761,18 +771,41 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
+                formatted_split = format ? format.call(split) : split.first
+                if split.size == 1 && original_bytesize == 0
+                  if format == nil && @compress != :text
+                    # The actual size of chunk is not determined until after chunk.append.
+                    # so, keep already processed 'split' content here.
+                    # (allow performance regression a bit)
+                    chunk.commit
+                  elsif format
```
ashie, line 781: What about `format == nil && @compress == :text`?

kenhys: Maybe `else` is enough.

lib/fluent/plugin/buffer.rb
```diff
@@ -352,6 +354,9 @@ def write(metadata_and_data, format: nil, size: nil, enqueue: false)
                 unstaged_chunks[metadata] ||= []
                 unstaged_chunks[metadata] << chunk
               end
+              if error and not error.empty?
```
ashie, line 357:
```diff
- not error.empty?
+ !error.empty?
```
is preferred in Fluend's code

kenhys: Yes, and it should be `if error && !error.empty?`, too.

ashie (approved):
Waiting CI...

ashie:
Thanks for your work!
TODO: Additional work is required for #1849

cosmo0920:
:tada:

PR #3558, opened 2021-11-16
Fix unstable command execution tests

ashie:
Fixes #

Sometimes some command execution tests failed on CI (especially on Drone CI, but occurs also on others).
It seems that it's caused by Ruby interpreter's launch cost and fewer timeout setting.
They seems suppressed by using `echo` command and extending the timeout.

**Docs Changes**:
None

**Release Note**: 
Same with the title

### DIFF
--- a/test/plugin_helper/test_child_process.rb
+++ b/test/plugin_helper/test_child_process.rb
@@ -319,12 +319,12 @@ class ChildProcessTest < Test::Unit::TestCase
 
   test 'can execute external command many times, which finishes immediately' do
     ary = []
-    arguments = ["-e", "puts 'okay'; STDOUT.flush rescue nil"]
+    arguments = ["okay"]
     Timeout.timeout(TEST_DEADLOCK_TIMEOUT) do
-      @d.child_process_execute(:t5, "ruby", arguments: arguments, interval: 1, mode: [:read]) do |io|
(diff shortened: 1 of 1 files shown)


test/plugin_helper/test_child_process.rb
```diff
@@ -335,12 +335,12 @@ def configure(conf)
 
   test 'can execute external command many times, with leading once executed immediately' do
     ary = []
-    arguments = ["-e", "puts 'okay'; STDOUT.flush rescue nil"]
+    arguments = ["okey"]
     Timeout.timeout(TEST_DEADLOCK_TIMEOUT) do
```
kenhys, line 339: This change okay => okey intentionally? ( to distingish test case ?)

kenhys (approved):
I didn't notice it, surely no need to be ruby.
If drone pass, we can merge it.l

test/plugin_helper/test_child_process.rb
ashie: No, I'll fix it.

ashie:
> If drone pass, we can merge it.

It will still fail because some other unstable tests are remained.
I'll continue the work to fix them.

test/plugin_helper/test_child_process.rb
ashie: Fixed.

ashie:
CI is still failed, but fixed tests aren't failed anymore.

PR #3560, opened 2021-11-22
Don't raise exception when each message size is smaller than chunk_limit_size

### LINKED ISSUE #1849: Large message overflow logic incorrect

kenhys:
Follow up #3553

In the previous versions, even though each record size is smaller
than chunk limit size, but whole message size exceeds
chunk limit size, BufferChunkOverflowError is raised unexpectedly.

For example, if chunk limit size is 1_280_000, process 3 event
stream (every 1_000_00 bytes), it throws an exception like this:

  Fluent::Plugin::Buffer::BufferChunkOverflowError(<a 1000025 bytes
  record (nth: 1) is larger than buffer chunk limit size, a 1000025
  bytes record (nth: 2) is larger than buffer chunk limit size>)

Now changed not to raise exception if it's record size is smaller
enough than chunk limit size.

The idea is based on that adding byte size is smaller than chunk
limit size, chunk should be unstaged and pushed into queue,
If not, it should be skipped like #3553.

Signed-off-by: Kentaro Hayashi <hayashi@clear-code.com>

Fixes #1849

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -767,7 +767,7 @@ module Fluent
             raise ShouldRetry unless chunk.writable?
             staged_chunk_used = true if chunk.staged?
 
-            original_bytesize = chunk.bytesize
+            original_bytesize = committed_bytesize = chunk.bytesize
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
@@ -778,12 +778,18 @@ module Fluent
                     # so, keep already processed 'split' content here.
                     # (allow performance regression a bit)
                     chunk.commit
+                    committed_bytesize = chunk.bytesize
                   else
                     big_record_size = formatted_split.bytesize
-                    if chunk.bytesize + big_record_size > @chunk_limit_size
+                    if big_record_size > @chunk_limit_size
+                      # Just skip to next split (current split is ignored)
                       errors << "a #{big_record_size} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
                       writing_splits_index += 1
                       next
+                    elsif chunk.bytesize + big_record_size > @chunk_limit_size
+                      # No doubt that the split is expected to cause size over, keep 'split' content here.
+                      chunk.commit
+                      committed_bytesize = chunk.bytesize
                     end
                   end
                 end
@@ -793,19 +799,29 @@ module Fluent
                 else
                   chunk.append(split, compress: @compress)
                 end
+                adding_bytes = chunk.bytesize - committed_bytesize
 
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
-                  adding_bytes = chunk.instance_eval { @adding_bytes } || "N/A" # 3rd party might not have 'adding_bytes'
                   chunk.rollback
-
+                  committed_bytesize = chunk.bytesize
                   if split.size == 1 && original_bytesize == 0
-                    # It is obviously case that BufferChunkOverflowError should be raised here,
-                    # but if it raises here, already processed 'split' or
-                    # the proceeding 'split' will be lost completely.
-                    # so it is a last resort to delay raising such a exception
-                    errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
-                    writing_splits_index += 1
-                    next
+                    if adding_bytes < @chunk_limit_size
+                      # As already processed content is kept after rollback, then unstaged chunk should be queued.
+                      # After that, re-process current split again.
+                      # New chunk should be allocated, to do it, modify @stage and so on.
+                      synchronize { @stage.delete(modified_metadata) }
+                      staged_chunk_used = false
+                      chunk.unstaged!
+                      break
+                    else
+                      # It is obviously case that BufferChunkOverflowError should be raised here,
+                      # but if it raises here, already processed 'split' or
+                      # the proceeding 'split' will be lost completely.
+                      # so it is a last resort to delay raising such a exception
+                      errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
+                      writing_splits_index += 1
+                      next
+                    end
                   end
 
                   if chunk_size_full?(chunk) || split.size == 1
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -771,26 +771,33 @@ module Fluent
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
-                formatted_split = format ? format.call(split) : split.first
-                if split.size == 1 && original_bytesize == 0
-                  if format == nil && @compress != :text
-                    # The actual size of chunk is not determined until after chunk.append.
-                    # so, keep already processed 'split' content here.
-                    # (allow performance regression a bit)
+                formatted_split = format ? format.call(split) : nil
+
+                if split.size == 1 # Check BufferChunkOverflowError
+                  determined_bytesize = nil
+                  if @compress != :text
+                    determined_bytesize = nil
+                  elsif formatted_split
+                    determined_bytesize = formatted_split.bytesize
+                  elsif split.first.respond_to?(:bytesize)
+                    determined_bytesize = split.first.bytesize
+                  end
+
+                  if determined_bytesize && determined_bytesize > @chunk_limit_size
+                    # It is a obvious case that BufferChunkOverflowError should be raised here.
+                    # But if it raises here, already processed 'split' or
+                    # the proceeding 'split' will be lost completely.
+                    # So it is a last resort to delay raising such a exception
+                    errors << "a #{determined_bytesize} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
+                    writing_splits_index += 1
+                    next
+                  end
+
+                  if determined_bytesize.nil? || chunk.bytesize + determined_bytesize > @chunk_limit_size
+                    # The split will (might) cause size over so keep already processed
+                    # 'split' content here (allow performance regression a bit).
                     chunk.commit
                     committed_bytesize = chunk.bytesize
-                  else
-                    big_record_size = formatted_split.bytesize
-                    if big_record_size > @chunk_limit_size
-                      # Just skip to next split (current split is ignored)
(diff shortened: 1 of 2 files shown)


kenhys:
updates about error message is split into #3562

kenhys:
This PR relies on `adding_bytes`, it might not be implemented 3rd party buffer plugin.

kenhys:
rebased with the recent master.

kenhys:
This PR will not be merged for v1.14.3.

cosmo0920:
> This PR will not be merged for v1.14.3.

This PR needs to be more testing required?

ashie:
Yes, we need further checking for this.
Skip v1.14.3.

lib/fluent/plugin/buffer.rb
```diff
@@ -799,13 +803,23 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
                   if split.size == 1 && original_bytesize == 0
-                    # It is obviously case that BufferChunkOverflowError should be raised here,
-                    # but if it raises here, already processed 'split' or
-                    # the proceeding 'split' will be lost completely.
-                    # so it is a last resort to delay raising such a exception
-                    errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
-                    writing_splits_index += 1
-                    next
+                    # FIXME: adding_bytes should be implemented as an official chunk API in the future
+                    if adding_bytes.is_a?(Integer) && adding_bytes < @chunk_limit_size
```
ashie, line 807: I'm not sure but `adding_bytes` might be able to calculate at caller side like this, because `adding_bytes` is cleared at `chunk.commit`, `chunk.rollback`, `chunk.purge`:

```diff
diff --git a/lib/fluent/plugin/buffer.rb b/lib/fluent/plugin/buffer.rb
index 5877af93..4df0dcb9 100644
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -767,7 +767,7 @@ module Fluent
             raise ShouldRetry unless chunk.writable?
             staged_chunk_used = true if chunk.staged?
 
-            original_bytesize = chunk.bytesize
+            original_bytesize = before_bytesize = chunk.bytesize
             begin
               while writing_splits_index < splits.size
                 split = splits[writing_splits_index]
@@ -778,6 +778,7 @@ module Fluent
                     # so, keep already processed 'split' content here.
                     # (allow performance regression a bit)
                     chunk.commit
+                    before_bytesize = chunk.bytesize
                   else
                     big_record_size = formatted_split.bytesize
                     if big_record_size > @chunk_limit_size
@@ -788,6 +789,7 @@ module Fluent
                     elsif chunk.bytesize + big_record_size > @chunk_limit_size
                       # No doubt that the split is expected to cause size over, keep 'split' content here.
                       chunk.commit
+                      before_bytesize = chunk.bytesize
                     end
                   end
                 end
@@ -797,14 +799,14 @@ module Fluent
                 else
                   chunk.append(split, compress: @compress)
                 end
+                adding_bytes = chunk.bytesize - before_bytesize
 
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
-                  adding_bytes = chunk.instance_eval { @adding_bytes } || "N/A" # 3rd party might not have 'adding_bytes'
                   chunk.rollback
+                  before_bytesize = chunk.bytesize
 
                   if split.size == 1 && original_bytesize == 0
-                    # FIXME: adding_bytes should be implemented as an official chunk API in the future
-                    if adding_bytes.is_a?(Integer) && adding_bytes < @chunk_limit_size
+                    if adding_bytes < @chunk_limit_size
                       # already processed content is kept after rollback, then unstaged chunk should be queued.
                       # After that, re-process current split again (new chunk should be allocated, to do it, modify @stage and so on)
                       synchronize { @stage.delete(modified_metadata) }
```

kenhys: Thanks, this suggestion is more good approach to do the job.

kenhys:
I've followed suggested approach (as before_bytesize is too general, so it was changed to represent what it is)

ashie:
I'm still checking this, especially the difference of `original_bytesize` and `committed_bytesize`...

lib/fluent/plugin/buffer.rb
```diff
@@ -793,19 +799,29 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
                   if split.size == 1 && original_bytesize == 0
-                    # It is obviously case that BufferChunkOverflowError should be raised here,
-                    # but if it raises here, already processed 'split' or
-                    # the proceeding 'split' will be lost completely.
-                    # so it is a last resort to delay raising such a exception
-                    errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
-                    writing_splits_index += 1
-                    next
+                    if adding_bytes < @chunk_limit_size
+                      # As already processed content is kept after rollback, then unstaged chunk should be queued.
```
ashie, line 809: I'm not sure but I think this block should be placed outer of `if split.size == 1 && original_bytesize == 0` block because probably this case can be occurred also on `split.size != 1` or `original_bytesize != 0`.

```diff
diff --git a/lib/fluent/plugin/buffer.rb b/lib/fluent/plugin/buffer.rb
index 7275d70c..f113f06b 100644
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -804,24 +804,23 @@ module Fluent
                 if chunk_size_over?(chunk) # split size is larger than difference between size_full? and size_over?
                   chunk.rollback
                   committed_bytesize = chunk.bytesize
-                  if split.size == 1 && original_bytesize == 0
-                    if adding_bytes < @chunk_limit_size
-                      # As already processed content is kept after rollback, then unstaged chunk should be queued.
-                      # After that, re-process current split again.
-                      # New chunk should be allocated, to do it, modify @stage and so on.
-                      synchronize { @stage.delete(modified_metadata) }
-                      staged_chunk_used = false
-                      chunk.unstaged!
-                      break
-                    else
-                      # It is obviously case that BufferChunkOverflowError should be raised here,
-                      # but if it raises here, already processed 'split' or
-                      # the proceeding 'split' will be lost completely.
-                      # so it is a last resort to delay raising such a exception
-                      errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
-                      writing_splits_index += 1
-                      next
-                    end
+                  if split.size == 1 && adding_bytes > @chunk_limit_size
+                    # It is obviously case that BufferChunkOverflowError should be raised here,
+                    # but if it raises here, already processed 'split' or
+                    # the proceeding 'split' will be lost completely.
+                    # so it is a last resort to delay raising such a exception
+                    errors << "concatenated/appended a #{adding_bytes} bytes record (nth: #{writing_splits_index}) is larger than buffer chunk limit size (#{@chunk_limit_size})"
+                    writing_splits_index += 1
+                    next
+                  end
+                  if adding_bytes < @chunk_limit_size
+                    # As already processed content is kept after rollback, then unstaged chunk should be queued.
+                    # After that, re-process current split again.
+                    # New chunk should be allocated, to do it, modify @stage and so on.
+                    synchronize { @stage.delete(modified_metadata) }
+                    staged_chunk_used = false
+                    chunk.unstaged!
+                    break
                   end
 
                   if chunk_size_full?(chunk) || split.size == 1
```

In addition, note that above patch includes the following change for detecting 1 big record:

```diff
- if split.size == 1 && original_bytesize == 0
+ if split.size == 1 && adding_bytes > @chunk_limit_size
```

I think it's more proper way to detect it.
(Without it, newly added tests are failed.)

ashie: > because probably this case can be occurred also on `split.size != 1` or `original_bytesize != 0`.

Hmm, probably I was wrong. It seems that it doesn't occur when `split.size != 1`.

ashie: I made it more understandable (for me...)

ashie (approved):

ashie:
Waiting CI...

lib/fluent/plugin/buffer.rb
```diff
@@ -793,19 +806,26 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
+                      writing_splits_index += 1
+                      next
+                    else
+                      # As already processed content is kept after rollback, then unstaged chunk should be queued.
+                      # After that, re-process current split again.
+                      # New chunk should be allocated, to do it, modify @stage and so on.
+                      synchronize { @stage.delete(modified_metadata) }
+                      staged_chunk_used = false
+                      chunk.unstaged!
+                      break
```
daipom, line 827: Why does this unstage the current chunk?

Issue #3193, opened 2020-12-09
in_http plugin: supporting http auth


kenhys:
**Is your feature request related to a problem? Please describe.**

This feature request is derived from the Fluentd Mailing list question.

* [http auth for in_http plugin](https://groups.google.com/g/fluentd/c/5dsZR4NCUII)

fluentd supports Basic Auth by http output plugin,

https://docs.fluentd.org/output/http#less-than-auth-greater-than-section

but it seems that http input plugin doesn't support Basic Auth.
Because of this inconsistency, it can't send directly between two fluentd instances.

**Describe the solution you'd like**

Implement Basic Auth support in input http plugin side.

**Describe alternatives you've considered**

N/A

Or give up to use Basic Auth and use an alternative method (TLS encryption)

**Additional context**

N/A

a-b-v:
Why do you use http between two fluentd instead forward ?

kenhys:
This issue is derived from google groups question:

http auth for in_http plugin
https://groups.google.com/g/fluentd/c/5dsZR4NCUII/m/2DnKjDFjAwAJ

Practically, maybe in forward is good enough but for the record, I've created this issue.

PR #3572, opened 2021-12-13
in_http: Verify Authroization token on HTTP headers

cosmo0920:

Fixes https://github.com/fluent/fluentd/issues/3193

`in_http` plugins does not provide `Authorization` header verification.
This PR adds Authorization headers handling in `in_http` plugin.
And this logic can be handle not only basic authentication but also other types of authentications.
ref: https://developer.mozilla.org/ja/docs/Web/HTTP/Headers/Authorization

**Docs Changes**:

Related doc PR is: https://github.com/fluent/fluentd-docs-gitbook/pull/382

**Release Note**: 

Same as title.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -84,6 +84,8 @@ module Fluent::Plugin
     config_param :dump_error_log, :bool, default: true
     desc 'Add QUERY_ prefix query params to record'
     config_param :add_query_params, :bool, default: false
+    desc 'Handle Authroization header authentication with specified secret'
+    config_param :authorization_token, :string, default: nil, secret: true
 
     config_section :parse do
       config_set_default :@type, 'in_http'
@@ -290,7 +292,7 @@ module Fluent::Plugin
       handler = Handler.new(conn, @km, method(:on_request),
                             @body_size_limit, @format_name, log,
                             @cors_allow_origins, @cors_allow_credentials,
-                            @add_query_params)
+                            @add_query_params, @authorization_token)
 
       conn.on(:data) do |data|
         handler.on_read(data)
@@ -368,7 +370,7 @@ module Fluent::Plugin
       attr_reader :content_type
 
       def initialize(io, km, callback, body_size_limit, format_name, log,
-                     cors_allow_origins, cors_allow_credentials, add_query_params)
+                     cors_allow_origins, cors_allow_credentials, add_query_params, authorization_token)
         @io = io
         @km = km
         @callback = callback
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_http.rb
```diff
@@ -84,6 +84,8 @@ class HttpInput < Input
     config_param :dump_error_log, :bool, default: true
     desc 'Add QUERY_ prefix query params to record'
     config_param :add_query_params, :bool, default: false
+    desc 'Handle Authroization header authentication with specified secret'
+    config_param :authorization_token, :string, default: nil, secret: true
```
kenhys, line 89: `:authorization_token` parameter name may be confused because it is  actually pair of  `<type> <credential>`.
`:authorization_fields` ? or something maybe better.

kenhys: It seems okay as is.

ashie:
Using a fixed `Authentication` header is almost nonsense, it seems that you need to learn the basis and purpose of authentication.
I think implementing basic auth and using it with TLS is enough for current Fluentd. In it, multi-user should be acceptable.
Implementation for it will be completely different from this patch, so I close this.

agup006:
Hey @ashie while this may not be the proper approach, I would ask that you still be respectful in feedback/comments. Please refer to the CNCF code of conduct that we use for the Fluentd project (https://github.com/cncf/foundation/blob/master/code-of-conduct.md)

ashie:
Sorry, I didn't mean to hurt your feelings.
BTW we did further discussion about it at private chat how to resolve your requirements.

cosmo0920:
Ah, it would be nice if we had done more discussing why this PR should be rejected.
But now, I didn't hurt for the declined decision. I didn't notice this PR is not fully implemented Digest authentication type and lack of considering like Log4jShell shocks.

bertramn:
So why was this PR denied? Looks like a valid feature to me.

cosmo0920:
This PR wasn't fully considered for digest authentication. The "fixed" authentication credentials makes vulnerabilities for the attackers. We wouldn't want to introduce such fragile feature.

ashie:
In addition, it doesn't support multi-users.

If we merge this it once, we need to keep backward compatibility persistently.
But proper implementation will conflict with this, so we will need to add switching mechanism and keep both implementations.
It's quite stressful for us.

bertramn:
Yes these auth features generally need to delegate the handling of the presented credentials to some pluggable handler (basic and digest auth) or an identity provider (bearer tokens). I stumbled onto this when looking for an ingestion endpoint for JSON based logs. I need to secure the http endpoint with some form of key that is unique to each log producer. The keys need to be manageable without restarting the endpoint process. Sort of like Slunk HEC or the thing that the others like logz.io or logdna do. Thanks for clarifying the reason!

agup006:
For splunk hec we do have https://github.com/calyptia/fluent-plugin-in-http-splunk-hec if that helps

PR #3543, opened 2021-11-01
Investigate stalled tests on Ruby 3.1

ashie:
Signed-off-by: Takuro Ashie <ashie@clear-code.com>

**Which issue(s) this PR fixes**: 
Fixes #

**What this PR does / why we need it**: 

**Docs Changes**:

**Release Note**:

PR #3577, opened 2021-12-21
test: Fix stalled tests on Ruby 3.1

ashie:
Fixes #3543

Several tests are stalled since Ruby 3.1, I found that it's caused by
the followig Ruby's change:

  https://github.com/ruby/ruby/commit/2d98593bf54a37397c6e4886ccc7e3654c2eaf85

It seems that there are some conditions to reproduce it:

  * `attr_reader` is defined in a class
  * `$log` is accessed by it
  * `assert` is called with a block
  * `assert` refers the `$log` via the `attr_reader`
  * ... (some other conditions seem needed but not clarified yet)

Here is the sample test case to reproduce it:

```ruby
require 'fluent/test'

class StallTest < Test::Unit::TestCase
  class Stall
    attr_reader :log
    def initialize
      @log = $log
    end
  end

  setup do
    Fluent::Test.setup
  end

  20.times do |i|
    test "stall #{i}" do
      assert { Stall.new.log }
    end
  end
end
```

Becasue all stalled tests don't need to use block, I remove these blocks
to avoid this issue.

**Docs Changes**:
None.

**Release Note**: 
Same with the title.

### DIFF
--- a/test/plugin/test_bare_output.rb
+++ b/test/plugin/test_bare_output.rb
@@ -83,7 +83,7 @@ class BareOutputTest < Test::Unit::TestCase
 
     @p.configure(config_element('ROOT', '', {'@log_level' => 'debug'}))
 
-    assert{ @p.log.object_id != original_logger.object_id }
+    assert(@p.log.object_id != original_logger.object_id)
     assert_equal Fluent::Log::LEVEL_DEBUG, @p.log.level
   end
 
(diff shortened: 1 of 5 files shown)


fujimotos (approved):

fujimotos:
I was initially confused by this patch, thinking "welp, isn't this patch
just removing spaces? Where's the actual diff?", and realized that the
fix is about converting `{}` to `()`. 
Maybe I need to buy a pair of glasses.

PR #3579, opened 2021-12-21
Make out_file append option work without compression

### LINKED ISSUE #3569: out_file: events except those in the first chunk are not written to files when append is true
### Describe the bug
Events except those in the first chunk are not written to files when `append` is true for `out_file` plugin.

### To Reproduce

1. Run `fluentd --setup` to generate `fluent.conf`.
2. Put the following content into `fluent.conf`.
```
<match hack.**>
  @type file
  path hack
  append true
  <format>
    @type json
  </format>
  <buffer>
    flush_mode immediate
  </buffer>
</match>
```
3. Run the following command in a terminal:
````
$ curl -X POST http://localhost:8888/hack.access  -d 'json={"no":"1"}'
$ cat hack.20211202.log
{"no":"1"}
$ curl -X POST http://localhost:8888/hack.access  -d 'json={"no":"2"}'
$ cat hack.20211202.log
{"no":"1"}
````

### Expected behavior

The content of `hack.20211202.log` should be:
```
{"no":"1"}
{"no":"2"}
```
but was:
```
{"no":"1"}
```

### Your Environment

- Fluentd version: 1.14.3
- TD Agent version: 1.14.3
- Operating system: macOS 10.15.5
- Kernel version: 19.5.0
- Ruby version: `ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin19]`

I have used Ruby `gem` to install Fluentd.

### Your Configuration

```apache
<match hack.**>
  @type file
  path hack
  append true
  <format>
    @type json
  </format>
  <buffer>
    flush_mode immediate
  </buffer>
</match>
```

### Your Error Log

```shell
No error log
```

### Additional context

 I have verified that `fluentd 1.14.3` which uses `ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-linux]` does not have this prolem on `CentOS Linux release 7.5.1804 (Core)`.  On macOS, I have tried two Ruby 2.x versions. Here is the result:

- `ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]` does not have this problem.
- `ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-darwin19]` __has__ this problem.

yaojingguo:
For Ruby 2.7.x and 3.0.x, out_file append option without compression
does not work. The cause is the following Ruby language bug:

https://bugs.ruby-lang.org/issues/18388.

This commit is a workaround for this problem.

See #3569.

Fixes #3569

**Docs Changes**:
none

**Release Note**: 
out_file: Work around for a bug of Ruby on macOS that `IO.copy_stream` doesn't work as expected with append mode

### DIFF
--- a/lib/fluent/env.rb
+++ b/lib/fluent/env.rb
@@ -33,4 +33,8 @@ module Fluent
   def self.linux?
     /linux/ === RUBY_PLATFORM
   end
+
+  def self.macos?
+    /darwin/ =~ RUBY_PLATFORM
+  end
 end
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -181,6 +181,13 @@ module Fluent::Plugin
       @dir_perm = system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION
       @file_perm = system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION
       @need_lock = system_config.workers > 1
+
+      # https://github.com/fluent/fluentd/issues/3569
+      @need_ruby_on_macos_workaround = false
+      if @append && Fluent.macos?
+        condition = Gem::Dependency.new('', [">= 2.7.0", "< 3.1.0"])
+        @need_ruby_on_macos_workaround = true if condition.match?('', RUBY_VERSION)
+      end
     end
 
     def multi_workers_ready?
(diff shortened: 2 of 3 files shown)


ashie:
Could you follow DCO? https://github.com/fluent/fluentd/pull/3579/checks?check_run_id=4594815458

yaojingguo:
Submitted a new commit with DCO.

ashie:
I've modified that the workaround to be applied only on macOS.
In addition, merge #3580 into this.

ashie (approved):

ashie:
Thanks!

PR #3589, opened 2022-01-05
v1.14.4

fujimotos:
* This is the pull request for upcoming v1.14.4 release.
* If anything is missing from my release note, please comment here.
  (I'm going to fast-forward this branch tomorrow)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+# v1.14.4
+
+## Release v1.14.4 - 2022/01/06
+
+### Enhancement
+
+* `in_tail`: Add option to skip long lines (`max_line_size`)
+  https://github.com/fluent/fluentd/pull/3565
+
+### Bug fix
+
+* Incorrect BufferChunkOverflowError when each event size is < `chunk_limit_size`
+  https://github.com/fluent/fluentd/pull/3560
+* On macOS with Ruby 2.7/3.0, `out_file` fails to write events if `append` is true.
+  https://github.com/fluent/fluentd/pull/3579
+* test: Fix unstable test cases
+  https://github.com/fluent/fluentd/pull/3574
+  https://github.com/fluent/fluentd/pull/3577
+
 # v1.14.3
 
 ## Release v1.14.3 - 2021/11/26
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+# v1.14.4
+
+## Release v1.14.4 - 2022/01/06
+
+### Enhancement
+
+* in_tail: add option to skip long lines (max_line_size)
+  https://github.com/fluent/fluentd/pull/3565
+
+### Bug fix
```
ashie, line 10: Please add #3560 too

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+
+* out_file: Make append option work without compression
```
ashie, line 12: This bug reproduces only on macOS, I want to add a note for it.
See also `Releaes Note` in https://github.com/fluent/fluentd/pull/3579#issue-1085787780

ashie (changes requested):
Please

CHANGELOG.md
kenhys: It is useful if we could explain a bit more. (when it should be used)

kenhys:
It looks good if the above 2 issues were fixed.

ashie (approved):

fujimotos:
So I was poking around the `CHANGELOG.md` this morning. I ended adding a few fixes:

 - The bug-fix section contains more detailed explanation about what was fixed.
    - e.g. `On macOS with Ruby 2.7/3.0, out_file fails to write events if append is true.` 
 - Use backticks to escape underscore characters.

CHANGELOG.md
fujimotos: I added the item to the release note in d1f33da.

CHANGELOG.md
fujimotos: d1f33da added the affected macOS and Ruby versions.

kenhys (approved):
LGTM.

fujimotos:
Merged. Thank you!

PR #3601, opened 2022-01-14
Apply connect_timeout also to TLS handshake

sekiyama58:
Fixes # N/A

When out_forward plugin uses TLS transport, if the TLS handshake takes a long time or hangs,
the flusher (or event_loop for heartbeat) thread waits forever even though connect_timeout is specified.

We encountered this problem in the setup that log-forwarder fluentd sends data to another fluentd cluster under a level 4 (TCP) load balancer.
The load balancer accepts TCP connections, but the backend fluentd (which needs to handle TLS) is too busy and unstable to accept the connections. This caused the TCP connection to stay open and the forwarder to hang.

This fixes the problem by applying connect_timeout to TLS handshake.
To avoid breaking existing environments configured to use the short connect_timeout, this patch counts the TLS handshake time independently of the TCP connection time in this patch.

**Docs Changes**:
N/A

**Release Note**: 
N/A


ashie (approved):
Good catch!

ashie:
Merged. Thanks!

PR #3604, opened 2022-01-19
Use more concrete error class on plugin not found

cosmo0920:

None.

Not found plugin is basically config error but we need to use another error class to customize this error for config validation.
In general, cloud service does not permit to install plugins via user operations.

This patch does not affect for error message for ordinary cases.
Perhaps, it might affect error handling but this should be a corner case.

**Docs Changes**:
No needed.

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/config/error.rb
+++ b/lib/fluent/config/error.rb
@@ -29,4 +29,16 @@ module Fluent
 
   class SetDefault < Exception
   end
+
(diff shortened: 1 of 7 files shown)


ashie (approved):
:+1:  Waiting CI...

cosmo0920:
Thanks!

PR #3585, opened 2021-12-27
Investigate issues on Ruby 3.1 Windows

ashie:
Signed-off-by: Takuro Ashie <ashie@clear-code.com>

**Which issue(s) this PR fixes**: 
Fixes #

**What this PR does / why we need it**: 

**Docs Changes**:

**Release Note**:

PR #3619, opened 2022-02-04
Tweak CI

ashie:
Fixes #

Tweak CI on GitHub Actions:
* Add Ruby 3.1 for GNU/Linux and macOS
  *  For windows, we have several known issues on RubyInstaller 3.1 so that we don't yet add for now: #3585
* Use latest RubyInstaller 3.0 & Fiddle

**Docs Changes**:
none

**Release Note**: 
none


ashie:
We'll drop Ruby 2.6 after it reaches EOL (2022-03-31).

Issue #3627, opened 2022-02-09
in_exec with out_exec in one config create file descriptors leak


LukasJerabek:
### Describe the bug

Hello, I think I have found the core issue of my problems with fluentd configuration, could someone have look on this please?
I have set everything according to https://docs.fluentd.org/installation/before-install.
Using configuration and two scripts below I can notice that if I do these for a while:
```
cd /proc/<fluentd PID>/ 
find fd | wc -l
```
Then I can see that number of file descriptors is rising infinitely eventually resulting in memory problems and "too many open files" error because of reaching fd limit (ulimit -n). This only occures if both in_exec and out_exec are used, since if I do following swaps, the number of file descriptors stabilizes. I can do either of these and the problems will disappear.

File descriptors count is stabilized if I swap match section for this:
```
<match hardware.memory>
  @type file
  append true
  path {{ home }}/fluentd/test_bug2.log
  add_path_suffix false
  <buffer []>
    @type memory
    flush_mode immediate
  </buffer>
  <format>
    @type json
  </format>
</match>
```

or swap source and filter section for this:
```
<source>
  @type sample
  sample {"mem_available":"1.5","swap_total":"1.5","swap_free":"1.2","time_local":"2022-02-09T10:00:00+0100"}
  tag hardware.memory
</source>
```

### To Reproduce

Install fluentd:
`gem install fluentd --version 1.14.5 --no-doc`
Create folder "fluentd" in home folder
Exchange {{ home }} or {{ user }} for your real values in every file below.
Put fluentd.conf (with configuration below), get_memory_metrics.sh, psql_memory.sh into {{ home }}/fluentd/
do chmod +x on get_memory_metrics.sh, psql_memory.sh
run {{ home }}/bin/fluentd _1.14.5_ -c {{ home }}/fluentd/fluentd.conf -vv

Now start checking the file descriptors in `/proc/<fluentd PID>/` with:
`find fd | wc -l`
Keep cheking it for some time, youll see the count of file descriptors never stops rising.

Hopefully I did not miss anything.

### Expected behavior

Number of file descriptors is not rising infinitely and not causing memory leaks.

### Your Environment

```markdown
- Fluentd version: 1.14.5 (installed as gem)
- TD Agent version: -
- Operating system: Rocky Linux 8
- Kernel version: 4.18.0-348.12.2.el8_5.x86_64
- Ruby version: 2.5.9
```

### Your Configuration

```apache
<system>
  process_name fluentd_monitor
</system>

[...]

PR #3634, opened 2022-02-15
in_exec: fix a file descriptor leaks

kenhys:

Follow up #3627

Every calling command from in_exec plugin, it causes
a file descriptor leak.

If interval is short enough or running Fluentd in long term,
it will reaches the limit of file descriptor - "too many open files".

In this commit, ensure to close IO object when it reaches EOF.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/in_exec.rb
+++ b/lib/fluent/plugin/in_exec.rb
@@ -95,6 +95,7 @@ module Fluent::Plugin
       else
         @parser.parse(io.read, &method(:on_record))
       end
+      io.close if io.eof?
     end
 
     def on_record(time, record)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_exec.rb
```diff
@@ -95,6 +95,7 @@ def run(io)
       else
         @parser.parse(io.read, &method(:on_record))
       end
+      io.close if io.eof?
```
ashie, line 98: Generally speaking, resources should be closed by the one who opened it.
I'm not sure but closing it here might not appropriate.

LukasJerabek:
Id just like to point out also here, that in #3627 I dont observe problems with in_exec itself, but only in combination of in_exec with out_exec.

LukasJerabek:
Bump.

ashie:
I can't reproduce the issue.
In addition, I don't think it's proper way as I mentioned before.

PR #3650, opened 2022-03-01
Declare null_value_pattern as regexp

tagomoris:
Because it is a regexp actually. This change provides flexibility and better document to users.

The `null_value_pattern` is defined as string currently, but it's actually compiled as a Regexp. But documents are written as string and it makes users confused.

**Docs Changes**:
Needed here: https://docs.fluentd.org/configuration/parse-section

**Release Note**: 
Probably no need to mention this. No compatibility will be broken by this change.


ashie (approved):
LGTM

ashie:
Thanks! I'll fix the document.

PR #3654, opened 2022-03-04
Switch to GitHub Discussions

kenhys:

Fixes #

Discuss.fluentd.org will not work as expected with
limitation of infrastructure itself and human resources,
it may be better transition from discuss.fluentd.org to GitHub
Discussions.

**Docs Changes**:

NOTE: fluent/fluentd-docs-gitbook has a link to it in issue template.

**Release Note**:


fujimotos (approved):

ashie (approved):
We need to announce it.

PR #3679, opened 2022-03-18
Make the default umask a string

syedriko:
Signed-off-by: Sergey Yedrikov <syedriko@redhat.com>

This is a correction to a previous commit that added the --umask command line parameter.
The default umask needs to be a string, otherwise the following happens:

2022-03-16 05:14:45 +0000 [error]: unexpected error error_class=ArgumentError error="wrong number of arguments (given 1, expected 0)"
2022-03-16 05:14:45 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.14.5/lib/fluent/supervisor.rb:714:in `to_i'

None

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -81,7 +81,6 @@ op.on('--group GROUP', "change group") {|s|
   opts[:chgroup] = s
 }
 
-opts[:chumask] = 0
 op.on('--umask UMASK', "change umask") {|s|
   opts[:chumask] = s
 }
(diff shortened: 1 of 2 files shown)


ashie (changes requested):
Ah, sorry I missed it.
In addition, default values should be placed in `Supervisor.default_options`

https://github.com/fluent/fluentd/blob/2036c4a6d791de24165d9575243eb71e5914b3b8/lib/fluent/command/fluentd.rb#L29

https://github.com/fluent/fluentd/blob/2036c4a6d791de24165d9575243eb71e5914b3b8/lib/fluent/supervisor.rb#L558-L583

ashie (approved):

ashie:
Thanks!

PR #3680, opened 2022-03-18
Add sigdump func to fluent-ctl

### LINKED ISSUE #3600: sigdump support for Windows

https://github.com/fluent/fluentd/pull/3131#issuecomment-1007690885
There is no way to trigger `Sigdump.dump` on Windows.

Add a new win32 event for it like #3131.

none.

### Additional context

_No response_

daipom:
Fixes #3600

Currently, [Sigdump](https://github.com/fluent/sigdump/) is available by sending the CONT signal, **without Windows**.
The main purpose of this PR is to make this function available for Windows too.

To do so, add the following features to `fluent-ctl` and rpc.

We can use the following methods to use Sigdump.

- command: `$ fluent-ctl dump {pid or svcname}`
- ~~http request to Fluentd: `/api/processes.dump`~~

**Windows**

- Fluentd outputs dump-files about the supervisor process and all worker processes to the system temp dir `C:\\Windows\\Temp`.
- We can specify `SIGDUMP_PATH` environment variable to change the path.
  - Automatically add each process-id to the path to avoid a writing conflict.

Examples of dump files on Windows.
```
C:\\Windows\\Temp\\fluentd-sigdump-{supervisor-pid}.log
C:\\Windows\\Temp\\fluentd-sigdump-{worker0-pid}.log
C:\\Windows\\Temp\\fluentd-sigdump-{worker1-pid}.log
...
```

Examples of specifying `SIGDUMP_PATH` by powershell on Windows.
```Powershell
$ $env:SIGDUMP_PATH="/sigdump/sigdump.log"

... [info]: fluent/log.rb:330:info: dump to /sigdump/sigdump-41544.log.
... [info]: #0 fluent/log.rb:330:info: dump to /sigdump/sigdump-21152.log.
... [info]: #1 fluent/log.rb:330:info: dump to /sigdump/sigdump-15656.log.
...
```

**UNIX-like**

- command: `$ fluent-ctl dump {pid}`
  - This is the same behavior as sending SIGCONT to the process as before. (Output the process's dump-file.)

- ~~http request to Fluentd: `/api/processes.dump`~~
  - ~~This is the same as Windows and outputs all processes' dump-files.~~
  - ~~However, we cannot use `SIGDUMP_PATH` to specify the path for this rpc command.~~
    - ~~Since each process-id is not added in this case and it causes a writing conflict.~~
    - ~~We need to fix [Sigdump library](https://github.com/fluent/sigdump).~~

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/397

**Release Note**: 
TODO

### DIFF
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -45,7 +45,7 @@ module Fluent
         restart: 128,
         flush: 129,
         reload: SERVICE_CONTROL_PARAMCHANGE,
-        # dump: 130? TODO support svc control
+        dump: 130,
       }
     else
       COMMAND_MAP = {
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -36,6 +36,7 @@ module Fluent
         restart: "HUP",
         flush: "USR1",
         reload: "USR2",
+        dump: "CONT",
       }
       WINSVC_CONTROL_CODE_MAP = {
         shutdown: SERVICE_CONTROL_STOP,
@@ -44,6 +45,7 @@ module Fluent
         restart: 128,
         flush: 129,
         reload: SERVICE_CONTROL_PARAMCHANGE,
+        # dump: 130? TODO support svc control
       }
     else
       COMMAND_MAP = {
@@ -51,6 +53,7 @@ module Fluent
         restart: :HUP,
         flush: :USR1,
         reload: :USR2,
+        dump: "CONT",
       }
     end
 
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -53,7 +53,7 @@ module Fluent
         restart: :HUP,
         flush: :USR1,
         reload: :USR2,
-        dump: "CONT",
+        dump: :CONT,
       }
     end
 
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -268,7 +268,7 @@ module Fluent
             end
           end
         ensure
-          events.each { |event| event.close }
+          events.each { |event| event[:win32_event].close }
         end
       end
     end
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -106,6 +106,11 @@ module Fluent
         end
         nil
       }
+      @rpc_server.mount_proc('/api/processes.dump') { |req, res|
+        $log.debug "fluentd RPC got /api/processes.dump request"
+        supervisor_sigcont_handler
+        nil
+      }
       @rpc_server.mount_proc('/api/plugins.flushBuffers') { |req, res|
         $log.debug "fluentd RPC got /api/plugins.flushBuffers request"
         if Fluent.windows?
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -215,39 +215,41 @@ module Fluent
       Thread.new do
         ipc = Win32::Ipc.new(nil)
         events = [
-          Win32::Event.new("#{@pid_signame}_STOP_EVENT_THREAD"),
-          Win32::Event.new("#{@pid_signame}"),
-          Win32::Event.new("#{@pid_signame}_HUP"),
-          Win32::Event.new("#{@pid_signame}_USR1"),
-          Win32::Event.new("#{@pid_signame}_USR2"),
+          {win32_event: Win32::Event.new("#{@pid_signame}_STOP_EVENT_THREAD"), action: :stop_event_thread},
+          {win32_event: Win32::Event.new("#{@pid_signame}"), action: :stop},
+          {win32_event: Win32::Event.new("#{@pid_signame}_HUP"), action: :hup},
+          {win32_event: Win32::Event.new("#{@pid_signame}_USR1"), action: :usr1},
+          {win32_event: Win32::Event.new("#{@pid_signame}_USR2"), action: :usr2},
         ]
         if @signame
           signame_events = [
-            Win32::Event.new("#{@signame}"),
-            Win32::Event.new("#{@signame}_HUP"),
-            Win32::Event.new("#{@signame}_USR1"),
-            Win32::Event.new("#{@signame}_USR2"),
+            {win32_event: Win32::Event.new("#{@signame}"), action: :stop},
+            {win32_event: Win32::Event.new("#{@signame}_HUP"), action: :hup},
+            {win32_event: Win32::Event.new("#{@signame}_USR1"), action: :usr1},
+            {win32_event: Win32::Event.new("#{@signame}_USR2"), action: :usr2},
           ]
           events.concat(signame_events)
         end
         begin
           loop do
-            idx = ipc.wait_any(events, Windows::Synchronize::INFINITE)
-            if idx > 0 && idx <= events.length
-              $log.debug("Got Win32 event \"#{events[idx - 1].name}\"")
+            ipc_idx = ipc.wait_any(events.map {|e| e[:win32_event]}, Windows::Synchronize::INFINITE)
+            event_idx = ipc_idx - 1
+
+            if event_idx >= 0 && event_idx < events.length
+              $log.debug("Got Win32 event \"#{events[event_idx][:win32_event].name}\"")
             else
-              $log.warn("Unexpected reutrn value of Win32::Ipc#wait_any: #{idx}")
+              $log.warn("Unexpected return value of Win32::Ipc#wait_any: #{ipc_idx}")
             end
-            case idx
-            when 2, 6
+            case events[event_idx][:action]
+            when :stop
               stop(true)
-            when 3, 7
+            when :hup
               supervisor_sighup_handler
-            when 4, 8
+            when :usr1
               supervisor_sigusr1_handler
-            when 5, 9
+            when :usr2
               supervisor_sigusr2_handler
-            when 1
+            when :stop_event_thread
               break
             end
           end
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -321,6 +321,18 @@ module Fluent
     def supervisor_sigcont_handler
       if Fluent.windows?
         $log.info "dump file. [pid:#{Process.pid}]"
+
+        # Sigdump outputs under `/tmp` dir without `SIGDUMP_PATH` specified
+        # but `/tmp` dir may not exist on Windows by default.
+        # Since this function is mainly for Windows, this case should be saved.
+        # (Don't want to couple tightly with the Sigdump library but want to save this case especially.)
+        unless ENV['SIGDUMP_PATH']
+          dump_dir = '/tmp'
+          unless Dir.exist?(dump_dir)
+            FileUtils.mkdir_p(dump_dir, mode: Fluent::DEFAULT_DIR_PERMISSION)
+          end
+        end
+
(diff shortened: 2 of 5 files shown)


daipom:
This is still in draft status. I will continue to work on it.

lib/fluent/command/ctl.rb
```diff
@@ -44,13 +45,15 @@ class Ctl
...
         reload: SERVICE_CONTROL_PARAMCHANGE,
+        # dump: 130? TODO support svc control
       }
     else
       COMMAND_MAP = {
         shutdown: :TERM,
         restart: :HUP,
         flush: :USR1,
         reload: :USR2,
+        dump: "CONT",
```
cosmo0920, line 56: Should we use `symbol` style here?

daipom: Thanks! I have unified by `symbol` here.

daipom:
I found the problem with UNIX-like system.
Currently `kill CONT` signal is sent infinitely.
I will fix this.

daipom:
I have tested on macOS too.
The failed tests are about `ChildProcessTest` and seem to have nothing to do with this fix.

daipom:
> I finished the implementation and tests on Windows and Ubuntu but I found a problem with this specification.
> 
> We can set `SIGDUMP_PATH` environment variable to specify the dump-filepath. (https://github.com/fluent/sigdump#usage)
> 
> However, this specified path is not be separated by each pid. So multiple dump-contents are written into the same file at the same time and it causes conflicts of writing.
> 
> I see two possible solutions.
> 
>     * Fix [Sigdump](https://github.com/fluent/sigdump) so that it adds pid to the path even with the environmental variable specified.
>       
>       * If any user uses the variable and the complete path systematically, this change may break the system...
> 
>     * Limit this feature to Windows only.

I would like to ask for opinion on this point.

lib/fluent/supervisor.rb
```diff
@@ -302,6 +322,37 @@ def supervisor_sigusr2_handler
...
+        $log.info "dump file. [pid:#{Process.pid}]"
+
+        # Sigdump outputs under `/tmp` dir without `SIGDUMP_PATH` specified
+        # but `/tmp` dir may not exist on Windows by default.
+        # Since this function is mainly for Windows, this case should be saved.
+        # (Don't want to couple tightly with the Sigdump library but want to save this case especially.)
+        unless ENV['SIGDUMP_PATH']
+          dump_dir = '/tmp'
+          unless Dir.exist?(dump_dir)
+            FileUtils.mkdir_p(dump_dir, mode: Fluent::DEFAULT_DIR_PERMISSION)
```
ashie, line 336: How about using OS's temp directory by default? (`%windir%\Temp`)

daipom: I have fixed it on https://github.com/fluent/fluentd/pull/3680/commits/65407e59a5200e21d1449815b4a310073cf27976.

In addition to using `%windir%/Temp` by default, now add pid to the path specified by `SIGDUMP_PATH` environment variable on Windows in order to avoid the writing conflicting.

Example.
```powershell
$ $env:SIGDUMP_PATH="/sigdump/sigdump.log"
$ fluentd -c "{conf}"
(sigdump action)
```

```
... [info]: fluent/log.rb:330:info: dump to /sigdump/sigdump-41544.log.
... [info]: #0 fluent/log.rb:330:info: dump to /sigdump/sigdump-21152.log.
... [info]: #1 fluent/log.rb:330:info: dump to /sigdump/sigdump-15656.log.
```

As for UNIX-like, in order to add the same specification, 
need to add a trap of SIGCONT to the worker processes
or to fix the Sigdump library.
However, it loses backward compatibility for users using 'SIGDUMP_PATH'
environment variable, so need to consider what to do.

lib/fluent/supervisor.rb
```diff
@@ -302,6 +322,37 @@ def supervisor_sigusr2_handler
       $log.error "Failed to reload config file: #{e}"
     end
 
+    def supervisor_sigcont_handler
```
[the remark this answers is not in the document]
daipom: Thanks,! I have fixed it.

lib/fluent/supervisor.rb
ashie: For UNIX-like, it would be better to leave it as is to keep compatibility.
On the other hand, it would be also better that modifying sigdump for specifying a directory.
In current implementation, sigdump assumes the path specified by `SIGDUMP_PATH` is a file.

https://github.com/fluent/sigdump/blob/c9fcf34ee58b529cb46c528b700f4ff491326d4a/lib/sigdump.rb#L134-L148

```ruby
  def self._open_dump_path(path, &block)
    case path
    when nil, ""
      path = "/tmp/sigdump-#{Process.pid}.log"
      File.open(path, "a", &block)
    when IO
      yield path
    when "-"
      yield STDOUT
    when "+"
      yield STDERR
    else
      File.open(path, "a", &block)
    end
  end
```

But when the path is ended with `/` or the patch exists and it's directory, it should be treated as directory.
When a directory is specified, it would be better that sigdump adds pid automatically.

daipom: Thanks for your advice.

> For UNIX-like, it would be better to leave it as is to keep compatibility.

Then the best solution is 

- Abolish trapping SIGCONT in the supervisor process on UNIX-like.
  - Don't change SIGCONT behavior on UNIX-like for compatibility.
- `/api/processes.dump` rpc command on UNIX-like sends SIGCONT to the supervisor process and all worker processes.
  - This is a new function, so we don't have to care about compatibility. 
  - With `SIGDUMP_PATH` specified, the writing confliction will occur. This problem should be fixed by improving the Sigdump library so that we can specify a directory path.

`/api/processes.dump` for UNIX-like
```ruby
Process.kill :CONT, $$
send_signal_to_workers(:CONT)
```

daipom: I have fixed this on https://github.com/fluent/fluentd/pull/3680/commits/9d0d1f331dda7a10d76a102b514096719cb8248e.

- SIGCONT behavior on UNIX-like remains as before.
- `/api/processes.dump` rpc command on UNIX-like sends SIGCONT to the supervisor process and all worker processes.
- Fix the Sigdump library later. Until then, we cannot use this rpc command with `SIGDUMP_PATH` on UNIX-like.

daipom:
On my Windows envrionment, `test_supervisor_event_dump_windows` fails under certain conditions.

`bundle exec rake test TESTOPTS="-t'SupervisorTest'"`
```
Failure: test_supervisor_event_dump_windows(SupervisorTest):
  class()
  Called 0 times.
  Expected 1 times.
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:74:in `block in dump_object_count'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:73:in `each_object'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:73:in `dump_object_count'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:18:in `block in dump'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:144:in `open'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:144:in `_open_dump_path'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:14:in `dump'
.../fluentd/lib/fluent/supervisor.rb:1128:in `dump_windows'
.../fluentd/lib/fluent/supervisor.rb:331:in `block in supervisor_dump_handler_for_windows'
```

`C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/sigdump-0.2.4/lib/sigdump.rb:74`
```ruby
73 ObjectSpace.each_object {|o|
74   c = o.class
```

This test succeeds when I execute only this test.

Also, with difficulty, the following line can be commented out so that the test succeeds.

`test_rpc_server_windows`
```diff
@@ -347,7 +347,7 @@ class SupervisorTest < ::Test::Unit::TestCase

     server.run_rpc_server

-    mock(server).restart(true) { nil }
+    # mock(server).restart(true) { nil }
     response = Net::HTTP.get(URI.parse("http://#{localhost}:24447/api/plugins.flushBuffers"))

     server.stop_rpc_server
```

ashie:
> On my Windows envrionment, `test_supervisor_event_dump_windows` fails under certain conditions.

Hmm, it's also occurred on CI:
https://github.com/fluent/fluentd/runs/5656988063?check_suite_focus=true

daipom:
> > On my Windows envrionment, `test_supervisor_event_dump_windows` fails under certain conditions.
> 
> Hmm, it's also occurred on CI: https://github.com/fluent/fluentd/runs/5656988063?check_suite_focus=true

Maybe I should not use `Sigdump` library itself, but the mock of it.

daipom:
Completes the whole process except for the documentation.
If there are no problems with the specification, I will re-test the whole and create a PR for the document.

lib/fluent/supervisor.rb
```diff
@@ -106,6 +107,16 @@ def run_rpc_server
...
+      @rpc_server.mount_proc('/api/processes.dump') { |req, res|
+        $log.debug "fluentd RPC got /api/processes.dump request"
+        if Fluent.windows?
+          supervisor_dump_handler_for_windows
+        else
+          Process.kill :CONT, $$
+          send_signal_to_workers(:CONT)
+        end
+        nil
+      }
```
ashie, line 119: It would be better that adding a config param to enable/disable this feature like `@enable_get_dump`, and its default value should be `false (disabled)`.
Or it would be better that we use existing `@enable_get_dump` as is.

ashie (changes requested):
`/api/processes.dump` should be disabled by default to prevent dumping sensitive information.

lib/fluent/supervisor.rb
ashie: > Or it would be better that we use existing `@enable_get_dump` as is.

It's not appropriate.
It's designed to use only for `/api/config.getDump` (get config via HTTP), and not used for `/api/config.dump` (dump config to log).

ref: https://www.clear-code.com/blog/2015/10/2.html

daipom: Oh, I see.
Is there a possibility that we don't have to turn off the API by the option since this doesn't send the content, but just outputs to the local files?

ashie: I want to remove this API completely for now, I don't think this API is really required.
I think it's okay to add it after someone actually requests it.

ashie: Since it's for debug usage, I think `fluent-ctl dump` is enough for now.

daipom: Agreed.
The implementation of the command accomplishes the main purpose of this PR, so I think the API can be removed.

ashie: I'll do it.

daipom: Thank you very much!

ashie: Removed in 4b18c6f34344d455faf75e407927871f1361f6b5

daipom:
Thank you for the fix.
I fixed the document: https://github.com/fluent/fluentd-docs-gitbook/pull/397

PR #3649, opened 2022-03-01
Fix wrong calcuration of retry interval and detecting retry limit

ashie:
**Which issue(s) this PR fixes**: 
Fixes #3609

**What this PR does / why we need it**: 

The correct total retry time should be:
```
c + c * b^1 + (...) + c*b^(k - 1)
```
But the previous implementation was:
```
c + c * b^0 + c * b^1 + (...) + c*b^(k - 1)    
```

where:

  * c: constant factor, @retry_wait
  * b: base factor, @retry_exponential_backoff_base
  * k: number of retry times, @max_retry_times
    
Although the first retry interval is already added at constructor of
ExponentialBackoffRetry, same value is added again unexpectedly on the
first calc_interval call (as `c * b^0`).

In addition, this PR also fixes wrong detection of retry limit.

**Docs Changes**:
https://docs.fluentd.org/configuration/buffer-section#retries-parameters is also wrong, should be fixed too.

**Release Note**: 
Same with the title

PR #3685, opened 2022-03-24
[For #3649] Add scenario test of RetryState

daipom:
This is about #3609.
Add scenario tests for #3649.

RetryStatement behavior is complicated, so scenario tests may be useful.

**Docs Changes**:
None.

**Release Note**: 
None.


ashie:
Thanks!

PR #3686, opened 2022-03-24
[For #3649] Fix timeout handling

daipom:
This is about #3609.
Fix timeout handling for #3649.

The current `limit?` usage is complicated, so I have tried to make it simple.

**Docs Changes**:
None.

**Release Note**: 
None.

### DIFF
--- a/lib/fluent/plugin_helper/retry_state.rb
+++ b/lib/fluent/plugin_helper/retry_state.rb
@@ -45,7 +45,7 @@ module Fluent
           @timeout = timeout
           @timeout_at = @start + timeout
           @has_reached_timeout = false
-          @has_timeouted = false
+          @has_timed_out = false
           @current = :primary
 
           if randomize_width < 0 || randomize_width > 0.5
@@ -128,10 +128,10 @@ module Fluent
 
           @next_time = calc_next_time
 
-          unless @has_reached_timeout
-            @has_reached_timeout = @next_time >= @timeout_at
+          if @has_reached_timeout
+            @has_timed_out = @next_time >= @timeout_at
           else
-            @has_timeouted = @next_time >= @timeout_at
+            @has_reached_timeout = @next_time >= @timeout_at
           end
 
           nil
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin_helper/retry_state.rb
```diff
@@ -44,6 +44,8 @@ def initialize(title, wait, timeout, forever, max_steps, randomize, randomize_wi
 
           @timeout = timeout
           @timeout_at = @start + timeout
+          @has_reached_timeout = false
+          @has_timeouted = false
```
ashie, line 48:
```suggestion
          @has_timed_out = false
```

lib/fluent/plugin_helper/retry_state.rb
```diff
@@ -123,32 +125,27 @@ def step
...
             @secondary_transition_steps = @steps
           end
+
           @next_time = calc_next_time
+
+          unless @has_reached_timeout
+            @has_reached_timeout = @next_time >= @timeout_at
+          else
+            @has_timeouted = @next_time >= @timeout_at
+          end
```
ashie, line 135: I don't prefer `unless` clause with `else` clause, double negative makes hard to read:
```suggestion
          if @has_reached_timeout
            @has_timeouted = @next_time >= @timeout_at
          else
            @has_reached_timeout = @next_time >= @timeout_at
          end
```

ashie (changes requested):

daipom:
Thanks for the review. I have reshaped codes, and confirmed the following success again.

- `$ bundle exec rake test TESTOPTS="-t'BufferedOutputRetryTest'"`
- `$ bundle exec rake test TESTOPTS="-t'RetryStateHelperTest'"`

lib/fluent/plugin_helper/retry_state.rb
daipom: fixed on 62b7002cdd09b452308e2ef869131f907022d904

lib/fluent/plugin_helper/retry_state.rb
daipom: fixed on 62b7002cdd09b452308e2ef869131f907022d904

ashie:
Thanks!

PR #3711, opened 2022-04-11
out_forward: Fix to update timeout of cached sockets

ashie:
None.

In `SocketCache`, timeout of sockets are set only on creating them and
never be updated. So that cached sockets (`@available_sockets`) are
always closed after `keepalive_timeout` is passed from its creation,
even if it was used within `keepalive_timeout` seconds.

**Docs Changes**:
Not needed.

**Release Note**: 
Same with the title

### DIFF
--- a/lib/fluent/plugin/out_forward/socket_cache.rb
+++ b/lib/fluent/plugin/out_forward/socket_cache.rb
@@ -50,6 +50,7 @@ module Fluent::Plugin
       def checkin(sock)
         @mutex.synchronize do
           if (s = @inflight_sockets.delete(sock))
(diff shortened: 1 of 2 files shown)


ashie:
I'm checking the history to confirm whether or not it's intentional, I think it's a bug though.

ashie:
> I'm checking the history to confirm whether or not it's intentional, I think it's a bug though. I'll add tests after I confirm it.

Added.
In addition, I found some problems in original test code, so I fixed them too.

cosmo0920 (approved):
Looks reasonable for me.

Issue #554, opened 2015-02-16
Configuration as yaml/json


warmfusion:
I've been trying to setup a FluentD puppet module that is able to build configuration from resource calls, however, the configuration syntax makes this very difficult.

Is it possible to use yaml or json for the configuration of this application as I would like to be able to
- Define sub-elements using a hierarchy of hashes
- Avoid having to declare 'footer' elements ( eg `</match>` )
## Example Configuration

```
fluentd::match:
    pattern: 'nginx.*'
    priority: 10
    options:
      type: copy
      store:
       - type: 'file'
         path: '/var/log/fluent/myapp'
         time_slice_format: '%Y%m%d'
         time_slice_wait: '10m'
         time_format: '%Y%m%dT%H%M%S%z'
         compress: 'gzip'
       - type: 'mongo'
         host: 'fluentd'
         port: '27017'
         database: 'fluentd'
         collection: 'test'
```

repeatedly:
Some users use fluentd with puppet: https://github.com/mmz-srf/puppet-fluentd
I'm not familiar with puppet but constructing yaml is easier? Dumping hash to yaml?

We don't have a plan to support these formats but
converting yaml to fluentd's configuration is one idea.

warmfusion:
Thanks @repeatedly, I tried to use that puppet module however due to the limitations of the configuration format it also has the same problems I did when configuring matchers/sources etc.

For example, it is not possible to create a matcher which uses the forwarder plugin with 2 servers and a secondary (as described in the [HA](http://docs.fluentd.org/articles/high-availability) and [Forwader](http://docs.fluentd.org/articles/out_forward) documentation)

This is due to the way the matcher config is [created from a template](https://github.com/mmz-srf/puppet-fluentd/blob/develop/templates/match.erb), which due to limitations in puppet, combined with limitations of the config format means sub-sections are hardcoded, duplicated and otherwise worked-around.

On the surface it appears that configuration is managed through yaml, however, this is only made possible by some complex coding under the hood in puppet which still isn't able to properly describe the configuration options available.

repeatedly:
> For example, it is not possible to create a matcher which uses the forwarder plugin with 2 servers and a secondary

From this spec, puppet-fluentd supports `<server>` sections: https://github.com/mmz-srf/puppet-fluentd/blob/develop/spec/defines/fluentd_match_spec.rb#L33
If `secondary` doesn't work, then adding `secondary` support to puppet-fluentd is good for other users.

[...]

PR #2818, opened 2020-02-05
yaml config

ganmacs:
Signed-off-by: Yuta Iwama <ganmacs@gmail.com>

**Which issue(s) this PR fixes**: 
ref: https://github.com/fluent/fluentd/issues/554

**What this PR does / why we need it**: 

I tried to create the feature of the config file in yaml .
I don't intend to replace this config format with the current one.

What do you think? @repeatedly @cosmo0920 @ashie 

* `!fluent/json` which is a yaml tag means that this value parser as json object

* `!fluent/s` which is a yaml tag means that this value parser as ruby's string

```yml
match: 
  key1: "ENV['ROOT']"
  key2: !fluent/s "ENV['ROOT']"
```
means

```
<match>
  key ENV['ROOT']
  key "ENV['ROOT']"
</match>
```

* `!include` which is a yaml tag is equivalent to `@include` in the current config.
* use `$` instead of `@` because `@` is reserved word in yaml

* worker example

```yml
worker:
  $arg: 1 
  config:
    - source:
    ....
```
means

```
<worker 1>
  <source>
    ....
  </source>
</worker>
```

* label example

```yml
label:
  $name: '@test'
  config:
    - match:
       ....
```

means

```
<label @test>
  <match>
    ....
  </match>
</label>
```

* source example

```yml
source:
  $type: dummy
```

means

```
<source>
  @type dummy
</source>
```

* filter and match example

```yml
filter:
  $tag: tag
  $type:  grep
  ...
match:
  $tag: tag
  $type: stdout
  buffer:
    $type: memory
```

means

```
<filter tag>
  @type grep
   ...
</filter>

<match tag>
  @type stdout
  <buffer>
    @type memory
  </buffer>
</match>
```

#### Example

```yml
system:
  workers: 2

config:
  - worker:
      $arg: 1
      config:
        - source:
            $type: dummy
            dummy: !fluent/json { "message": "hello" }
            tag: dummy1.t
        - filter:
            $type: record_transformer
            $tag: dummy1.*
            record:
              - hostname: !fluent/s "#{Socket.gethostname}"
                tag: ${tag}
        - match:
            $type: stdout
            $tag: dummy1.*
            buffer:
              $type: memory
              flush_interval: 1s
  - worker:
      $arg: 1
      config:
        - source:
            $type: dummy
            $label: '@test'
            tag: test1
        - source:
            $type: dummy
            $label: '@test'
            tag: test2
        - label:
            $name: '@test'
            config:
              - match:
                  $tag: test2
[...]

PR #3712, opened 2022-04-11
Handle YAML configuration format on configuration file

cosmo0920:
Fixes https://github.com/fluent/fluentd/issues/554
Revised #2818

Recently, fluent-bit supports YAML configuration in https://github.com/fluent/fluent-bit/pull/4621.

Now, it is reasonable to handle YAML configuration format on Fluentd, too.
This is because YAML config format is more kubernetes friendly and machine friendly.
It's time to discuss to support YAML configuration format again.
 
**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/402

**Release Note**: 

Same as title

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -126,6 +126,14 @@ op.on('--without-source', "invoke a fluentd without input plugins", TrueClass) {
   opts[:without_source] = b
 }
 
+op.on('--config-file-type VALU', 'file type of fluentd. yaml or yml') { |s|
+  if (s == 'yaml') || (s == 'yml')
+    opts[:config_file_type] = s.to_sym
+  else
+    usage '--config-file-type accepts yaml or yml'
+  end
+}
+
 op.on('--use-v1-config', "Use v1 configuration format (default)", TrueClass) {|b|
   opts[:use_v1_config] = b
 }
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -126,11 +126,13 @@ op.on('--without-source', "invoke a fluentd without input plugins", TrueClass) {
   opts[:without_source] = b
 }
 
-op.on('--guess-config-file-type VALU', 'guessing file type of fluentd configuration. yaml or yml') { |s|
+op.on('--config-file-type VALU', 'guessing file type of fluentd configuration. yaml/yml or guess') { |s|
   if (s == 'yaml') || (s == 'yml')
-    opts[:guess_config_file_type] = s.to_sym
+    opts[:config_file_type] = s.to_sym
+  elsif (s == 'guess')
+    opts[:config_file_type] = s.to_sym
   else
-    usage '--guess-config-file-type accepts yaml or yml'
+    usage '--config-file-type accepts yaml/yml or guess'
   end
 }
 
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -126,16 +126,12 @@ op.on('--without-source', "invoke a fluentd without input plugins", TrueClass) {
   opts[:without_source] = b
 }
 
-op.on('--config-file-type VALU', 'file type of fluentd. yaml or yml') { |s|
+op.on('--guess-config-file-type VALU', 'guessing file type of fluentd configuration. yaml or yml') { |s|
   if (s == 'yaml') || (s == 'yml')
-    opts[:config_file_type] = s.to_sym
+    opts[:guess_config_file_type] = s.to_sym
   else
-    usage '--config-file-type accepts yaml or yml'
+    usage '--guess-config-file-type accepts yaml or yml'
   end
-}
-
-op.on('--guess-config-file-type', "Guess config file type that is yaml or not (default)", TrueClass) {|b|
-  opts[:guess_config_file_type] = b
 }
 
 op.on('--use-v1-config', "Use v1 configuration format (default)", TrueClass) {|b|
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -134,6 +134,10 @@ op.on('--config-file-type VALU', 'file type of fluentd. yaml or yml') { |s|
   end
 }
 
+op.on('--guess-config-file-type', "Guess config file type that is yaml or not (default)", TrueClass) {|b|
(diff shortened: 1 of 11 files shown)


lib/fluent/config.rb
```diff
@@ -25,7 +26,18 @@ module Config
     # @param additional_config [String] config which is added to last of config body
     # @param use_v1_config [Bool] config is formatted with v1 or not
     # @return [Fluent::Config]
-    def self.build(config_path:, encoding: 'utf-8', additional_config: nil, use_v1_config: true)
+    def self.build(config_path:, encoding: 'utf-8', additional_config: nil, use_v1_config: true, type: nil, guess_type: nil)
+      if guess_type
+        config_fext = File.extname(config_path.dup)
```
ashie, line 31: I prefer `config_file_ext` or `config_ext`.

ashie: > `config_path.dup`

`.dup` doesn't seem needed.

cosmo0920:
@ashie @kenhys How's it going to review this PR?

ashie:
I think #3535 and this are featured functions of next release, v1.15.
#3535 has been landed just now, so I'll start to review this from now :smiling_face_with_tear:

ashie:
(Although I already reviewed it roughly and it seems fine, I'll check the details more.)

lib/fluent/command/fluentd.rb
```diff
@@ -126,6 +126,18 @@
...
 
+op.on('--config-file-type VALU', 'file type of fluentd. yaml or yml') { |s|
+  if (s == 'yaml') || (s == 'yml')
+    opts[:config_file_type] = s.to_sym
+  else
+    usage '--config-file-type accepts yaml or yml'
+  end
+}
+
+op.on('--guess-config-file-type', "Guess config file type that is yaml or not (default)", TrueClass) {|b|
```
ashie, line 137: How about removing this and add `guess` to `--config-file-type` (and make it default).

cosmo0920: Sure, I'll change them.

cosmo0920: Fixed in https://github.com/fluent/fluentd/pull/3712/commits/49ec02315311c94ba3550c0b752ad24deaaaee0e

ashie: :+1:

ashie: Ah, sorry it's a bit different what I thought.
I intended adding `:guess` (`--config-file-type=guess`), how about this?

cosmo0920: Ah, I see. I'll try to implement what you thought.

ashie (changes requested):

lib/fluent/command/fluentd.rb
cosmo0920: Really fixed in https://github.com/fluent/fluentd/pull/3712/commits/6574387ef2c78363dc23640df14d9886d5606bba

lib/fluent/config/yaml_parser/parser.rb
```diff
@@ -0,0 +1,166 @@
...
+    module YamlParser
+      class Parser
+        def initialize(config, indent: 2)
+          @base_indent = indent
+          @config = config
+        end
+
+        def build
+          s = @config['system'] && system_config_build(@config['system'])
+          c = @config['config'] && config_build(@config['config'], root: true)
```
ashie, line 30: Always requires `config` parent for `label`, `source`, `filter`, `match` and `worker` is bother and doesn't compatible with Fluentd's format. It makes hard to understand how to convert.

cosmo0920: This format limitations is coming from _YAML format specification and its parser simplicity_.
Fluentd config format does not have 1-to-1 correspondence for YAML config format for now.
It is quite hard to implement to arbitrary place handling for `label`, `source`, `filter`, `match` and `worker`. directives.

I didn't find no mono morphic way from Fluentd config format to YAML format.

cosmo0920: FYI: Fluent Bit uses similar approach to handle data pipeline:

https://github.com/fluent/fluent-bit/pull/4734

Fluent Bit ini like format should use `[INPUT]`, `[OUTOUT]`, `[FILTER]` sections as a top level settings on the original config implementation but its YAML format does not use them as top level objects.

cosmo0920: And we already prepared from Fluentd configuration format to YAML format translation documents here:
https://github.com/fluent/fluentd-docs-gitbook/pull/402

cosmo0920: Our assumed use-case is:
This format is mainly used from Golang's struct Marshal/Unmarshal not by writing by hands.

ashie: I got it. For example, the following is invalid as yaml:
```yml
- label:
    $name: "@FLUENT_LOG"
   - match:
       $type: "stdout"
       $tag: tag1.**
  - match:
      $type: "null"
      $tag: tag2.**
```

cosmo0920:
Any concerns?

ashie (approved):

ashie:
:+1:

cosmo0920:
Thanks!!

daipom:
https://github.com/fluent/fluentd/issues/4463#issuecomment-2054279009

> it appears that YAML format doesn't support the $log_level

If someone remembers something about this, please let me know.

lib/fluent/config/yaml_parser/parser.rb
```diff
@@ -0,0 +1,166 @@
...
+          end
+
+          if (v = config.delete('$label'))
+            sb.add_line('@label', v)
+          end
+
+          if (v = config.delete('$id'))
+            sb.add_line('@id', v)
+          end
+
```
cosmo0920, line 140: @daipom Probably, we should add handler for `$log_level` here. Do you have a bandwidth for this?

daipom: Thanks!

I don't think this is a serious problem.
I will work on other issues for a while.

cosmo0920: Sure. Fair enough.

PR #3766, opened 2022-05-30
Support Ruby 3.1 on Windows

ashie:
Fixes #3585

Currently in_tail is broken on Ruby 3.1 on Windows.
In addition, some out_forward tests are broken too.
This PR fixes these issues.
In addition, this PR add CI for Ruby 3.1 and head on Windows

**Docs Changes**:
None.

**Release Note**: 
Same with the title.

### DIFF
--- a/lib/fluent/plugin/file_wrapper.rb
+++ b/lib/fluent/plugin/file_wrapper.rb
@@ -16,8 +16,45 @@
 
 module Fluent
   module FileWrapper
-    def self.open(*args)
-      io = WindowsFile.new(*args).io
+    include File::Constants
+    RUBY_3_1_OR_LATER = Gem::Version.create(RUBY_VERSION) >= Gem::Version.create('3.1.0')
+
+    def self.mode2flags(mode)
+      # Always need BINARY to enable SHARE_DELETE
+      # https://bugs.ruby-lang.org/issues/11218
+      # https://github.com/ruby/ruby/blob/d6684f063bc53e3cab025bd39526eca3b480b5e7/win32/win32.c#L6332-L6345
+      flags = BINARY | SHARE_DELETE
+      case mode.delete("b")
+      when "r"
+        flags |= RDONLY
+      when "r+"
+        flags |= RDWR
+      when "w"
+        flags |= WRONLY | CREAT | TRUNC
+      when "w+"
+        flags |= RDWR | CREAT | TRUNC
+      when "a"
+        flags |= WRONLY | CREAT | APPEND
+      when "a+"
+        flags |= RDWR | CREAT | APPEND
+      else
+        raise Errno::EINVAL.new("Unsupported mode by Fluent::FileWrapper: #{mode}")
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


ashie:
Although I replaced `WindowsFile.open` with `File.open`, `WindowsFile.stat` can't replace with `File.stat`:

* They returns different number for `ino`, it breaks backward compatibility of position file.
* `File.stat` doesn't support delete_pending.

cosmo0920 (commented):
From [the comment](https://github.com/fluent/fluentd/pull/3585#issuecomment-1140903514) and the CI result on Ruby 3.1 on Windows, it looks good to me! :muscle:

cosmo0920 (approved):

lib/fluent/plugin/file_wrapper.rb
```diff
@@ -16,8 +16,34 @@
...
+        flags |= RDWR | CREAT | APPEND
+      else
+        raise Errno::EINVAL.new("Unsupported mode by Fluent::FileWrapper: #{mode}")
+      end
+    end
+
+    def self.open(path, mode='r')
+      # inject File::Constants::SHARE_DELETE
+      # https://github.com/fluent/fluentd/pull/3585#issuecomment-1101502617
+      io = File.open(path, mode2flags(mode))
```
fujimotos, line 46: My Win32-fu is somewhat rusty, but won't this make Fluentd acquire exclusive
read/write lock on the target file? I'm thinking a scenario like this:

1. Fluentd started watching on `access.log`, and while doing so,
    it acquired read/write lock on the file.
2. Nginx started and it tried to open `access log` to record requests.
3. ???

To avoid the issue above, we probably need to set  `FILE_SHARE_READ` and
`FILE_SHARE_WRITE` flags (or will Ruby set these flags by default? I haven't
fully trace the file opening code in Ruby 3.x...)

ashie: > To avoid the issue above, we probably need to set `FILE_SHARE_READ` and `FILE_SHARE_WRITE` flags (or will Ruby set these flags by default? I haven't fully trace the file opening code in Ruby 3.x...)

They are always added:
https://github.com/ruby/ruby/blob/c7bbed299493d0fe5323916b83dab04021bc570f/win32/win32.c#L6446
```c
	h = CreateFileW(file, access, FILE_SHARE_READ | FILE_SHARE_WRITE | share_delete, &sec, create, attr, NULL);
```

fujimotos (approved):

lib/fluent/plugin/file_wrapper.rb
fujimotos: OK! LGTM.

ashie:
Thanks for your review!
I'll merge this after checking some more nitpicks.

ashie:
CI on Ruby 2.7 Windows always fails, but it's not concerned with this pull request.
I've checked failed rotate tests and found a bug in the tests.
I'll create a new pull request to fixed them.

PR #3768, opened 2022-05-31
Add new system-config to set interval to restart workers

### LINKED ISSUE #3749: Add a new system option "delay_start_worker"

When a worker process died for some reason, Fluentd always tries
to restart the worker as soon as possible. 

This "always-restart-immdeiately" policy does not always work fine.
For example, think the following cases:

* The worker process was killed because the host OS is almost running
   out of the available memory at the moment.

* The worker process was killed because the host OS tries to perform
  operation but Fluentd locked the resource (e.g. file locks).

Systemd provides `RestartSec` option that allows to wait for a few
seconds before restarting a service to solve the same issue.

https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartSec=

It would be better if Fluentd's supervisor can provide a similar option too.

NA

### Additional context

The underlying serverengine has a feature called "delayed_start_worker".
The option described above can be iimplemented on it (we need to tweak the
serverengine as well, though)

https://github.com/treasure-data/serverengine/blob/master/lib/serverengine/multi_worker_server.rb#L132-L144

daipom:
Fixes #3749

Add new option `restart_worker_interval` to system-config.
This enables workers to restart with an interval.

Please see #3749 for why this feature is necessary.

This needs the fix of ServerEngine: https://github.com/treasure-data/serverengine/pull/120

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/407

**Release Note**: 
Add `restart_worker_interval` parameter in `<system>` directive to set interval to restart workers that has stopped for some reason.

**Operation example**:

Config

```
<system>
  workers 3
  restart_worker_interval 1m
</system>
```

Operation

* Worker-A, B, C starts.
* Kill worker-A and worker-B at the same time (set this point as 0s-point).
* Kill worker-C at 15s-point.
* Worker-A and worker-B restart at 60s-point.
* Worker-C restarts at 75s-point.

**TODO**

* [x] Wait for the merge of https://github.com/treasure-data/serverengine/pull/120
* [x] Add some tests
* [x] Make the document's fix

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -493,7 +493,8 @@ module Fluent
         config_path: path,
         main_cmd: params['main_cmd'],
         signame: params['signame'],
-        disable_shared_socket: params['disable_shared_socket']
+        disable_shared_socket: params['disable_shared_socket'],
+        restart_worker_interval: params['restart_worker_interval'],
       }
       if daemonize
         se_config[:pid_path] = pid_path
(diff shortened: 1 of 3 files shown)


daipom:
Since the ServerEngine side tests the restarting behavior,
it is enough to test only the configuration parsing on the Fluentd side.

This completes the whole process now.

daipom:
Thanks for merging.

PR #3783, opened 2022-06-13
test_in_tail: Try to fix unstable test_shutdown_timeout

ashie:
None

test_shutdown_timeout in test_in_tail is a bit unstable on Drone.io and macOS.
Probably the cause is that recording start time is too early.

**Docs Changes**:
None

**Release Note**: 
None


ashie:
> It seems that sometimes timeout occurs.

Fixed.

PR #3808, opened 2022-07-01
Add support for concurrent append in out_file

### LINKED ISSUE #3805: out_file: Conflicts occur when appending to the same file with multiple workers
### Describe the bug

`out_file` can corrupt the output file if `append true` and `workers` > 1.

* The full reproduction step is attached.
* I make use of `flush_at_shutdown` in the attached procedure to make the reproduction easier.
* However, it should occur on normal writes as well.

We should implement a proper lock mechanism as we do for `append false`

### To Reproduce

**fluentd.conf**

```xml
<system>
  workers 8
</system>

<source>
  @type dummy
  tag test.log
  rate 100
  dummy {"message": "This is the test message"}
</source>

<match test.**>
  @type file
  path /test/fluentd/log/test
  append true
  <buffer>
    @type "file"
    path /test/fluentd/buffer
    flush_mode interval
    flush_interval 60m
    flush_at_shutdown true
  </buffer>
</match>
```

**Script to check the output file**

`check.py`

```python
import sys
import json

all_line_counts = 0
broken_line_counts = 0

for line in sys.stdin:
    try:
        json.loads(line.strip().split(maxsplit=2)[2])
    except Exception:
        broken_line_counts += 1
        print(all_line_counts, line)
    all_line_counts += 1

print(f"{broken_line_counts} / {all_line_counts} lines are broken.")
```

**Operation**

1. Run the fluentd with the config
2. Wait a few seconds
3. Stop the fluentd (This flushes all buffers)
4. Check the output file: `$ python3 check.py < {the output filepath}`

Then, you can see some lines broken.

```
4272 2022-06-30T17:01:37+09:00	test.log	{"message":"T2022-06-30T17:01:17+09:00	test.log	{"message":"This is the test message"}

4493 2022-06-30T17:01:19+09:00	testhis is the test message"}

5157 2022-06-30T17:01:21+09:00	test.log	{"message":"This is the t-06-30T17:01:39+09:00	test.log	{"message":"This is the test message"}

5223 .log	{"message":"This is the test message"}

5666 2022-06-30T17:01est message"}

6109 2022-06-30T17:01:26+09:00	test.log	{"message"::23+09:00	test.log	{"message":"This is the test message"}

6773 2022-06-30T17:01:19+09:00	test"This is the test message"}

7437 2022-06-30T17:01:21+09:00	test.log	{"message":"This is the t22-06-30T17:01:28+09:00	test.log	{"message":"This is the test message"}

8101 2022-06-30T17:01og	{"message":"This is the test message"}

8765 2022-06-30T17:01:26+09:00	test.log	{"message":t message"}

9208 2022-06-30T17:01:37+09:00	test.log	{"message":"T"This is the test message"}

10094 2022-06-30T17:01:30+09:00	test.lhis is the test message"}

10382 og	{"message":"This is the test message"}

10603 2022-06-30T17:01:32+09:00	test.log	{"message":"This is the tes-06-30T17:01:39+09:00	test.log	{"message":"This is the test message"}

10669 t message"}

12285 2022-06-30T17:01:19+09:00	test:24+09:00	test.log	{"message":"This is the test message"}

12506 2022-06-30T17:01:26+09:00	test.log	{"message":.log	{"message":"This is the test message"}

12727 2022-06-30T17:01:21+09:00	test.log	{"message":"This is the t"This is the test message"}

12949 20est message"}

13392 2022-06-30T17:01:30+09:00	test.l:24+09:00	test.log	{"message":"This is the test message"}

13613 2022-06-30T17:01:26+09:00	test.log	{"message":og	{"message":"This is the test message"}

13834 2022-06-30T17:01:32+09:00	test.log	{"message":"This is the tes"This is the test message"}

14056 20t message"}

14499 2022-06-30T17:01:30+09:00	test.l5+09:00	test.log	{"message":"This is the test message"}

14720 2022-06-30T17:01:37+09:00	test.log	{"message":"Tog	{"message":"This is the test message"}

15163 2022t message"}

26 / 18240 lines are broken.
```

### Expected behavior

`out_file` outputs all lines correctly,
or the configuration is validated.

### Your Environment

```markdown
- Fluentd version: fluentd 1.15.0
- Operating system: Ubuntu 20.04.04 LTS


$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal


- Kernel version: 5.13.0-44-generic
```

### Your Configuration

```apache
I Wrote it in `To Reproduce` section.
```

### Your Error Log

```shell
None.
```

### Additional context

* One solution is to use "`worker_id`" to make the output path unique, but it's arguably suboptimal.
* We should ensure the exclusive access by implementing a file locking.
* I'm going to fix this problem.

fujimotos:
Fixes #3805

The basic problem is that Fluentd can run a plugin in multiple workers,
but there is no mechanism that allows workers to coordinate each other.

For this reason, plugins are often suffers race condition issues on multi-worker
mode (e.g. #3805, https://github.com/fluent/fluent-plugin-s3/issues/391).

This can be simply resolvable by implementing a inter-worker lock.
Here is how I implemented it:

 * ServerModule prepares a lockfile directory and share that directory
   path with its workers (through ENV).
    
 * Now, Workers can define critical sessions by creating a lockfile
   in that directory.
    
 * The actual lock was held by flock, so it's automatically freed by
   OS even when Worker gets killed in the middle of critical session.

WIth this merged, we will be able to run `out_file` without race
conditions among workers.

**Docs Changes**:

Required

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -51,6 +51,12 @@ module Fluent
         @_fluentd_worker_id
       end
 
+      def fluentd_lockdir
+        return @_fluentd_lockdir if @_fluentd_lockdir
+        @_fluentd_lockdir = ENV['FLUENTD_LOCKDIR']
+        @_fluentd_lockdir
+      end
+
       def configure(conf)
         if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
           workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
@@ -68,6 +74,19 @@ module Fluent
 
       def multi_workers_ready?
         true
+      end
+
+      def acquire_worker_lock(name, &block)
+        if fluentd_lockdir.nil?
+          raise RuntimeError, "fail to create lockfile on '#{fluentd_lockdir}'"
+        end
+
+        name = name.gsub(/[^a-zA-Z0-9]/, "_")
+        lockfile = "fluentd-#{name}.lock"
+        File.open(File.join(fluentd_lockdir, lockfile), "w") do |f|
+          f.flock(File::LOCK_EX)
+          block.call()
+        end
       end
 
       def string_safe_encoding(str)
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -50,6 +50,9 @@ module Fluent
       @rpc_server = nil
       @counter = nil
 
+      @fluentd_lockdir = Dir.mktmpdir("fluentd-lock-")
+      ENV['FLUENTD_LOCKDIR'] = @fluentd_lockdir
+
       if config[:rpc_endpoint]
         @rpc_endpoint = config[:rpc_endpoint]
         @enable_get_dump = config[:enable_get_dump]
@@ -79,7 +82,13 @@ module Fluent
       stop_windows_event_thread if Fluent.windows?
       stop_rpc_server if @rpc_endpoint
       stop_counter_server if @counter
+      cleanup_lockdir
       Fluent::Supervisor.cleanup_resources
+    end
+
+    def cleanup_lockdir
+      FileUtils.rm(Dir.glob(File.join(@fluentd_lockdir, "fluentd-*.lock")))
+      FileUtils.rmdir(@fluentd_lockdir)
     end
 
     def run_rpc_server
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -217,7 +217,13 @@ module Fluent::Plugin
                end
 
       if @append
-        writer.call(path, chunk)
+        if @need_lock
+          acquire_worker_lock(path) do
+            writer.call(path, chunk)
+          end
+        else
+          writer.call(path, chunk)
+        end
       else
         find_filepath_available(path, with_lock: @need_lock) do |actual_path|
           writer.call(actual_path, chunk)


daipom:
Thanks for the implementation!
I think adding per-worker locking as a common feature of Fluentd would be a very elegant solution.
I will see this in detail.

lib/fluent/plugin/base.rb
```diff
@@ -51,6 +51,12 @@ def fluentd_worker_id
         @_fluentd_worker_id
       end
 
+      def fluentd_lockdir
+        return @_fluentd_lockdir if @_fluentd_lockdir
+        @_fluentd_lockdir = ENV['FLUENTD_LOCKDIR']
+        @_fluentd_lockdir
```
kou, line 57: We can simplify this:

```suggestion
        @_fluentd_lockdir ||= ENV['FLUENTD_LOCKDIR']
```

lib/fluent/plugin/base.rb
```diff
@@ -70,6 +76,19 @@ def multi_workers_ready?
         true
       end
 
+      def acquire_worker_lock(name, &block)
+        if fluentd_lockdir.nil?
+          raise RuntimeError, "fail to create lockfile on '#{fluentd_lockdir}'"
```
kou, line 81:
```suggestion
          raise RuntimeError, "failed to create lock file becuase FLUENTD_LOCKDIR isn't set'"
```

It may be better that we use an appropriate exception class instead of a general propose `RuntimeError`.

lib/fluent/plugin/base.rb
```diff
@@ -70,6 +76,19 @@ def multi_workers_ready?
...
+        end
+
+        name = name.gsub(/[^a-zA-Z0-9]/, "_")
+        lockfile = "fluentd-#{name}.lock"
+        File.open(File.join(fluentd_lockdir, lockfile), "w") do |f|
+          f.flock(File::LOCK_EX)
+          block.call()
```
kou, line 88:
```suggestion
          yield
```

lib/fluent/supervisor.rb
```diff
@@ -50,6 +50,9 @@ def before_run
       @rpc_server = nil
       @counter = nil
 
+      @fluentd_lockdir = Dir.mktmpdir("fluentd-lock-")
```
kou, line 53: How about using a block instead of manual cleanup?

```ruby
Dir.mktmpdir("fluentd-lock-") do |lockdir|
  ENV["FLUENTD_LOCKDIR"] = lockdir
  ...
end
```

lib/fluent/plugin/base.rb
daipom: If using `yield`, then we don't need `$block` argument.

lib/fluent/plugin/base.rb
daipom: If using `RuntimeError` after all, then should we write it just as follows?

```ruby
raise "fail to create lock file because FLUENTD_LOCKDIR isn't set'"
```

lib/fluent/plugin/base.rb
daipom: https://github.com/fluent/fluentd/pull/3808/files#diff-a011588d74f931c0c2154891220af14f9181f28eb70b978d2e017866dfba3081R31-R38

I don't know if this is really necessary, but should we set nil to `@_fluentd_lockdir` here?

lib/fluent/supervisor.rb
```diff
@@ -79,9 +82,15 @@ def after_run
       stop_windows_event_thread if Fluent.windows?
       stop_rpc_server if @rpc_endpoint
       stop_counter_server if @counter
+      cleanup_lockdir
       Fluent::Supervisor.cleanup_resources
     end
 
+    def cleanup_lockdir
+      FileUtils.rm(Dir.glob(File.join(@fluentd_lockdir, "fluentd-*.lock")))
+      FileUtils.rmdir(@fluentd_lockdir)
```
daipom, line 91:
```suggestion
      FileUtils.rmdir(@fluentd_lockdir)
    rescue StandardError
      # do nothing
```
Should we rescue exceptions for subsequent processing?

daipom:
The overall implementation direction is fine for me.
I have commented on the points I noticed.

lib/fluent/supervisor.rb
kou: We can use `FileUtils.rm_rf(@fluentd_lockdir)` but we don't need to use this method if we use `Dir.mktmpdir do ... end`. `Dir.mktmpdir do ... end` removes the temporary directory automatically.

lib/fluent/plugin/base.rb
kou: Putting `@_fluentd_lockdir = ENV['FLUENTD_LOCKDIR']` into `#initialize` is better.

fujimotos: Thank you for your review. Fixed by a538b9bcc6561aac500823f1b7332edae8c2d0cb

lib/fluent/plugin/base.rb
fujimotos: Fixed in a538b9bcc6561aac500823f1b7332edae8c2d0cb.

>It may be better that we use an appropriate exception class instead of a general propose RuntimeError

I defined "InvalidLockDirectory" error for that purpose.

lib/fluent/supervisor.rb
daipom: Oh, I'm sorry. We should use the block of `mktmpdir` and we don't need this change.

lib/fluent/supervisor.rb
fujimotos: 75ef92f787106d changed it so that Fluentd makes use of the automatic cleanup.
It is working on my machine.

Let's see it works nciely on other platforms as well...

lib/fluent/supervisor.rb
fujimotos: Thank you! I fixed it in 75ef92f787106de0a054174c2423a7f2fee00350, in such a way that we don't need to do any manual clean up any more.

lib/fluent/plugin/base.rb
fujimotos: Fixed. Thank you.

lib/fluent/plugin/base.rb
```diff
@@ -70,6 +74,19 @@ def multi_workers_ready?
         true
       end
 
+      def acquire_worker_lock(name, &block)
```
ashie, line 77:
```suggestion
      def acquire_worker_lock(name)
```

lib/fluent/plugin/base.rb
```diff
@@ -51,6 +51,10 @@ def fluentd_worker_id
         @_fluentd_worker_id
       end
 
+      def fluentd_lockdir
+        @_fluentd_lockdir ||= ENV['FLUENTD_LOCKDIR']
```
ashie, line 55: It should be initialized at `initialize`.

fujimotos: Addeed initialization (in the same way with fluentd_worker_id).

lib/fluent/plugin/base.rb
fujimotos: Thank you. Fixed.

lib/fluent/plugin/base.rb
kou: Why don't you put `@fluentd_lockdir = ENV['FLUENTD_LOCKDIR']`, remove this method and use `@fluentd_lockdir` directly?

BTW, `lock_dir` may be better than `lockdir` because `plugin_root_dir` uses `_` before `dir`.

daipom: Hmm, `@_fluentd_worker_id` is designed that way.
Does this mean that the environment variable can be set afterwards?

daipom:
I have confirmed that this works correctly on Windows too. 

* Windows10 (10.0.19043)

The results of the reproducing way in #3805 are as follows. (3 times each)

* This branch: d31631ec088c74cf7f190ed2639658394086e2c9

```
0 / 7706 lines are broken.
0 / 18388 lines are broken.
0 / 9933 lines are broken.
```

* Fluentd v1.15.0: 79ead40de9dec33afc2a9bf6cc25f5ec2d65a482

```
3 / 6786 lines are broken.
11 / 12245 lines are broken.
10 / 12283 lines are broken.
```

lib/fluent/plugin/base.rb
fujimotos: > Why don't you put @fluentd_lockdir = ENV['FLUENTD_LOCKDIR'], remove this method and use @fluentd_lockdir directly?

I fixed the initialization part following this.

fujimotos:
Unit test added.

ashie:
Hmm, behavior of GitHub seems strange.
I just merged this with merge button and it seems correctly applied to master branch, but this PR isn't closed yet.

ashie:
Probably it's caused by this incident of GitHub: https://www.githubstatus.com/incidents/dsj40hwxnldc

ashie:
The merged commits aren't broken, So I close this PR.
Sorry for that.

fujimotos:
## Testing with NFS storages

I've set up a NFS server and spent this morning testing this fix.

I configured Fluentd to perform an intensive concurrent writes, and run
the same test using v1.15.0 and master HEAD (20 times each). Here is
the result:

| #       | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |
| ------- | - | - | - | - | - | - | - | - | - | - |
| v1.15.0 | o | x | o | x | o | o | x | o | x | x |
| (cont)  | x | o | o | o | o | o | o | o | x | o |
| master  | o | o | o | o | o | o | o | o | o | o |
| (cont)  | o | o | o | o | o | o | o | o | o | o |

`o = No corruption`, `x = Corrupted output`

To put it short, v1.15.0 corrupted the output 7 times out of 20, while
the master HEAD never did. So I believe this fix works fine not only
on local disks, but also on NFS storages.

## Test Procedure Details

**System**

* Ubuntu 20.04 on AWS
* Created NFSv4 server on `/opt/nfs`.
* NFS Server was `nfs-kernel-server v1:1.3.4-2.5ubuntu3.4`

**Procedure**

1. Launch Fluentd with `bundle exec fluentd -c fluent.conf`
2. Run `send-record.sh`
3. Shutdown Fluentd
4. Run `check.py` on `/opt/nfs/test.*.log`

**fluentd.conf**

```xml
<system>
  workers 8
</system>
<source>
  @type forward
</source>
<match **>
  @type file
  path "/opt/nfs/test.log"
  append true
  <buffer>
    @type "file"
    path "buffer/"
    flush_mode interval
    flush_interval 60m
    flush_at_shutdown true
  </buffer>
</match>
```

**send-records.sh**

```bash
msg=`yes 1 | head -n 65536 | tr -d "\n"`
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
echo "{\"message\": \"$msg\"}" | bundle exec fluent-cat test.log &
wait
```

**check.py**

```python3
import sys
import json

i = 0
for line in sys.stdin:
    try:
        json.loads(line.strip().split(maxsplit=2)[2])
    except Exception:
        print(i, line)
        raise
    i += 1
print(i)
```

Issue #3614, opened 2022-02-01
Fluentd in_tail "unreadable" file causes "following tail of <file>" to stop and no logs pushed


salavessa:
### Describe the bug

After a warning of an "unreadable" (likely due to rotation), no more logs were pushed (in_tail + pos_file).
Reloading config or restarting fluentd sorts the issue.
All other existing files being tracked continued to work as expected.

### To Reproduce

Not able to reproduce at will.

### Expected behavior

Logs to be pushed as usual after file rotation as fluentd recovers from the temporary "unreadable" file.

### Your Environment

```markdown
- Fluentd version: 1.14.4
- TD Agent version: 4.2.0
- Operating system: Ubuntu 20.04.3 LTS
- Kernel version: 5.11.0-1022-aws
```

### Your Configuration

```apache
<source>
  @type tail
  path /var/log/containers/*.log
  pos_file /var/log/es-containers.log.pos
  tag kubernetes.*
  <parse>
    @type regexp
    expression /^(?<time>[^ ]+) (?<stream>[^ ]+) (?<logtag>[^ ]+) (?<log>.+)$/
    time_key time
    time_format %Y-%m-%dT%H:%M:%S.%N%z
  </parse>
  read_from_head true
</source>
```

### Your Error Log

```shell
Relevant log entries with some context. When "detected rotation of ..." isn't followed by a "following tail of ..." then log file contents aren't being processed/pushed:

2022-02-01 01:26:33 +0000 [info]: #0 detected rotation of /var/log/containers/gateway-mtcw6_atom_pcms-gateway-e54ea9ee115f1127682b0d92e37404e4b1693c14edeea93e485f70b3110eecfc.log; waiting 5 seconds
2022-02-01 01:26:33 +0000 [info]: #0 following tail of /var/log/containers/gateway-mtcw6_atom_pcms-gateway-e54ea9ee115f1127682b0d92e37404e4b1693c14edeea93e485f70b3110eecfc.log
2022-02-01 01:32:53 +0000 [warn]: #0 /var/log/containers/gateway-mtcw6_atom_pcms-gateway-e54ea9ee115f1127682b0d92e37404e4b1693c14edeea93e485f70b3110eecfc.log unreadable. It is excluded and would be examined next time.
[...]

PR #3829, opened 2022-07-25
in_tail: Show more information on skipping update_watcher

ashie:
None

In #3614 some users reported that in_tail rarely stops tailing. It seems
it's caused when skipping update_watcher due to unexpected duplicate
pos_file entries. The root cause is unknown yet. To investigate it, show
more infomation when it happens. Ideally it should't happen so that log
level "warn" is desired for it (I can't reproduce this issue yet).

**Docs Changes**:
None

**Release Note**: 
Same with the title


cosmo0920 (approved):

ashie:
Thanks for your review!

kashok7474:
@ashie Is there any update on this issue?
we are using fluentd 1.13.3 and observing similar issue, before upgrading fluentd version want to ensure it will work.

PR #3857, opened 2022-08-13
Add a new system config 'enable_jit'

fujimotos:

N/A

This adds a new option to enable Ruby JIT in worker processes.

Here is an example configuration:

```xml
<system>
  workers 3
  enable_jit true
</system>
```

The exact JIT strategy (in particular, mjit or yjit) will be chosen
by Ruby as it sees fit. See the following ticket for details:

https://bugs.ruby-lang.org/issues/18349

**Docs Changes**:

https://github.com/fluent/fluentd-docs-gitbook/pull/422

**Release Note**: 

Add a new system configuration `enable_jit`

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -1104,6 +1104,11 @@ module Fluent
         fluentd_spawn_cmd << '-Eascii-8bit:ascii-8bit'
       end
 
+      if @system_config.enable_jit
+        $log.info "enable Ruby JIT for workers (--jit)"
+        fluentd_spawn_cmd << '--jit'
+      end
+
       # Adding `-h` so that it can avoid ruby's command blocking
       # e.g. `ruby -Eascii-8bit:ascii-8bit` will block. but `ruby -Eascii-8bit:ascii-8bit -h` won't.
       _, e, s = Open3.capture3(*fluentd_spawn_cmd, "-h")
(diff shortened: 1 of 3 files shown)


fujimotos:
The good news is that Ruby JIT made Fluentd 10%-20% faster on my Ubuntu
22.04. Here is the summary of my experiments:

jit

The not-so-good news is that it's still somewhat experimental. In fact,
I observed some segmentation fault with Ruby 3.1 + YJIT.

While the particular SEGV bug seems to be already fixed in Ruby's
latest master, I think we should use this feature wisely.

daipom:
Thanks for the fix!
I'm not sure about Ruby JIT, but I don't see any problem with this fix.

daipom (approved):
Thanks for the fix!
I'm not sure about Ruby JIT, but I don't see any problem with this fix.

lib/fluent/supervisor.rb
```diff
@@ -1104,6 +1104,11 @@ def build_spawn_command
         fluentd_spawn_cmd << '-Eascii-8bit:ascii-8bit'
       end
 
+      if @system_config.enable_jit
+        $log.info "enable Ruby JIT for workers"
+        fluentd_spawn_cmd << '--jit'
```
kenhys, line 1109: It may be obvious, but it is better to mention --jit (YJIT since Ruby 3.1) because the default meaning is changed.

kenhys (approved):
LGTM.

lib/fluent/supervisor.rb
fujimotos: Added. Thanks.

ashie (approved):
LGTM

It might be better that add a command line option for this, we can add it by another pull request if it's really required.

k0kubun:
> Here is the summary of my experiments (with tail + ltsv parser)

@fujimotos Do you happen to have the benchmark script still? I'm interested in running the same thing if possible.

PR #3864, opened 2022-08-19
Make the worker lock work in '--daemon' mode as well

fujimotos:
Signed-off-by: Fujimoto Seiji <fujimoto@ceptord.net>

Fixes https://github.com/fluent/fluent-package-builder/issues/411

75ef92f introduced automatic clean up as follows:

    Dir.mktmpdir("fluentd-lock-") do |lock_dir|
        ENV['FLUENTD_LOCK_DIR'] = lock_dir
        se.run
    end

This does not work with `--daemon`, because `se.run` will not
block with that option. Fix this issue by reverting to the manual
clean up method in the initial version.

**Docs Changes**:

N/A

**Release Note**: 

`out_file: Fix append mode with --daemon flag.`

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -50,6 +50,9 @@ module Fluent
       @rpc_server = nil
       @counter = nil
 
+      @fluentd_lock_dir = Dir.mktmpdir("fluentd-lock-")
+      ENV['FLUENTD_LOCK_DIR'] = @fluentd_lock_dir
+
       if config[:rpc_endpoint]
         @rpc_endpoint = config[:rpc_endpoint]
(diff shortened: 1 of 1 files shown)


ashie (approved):
LGTM

fujimotos:
NOTE: I test this patch more (probably for some hours) and will merge this myself.

daipom:
Thanks for the fix!
I'm sorry for overlooking that `se.run` doesn't block with this option.

fujimotos:
OK. I can confirm that this patch works both on Linux and Windows.

In particular, I built a custom td-agent package that contains this patch
and confirmed the package on both platforms (RHEL and Windows).

So I believe this is releasable now.

PR #3866, opened 2022-08-22
winsvc: require serverengine

ashie:
Follow up for #3849

In #3849, getting ruby's path was replaced with
`ServerEngine.ruby_bin_path` but `require 'serverengine'` was missing.
As a result it fails to launch fluentd service.

**Docs Changes**:
None

**Release Note**: 
None (Same with #3849)

### DIFF
--- a/lib/fluent/winsvc.rb
+++ b/lib/fluent/winsvc.rb
@@ -19,6 +19,8 @@ begin
   require 'optparse'
   require 'win32/daemon'
   require 'win32/event'
+  require 'win32/registry'
+  require 'serverengine'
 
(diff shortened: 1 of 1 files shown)


cosmo0920 (approved):
Looks reasonable. Thanks for the quick fix! :+1:

cosmo0920:
BTW, can we add register/unregister e2e test on CI task? If not possible for now, testing manually is also OK for us.

ashie:
I think at least `service_main_start` should be tested.
I'm considering how to test it, I'll post another PR if I done it.

Issue #3869, opened 2022-08-22
CI: Fail on Ruby 3.2 preview2 on Windows


ashie:
### Describe the bug

CI fails with `ruby 3.2.0dev (2022-08-22 master 9f3140a42e) [x64-mingw-ucrt]` on Windows:
https://github.com/fluent/fluentd/runs/7945647354?check_suite_focus=true

It succeeded with ruby 3.2.0dev (2022-08-19 master 35c794b26d) [x64-mingw-ucrt] on Windows though:
https://github.com/fluent/fluentd/runs/7913475855?check_suite_focus=true

### To Reproduce

Run CI with `ruby 3.2.0dev (2022-08-22 master 9f3140a42e) [x64-mingw-ucrt]` on Windows

### Expected behavior

The test should succeed.

### Your Environment

```markdown
- Fluentd version: 1.15.2
- TD Agent version:
- Operating system: Windows
- Kernel version:
```

### Your Configuration

```apache
Changes in Ruby: https://github.com/ruby/ruby/compare/35c794b..9f3140a
```

### Your Error Log

```shell
2022-08-22T06:26:48.0261596Z   keepalive: 
2022-08-22T06:26:48.7718954Z     test: Do not create connection per send_data:	.: (0.745608)
2022-08-22T06:26:49.0457212Z     test: create timer of purging obsolete sockets:	.: (0.273691)
2022-08-22T06:26:49.0458376Z     with require_ack_response: 
2022-08-22T06:26:49.8032564Z #<Thread:0x000001febc53fe78@event_loop D:/a/fluentd/fluentd/lib/fluent/plugin_helper/thread.rb:70 run> terminated with exception (report_on_exception is true):
2022-08-22T06:26:49.8044023Z D:/rubyinstaller-head-x64/lib/ruby/gems/3.2.0+2/gems/cool.io-1.7.1/lib/cool.io/io.rb:35:in `attach': expected loop to be an instance of Coolio::Loop, not nil (ArgumentError)
2022-08-22T06:26:49.8044554Z 
2022-08-22T06:26:49.8044905Z       @_read_watcher.attach(loop)
2022-08-22T06:26:49.8045358Z                             ^^^^
2022-08-22T06:26:49.8047633Z 	from D:/rubyinstaller-head-x64/lib/ruby/gems/3.2.0+2/gems/cool.io-1.7.1/lib/cool.io/io.rb:35:in `attach'
[...]

PR #3883, opened 2022-09-01
Fix Errno::EACCES which occurs while testing on Ruby 3.2

ashie:
ref #3869

`FileUtils.rm_rf` of Ruby 3.2 no longer ignore `Errno::EACCES`:
https://github.com/ruby/ruby/commit/983115cf3c8f75b1afbe3274f02c1529e1ce3a81

But it will occur on Windows test because of delete pending state.
Most of them can be fixed by adding `File::Constants::SHARE_DELETE` on calling `File.open`.
`Fluent::FileWrapper.open` will do it, so I replace `File.open` in failed test with it.

But tests including `LogDevice` can't use it easily.
For such tests, simply ignore `Errno::EACCES` as ever.
(To avoid using a same temp file between different tests, create random temp directory instead.)

Please note that remaining 1 test on Ruby 3.2 Windows still fails because there is another issue: https://github.com/fluent/fluentd/issues/3869#issue-1346074285
I'll investigate it later.

**Docs Changes**:
None

**Release Note**: 
Same with the title


test/test_log.rb
```diff
@@ -21,6 +24,13 @@ def teardown
     @log_device.reset
     Timecop.return
     Thread.current[:last_repeated_stacktrace] = nil
+    begin
+      FileUtils.rm_rf(@tmp_dir)
+    rescue Errno::EACCES
+      # It may occur on Windows because of delete pending state due to delayed GC.
+      # Ruby 3.2 or later doesn't ignore Errno::EACCES:
+      # https://github.com/ruby/ruby/commit/983115cf3c8f75b1afbe3274f02c1529e1ce3a81
+    end
```
cosmo0920, line 33: :+1:

test/test_supervisor.rb
```diff
@@ -22,17 +24,29 @@ def config
...
+  end
+
+  def teardown
+    begin
+      FileUtils.rm_rf(@tmp_dir)
+    rescue Errno::EACCES
+      # It may occur on Windows because of delete pending state due to delayed GC.
+      # Ruby 3.2 or later doesn't ignore Errno::EACCES:
+      # https://github.com/ruby/ruby/commit/983115cf3c8f75b1afbe3274f02c1529e1ce3a81
+    end
```
cosmo0920, line 44: :+1:

cosmo0920 (approved):
Looks good. And I didn't notice that Ruby 3.2 won't ignore Errno::EACCES anymore.

PR #3899, opened 2022-09-21
Add method for testing filtered_with_time

nicklas-dohrn:
This change allows plugin developers to test their filter_with_time plugins without sideloading a new test driver.

Fixes #

**Docs Changes**:

**Release Note**:


lib/fluent/test/driver/filter.rb
```diff
@@ -37,6 +37,10 @@ def filtered_records
           @filtered.map {|_time, record| record }
         end
 
+        def filtered_time
+          @filtered.map {|_time, record| _time }
```
ashie, line 41:
```suggestion
          @filtered.map {|time, _record| time }
```

The prefix `_` means `unused` so that it should be added to `record` not `time` for this case.

ashie (changes requested):

lib/fluent/test/driver/filter.rb
ashie: I've fixed it.

PR #3922, opened 2022-10-20
Fix a test failure on ruby 3.2.0dev (2022-10-20)

ashie:
N/A

A test fails on latest Ruby-3.2dev.

The expected error massage pattern was

```
in_buggy.rb:5: syntax error, unexpected end-of-input, expecting
```

but the actual error message on Ruby 3.2 is

```
in_buggy.rb:5: syntax error, unexpected end-of-input (SyntaxError)
```

**Docs Changes**:
N/A

**Release Note**: 
N/A


ashie:
I've confirmed that the failed test has been fixed on CI.

PR #3939, opened 2022-10-31
Fix "unable to rotate log files on Windows" bug

### LINKED ISSUE #3921: Windows: log-rotate setting by config file does not work correctly
### Describe the bug

https://docs.fluentd.org/deployment/logging#log-rotation-setting

> NOTE: When --log-rotate-age is specified on Windows, log files are separated into log-supervisor-0.log, log-0.log, ..., log-N.log where N is generation - 1 due to the system limitation. Windows does not permit delete and rename files simultaneously owned by another process.

> NOTE: When --log-rotate-size is specified on Windows, log files are separated into log-supervisor-0.log, log-0.log, ..., log-N.log where N is generation - 1 due to the system limitation. Windows does not permit delete and rename files simultaneously owned by another process.

As this document says, If rotation is enabled, each process must output a different log file on Windows.

However, when using [log-rotate setting of `system` directive](https://docs.fluentd.org/deployment/system-config#less-than-log-greater-than-section),  the supervisor process outputs its log to `log-0.log`, not `log-supervisor-0.log`.

When using command line options such as `--log-rotate-age`, the following code in `command/fluentd.rb` fixes the path for the supervisor.

* https://github.com/fluent/fluentd/blob/v1.15.2/lib/fluent/command/fluentd.rb#L348-L356

This process seems to be missing for log-rotate setting of `system` directive.

### To Reproduce

Just run the fluentd with the following environment and the following config.

```cmd
> bundle exec fluentd -c fluentd.conf -o fluentd.log
```

### Expected behavior

First, the following two files are created.

* `fluentd-0.log`
* `fluentd-supervisor-0.log`

Then, they will be rotated without an error.

### Your Environment

```markdown
- Fluentd version: 1.15.2
- TD Agent version:
- Operating system: Windows 10
- Kernel version:
```

### Your Configuration

```apache
<system>
  <log>
    rotate_age 5
    rotate_size 2000
  </log>
</system>

<source>
  @type sample
  tag test.log
  rate 100
  dummy {"message": "This is the test message"}
</source>

<match test.**>
  @type stdout
</match>
```

### Your Error Log

Immediately the following error begins to output to the console.

```shell
> bundle exec fluentd -c fluentd.conf -o fluentd.log
log shifting failed. Permission denied @ rb_file_s_rename - (.\fluent\fluentd-0.log, .\fluent\fluentd-0.log.0)
log writing failed. closed stream
log shifting failed. closed stream
...
```

### Additional context

When using command line options, this problem does not occur.

```cmd
> bundle exec fluentd -c fluentd.conf -o fluentd.log --log-rotate-age 5
```

fujimotos:

Fixes #3921

```
Suppose we set up the log rotation on Windows as follows:

    <system>
      <log>
        rotate_age 5
      </log>
    </sytem>

Currently this does not work, because the supervisor and worker will
share the same log path, and they won't coordinate with each other.

This makes it sure that each process has a log file for themselves
on Windows.

```

**Docs Changes**:

N/A

**Release Note**: 

Fix log initializer to correctly create per-process files on Windows

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -345,17 +345,6 @@ end
 exit 0 if early_exit
 
 if opts[:supervise]
-  if Fluent.windows?
-    if opts[:log_path] && opts[:log_path] != "-"
-      if opts[:log_rotate_age] || opts[:log_rotate_size]
-        require 'pathname'
-
-        log_path = Pathname(opts[:log_path]).sub_ext("-supervisor#{Pathname(opts[:log_path]).extname}").to_s
-        opts[:log_path] = log_path
-      end
-    end
-  end
-
   supervisor = Fluent::Supervisor.new(opts)
   supervisor.configure(supervisor: true)
   supervisor.run_supervisor(dry_run: opts[:dry_run])
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -525,10 +525,22 @@ module Fluent
         @log_rotate_size = log_rotate_size
       end
 
-      def worker_id_suffixed_path(worker_id, path)
-        require 'pathname'
+      # Create a unique path for each process.
+      #
+      # >>> per_process_path(:worker, 1, "C:/tmp/test.log")
+      # C:/tmp/test-1.log
+      # >>> per_process_path(:supervisor, 0, "C:/tmp/test.log")
+      # C:/tmp/test-supervisor-0.log
+      def self.per_process_path(path, process_type, worker_id)
+        path = Pathname(path)
+        ext = path.extname
 
-        Pathname(path).sub_ext("-#{worker_id}#{Pathname(path).extname}").to_s
+        if process_type == :supervisor
+          suffix = "-#{process_type}-0#{ext}"  # "-0" for backword compatibility.
+        else
+          suffix = "-#{worker_id}#{ext}"
+        end
+        return path.sub_ext(suffix).to_s
       end
 
       def init(process_type, worker_id)
@@ -540,13 +552,19 @@ module Fluent
             FileUtils.mkdir_p(File.dirname(@path))
           end
 
-          @logdev = if @log_rotate_age || @log_rotate_size
-                     Fluent::LogDeviceIO.new(Fluent.windows? ?
-                                               worker_id_suffixed_path(worker_id, @path) : @path,
-                                             shift_age: @log_rotate_age, shift_size: @log_rotate_size)
-                   else
-                     File.open(@path, "a")
-                   end
+          if @log_rotate_age || @log_rotate_size
+            # We need to prepare a unique path for each worker since
+            # Windows locks files.
+            if Fluent.windows?
+                path = LoggerInitializer.per_process_path(@path, process_type, worker_id)
+            else
+                path = @path
+            end
+            @logdev = Fluent::LogDeviceIO.new(path, shift_age: @log_rotate_age, shift_size: @log_rotate_size)
+          else
+            @logdev = File.open(@path, "a")
+          end
+
           if @chuser || @chgroup
             chuid = @chuser ? ServerEngine::Privilege.get_etc_passwd(@chuser).uid : nil
             chgid = @chgroup ? ServerEngine::Privilege.get_etc_group(@chgroup).gid : nil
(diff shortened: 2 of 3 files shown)


daipom:
Thanks for the fix!

lib/fluent/supervisor.rb
```diff
@@ -525,10 +525,22 @@ def initialize(path, level, chuser, chgroup, opts, log_rotate_age: nil, log_rota
...
+      def self.per_process_path(path, process_type, worker_id)
+        path = Pathname(path)
+        ext = path.extname
+
+        if process_type == :supervisor
+          suffix = "-#{process_type}-0#{ext}"  # "-0" for backword compatibility.
+        else
+          suffix = "-#{worker_id}#{ext}"
+        end
+        return path.sub_ext(suffix).to_s
```
daipom, line 543: Considering the surrounding code style, it seems better to omit this 'return'.

lib/fluent/command/fluentd.rb
```diff
@@ -345,17 +345,6 @@
 exit 0 if early_exit
 
 if opts[:supervise]
-  if Fluent.windows?
-    if opts[:log_path] && opts[:log_path] != "-"
-      if opts[:log_rotate_age] || opts[:log_rotate_size]
-        require 'pathname'
-
-        log_path = Pathname(opts[:log_path]).sub_ext("-supervisor#{Pathname(opts[:log_path]).extname}").to_s
-        opts[:log_path] = log_path
```
daipom, line 354: I wondered if this value is used for anything else, so I checked it.

This `log_path` value will be used in 2 other locations, but this change does not affect them.
So this looks fine.

Firstly, the config passed to ServerEngine has this value

* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/supervisor.rb#L474

However, ServerEnigne does not use this value and neither do ServerModule and WorkerModule.
So we can stop passing this `log_path` value to ServerEngine, but it has nothing to do with this issue.

Secondly, the `options` method uses this value.

* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/supervisor.rb#L727-L735

The `in_monitor_agent` plugin will use this.

* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/plugin/in_monitor_agent.rb#L403-L410 
* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/plugin/in_monitor_agent.rb#L399-L401
* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/plugin/in_monitor_agent.rb#L64-L83
* https://github.com/fluent/fluentd/blob/a3b1cdaa8c171f51e2b3d37e7f83b263243a236b/lib/fluent/plugin/in_monitor_agent.rb#L216-L217

We can get the `log_path` value by `api/config` or `api/config.json` of `in_monitor_agent`.

However, this path is only for workers.
We can't know supervisor's `log_path` since this is just one plugin's function.

The previous logic overwrote only the supervisor's `log_path`.
So this fix does not affect this plugin.

We can fix this to allow this plugin to display the actual `log_path` of each worker, but I think this should be separated from this issue as well.

lib/fluent/supervisor.rb
daipom:
```suggestion
        path.sub_ext(suffix).to_s
```

daipom (approved):
I can confirm that it works as expected in my environment!
Thanks for the fix!

lib/fluent/supervisor.rb
```diff
@@ -565,6 +583,7 @@ def init(process_type, worker_id)
         $log = Fluent::Log.new(logger, @opts)
         $log.enable_color(false) if @path
         $log.enable_debug if @level <= Fluent::Log::LEVEL_DEBUG
+        $log.info "init #{process_type} logger", path: path, rotate_age: @log_rotate_age, rotate_size: @log_rotate_size
```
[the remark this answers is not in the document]
daipom: Great!

ashie (approved):
LGTM
Thanks!

PR #3942, opened 2022-11-01
out_file: Avoid performing multi-worker check in configure()

### LINKED ISSUE #3941: out_file: start-up check fails with <worker 0-N> directive
### Describe the bug

When `out_file` is configured with `append=true` and `workers > 1`,
it tries to confirm that the inter-worker lock is available.

This check can cause an unexpected exception during startup if it's
combined with ranged worker directive (such as `<worker 0-2>`)

### To Reproduce

1. Save the following configuration as `fluentd.conf`
2. Run `fluentd -c fluentd.conf`

### Expected behavior

Fluentd should start up without errors.

### Your Environment

```markdown
- Fluentd version: v1.15.2
- TD Agent version: N/A
- Operating system: N/A
- Kernel version: N/A
```

### Your Configuration

```apache
<system>
  workers 2
</system>

<worker 0-1>
  <source>
    @type dummy
    tag test.log
  </source>
  
  <match test.**>
    @type file
    path output.log
    append true
    <buffer>
      @type memory
    </buffer>
  </match>
</worker>
```

### Your Error Log

```shell
out_file.rb:193:in `configure': must set FLUENTD_LOCK_DIR on multi-worker append mode (Fluent::InvalidLockDirectory)
```

### Additional context

_No response_

fujimotos:
Fixes #3941

    The check did not work well with `<worker 0-N>` directive for
    the following reasons:
    
     * When Fluentd starts up, supervisor performs sanity check on
       configuration by creating actual instances of plugins.
    
     * This check precedes SeverModule, which prepares various assets
       for workers (lockdir, RPC endpoints and shared sockets etc.)
    
     * Fluent::Plugin::Base dynamically overrides `system_config.workers`
       during start-up, and leaves the value > 1 only when `<worker 0-N>`
       directive is used.
    
    Fix this issue by avoiding faulty sanity check in `configure()`.
    
**Docs Changes**:

N/A

**Release Note**: 

out_file: Fix the multi-worker check with `<worker 0-N>` directive

### DIFF
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -188,10 +188,6 @@ module Fluent::Plugin
         condition = Gem::Dependency.new('', [">= 2.7.0", "< 3.1.0"])
         @need_ruby_on_macos_workaround = true if condition.match?('', RUBY_VERSION)
       end
-
-      if @need_lock && @append && @fluentd_lock_dir.nil?
-        raise Fluent::InvalidLockDirectory, "must set FLUENTD_LOCK_DIR on multi-worker append mode"
-      end
     end
 
     def multi_workers_ready?


fujimotos:
@ashie Can you check this PR before the next release? If it looks good,
I want to included this in td-agent v4.4.2 along with #3939.

ashie:
> > * Fluent::Plugin::Base dynamically overrides `system_config.workers`
> >   during start-up, and leaves the value > 1 only when `<worker 0-N>`
> >   directive is used.
> 
> https://github.com/fluent/fluentd/blob/a7771835d65dea8e49d284a148d94cecb1abed19/lib/fluent/plugin/base.rb#L56-L62

`conf.target_worker_ids` is set only when `<worker 0-N>` is used:
https://github.com/fluent/fluentd/blob/a7771835d65dea8e49d284a148d94cecb1abed19/lib/fluent/root_agent.rb#L78-L121

ashie:
When `<worker 0-N>` isn't specified `system_config.workers` is `1` on sanity check of supervisor so that this check doesn't return error.

ashie (approved):
LGTM
We may need to address confusing `system_config.workers` issue later.

ashie:
Thanks!

daipom:
Thanks for the fix!

PR #3955, opened 2022-11-11
in_tail: Fix unhandled exception due to NFS race conditions

fujimotos:

N/A

```
While we're reading from a remote file on NFS, the same file can
be concurrently deleted by another user. POSIX defines a special
error code (ESTALE) specifically for that case.

Sadly Fluentd just loudly crashes on such cases:

    2022-01-01 00:00:00 [error]: Stale file handler @ io_getpartial

This teaches Fluentd to treat ESTALE same as EOF.

```

**Docs Changes**:

N/A

**Release Note**: 

in_tail: Fix unhandled exception due to NFS race conditions

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -1097,6 +1097,9 @@ module Fluent::Plugin
                   end
                 rescue EOFError
                   @eof = true
+                rescue Errno::ESTALE
+                  @log.info "remote file was deleted during read", path: path
+                  @eof = true
                 end
               end
 


daipom (approved):
LGTM. Thanks for the fix!

lib/fluent/plugin/in_tail.rb
```diff
@@ -1095,7 +1095,7 @@ def handle_notify
                       break
                     end
                   end
-                rescue EOFError
+                rescue EOFError, Errno::ESTALE
```
ashie, line 1098: Isn't it needed to output some warning or info log message for this?

fujimotos: > Isn't it needed to output some warning or info log message for this?

Pushed a new commit 331ac72 to incorporate that feedback. Added a
sepearate `reque` and called `log.info` within that block.

lib/fluent/plugin/in_tail.rb
```diff
@@ -1097,6 +1097,9 @@ def handle_notify
                   end
                 rescue EOFError
                   @eof = true
+                rescue Errno::ESTALE
+                  @log.info "remote file was deleted during read", path: path
```
ashie, line 1101:
```suggestion
                  @log.info "remote file was deleted during read", path: @path
```
A local variable or method with the name `path` doesn't seem exist.

ashie (changes requested):

PR #4013, opened 2023-01-17
CI: Remove Drone.io

ashie:
N/A

Last 2 months it always stalls without any output, I couldn't find the reason in any logs & documents. In addition, we rarely detected any real issues with Arm architecture there in fact.
I think it's time to remove it.

**Docs Changes**:
N/A

**Release Note**: 
N/A


cosmo0920 (approved):
Yup. We can safely remove this. :+1:

In my opinion, ARM64 specific issue is not observed for a quite long times.

ashie:
Thanks!

ashie:
I've disabled webhook for done.io.

Issue #3970, opened 2022-11-25
Add more checks for buffer corruption on startup


fujimotos:
### Is your feature request related to a problem? Please describe.

Currently when starting up Fluentd outputs, we try to check if each buffer chunk
is non-empty, and if it has some bytes, we assume it contains valid data.

It turned out that this operation model has a few issues:

* If the data was indeed corrupted, the behavior is undefined. It likely causes
  many kinds of errors in various parts of the pipeline.
* It is also hard to tell which chunks was corrupted from td-agent.log.
  This is important because users probably want to recover the lost data.

We should perform more rigorous buffer checks on startup,
so that Fluentd can handle corrupted chunks gracefully.

### Describe the solution you'd like

* Perform more sanity checks on buffer chunks on startup.
* Emit more error logs regarding the corrupted chunks.

### Describe alternatives you've considered

N/A

daipom:
Thanks for summarizing this issue!
I'm willing to consider this issue in December.
Here are my impressions at present.

> It is also hard to tell which chunks was corrupted from td-agent.log.

I feel the next log should be the info level since these chunks could be a problem in the case of abnormal system termination, such as a machine power failure.

* https://github.com/fluent/fluentd/blob/981decb99c0cad186c0b12d4b2587d92b426315c/lib/fluent/plugin/buf_file.rb#L148

> Perform more sanity checks on buffer chunks on startup.

We have the check in `FileChunk` initialization, but it can only detect the corruption of the meta data.
If the chunk body file is corrupt, it proceeds to the next logic.

* https://github.com/fluent/fluentd/blob/981decb99c0cad186c0b12d4b2587d92b426315c/lib/fluent/plugin/buffer/file_chunk.rb#L339-L345

`in_forward` has the feature of checking the stream, but it is disabled by default, so we may not notice that corrupted data is being sent. 

* https://github.com/fluent/fluentd/blob/981decb99c0cad186c0b12d4b2587d92b426315c/lib/fluent/plugin/in_forward.rb#L367-L377

daipom:
I am examining this issue.

The most important thing is to detect file corruption at abnormal system termination, such as a machine power failure.
To handle this, we should improve the process of loading the existing chunk files at startup.

> I feel the next log should be the info level since these chunks could be a problem in the case of abnormal system termination, such as a machine power failure.
> 

[...]

PR #4025, opened 2023-01-26
buffer: backup broken file chunk

daipom:
Partial fix for #3970

Backup feature was implemented in #1952, but it didn't support handling broken file chunks found in resuming buffer.
This extends the backup feature to support it.

When a file is corrupted due to power failure or other reasons, backing up the file allows us to guess which data was corrupted.

Example of a forwarder:

```xml
<system>
  root_dir /test/fluentd/forwarder
</system>

<source>
  @type sample
  tag test
</source>

<match test.**>
  @id test_id
  @type forward
  <buffer tag,time,file>
    @type file
    path /test/fluentd/forwarder/buffer
    timekey 24h
    flush_mode interval
    flush_interval 10s
  </buffer>
  <server>
    host localhost
    port 24224
  </server>
</match>
```

We can reproduce file corruption due to power failure as follows.
(Especially in Windows, a phenomenon in which the data being edited is turned to a sequence of zero is confirmed.)

* Stop the fluentd.
* Fill in the second half of the chunk data with zeros.
* Break the meta-data completely.

```console
$ truncate -s 80 buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log
$ truncate -s 160 buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log
$ head -c 89 /dev/zero > buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log.meta
```

Then, restart the fluentd, and it backups the chunk.
The logs of fluentd are as follows.

```
[info]: #0 fluent/log.rb:330:info: starting fluentd worker pid=891085 ppid=891065 worker=0
[debug]: #0 [test_id] restoring buffer file: path = /test/fluentd/forwarder/buffer/buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log
[error]: #0 [test_id] found broken chunk file during resume. path="/test/fluentd/forwarder/buffer/buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log" mode=:staged err_msg="staged meta file is broken. no implicit conversion of Symbol into Integer"
[warn]: #0 [test_id] bad chunk is moved to /test/fluentd/forwarder/backup/worker0/test_id/5f32232e76a4d1bdfdbeed36c384b03b.log
```

We can check the backup file as follows.

```rb
require "fleunt/event"
chunk_data = File.read("buffer.b5f32232e76a4d1bdfdbeed36c384b03b.log")
Fluent::MessagePackEventStream.new(chunk_data, nil, 0).each {|time, record| p time, record}

2023-01-26 12:18:12.017325082 +0900
{"message"=>"sample"}
2023-01-26 12:18:13.020290925 +0900
{"message"=>"sample"}
2023-01-26 12:18:14.022211281 +0900
{"message"=>"sampl\u0000"}
0
nil
0
nil
...
```

**Docs Changes**:
TODO

**Release Note**: 
TODO

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -195,8 +195,23 @@ module Fluent
       end
 
       def handle_broken_files(path, mode, e)
-        log.error "found broken chunk file during resume. Deleted corresponding files:", :path => path, :mode => mode, :err_msg => e.message
-        # After support 'backup_dir' feature, these files are moved to backup_dir instead of unlink.
+        log.error "found broken chunk file during resume.", :path => path, :mode => mode, :err_msg => e.message
+        unique_id = Fluent::Plugin::Buffer::FileChunk.unique_id_from_path(path)
+        if @disable_chunk_backup
+          log.warn "disable_chunk_backup is true. #{dump_unique_id_hex(unique_id)} chunk is thrown away"
+          return
+        end
+        backup(unique_id) { |f|
+          File.open(path, 'rb') { |chunk|
+            chunk.set_encoding(Encoding::ASCII_8BIT)
+            chunk.sync = true
+            chunk.binmode
+            IO.copy_stream(chunk, f)
+          }
+        }
+      rescue => error
+        log.error "backup failed. Delete corresponding files.", :err_msg => error.message
+      ensure
         File.unlink(path, path + '.meta') rescue nil
       end
 
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -197,10 +197,6 @@ module Fluent
       def handle_broken_files(path, mode, e)
         log.error "found broken chunk file during resume.", :path => path, :mode => mode, :err_msg => e.message
         unique_id = Fluent::Plugin::Buffer::FileChunk.unique_id_from_path(path)
-        if @disable_chunk_backup
-          log.warn "disable_chunk_backup is true. #{dump_unique_id_hex(unique_id)} chunk is thrown away"
-          return
-        end
         backup(unique_id) { |f|
           File.open(path, 'rb') { |chunk|
             chunk.set_encoding(Encoding::ASCII_8BIT)
@@ -212,6 +208,7 @@ module Fluent
       rescue => error
         log.error "backup failed. Delete corresponding files.", :err_msg => error.message
       ensure
+        log.warn "disable_chunk_backup is true. #{dump_unique_id_hex(unique_id)} chunk is thrown away." if @disable_chunk_backup
         File.unlink(path, path + '.meta') rescue nil
       end
 
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -66,6 +66,9 @@ module Fluent
       desc 'Compress buffered data.'
       config_param :compress, :enum, list: [:text, :gzip], default: :text
 
+      desc 'If true, chunks are thrown away when unrecoverable error happens'
+      config_param :disable_chunk_backup, :bool, default: false
+
       Metadata = Struct.new(:timekey, :tag, :variables, :seq) do
         def initialize(timekey, tag, variables)
           super(timekey, tag, variables, 0)
@@ -901,6 +904,18 @@ module Fluent
         end
 
         { 'buffer' => stats }
+      end
+
+      def backup(chunk_unique_id)
+        unique_id = dump_unique_id_hex(chunk_unique_id)
+        safe_owner_id = owner.plugin_id.gsub(/[ "\/\\:;|*<>?]/, '_')
+        backup_base_dir = system_config.root_dir || DEFAULT_BACKUP_DIR
+        backup_file = File.join(backup_base_dir, 'backup', "worker#{fluentd_worker_id}", safe_owner_id, "#{unique_id}.log")
+        backup_dir = File.dirname(backup_file)
+
+        log.warn "bad chunk is moved to #{backup_file}"
+        FileUtils.mkdir_p(backup_dir, mode: system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION) unless Dir.exist?(backup_dir)
+        File.open(backup_file, 'ab', system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION) { |f| yield f }
       end
 
       private
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -908,6 +908,12 @@ module Fluent
 
       def backup(chunk_unique_id)
         unique_id = dump_unique_id_hex(chunk_unique_id)
+
+        if @disable_chunk_backup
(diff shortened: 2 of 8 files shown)


daipom:
I'm fixing `TestFluentPluginConfigFormatter`...

daipom:
Hmm, `disable_chunk_backup` option is still in `buffer` section, but I moved it into the `buffer` plugin from `output` plugin.

I thought this was OK since this doesn't change the config shape, but it seems this is not good.
This will make the `output` plugins directly using this option unworkable.
I will add the getter for the compatibility.

lib/fluent/plugin/buffer.rb
```diff
@@ -903,6 +906,18 @@ def statistics
...
+        unique_id = dump_unique_id_hex(chunk_unique_id)
+        safe_owner_id = owner.plugin_id.gsub(/[ "\/\\:;|*<>?]/, '_')
+        backup_base_dir = system_config.root_dir || DEFAULT_BACKUP_DIR
+        backup_file = File.join(backup_base_dir, 'backup', "worker#{fluentd_worker_id}", safe_owner_id, "#{unique_id}.log")
+        backup_dir = File.dirname(backup_file)
+
+        log.warn "bad chunk is moved to #{backup_file}"
+        FileUtils.mkdir_p(backup_dir, mode: system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION) unless Dir.exist?(backup_dir)
+        File.open(backup_file, 'ab', system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION) { |f| yield f }
+      end
```
ashie, line 919: Memo: It seems be inspired by https://github.com/fluent/fluentd/blob/bf7498d6ff2805bfa517bbd6f7c17103c7d879d4/lib/fluent/plugin/output.rb#L1242-L1259

daipom: Yes, I need to move the backup logic to `buffer` in order to backup broken chunk files in `buf_file` and `buf_file_single`

lib/fluent/plugin/buf_file.rb
```diff
@@ -195,8 +195,23 @@ def generate_chunk(metadata)
...
 
       def handle_broken_files(path, mode, e)
-        log.error "found broken chunk file during resume. Deleted corresponding files:", :path => path, :mode => mode, :err_msg => e.message
-        # After support 'backup_dir' feature, these files are moved to backup_dir instead of unlink.
+        log.error "found broken chunk file during resume.", :path => path, :mode => mode, :err_msg => e.message
+        unique_id = Fluent::Plugin::Buffer::FileChunk.unique_id_from_path(path)
+        if @disable_chunk_backup
+          log.warn "disable_chunk_backup is true. #{dump_unique_id_hex(unique_id)} chunk is thrown away"
+          return
+        end
```
ashie, line 203: It might be better to move into `backup` method as same as `Fluent::Output::backup_chunk`, so that:
* `@disable_chunk_backup` is only referred in lib/fluent/buffer.rb, it makes easy to read.
* If other child classes are added, they can use same implementation.

daipom: Indeed! This logic should be in the parent `buffer` class. I will fix it.

daipom: I just remembered why I didn't do that.

The function of throwing away the chunks should be in the child class, so it was difficult to move this warning message to the parent class.

But I think I can do something good about this. I will try.

daipom: I will fix this based on the following idea.

The parent class `Buffer` should have the logic not to back up with `@disable_chunk_backup` setting.

On the other hand, the logic of throwing chunks is for child classes, so the warning message should remain in them.

daipom: I have fixed this @ 6152b8fc111adb1e137546ab6c1bc5958e3682bc
I separated the warning message into the parent class and the child classes.

```
 [info]: #0 starting fluentd worker pid=773861 ppid=773841 worker=0
 [error]: #0 [test_id] found broken chunk file during resume. path="/test/fluentd/buffer/buffer.b5f48ea187091b2e4a46d3b1fa7e0a730.log" mode=:staged err_msg="staged meta file is broken. no implicit conversion of Symbol into Integer"
 [warn]: #0 [test_id] disable_chunk_backup is true. 5f48ea187091b2e4a46d3b1fa7e0a730 chunk is not backed up.
 [warn]: #0 [test_id] disable_chunk_backup is true. 5f48ea187091b2e4a46d3b1fa7e0a730 chunk is thrown away.
 [info]: #0 fluentd worker is now running worker=0
```

lib/fluent/plugin/output.rb
```diff
@@ -435,6 +435,12 @@ def configure(conf)
         self
       end
 
+      def keep_buffer_config_compat
+        # Need this to call `@buffer_config.disable_chunk_backup` just as before,
+        # since some plugins may use this option in this shape.
```
cosmo0920, line 440: > in this shape

in this way?

daipom: Thanks! Was it strange English? I will fix this.

daipom: Fixed. Thanks!

cosmo0920: > Thanks! Was it strange English? I will fix this.

Yes. Normally, "(how to) do it like this" can be replaced with "in this way". o/

daipom: I see! Thanks!

test/plugin/test_buf_file.rb
```diff
@@ -1232,44 +1230,85 @@ def compare_log(plugin, msg)
...
       c2id, _ = create_second_chunk('b')
 
-      @p.start
+      Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
+        @p.start
+      end
+
       compare_staged_chunk(@p.stage, c2id, '2016-04-17 14:01:00 -0700', 3, :staged)
       compare_log(@p, 'staged file chunk is empty')
+      assert_false File.exist?(p1)
```
ashie, line 1245: Power assert style (`assert { File.exist?(p1) }`) is preferred:
https://github.com/fluent/fluentd/pull/4034#discussion_r1096937376

daipom: Thanks! Yes! Power assertion is great! I have fixed entirely assertions related to this fix.

lib/fluent/plugin/output.rb
```diff
@@ -378,6 +377,7 @@ def configure(conf)
           buffer_conf = conf.elements(name: 'buffer').first || Fluent::Config::Element.new('buffer', '', {}, [])
           @buffer = Plugin.new_buffer(buffer_type, parent: self)
           @buffer.configure(buffer_conf)
+          keep_buffer_config_compat
```
ashie, line 380: I'm considering how to do it more gracefully, but I don't yet get good idea.
I'll merge this as is, after considering a few days...

daipom: I see...

 Is there a possibility that it would be better not to transfer the option to `Buffer`?

daipom:
Thanks for your review!

PR #4027, opened 2023-01-26
buffer: warning message for restoring buffer with flush_at_shutdown

daipom:
Partial fix for #3970

When `flush_at_shutdown` is `true`, the remaining chunk files during resuming are possibly broken since there may be a power failure or similar failure.

Without this log, even if we notice that we have received corrupted data on the destination server, we can not know which chunks may have been corrupted.

**Docs Changes**:
Not needed.

**Release Note**: 
Add warning messages for restoring buffer with `flush_at_shutdown true`.

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -145,7 +145,11 @@ module Fluent
         Dir.glob(escaped_patterns(patterns)) do |path|
           next unless File.file?(path)
 
-          log.debug { "restoring buffer file: path = #{path}" }
+          if owner.respond_to?(:buffer_config) && owner.buffer_config&.flush_at_shutdown
+            log.warn { "restoring buffer file: path = #{path}" }
+          else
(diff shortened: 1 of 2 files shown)


ashie (approved):
LGTM

Is it O.K. to merge this before #4025 ?

daipom:
Thanks for your review!

Yes, please!

ashie (changes requested):
Sorry, one more thing.

> When flush_at_shutdown is true, the remaining chunk files during resuming are possibly broken since there may be a power failure or similar failure.

Could you add this information as the commit message or comment in the code?

daipom:
Indeed! I will add the comment to the code.

daipom:
I have added the comment at 0807567b538f39db5a6e9be9e87b94f9c269f6e0.

ashie (approved):
Thanks!

daipom:
Thanks for merging!

PR #4034, opened 2023-01-31
Improve handling of SIGCONT after SIGTERM

### LINKED ISSUE #3932: Provide option to ignore SIGCONT

When running under systemd, fluentd will produce a sigdump file even when a normal exit occurs.

Systemd sends a SIGCONT tot he process after the SIGTERM, which causes the sigdump file to be generated, putting unnecessary files in /tmp

An option in the fluentd config file to disable producing sigdump files

We have adjusted settings for our systemd unit file, but systemd always sends the sigcont.

### Additional context

_No response_

abetomo:

Ideas for #3932 fixes.

After SIGTERM, disable SIGCONT handler.
This will prevent sigdump files from being generated.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -190,6 +190,12 @@ module Fluent
         $log.debug 'fluentd supervisor process got SIGUSR2'
         supervisor_sigusr2_handler
       end
+
+      term_handler = trap :TERM do
+        # After SIGTERM, disable the sigdump file by ignoring SIGCONT.
+        trap(:CONT, nil)
+        term_handler.call
+      end
     end
 
     if Fluent.windows?
@@ -922,6 +928,9 @@ module Fluent
       end
 
       trap :TERM do
+        # After SIGTERM, disable the sigdump file by ignoring SIGCONT.
+        trap(:CONT, nil)
+
         $log.debug "fluentd main process get SIGTERM"
         unless @finished
           @finished = true
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -194,7 +194,7 @@ module Fluent
       term_handler = trap :TERM do
         # After SIGTERM, disable the sigdump file by ignoring SIGCONT.
         trap(:CONT, nil)
-        term_handler.call
+        term_handler.call if term_handler.is_a?(Proc)
       end
     end
 
--- a/test/test_supervisor.rb
+++ b/test/test_supervisor.rb
@@ -324,25 +324,21 @@ class SupervisorTest < ::Test::Unit::TestCase
     File.delete(@sigdump_path) if File.exist?(@sigdump_path)
   end
 
-  # TODO Sending SIGTERM terminates the test itself.
-  # I think we have to figure out a good way.
-  # def test_term_cont_in_supervisor_signal_handler
-  #   omit "Windows cannot handle signals" if Fluent.windows?
-  #
-  #   server = DummyServer.new
-  #   server.install_supervisor_signal_handlers
-  #   begin
-  #     Process.kill :TERM, $$
-  #     Process.kill :CONT, $$
-  #   rescue
-  #   end
-  #
-  #   sleep 1
-  #
-  #   assert_not File.exist?(@sigdump_path)
-  # ensure
-  #   File.delete(@sigdump_path) if File.exist?(@sigdump_path)
-  # end
+  def test_term_cont_in_supervisor_signal_handler
+    omit "Windows cannot handle signals" if Fluent.windows?
+
+    server = DummyServer.new
+    server.install_supervisor_signal_handlers
+    begin
+      Process.kill :TERM, $$
+      Process.kill :CONT, $$
+    rescue
+    end
+
+    assert_false File.exist?(@sigdump_path)
+  ensure
+    File.delete(@sigdump_path) if File.exist?(@sigdump_path)
+  end
 
   def test_windows_shutdown_event
     omit "Only for Windows platform" unless Fluent.windows?
--- a/test/test_supervisor.rb
+++ b/test/test_supervisor.rb
@@ -32,6 +32,7 @@ class SupervisorTest < ::Test::Unit::TestCase
     @tmp_dir = tmp_dir
     @tmp_root_dir = File.join(@tmp_dir, 'root')
     FileUtils.mkdir_p(@tmp_dir)
+    @sigdump_path = "/tmp/sigdump-#{$$}.log"
   end
 
   def teardown
@@ -191,7 +192,7 @@ class SupervisorTest < ::Test::Unit::TestCase
     end
   end
 
-  def test_main_process_signal_handlers
+  def test_usr1_in_main_process_signal_handlers
     omit "Windows cannot handle signals" if Fluent.windows?
 
     create_info_dummy_logger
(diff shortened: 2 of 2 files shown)


daipom (commented):
I am not familiar with SIGNAL but I think this fix looks nice!

It looks like this works well with ServerEngine.

* https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/server.rb#L102-L114
* https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/worker.rb#L63-L77

lib/fluent/supervisor.rb
```diff
@@ -190,6 +190,12 @@ def install_supervisor_signal_handlers
         $log.debug 'fluentd supervisor process got SIGUSR2'
         supervisor_sigusr2_handler
       end
+
+      term_handler = trap :TERM do
+        # After SIGTERM, disable the sigdump file by ignoring SIGCONT.
+        trap(:CONT, nil)
+        term_handler.call
```
daipom, line 197: This `term_handler.call()` will call `Server.stop()` in ServerEngine.
I think this is a nice way.

* https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/server.rb#L103

abetomo:
Thanks for the review.
I will add any missing tests, etc.

daipom (commented):
At this point, the followings are what I have noticed.

lib/fluent/supervisor.rb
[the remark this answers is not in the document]
abetomo: Actually, `term_handler` may be a string.
I will add a check to see if it is a `Proc`.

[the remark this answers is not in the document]
daipom: I see! This can be `String` or `Proc` or `nil according to the following document.

https://docs.ruby-lang.org/ja/latest/method/Signal/m/trap.html

> I will add a check to see if it is a Proc.

So this looks good!

kou:
How about just disabling dumping after stop instead of changing signal handlers?

```diff
diff --git a/lib/fluent/supervisor.rb b/lib/fluent/supervisor.rb
index b372f922..7dd5fc80 100644
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -355,6 +355,10 @@ module Fluent
       { conf: @fluentd_conf }
     end
 
+    def dump
+      super unless @stop
+    end
+
     private
 
     def reopen_log
@@ -413,6 +417,10 @@ module Fluent
     def after_start
       (config[:worker_pid] ||= {})[@worker_id] = @pm.pid
     end
+
+    def dump
+      super unless @stop
+    end
   end
 
   class Supervisor
@@ -754,12 +762,6 @@ module Fluent
     end
 
     def run_worker
-      begin
-        require 'sigdump/setup'
-      rescue Exception
-        # ignore LoadError and others (related with signals): it may raise these errors in Windows
-      end
-
       Process.setproctitle("worker:#{@system_config.process_name}") if @process_name
 
       if @standalone_worker && @system_config.workers != 1
@@ -940,6 +942,10 @@ module Fluent
         trap :USR2 do
           reload_config
         end
+
+        trap :CONT do
+          dump_non_windows
+        end
       end
     end
 
@@ -969,7 +975,7 @@ module Fluent
             reload_config
           when "DUMP"
             $log.debug "fluentd main process get #{cmd} command"
-            dump
+            dump_windows
           else
             $log.warn "fluentd main process get unknown command [#{cmd}]"
           end
@@ -1020,7 +1026,15 @@ module Fluent
       end
     end
 
-    def dump
+    def dump_non_windows
+      begin
+        Sigdump.dump unless @finished
+      rescue => e
+        $log.error("failed to dump: #{e}")
+      end
+    end
+
+    def dump_windows
       Thread.new do
         begin
           FluentSigdump.dump_windows
```

abetomo:
@kou Thanks. I will try it.

test/test_supervisor.rb
```diff
@@ -260,6 +307,43 @@ def test_supervisor_signal_handler
...
+    end
+
+    sleep 1
+
+    assert_true File.exist?(@sigdump_path)
+  ensure
+    File.delete(@sigdump_path) if File.exist?(@sigdump_path)
+  end
+
+  # TODO Sending SIGTERM terminates the test itself.
```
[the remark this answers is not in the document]
abetomo: You are right. I was mistaken.

abetomo:
@daipom What do you think about https://github.com/fluent/fluentd/pull/4034#issuecomment-1415431569?
I like this idea.
After stopping, it is easy to understand that sigdump is disabled.

ashie:
I prefer @kou's idea, too.

daipom:
> @daipom What do you think about [#4034 (comment)](https://github.com/fluent/fluentd/pull/4034#issuecomment-1415431569)? I like this idea. After stopping, it is easy to understand that sigdump is disabled.

I think so too!

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
 
+  def test_cont_in_main_process_signal_handlers
+    omit "Windows cannot handle signals" if Fluent.windows?
+
+    opts = Fluent::Supervisor.default_options
+    sv = Fluent::Supervisor.new(opts)
+    sv.send(:install_main_process_signal_handlers)
+
+    begin
+      Process.kill :CONT, $$
```
kou, line 225: `Process.pid` is preferred than `$$`  for readability.

```suggestion
      Process.kill :CONT, Process.pid
```

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+    rescue
+    end
```
kou, line 227: Can we remove this `begin ... rescue ... end`?

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+
+    sleep 1
+
+    assert_true File.exist?(@sigdump_path)
```
kou, line 231: Power assert style is preferred for better failure message:

```suggestion
    assert do
      File.exist?(@sigdump_path)
    end
```

`assert_true ...` shows just `File.exist?(@sigdump_path)` is not `true` on failure.
But `assert {...}` shows not only `File.exist?(@sigdump_path)` value but also `@sigdump_path` and so on.

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+
+    begin
+      Process.kill :TERM, $$
+      Process.kill :CONT, $$
+    rescue
+    end
+
+    sleep 1
+
+    debug_msg = '[debug]: fluentd main process get SIGTERM' + "\n"
```
kou, line 253:
```suggestion
    debug_msg = "[debug]: fluentd main process get SIGTERM\n"
```

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+
+    opts = Fluent::Supervisor.default_options
+    sv = Fluent::Supervisor.new(opts)
+    sv.send(:install_main_process_signal_handlers)
+
+    begin
+      Process.kill :TERM, $$
+      Process.kill :CONT, $$
+    rescue
+    end
```
kou, line 249: Can we remove this `begin ... rescue ... end`?

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+
+    sleep 1
+
+    debug_msg = '[debug]: fluentd main process get SIGTERM' + "\n"
+    logs = $log.out.logs
+    assert{ logs.any?{|log| log.include?(debug_msg) } }
```
kou, line 255:
```suggestion
    assert{ logs.any?{|log| log.include?(debug_msg) } }
```

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+
+    assert_false File.exist?(@sigdump_path)
```
kou, line 257:
```suggestion
    assert do
      not File.exist?(@sigdump_path)
    end
```

test/test_supervisor.rb
```diff
@@ -213,6 +214,52 @@ def test_main_process_signal_handlers
...
+  ensure
+    $log.out.reset if $log && $log.out && $log.out.respond_to?(:reset)
```
kou, line 259:
```suggestion
    $log.out.reset if $log&.out&.respond_to?(:reset)
```

test/test_supervisor.rb
abetomo: It seems to be the same code.
I would like to know if there is a better way.

kou: Ah, sorry. This is garbage.

abetomo:
Thank you for your comments and review.

I'll fix it with @kou's idea.
Close this PR and open another PR with a suitable title.

#4043

PR #4043, opened 2023-02-07
Disable sigdump after stop

### LINKED ISSUE #3932: Provide option to ignore SIGCONT

abetomo:
Fixes #3932

Don't generate sigdump files after stopping.
https://github.com/fluent/fluentd/pull/4034#issuecomment-1415431569

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -355,6 +355,10 @@ module Fluent
       { conf: @fluentd_conf }
     end
 
+    def dump
+      super unless @stop
+    end
+
     private
 
     def reopen_log
@@ -412,6 +416,10 @@ module Fluent
 
     def after_start
       (config[:worker_pid] ||= {})[@worker_id] = @pm.pid
+    end
+
+    def dump
+      super unless @stop
     end
   end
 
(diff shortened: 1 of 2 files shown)


kou (approved):
+1

ashie (approved):
LGTM

daipom (approved):
LGTM. Thanks for the fix!

I have checked that Fluentd as a Windows Service can dump correctly by `$ fluent-ctl dump`.

What about `Docs Changes` and `Release Note`?

`Docs Changes` is the fix of https://github.com/fluent/fluentd-docs-gitbook corresponding to this fix.
I think we don't need it this time.

`Release Note` is used for https://github.com/fluent/fluentd/blob/master/CHANGELOG.md.
This is often the same as the title of the PR, but I think it is worth considering a little more this time.

ashie:
Merged. Thanks!

> Docs Changes is the fix of https://github.com/fluent/fluentd-docs-gitbook corresponding to this fix.
> I think we don't need it this time.

I also think it's not needed for this.

abetomo:
> Release Note:

How about "Change to not generate a sigdump file after receiving a SIGTERM signal on non-Windows".

daipom:
> > Release Note:
> 
> How about "Change to not generate a sigdump file after receiving a SIGTERM signal on non-Windows"?

Great!

PR #4050, opened 2023-02-10
Clean up after unit tests

### LINKED ISSUE #4048: Files used in unit tests remain in `/tmp`
### Describe the bug

After `rake test`, the following files remained in `/tmp`.

* `/tmp/SERVERENGINE_SOCKETMANAGER_*`
* `/tmp/fluentd-lock-*`
* `/tmp/fluent/`

### To Reproduce

1. `rake test`
2. `ls /tmp`

### Expected behavior

Files created by tests should be deleted.

### Your Environment

```markdown
- Fluentd version: 1.15.3
- Operating system: Ubuntu 22.04
- Kernel version: 5.15.0-58-generic
```

### Your Configuration

```apache
-
```

### Your Error Log

```shell
-
```

### Additional context

_No response_

abetomo:

Fixes #4048

After unit testing, clean up in /tmp.

(If you cannot terminate with SIGTERM and terminate with SIGKILL, some files cannot be cleaned up. https://github.com/fluent/fluentd/pull/4050#issuecomment-1427932016 https://github.com/fluent/fluentd/pull/4050#issuecomment-1430665345)

**Docs Changes**:

None

**Release Note**: 

None

### DIFF
--- a/lib/fluent/test/driver/base.rb
+++ b/lib/fluent/test/driver/base.rb
@@ -173,7 +173,7 @@ module Fluent
 
           if @socket_manager_server
             @socket_manager_server.close
-            if @socket_manager_server.is_a?(String) && File.exist?(@socket_manager_path)
+            if @socket_manager_path.is_a?(String) && File.exist?(@socket_manager_path)
               FileUtils.rm_f @socket_manager_path
             end
           end
--- a/test/plugin/test_in_tcp.rb
+++ b/test/plugin/test_in_tcp.rb
@@ -226,8 +226,8 @@ class TcpInputTest < Test::Unit::TestCase
 
       assert_equal 1, d.instance.log.logs.count { |l| l =~ /anonymous client/ }
       assert_equal 0, d.events.size
-
-      d.instance_shutdown
+    ensure
+      d.instance_shutdown if d&.instance
     end
   end
 
--- a/test/plugin/test_in_tcp.rb
+++ b/test/plugin/test_in_tcp.rb
@@ -226,6 +226,8 @@ class TcpInputTest < Test::Unit::TestCase
 
       assert_equal 1, d.instance.log.logs.count { |l| l =~ /anonymous client/ }
       assert_equal 0, d.events.size
+
+      d.instance_shutdown
     end
   end
 
--- a/test/plugin/test_output.rb
+++ b/test/plugin/test_output.rb
@@ -803,7 +803,10 @@ class OutputTest < Test::Unit::TestCase
     end
 
     test 'output plugin will call #try_write for plugin supports delayed commit only to flush buffer chunks' do
+      tmp_dir = File.join(__dir__, '../tmp/test_output')
+
       i = create_output(:delayed)
+      i.system_config_override(root_dir: tmp_dir) # Backup files are generated in `tmp_dir`.
       try_write_called = false
       i.register(:try_write){|chunk| try_write_called = true; commit_write(chunk.unique_id) }
 
@@ -820,6 +823,8 @@ class OutputTest < Test::Unit::TestCase
       assert try_write_called
 
       i.stop; i.before_shutdown; i.shutdown; i.after_shutdown; i.close; i.terminate
+    ensure
+      FileUtils.rm_rf(tmp_dir)
     end
 
     test '#prefer_delayed_commit (returns false) decides delayed commit is disabled if both are implemented' do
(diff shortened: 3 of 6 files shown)


lib/fluent/test/driver/base.rb
```diff
@@ -173,7 +173,7 @@ def instance_shutdown(log: Logger.new($stdout))
 
           if @socket_manager_server
             @socket_manager_server.close
-            if @socket_manager_server.is_a?(String) && File.exist?(@socket_manager_path)
+            if @socket_manager_path.is_a?(String) && File.exist?(@socket_manager_path)
```
daipom, line 176: Wow, thanks! This is a typo!

test/plugin/test_in_tcp.rb
```diff
@@ -226,6 +226,8 @@ def create_tcp_socket(host, port, &block)
 
       assert_equal 1, d.instance.log.logs.count { |l| l =~ /anonymous client/ }
       assert_equal 0, d.events.size
+
+      d.instance_shutdown
```
daipom, line 230: Indeed. Need it since the driver ran with `shutdown: false`.

ashie: I think we shouldn't use `shutdown: false` as possible although I'm not sure whether it's needed here or not.
FYI sometimes I found some needless `shutdown: false` in other tests (probably copy&pasted).

daipom: When terminating a plugin on shutdown, the `reset` of the logger is called and it erases the logs.

So this test sets `shutdown: false`.

But somehow the data of `driver.logs` is still alive after terminating...

daipom: So I think using `driver.logs` is the correct way, but I need more time to check how it works.

test/plugin/test_output.rb
```diff
@@ -820,6 +823,7 @@ def waiting(seconds)
       assert try_write_called
 
       i.stop; i.before_shutdown; i.shutdown; i.after_shutdown; i.close; i.terminate
+      FileUtils.rm_rf(tmp_dir)
```
ashie, line 826: It should be done in `ensure` block or `teardown` to make sure to delete it even if assertions fails.

test/plugin/test_in_tcp.rb
abetomo: Is it better to fix it in this PR?
Or should I fix it in another PR?

abetomo:
With the committed changes and the following changes, all files were cleaned.

```diff
diff --git a/test/command/test_fluentd.rb b/test/command/test_fluentd.rb
index 461e19ef..9d62cd40 100644
--- a/test/command/test_fluentd.rb
+++ b/test/command/test_fluentd.rb
@@ -85,12 +85,12 @@ class TestFluentdCommand < ::Test::Unit::TestCase
         yield pid, io
         # p(here: "execute command", pid: pid, worker_pids: @worker_pids)
       ensure
-        Process.kill(:KILL, pid) rescue nil
+        Process.kill(:TERM, pid) rescue nil
         if @supervisor_pid
-          Process.kill(:KILL, @supervisor_pid) rescue nil
+          Process.kill(:TERM, @supervisor_pid) rescue nil
         end
         @worker_pids.each do |cpid|
-          Process.kill(:KILL, cpid) rescue nil
+          Process.kill(:TERM, cpid) rescue nil
         end
         # p(here: "execute command", pid: pid, exist: process_exist?(pid), worker_pids: @worker_pids, exists: @worker_pids.map{|i| process_exist?(i) })
         Timeout.timeout(10){ sleep 0.1 while process_exist?(pid) }
```

When you exit with SIGTERM, the shutdown process runs and the files are deleted.
However, the test was not stable because the process could not terminate in some cases.

test/plugin/test_in_tcp.rb
ashie: It's fine as is (`shutdown: false`).
Instead it should be run in ensure block to make sure to stop:
```suggestion
  ensure
      d.instance_shutdown if d && d.instance
```

ashie:
How about trying `SIGTERM` first then `SIGKILL` after timeout?
Although we should make it stable but not needed to do it in this PR.

abetomo:
@ashie  Thanks!
I’ll give it a try.

ashie (approved):
LGTM

ashie:
Retrying some unstable tests not concerned with this PR...

daipom (approved):
LGTM. Thanks for the fix!

ashie:
Thanks!

abetomo:
Thanks for your review!

test/plugin/test_in_tcp.rb
daipom: Now I understand this.
I made an issue.

* #4136

daipom: I made a PR

* #4138

Issue #4037, opened 2023-02-03
Logger: Log format setting is not applied to the first log


daipom:
### Describe the bug

I noticed this issue when I was checking #4011.

The log of initializing logger is added in Fluentd v1.15.3 at #3939:
Fluentd outputs this log first.

```
2023-02-03 19:24:28 +0900 [info]: fluent/log.rb:330:info: init supervisor logger path=nil rotate_age=nil rotate_size=nil
```

However, the [log format](https://docs.fluentd.org/deployment/logging#log-format) is not applied to this log.

* https://github.com/fluent/fluentd/blob/v1.15.3/lib/fluent/supervisor.rb#L451-L452
* https://github.com/fluent/fluentd/blob/v1.15.3/lib/fluent/supervisor.rb#L586

At the point of `LoggerInitializer::init()`, the format is not applied yet, so the logger must not output logs.

### To Reproduce

Run Fluentd v1.15.3 by the following config and see the first log.

```console
$ bundle exec fluentd -c /path/to/fluent.conf
```

### Expected behavior

[Log format setting](https://docs.fluentd.org/deployment/logging#log-format) is applied to the all logs of Fluentd.

### Your Environment

- Fluentd version: v1.15.3
- Operating system:
  ```console
  $ cat /etc/os-release
  NAME="Ubuntu"
  VERSION="20.04.5 LTS (Focal Fossa)"
  ID=ubuntu
  ID_LIKE=debian
  PRETTY_NAME="Ubuntu 20.04.5 LTS"
  VERSION_ID="20.04"
  HOME_URL="https://www.ubuntu.com/"
  SUPPORT_URL="https://help.ubuntu.com/"
  BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
  PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
  VERSION_CODENAME=focal
  UBUNTU_CODENAME=focal
  ```
- Kernel version:
  ```console
  $ uname -r
  5.15.0-58-generic
  ```

### Your Configuration

```apache
<system>
  <log>
    format json
    time_format %Y-%m-%dT%H:%M:%S%z
  </log>
</system>

<source>
  @type sample
  tag test
</source>

<match test.**>
  @type stdout
</match>
```

### Your Error Log

```shell
[...]

PR #2663, opened 2019-10-23
Support logrotate on Windows correctly. Fix #2446

cosmo0920:
Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>

**Which issue(s) this PR fixes**: 
Fixes #2446

**What this PR does / why we need it**: 

In Windows multi-process environment, we cannot rename a file which is opened by another process.

Then, we should do this:

* Separate log file for each process on Windows

**Docs Changes**:

I'll add a note in https://docs.fluentd.org/deployment/logging#log-rotation-setting.

**Release Note**: 

Same as title.

PR #4054, opened 2023-02-13
Windows: Fix a bug that the wrong log file is reopened with log ratote setting when flushing or graceful reloading

daipom:
When log rotation is enabled on Windows, the log path is separated for each process.
But, the new path value is not applied to the instance variable, so it breaks the following behavior, since #2663.

https://github.com/fluent/fluentd/blob/bf7498d6ff2805bfa517bbd6f7c17103c7d879d4/lib/fluent/supervisor.rb#L576-L580

https://github.com/fluent/fluentd/blob/bf7498d6ff2805bfa517bbd6f7c17103c7d879d4/lib/fluent/supervisor.rb#L601-L606

I can't assume that `File.chown` is used on Windows, but `reopen` must be fixed.
(`--user` and `--group` are not intended to be used on Windows, right? If so, I will add some comments to [the doc](https://docs.fluentd.org/deployment/command-line-option) and code.)

* #4037

I thought this should be fixed in a separate PR, so I made this PR first.

This bug breaks the function of reopening the log file with log rotation enabled on Windows, since

* #2663

We need this fix to fix this problem.

**How to Reproduce**

* Prepare config
```xml
<system>
  <log>
    rotate_age 5
    rotate_size 2000
  </log>
</system>

 <source>
   @type sample
   tag test
 </source>
 
 <match test.**>
   @type stdout
 </match>
```

* Run Fluentd
```console
$ bundle exec fluentd -c /test/fluent.conf -o /test/log/fluent.log
```

* We can find the log files
```
fluent-supervisor-0.log
fluent-0.log
fluent-0.log.0
...
```

* Flush([SIGUSR1](https://docs.fluentd.org/deployment/signals#sigusr1))
  * We can check the pid of the supervisor process in `fluent-supervisor-0.log`.
  * `starting fluentd-1.15.3 pid=4208 ruby="2.6.6"`
```console
$ fluent-ctl flush {pid of supervisor process}
```

* Then Fluentd reopens `fluent.log` for the worker process, although this should be `fluent-0.log`.

We can find a similar problem in graceful reloading([SIGUSR2](https://docs.fluentd.org/deployment/signals#sigusr2)), but this often causes an error in rotating supervisor log file.
This seems to have another bug, I would like to exclude this bug for now.

```console
$ fluent-ctl reload {pid of supervisor process} (another shell)
log shifting failed. Permission denied @ rb_file_s_rename - (.\fluent\log\fluent-supervisor-0.log, .\fluent\log\fluent-supervisor-0.log.0)
log writing failed. closed stream
```

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -563,12 +563,8 @@ module Fluent
           if @log_rotate_age || @log_rotate_size
             # We need to prepare a unique path for each worker since
             # Windows locks files.
-            if Fluent.windows?
-                path = LoggerInitializer.per_process_path(@path, process_type, worker_id)
-            else
-                path = @path
-            end
-            @logdev = Fluent::LogDeviceIO.new(path, shift_age: @log_rotate_age, shift_size: @log_rotate_size)
+            @path = LoggerInitializer.per_process_path(@path, process_type, worker_id) if Fluent.windows?
+            @logdev = Fluent::LogDeviceIO.new(@path, shift_age: @log_rotate_age, shift_size: @log_rotate_size)
           else
             @logdev = File.open(@path, "a")
           end
@@ -591,7 +587,7 @@ module Fluent
         $log = Fluent::Log.new(logger, @opts)
         $log.enable_color(false) if @path
         $log.enable_debug if @level <= Fluent::Log::LEVEL_DEBUG
-        $log.info "init #{process_type} logger", path: path, rotate_age: @log_rotate_age, rotate_size: @log_rotate_size
+        $log.info "init #{process_type} logger", path: @path, rotate_age: @log_rotate_age, rotate_size: @log_rotate_size
       end
 
       def stdout?
--- a/test/test_logger_initializer.rb
+++ b/test/test_logger_initializer.rb
@@ -7,13 +7,13 @@ class LoggerInitializerTest < ::Test::Unit::TestCase
   def setup
     @stored_global_logger = $log
     Dir.mktmpdir do |tmp_dir|
-      begin
-        @tmp_dir = Pathname(tmp_dir)
-        yield
-      ensure
-        $log = @stored_global_logger
-      end
+      @tmp_dir = Pathname(tmp_dir)
+      yield
     end
+  end
+
+  def teardown
+    $log = @stored_global_logger
   end
 
   test 'when path is given' do
--- a/test/test_logger_initializer.rb
+++ b/test/test_logger_initializer.rb
@@ -22,6 +22,7 @@ class LoggerInitializerTest < ::Test::Unit::TestCase
     assert_nothing_raised do
       logger.init(:supervisor, 0)
     end
+    $log.out.close
 
     assert_true File.exist?(TMP_DIR)
   end
@@ -38,8 +39,9 @@ class LoggerInitializerTest < ::Test::Unit::TestCase
     assert_nothing_raised do
       logger.init(:supervisor, 0)
     end
-
     logger.apply_options(log_dir_perm: 0o777)
+    $log.out.close
+
     assert_true File.exist?(TMP_DIR)
     assert_equal 0o777, (File.stat(TMP_DIR).mode & 0xFFF)
   end
--- a/test/test_logger_initializer.rb
+++ b/test/test_logger_initializer.rb
@@ -1,42 +1,40 @@
 require_relative 'helper'
 require 'fluent/supervisor'
 require 'fileutils'
+require 'pathname'
 
 class LoggerInitializerTest < ::Test::Unit::TestCase
-  TMP_DIR = File.expand_path(File.dirname(__FILE__) + "/tmp/logger_initializer#{ENV['TEST_ENV_NUMBER']}")
-
-  setup do
-    FileUtils.rm_rf(TMP_DIR) rescue nil
+  def setup
     @stored_global_logger = $log
-  end
-
-  teardown do
-    $log = @stored_global_logger
+    Dir.mktmpdir do |tmp_dir|
+      begin
+        @tmp_dir = Pathname(tmp_dir)
+        yield
+      ensure
+        $log = @stored_global_logger
+      end
+    end
   end
 
   test 'when path is given' do
-    assert { not File.exist?(TMP_DIR) }
-
-    path = File.join(TMP_DIR, 'fluent_with_path.log')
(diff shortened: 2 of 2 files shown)


daipom:
Oh, I have to add tests.

lib/fluent/supervisor.rb
```diff
@@ -563,12 +563,8 @@ def init(process_type, worker_id)
           if @log_rotate_age || @log_rotate_size
             # We need to prepare a unique path for each worker since
             # Windows locks files.
-            if Fluent.windows?
-                path = LoggerInitializer.per_process_path(@path, process_type, worker_id)
-            else
-                path = @path
-            end
-            @logdev = Fluent::LogDeviceIO.new(path, shift_age: @log_rotate_age, shift_size: @log_rotate_size)
+            @path = LoggerInitializer.per_process_path(@path, process_type, worker_id) if Fluent.windows?
```
ashie, line 566: `@path` should be set at the head of this method.

lib/fluent/supervisor.rb
```diff
@@ -591,7 +587,7 @@ def init(process_type, worker_id)
         $log = Fluent::Log.new(logger, @opts)
         $log.enable_color(false) if @path
         $log.enable_debug if @level <= Fluent::Log::LEVEL_DEBUG
-        $log.info "init #{process_type} logger", path: path, rotate_age: @log_rotate_age, rotate_size: @log_rotate_size
+        $log.info "init #{process_type} logger", path: @path, rotate_age: @log_rotate_age, rotate_size: @log_rotate_size
```
ashie, line 590: Memo: I've confirmed that it always outputs `path=nil` even if `--log=/path/to/log` is set:
```
2023-02-13 18:38:40 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
```

ashie (approved):
Good catch!

ashie:
I'll wait merging this until you add tests for it.

lib/fluent/supervisor.rb
daipom: Yes, can I fix it in the next PR that I'm planning to fix #4037?

I plan to fix it so that the path is passed from outside `init()`, in order to avoid logging inside the `LoggerInitializer`.

lib/fluent/supervisor.rb
daipom: Really? I don't understand it. I will check it.

daipom:
> I'll wait merging this until you add tests for it.

Thanks. Please.

lib/fluent/supervisor.rb
ashie: Ah, sorry. I was talking about the case without this fix.

daipom: I see! When we don't enable log rotate, this always outputs `path=nil`.

daipom: This PR fixes this problem too.

lib/fluent/supervisor.rb
ashie: I see, please go ahead.

daipom:
I added tests.

I confirmed one of those tests failed on master on Windows and this PR fixed it.

test/test_logger_initializer.rb
```diff
@@ -5,11 +5,8 @@
 class LoggerInitializerTest < ::Test::Unit::TestCase
   TMP_DIR = File.expand_path(File.dirname(__FILE__) + "/tmp/logger_initializer#{ENV['TEST_ENV_NUMBER']}")
 
-  teardown do
-    begin
-      FileUtils.rm_rf(TMP_DIR)
-    rescue => _
-    end
+  setup do
+    FileUtils.rm_rf(TMP_DIR) rescue nil
```
ashie, line 9: > to allow us to confirm the data after executing a test.

Is it really needed?
Generally speaking tmp files shouldn't exist after test so that it should be removed at `teardown`.
(It's OK to do in also `setup` to make sure.)

daipom: Oh, I see.
I thought that tests should not leave tmp data outside the working directory, but can leave them inside.

Indeed, many other tests delete tmp files at `teardown`.

daipom: > I thought that tests should not leave tmp data outside the working directory, but can leave them inside.

I found out that what I was thinking was wrong! I learned!

I will fix `setup`! Thanks!

daipom: I have fixed this to make sure not to leave tmp files.

daipom:
I wonder why using `mock.proxy`, not just `mock`.

* https://github.com/fluent/fluentd/pull/4054/files#diff-eddebe316e0a90606efd872b0fcae79fa441e6edba5b3bba0571bef9410368d8R37

When I change as follows, the test fails. I don't understand what the difference is.
What does the proxy being not passed a block mean?

```diff
- mock.proxy(File).chmod(0o777, path.parent.to_s).once
+ mock(File).chmod(0o777, path.parent.to_s).once
```

ashie:
Probably it intends to check `File.chmod` is called just once.
And need to call real `File.chmod` for following checks.

daipom:
> need to call real `File.chmod` for following checks.

Oh, thanks! Just simple `mock(File).chmod` destroys the original behavior of `File.chmod`, so we have to use `proxy` here.

test/test_logger_initializer.rb
```diff
@@ -1,46 +1,120 @@
...
-      FileUtils.rm_rf(TMP_DIR)
-    rescue => _
+  def setup
+    @stored_global_logger = $log
+    Dir.mktmpdir do |tmp_dir|
+      begin
+        @tmp_dir = Pathname(tmp_dir)
+        yield
+      ensure
+        $log = @stored_global_logger
```
ashie, line 14: I prefer to place it outside of `Dir.mktmpdir` or in `teardown` to correspond nesting level of replacing `$log`.
(Although probably it's harmless since `Dir.mktmpdir` rarely fail.)

daipom: Indeed! I will fix it.

daipom: Fixed! I placed it in `teardown`. Thanks!

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4064, opened 2023-02-19
Fix problem that some system configs are not reflected

daipom:

* #4037

I thought this should be fixed in a separate PR, so I made this PR first.

Options explicitly specified in command line should be preferred to options specified in system config.

* https://github.com/fluent/fluentd/pull/1398
* https://github.com/fluent/fluentd/pull/2674

However, options that have a default value, such as `suppress_repeated_stacktrace`, are preferred even when they are not specified in the command line.

Because of this, the following setting is not reflected.

```xml
<system>
  suppress_repeated_stacktrace false
</system>
```

> Options explicitly specified in command line should be preferred to options specified in system config.

In order to realize this specification, we should pass only options that are explicitly specified in the command line to `Superviser.new()`.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -158,15 +158,15 @@ op.on('--enable-size-metrics', "Enable plugin record size metrics on fluentd", T
 }
 
 op.on('-v', '--verbose', "increase verbose level (-v: debug, -vv: trace)", TrueClass) {|b|
-  if b
-    cmd_opts[:log_level] = [default_opts[:log_level] - 1, Fluent::Log::LEVEL_TRACE].max
-  end
+  return unless b
+  cur_level = cmd_opts.fetch(:log_level, default_opts[:log_level])
+  cmd_opts[:log_level] = [cur_level - 1, Fluent::Log::LEVEL_TRACE].max
 }
 
 op.on('-q', '--quiet', "decrease verbose level (-q: warn, -qq: error)", TrueClass) {|b|
-  if b
-    cmd_opts[:log_level] = [default_opts[:log_level] + 1, Fluent::Log::LEVEL_ERROR].min
-  end
+  return unless b
+  cur_level = cmd_opts.fetch(:log_level, default_opts[:log_level])
+  cmd_opts[:log_level] = [cur_level + 1, Fluent::Log::LEVEL_TRACE].max
 }
 
 op.on('--suppress-config-dump', "suppress config dumping when fluentd starts", TrueClass) {|b|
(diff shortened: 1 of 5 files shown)


lib/fluent/command/fluentd.rb
```diff
@@ -26,205 +26,206 @@
...
+    cmd_opts[:log_level] = [default_opts[:log_level] - 1, Fluent::Log::LEVEL_TRACE].max
   end
 }
 
 op.on('-q', '--quiet', "decrease verbose level (-q: warn, -qq: error)", TrueClass) {|b|
   if b
-    opts[:log_level] = [opts[:log_level] + 1, Fluent::Log::LEVEL_ERROR].min
+    cmd_opts[:log_level] = [default_opts[:log_level] + 1, Fluent::Log::LEVEL_ERROR].min
   end
 }
```
daipom, line 171: It seems possible that I'm wrong here.
I will check this. Maybe we don't have the corresponding tests.

daipom:
https://github.com/fluent/fluentd/pull/4064/files#diff-6042a07157b33b2d729ab3d10ed6bba89ad5f71aaed2be1aeb22a179a45ebbaeR1107-R1111

We can discard this logic now, and we should have the corresponding test here too,

lib/fluent/command/fluentd.rb
daipom: Fixed.

daipom:
> https://github.com/fluent/fluentd/pull/4064/files#diff-6042a07157b33b2d729ab3d10ed6bba89ad5f71aaed2be1aeb22a179a45ebbaeR1107-R1111
> 
> We can discard this logic now, and we should have the corresponding test here too,

Fixed.

ashie (approved):
LGTM

ashie:
Sorry for the delay, I missed the additional fixes...

daipom:
Thanks for your review!

lib/fluent/command/fluentd.rb
```diff
@@ -26,205 +26,206 @@
...
+  cmd_opts[:log_level] = [cur_level - 1, Fluent::Log::LEVEL_TRACE].max
 }
 
 op.on('-q', '--quiet', "decrease verbose level (-q: warn, -qq: error)", TrueClass) {|b|
-  if b
-    opts[:log_level] = [opts[:log_level] + 1, Fluent::Log::LEVEL_ERROR].min
-  end
+  return unless b
+  cur_level = cmd_opts.fetch(:log_level, default_opts[:log_level])
+  cmd_opts[:log_level] = [cur_level + 1, Fluent::Log::LEVEL_TRACE].max
```
ashie, line 169: Oops, I overlooked it...
#4090

daipom: Sorry for my simple mistake.

PR #4065, opened 2023-02-19
Refactor logger initialization

daipom:

* https://github.com/fluent/fluentd/issues/4037#issuecomment-1427366287

This fix should be done before fixing this issue.

In the supervisor process, Fluentd initializes the logger twice, at `Supervisor::configure()` and `Supervisor.load_config()`.

`Supervisor.load_config()` is called in ServerEngine's reloading function, but Fluentd doesn't use the function even when `SIGHUP` or `SIGUSR2`.
So I can't see the reason for initializing the logger in that callback.

This also fixes a problem that `ignore_same_log_interval` is not passed to the `Supervisor.load_config()` and that option is not reflected to the logger of the supervisor process.

**Docs Changes**:
Not needed.

**Release Note**: 
* Fix bug that `ignore_same_log_interval` option is not reflected in the supervisor process.
* Fix bug that `suppress_repeated_stacktrace` from SystemConfig is not reflected in the worker processes.

We should consider the release note with

* #4064

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -436,14 +436,6 @@ module Fluent
         return params['pre_conf']
       end
 
-      log_level = params['log_level']
-
-      chuser = params['chuser']
-      chgroup = params['chgroup']
-      chumask = params['chumask']
-
-      command_sender = Fluent.windows? ? "pipe" : "signal"
-
       # ServerEngine's "daemonize" option is boolean, and path of pid file is brought by "pid_path"
       pid_path = params['daemonize']
       daemonize = !!params['daemonize']
@@ -461,11 +453,11 @@ module Fluent
         root_dir: params['root_dir'],
         logger: $log,
         log: $log.out,
-        log_level: log_level,
+        log_level: params['log_level'],
         logger_initializer: params['logger_initializer'],
-        chuser: chuser,
-        chgroup: chgroup,
-        chumask: chumask,
+        chuser: params['chuser'],
+        chgroup: params['chgroup'],
+        chumask: params['chumask'],
         daemonize: daemonize,
         rpc_endpoint: params['rpc_endpoint'],
         counter_server: params['counter_server'],
@@ -476,7 +468,7 @@ module Fluent
                                  WorkerModule.name,
                                  path,
                                  JSON.dump(params)],
-        command_sender: command_sender,
+        command_sender: Fluent.windows? ? "pipe" : "signal",
         fluentd_conf: params['fluentd_conf'],
         conf_encoding: params['conf_encoding'],
         inline_config: params['inline_config'],
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -437,28 +437,10 @@ module Fluent
       end
 
       log_level = params['log_level']
-      suppress_repeated_stacktrace = params['suppress_repeated_stacktrace']
-      ignore_repeated_log_interval = params['ignore_repeated_log_interval']
-      ignore_same_log_interval = params['ignore_same_log_interval']
 
-      log_path = params['log_path']
       chuser = params['chuser']
       chgroup = params['chgroup']
       chumask = params['chumask']
-      log_rotate_age = params['log_rotate_age']
-      log_rotate_size = params['log_rotate_size']
-
-      log_opts = {suppress_repeated_stacktrace: suppress_repeated_stacktrace, ignore_repeated_log_interval: ignore_repeated_log_interval,
-                  ignore_same_log_interval: ignore_same_log_interval}
-      logger_initializer = Supervisor::LoggerInitializer.new(
-        log_path, log_level, chuser, chgroup, log_opts,
-        log_rotate_age: log_rotate_age,
-        log_rotate_size: log_rotate_size
-      )
-      # this #init sets initialized logger to $log
-      logger_initializer.init(:supervisor, 0)
-      logger_initializer.apply_options(format: params['log_format'], time_format: params['log_time_format'])
-      logger = $log
 
       command_sender = Fluent.windows? ? "pipe" : "signal"
 
@@ -477,17 +459,13 @@ module Fluent
         unrecoverable_exit_codes: [2],
         stop_immediately_at_unrecoverable_exit: true,
         root_dir: params['root_dir'],
-        logger: logger,
-        log: logger.out,
-        log_path: log_path,
+        logger: $log,
+        log: $log.out,
         log_level: log_level,
-        logger_initializer: logger_initializer,
+        logger_initializer: params['logger_initializer'],
         chuser: chuser,
         chgroup: chgroup,
         chumask: chumask,
-        suppress_repeated_stacktrace: suppress_repeated_stacktrace,
-        ignore_repeated_log_interval: ignore_repeated_log_interval,
-        ignore_same_log_interval: ignore_same_log_interval,
         daemonize: daemonize,
         rpc_endpoint: params['rpc_endpoint'],
         counter_server: params['counter_server'],
@@ -601,11 +579,13 @@ module Fluent
         self
       end
 
-      def apply_options(format: nil, time_format: nil, log_dir_perm: nil, ignore_repeated_log_interval: nil, ignore_same_log_interval: nil)
+      def apply_options(format: nil, time_format: nil, log_dir_perm: nil,
+                        ignore_repeated_log_interval: nil, ignore_same_log_interval: nil, suppress_repeated_stacktrace: nil)
         $log.format = format if format
         $log.time_format = time_format if time_format
         $log.ignore_repeated_log_interval = ignore_repeated_log_interval if ignore_repeated_log_interval
         $log.ignore_same_log_interval = ignore_same_log_interval if ignore_same_log_interval
+        $log.suppress_repeated_stacktrace = suppress_repeated_stacktrace unless suppress_repeated_stacktrace.nil?
 
         if @path && log_dir_perm
           File.chmod(log_dir_perm || Fluent::DEFAULT_DIR_PERMISSION, File.dirname(@path))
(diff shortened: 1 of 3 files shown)


daipom:
I rebased this to apply #4064.

Please tell me if we need to investigate more about ServerEngine's reloading function.
If so, I will summarize Fluentd and  ServerEngine's behavior related to this.

daipom:
I am concerned that the `Ruby 2.7 on macos-latest` test fails.

https://github.com/fluent/fluentd/actions/runs/4320183945/jobs/7540167787

```
 Failure: test_unwatched_files_should_be_removed(TailInputTest::path)
/Users/runner/work/fluentd/fluentd/test/plugin/test_in_tail.rb:1616:in `test_unwatched_files_should_be_removed'
     1613:       waiting(20) { sleep 0.1 until Dir.glob("#{@tmp_dir}/*.txt").size == 0 } # Ensure file is deleted on Windows
     1614:       waiting(5) { sleep 0.1 until d.instance.instance_variable_get(:@tails).keys.size <= 0 }
     1615: 
  => 1616:       assert_equal(
     1617:         {
     1618:           files: [],
     1619:           tails: []
<{:files=>[], :tails=>[]}> expected but was
<{:files=>[],
 :tails=>
  ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/dcebe801d2a39d7ba0fc/tail.txt"]}>

diff:
? {:files=>[], :tails=>[]}
+  :tails=>
+   ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/dcebe801d2a39d7ba0fc/tail.txt"]}
```

ashie:
It's a known unstable test.
We should make it stable but not related with this change.

daipom:
> It's a known unstable test. We should make it stable but not related with this change.

Thanks, I have created the issue:

* #4079

ashie:
> This also fixes a problem that `ignore_same_log_interval` is not passed to the `Supervisor.load_config()` and that option is not reflected to the logger of the supervisor process.

You mean `suppress_repeated_stacktrace`?
According to your commit, the addressed parameter seems `suppress_repeated_stacktrace`, not `ignore_same_log_interval`.

ashie:
> `Supervisor.load_config()` is called in ServerEngine's reloading function, but Fluentd doesn't use the function even when `SIGHUP` or `SIGUSR2`. So I can't see the reason for initializing the logger in that callback.

Sorry, I can't remember the code around it soon.
Could you point out the rationale?

daipom:
> > This also fixes a problem that `ignore_same_log_interval` is not passed to the `Supervisor.load_config()` and that option is not reflected to the logger of the supervisor process.
> 
> You mean `suppress_repeated_stacktrace`? According to your commit, the addressed parameter seems `suppress_repeated_stacktrace`, not `ignore_same_log_interval`.

No. Originally, `ignore_same_log_interval` is not passed here.

https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L870-L896

So `Fluent::Log::ignore_same_log_interval` is always `nil` after `Supervisor.load_config()`

* https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L442
* https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L451-L457

daipom:
The reason why I fixed the codes about `suppress_repeated_stacktrace` is that originally the `suppress_repeated_stacktrace` value **from the `SytemConfig`** was applied in `Supervisor.load_config()`.

https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L887

Since this PR removes the initialization process of the logger from `Supervisor.load_config()`, now we need to apply the `suppress_repeated_stacktrace` value **from the `SytemConfig`**  in here.

https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L820-L826

Note: The following logic applies the `suppress_repeated_stacktrace` value **from the command line options**.

https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/supervisor.rb#L699-L705

daipom:
Thus, I thought I didn't change the specification of `suppress_repeated_stacktrace`,,, but now I'm concerned about if `suppress_repeated_stacktrace` of a worker process has been changed...
(Originally, was it applied to a worker process correctly??)
I will look into this.

daipom:
> > `Supervisor.load_config()` is called in ServerEngine's reloading function, but Fluentd doesn't use the function even when `SIGHUP` or `SIGUSR2`. So I can't see the reason for initializing the logger in that callback.
> 
> Sorry, I can't remember the code around it soon. I'm sorry to bother you but could you point out the rationale?

Sure! Me too, sorry for the lack of explanation.
I will add more explanation about this.

daipom:
> > > `Supervisor.load_config()` is called in ServerEngine's reloading function, but Fluentd doesn't use the function even when `SIGHUP` or `SIGUSR2`. So I can't see the reason for initializing the logger in that callback.
> > 
> > 
> > Sorry, I can't remember the code around it soon. I'm sorry to bother you but could you point out the rationale?
> 
> Sure! Me too, sorry for the lack of explanation. I will add more explanation about this.

# About ServerEngine logic

* Initialization of `Daemon`
  * `Supervisor.load_config() ` is called at this initialization process.
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/daemon.rb#L32
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/config_loader.rb#L23-L37
* `ConfigLoader::reload_config()`
  * The question is whether this can be called outside of initialization.
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/config_loader.rb#L42-L58
* `Server::restart()` or `Server::reload()` are the trigger for `ConfigLoader::reload_config()`.
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/server.rb#L59-L71
* They can be called by SIGNAL.
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/server.rb#L78-L116
  * `USR1`, `USR2`, `HUP`
* Fluentd supervisor process overwrite all these SIGNALs, and does NOT use this ServerEngine feature.
  * https://github.com/fluent/fluentd/blob/248160f22228672ad7bcbcb6ae68dbff9627a98b/lib/fluent/supervisor.rb#L176-L193

I'll check the Windows just to be sure.

daipom:
On Windows, Fluentd overwrites `Server::restart()` and `Server:reload()`, so it does NOT use `ConfigLoader::reload_config()`.

https://github.com/fluent/fluentd/blob/248160f22228672ad7bcbcb6ae68dbff9627a98b/lib/fluent/supervisor.rb#L195-L211

daipom:
On Ubuntu focal, I have confirmed that sending `USR1`, `USR2`, `HUP` to the supervisor process does not call `Supervisor.load_config()`.

On Windows, I have confirmed that using `fluent-ctl flush/restart/reload` to the supervisor process does not call it.

daipom:
I'm checking the history...

* https://github.com/fluent/fluentd/pull/880

daipom:
It seems that `Supervisor.load_config()` is used by `USR2` and `HUP` at the point of #880, but after #2716, Fluentd completely stopped using ServerEngine's `reload_config()` feature.

After #2716, there seems to be no longer any need to initialize the logger in `Supervisor.load_config()`.

daipom:
I tried Fluentd v1.8.1, which does not include #2716, and I confirmed that sending `USR2` to the supervisor process call `Supervisor.load_config()`.

From the above, I conclude as follows.

* `Supervisor.load_config()` was used for reloading, but not anymore. (since #2716)
  * Therefore, in the past, we could update the logger on reloading by initializing it there, but now it is meaningless.
* So currently `Supervisor.load_config()` is only for the initialization of `ServerEngine`.
* Supervisor and worker process of Fluentd have their own logger for their process, so all Fluentd has to do about logging is passing its logger to `ServerEnigne`.
  * Then `ServerEnigne` just uses it.
  * https://github.com/treasure-data/serverengine/blob/v2.3.1/lib/serverengine/config_loader.rb#L62-L68

lib/fluent/supervisor.rb
```diff
@@ -437,28 +437,10 @@ def self.load_config(path, params = {})
...
 
       log_level = params['log_level']
-      suppress_repeated_stacktrace = params['suppress_repeated_stacktrace']
-      ignore_repeated_log_interval = params['ignore_repeated_log_interval']
-      ignore_same_log_interval = params['ignore_same_log_interval']
 
-      log_path = params['log_path']
       chuser = params['chuser']
       chgroup = params['chgroup']
       chumask = params['chumask']
```
ashie, line 443: These remaining local variables doesn't seem need too, passing them directly to `se_config` will make the code makes simple.

ashie (commented):
Thanks for your detailed explanation! I've understood the issue.
It looks good to me.
I've added a nitpick comment in lines.

lib/fluent/supervisor.rb
daipom: Indeed! I will fix it!

daipom: Fixed!

lib/fluent/supervisor.rb
```diff
@@ -424,6 +424,10 @@ def dump
   end
 
   class Supervisor
+    # For ServerEngine's `reload_config`.
+    # This is called only at the initilization of the supervisor process,
+    # since Fluentd overwrites all related SIGNAL(HUP,USR1,USR2) and have own
+    # reloading feature.
     def self.load_config(path, params = {})
```
daipom, line 431: Perhaps we should rename or move this method.
It is hard to understand what this method is.
I will do it in another PR.

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

Issue #2289, opened 2019-02-08
Support <worker N-M> syntax


repeatedly:
This feature improves multi process usage.
Currently, `<worker>` section takes only 1 number and it can't define multiple input plugins to share port.
`<worker N-M>` resolves the problem like below.

```
<worker 0-4>
  <source>
    @type tcp
  </source>
  <match pattern>
    # ...
  </match>
</worker>

<worker 5-10>
  <source>
    @type forward
  </source>
  <match pattern>
    # ...
  </match>
</worker>
```

This patten is popular in large scale deployment so adding this syntax helps users.

PR #2651, opened 2019-10-15
Fix dry run mode

ganmacs:
**Which issue(s) this PR fixes**: 
ref #2624

**What this PR does / why we need it**: 

#### Show better log

<details>

Example config
```
<system>
  workers 2
</system>

<worker 0>
  <source>
    @type fake
    secret very-secret-string0
    test 1
  </source>

  <match>
    @type null
  </match>
</worker>

<worker 1>
  <source>
    @type fake
    secret very-secret-string1
  </source>
  <match>
    @type null
  </match>
</worker>
```

#### normal mode

```
$ fluentd -c example/debug/secret.conf --dry-run
```

Before (supervisor and worker show same log about `@type fake`)

```
2019-10-15 15:16:44 +0900 [info]: parsing config file is succeeded path="example/debug/secret.conf"
2019-10-15 15:16:44 +0900 [info]: adding match pattern="**" type="null"
2019-10-15 15:16:44 +0900 [info]: adding match pattern="**" type="null"
2019-10-15 15:16:44 +0900 [info]: adding source type="fake"
2019-10-15 15:16:44 +0900 [info]: adding source type="fake"
2019-10-15 15:16:44 +0900 [info]: using configuration file: <ROOT>
  <system>
    workers 2
  </system>
  <worker 0>
    <source>
      @type fake
      secret xxxxxx
      test 1
    </source>
    <match>
      @type null
    </match>
  </worker>
  <worker 1>
    <source>
      @type fake
      secret xxxxxx
    </source>
    <match>
      @type null
    </match>
  </worker>
</ROOT>
2019-10-15 15:16:44 +0900 [warn]: parameter 'test' in <source>
  @type fake
  secret xxxxxx
  test 1
</source> is not used.
2019-10-15 15:16:44 +0900 [info]: starting fluentd-1.7.3 pid=91300 ruby="2.6.3"
2019-10-15 15:16:44 +0900 [info]: spawn command to main:  cmdline=["/Users/yuta.iwama/.rbenv/versions/2.6.3/bin/ruby", "-Eascii-8bit:ascii-8bit", "-rbundler/setup", "bin/fluentd", "-c", "example/debug/secret.conf", "-p", "plugins", "--under-supervisor"]
2019-10-15 15:16:45 +0900 [info]: #1 adding match pattern="**" type="null"
2019-10-15 15:16:45 +0900 [info]: gem 'fluentd' version '1.7.3'
[...]

PR #2292, opened 2019-02-13
Support multi workers assign syntax on `<worker>` section. Fix #2289

cosmo0920:
This PR contains a naive `<worker N-M>` syntax implementation.
Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>

PR #4066, opened 2023-02-20
Fix value of `system_config.workers` at `run_configure`

### LINKED ISSUE #4051: `system_config.workers` value is wrong when validating config on launching
### Describe the bug

`system_config.workers` value is wrong when validating config on launching.

When launching the supervisor process, Fluentd validates the config as follows.
(It may be difficult to understand what this code is doing. This just validates the config before launching each worker)

https://github.com/fluent/fluentd/blob/e89092ce1132a933c12bb23fe8c9323c07ca81f5/lib/fluent/supervisor.rb#L731

This calls `configure()` of each plugin, and tries to find any config error.

However, `system_config.workers` value is wrong when we use the multiple workers setting in `system` directive.

Interestingly, in the case `<worker 0-N>`, the value is correct.

* This caused a bug about #3942

A detailed code survey can be found here.

This can cause bugs in config validation in the future.
We should fix this.

* Note: The problem is similar to #3469. We may consider fixing it at the same time.

### To Reproduce

Use a plugin to print out `system_config.workers` in `configure()`.

```ruby
def configure(conf)
  super
  p system_config.workers
end
```

I fixed `out_stdout` to reproduce.

```diff
--- a/lib/fluent/plugin/out_stdout.rb
+++ b/lib/fluent/plugin/out_stdout.rb
@@ -51,6 +51,8 @@ module Fluent::Plugin
 
       super
 
+      p system_config.workers
+
       @formatter = formatter_create
     end
```

Then you can see the value is `1`, although you set `workers 2` in `system` directive.

### Expected behavior

The actual number of workers is printed out.

### Your Environment

```markdown
- Fluentd version: 1.15.3
- Operating system: "Ubuntu 20.04.5 LTS"
- Kernel version: 5.15.0-58-generic
```

### Your Configuration

```apache
<system>
  workers 2
</system>

<source>
  @type sample
  tag test
</source>

<match test.**>
  @type stdout
</match>
```

### Your Error Log

```shell
2023-02-10 18:47:59 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2023-02-10 18:47:59 +0900 [info]: parsing config file is succeeded path="../work/fluent.conf"
2023-02-10 18:47:59 +0900 [info]: gem 'fluentd' version '1.15.3'
1 <--- WRONG!!
2023-02-10 18:47:59 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [info]: using configuration file: <ROOT>
  <system>
    workers 2
  </system>
  <source>
    @type sample
    tag "test"
  </source>
  <match test.**>
    @type stdout
  </match>
</ROOT>
2023-02-10 18:47:59 +0900 [info]: starting fluentd-1.15.3 pid=677703 ruby="2.7.5"
2023-02-10 18:47:59 +0900 [info]: spawn command to main:  cmdline=["/home/daipom/.rbenv/versions/2.7.5/bin/ruby", "-r/home/daipom/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/bundler-2.2.31/lib/bundler/setup", "-Eascii-8bit:ascii-8bit", "/home/daipom/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/bin/fluentd", "-c", "../work/fluent.conf", "--under-supervisor"]
2023-02-10 18:47:59 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2023-02-10 18:47:59 +0900 [info]: #0 init worker0 logger path=nil rotate_age=nil rotate_size=nil
2023-02-10 18:47:59 +0900 [info]: #1 init workers logger path=nil rotate_age=nil rotate_size=nil
2023-02-10 18:47:59 +0900 [info]: adding match pattern="test.**" type="stdout"
2
2023-02-10 18:47:59 +0900 [info]: adding source type="sample"
2
2023-02-10 18:47:59 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [warn]: #1 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [info]: #0 starting fluentd worker pid=677723 ppid=677703 worker=0
2023-02-10 18:47:59 +0900 [warn]: #1 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2023-02-10 18:47:59 +0900 [info]: #1 starting fluentd worker pid=677724 ppid=677703 worker=1
2023-02-10 18:47:59 +0900 [info]: #0 fluentd worker is now running worker=0
2023-02-10 18:47:59 +0900 [info]: #1 fluentd worker is now running worker=1
2023-02-10 18:48:00.003808793 +0900 test: {"message":"sample"}
2023-02-10 18:48:00.003945149 +0900 test: {"message":"sample"}
```

### Additional context

_No response_

abetomo:
Fixes #4051

Override `system_config` only if `conf.target_worker_ids.size >= 1`.

* If `<worker N>` or `<worker 0-N>` is not set, `conf.target_worker_ids` is empty.
* If `<system> workers N </system>` is not set, `system_config.workers` defaults to 1.

**Docs Changes**:

**Release Note**: 

* Fix value of `system_config.workers` at `run_configure`
* Change argument type of `Fluent::Plugin::Base::configure()` to `Fluent::Config::Element` only.

### DIFF
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -52,13 +52,13 @@ module Fluent
         @_fluentd_worker_id
       end
 
-      # @param conf [Fluent::Config::Element, Hash] configuration
       def configure(conf)
-        if conf.is_a?(Fluent::Config::Element)
-          if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
-            system_config_override(workers: conf.target_worker_ids.size)
-          end
+        raise ArgumentError, "BUG: type of conf must be Fluent::Config::Element, but #{conf.class} is passed." unless conf.is_a?(Fluent::Config::Element)
+
+        if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
+          system_config_override(workers: conf.target_worker_ids.size)
         end
+
         super(conf, system_config.strict_config_value)
         @_state ||= State.new(false, false, false, false, false, false, false, false, false)
         @_state.configure = true
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -55,7 +55,7 @@ module Fluent
       def configure(conf)
         raise ArgumentError, "BUG: type of conf must be Fluent::Config::Element, but #{conf.class} is passed." unless conf.is_a?(Fluent::Config::Element)
 
-        if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
+        if conf.for_this_worker? || (Fluent::Engine.supervisor_mode && !conf.for_every_workers?)
           system_config_override(workers: conf.target_worker_ids.size)
         end
 
--- a/lib/fluent/plugin/base.rb
+++ b/lib/fluent/plugin/base.rb
@@ -52,14 +52,12 @@ module Fluent
         @_fluentd_worker_id
       end
 
+      # @param conf [Fluent::Config::Element, Hash] configuration
       def configure(conf)
-        if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
-          workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
-                      conf.target_worker_ids.size
-                    else
-                      1
-                    end
-          system_config_override(workers: workers)
+        if conf.is_a?(Fluent::Config::Element)
+          if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
+            system_config_override(workers: conf.target_worker_ids.size)
+          end
         end
         super(conf, system_config.strict_config_value)
         @_state ||= State.new(false, false, false, false, false, false, false, false, false)
--- a/test/plugin/test_base.rb
+++ b/test/plugin/test_base.rb
@@ -146,4 +146,148 @@ class BaseTest < Test::Unit::TestCase
       end
     end
   end
+
+  sub_test_case 'system_config.workers value after configure' do
+    sub_test_case 'with <system> workers 3 </system>' do
+      setup do
+        system_config = Fluent::SystemConfig.new
+        system_config.workers = 3
+        stub(Fluent::Engine).system_config { system_config }
+      end
+
+      sub_test_case 'supervisor_mode is true' do
+        setup do
+          stub(Fluent::Engine).supervisor_mode { true }
+          stub(Fluent::Engine).worker_id { -1 }
+        end
+
+        test 'without <worker> directive' do
+          conf = config_element()
+          conf.set_target_worker_ids([])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 3 }
+        end
+
+        test 'with <worker 0>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test 'with <worker 0-1>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0, 1])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 2 }
+        end
+
+        test 'with <worker 0-2>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0, 1, 2])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 3 }
+        end
+
+        test '`conf` is `Hash`' do
+          @p.configure({})
+          assert{ @p.system_config.workers == 3 }
+        end
+      end
+
+      sub_test_case 'supervisor_mode is false' do
+        setup do
+          stub(Fluent::Engine).supervisor_mode { false }
+          stub(Fluent::Engine).worker_id { 0 }
+        end
+
+        test 'without <worker> directive' do
+          conf = config_element()
+          conf.set_target_worker_ids([])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 3 }
+        end
+
+        test 'with <worker 0>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test 'with <worker 0-1>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0, 1])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 2 }
+        end
+
+        test 'with <worker 0-2>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0, 1, 2])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 3 }
+        end
+
+        test '`conf` is `Hash`' do
+          @p.configure({})
+          assert{ @p.system_config.workers == 3 }
+        end
+      end
+    end
+
+    sub_test_case 'without <system> directive' do
+      sub_test_case 'supervisor_mode is true' do
+        setup do
+          stub(Fluent::Engine).supervisor_mode { true }
+          stub(Fluent::Engine).worker_id { -1 }
+        end
+
+        test 'without <worker> directive' do
+          conf = config_element()
+          conf.set_target_worker_ids([])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test 'with <worker 0>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test '`conf` is `Hash`' do
+          @p.configure({})
+          assert{ @p.system_config.workers == 1 }
+        end
+      end
+
+      sub_test_case 'supervisor_mode is false' do
+        setup do
+          stub(Fluent::Engine).supervisor_mode { false }
+          stub(Fluent::Engine).worker_id { 0 }
+        end
+
+        test 'without <worker> directive' do
+          conf = config_element()
+          conf.set_target_worker_ids([])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test 'with <worker 0>' do
+          conf = config_element()
+          conf.set_target_worker_ids([0])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 1 }
+        end
+
+        test '`conf` is `Hash`' do
+          @p.configure({})
+          assert{ @p.system_config.workers == 1 }
+        end
+      end
+    end
+  end
 end
--- a/test/plugin/test_base.rb
+++ b/test/plugin/test_base.rb
@@ -147,6 +147,12 @@ class BaseTest < Test::Unit::TestCase
     end
   end
 
+  test '`ArgumentError` when `conf` is not `Fluent::Config::Element`' do
+    assert_raise ArgumentError.new('BUG: type of conf must be Fluent::Config::Element, but Hash is passed.') do
+      @p.configure({})
+    end
+  end
+
   sub_test_case 'system_config.workers value after configure' do
     sub_test_case 'with <system> workers 3 </system>' do
       setup do
@@ -188,11 +194,6 @@ class BaseTest < Test::Unit::TestCase
           @p.configure(conf)
           assert{ @p.system_config.workers == 3 }
         end
-
-        test '`conf` is `Hash`' do
-          @p.configure({})
-          assert{ @p.system_config.workers == 3 }
-        end
       end
 
       sub_test_case 'supervisor_mode is false' do
@@ -228,11 +229,6 @@ class BaseTest < Test::Unit::TestCase
           @p.configure(conf)
           assert{ @p.system_config.workers == 3 }
         end
-
-        test '`conf` is `Hash`' do
-          @p.configure({})
-          assert{ @p.system_config.workers == 3 }
-        end
       end
     end
 
@@ -256,11 +252,6 @@ class BaseTest < Test::Unit::TestCase
           @p.configure(conf)
           assert{ @p.system_config.workers == 1 }
         end
-
-        test '`conf` is `Hash`' do
-          @p.configure({})
-          assert{ @p.system_config.workers == 1 }
-        end
       end
 
       sub_test_case 'supervisor_mode is false' do
@@ -280,11 +271,6 @@ class BaseTest < Test::Unit::TestCase
           conf = config_element()
           conf.set_target_worker_ids([0])
           @p.configure(conf)
-          assert{ @p.system_config.workers == 1 }
-        end
-
-        test '`conf` is `Hash`' do
-          @p.configure({})
           assert{ @p.system_config.workers == 1 }
         end
       end
--- a/test/plugin/test_base.rb
+++ b/test/plugin/test_base.rb
@@ -154,6 +154,18 @@ class BaseTest < Test::Unit::TestCase
   end
 
   sub_test_case 'system_config.workers value after configure' do
+    def assert_system_config_workers_value(data)
+      conf = config_element()
+      conf.set_target_worker_ids(data[:target_worker_ids])
+      @p.configure(conf)
+      assert{ @p.system_config.workers == data[:expected] }
+    end
+
+    def stub_supervisor_mode
+      stub(Fluent::Engine).supervisor_mode { true }
+      stub(Fluent::Engine).worker_id { -1 }
+    end
+
     sub_test_case 'with <system> workers 3 </system>' do
       setup do
         system_config = Fluent::SystemConfig.new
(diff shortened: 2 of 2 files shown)


daipom:
Thanks for the fix!

The status of this PR is Draft. Please let me know which part is TODO.

abetomo:
Thanks for your comments!
I will add some tests.

daipom:
> I will add some tests.

Thanks! We need tests for this issue!

In addition, this problem is difficult. Hard to know what was wrong and how we should fix it.
I want to clarify it.

The specification history related to this is as follows.

* #1507
* #2292

What went wrong in which process?

abetomo:
#2651 is related.

If there is no `<worker>` directive, `conf.for_this_worker?` is false.
However, if `Fluent::Engine.supervisor_mode` is true, it will overwrite `system_config.workers` regardless of `conf.for_this_worker?`.

Since `conf.target_worker_ids` is empty when there is no `<worker>` directive, this fix checks for it and does not override `system_config.workers`.

daipom:
Thanks so much!
It certainly seems that #2651 has something to do with this problem.

There was nothing wrong at the point of #1507 and #2292.
(The method of rewriting the value of the system config in the process seems a bit too complicated, but I guess it was inevitable since `SystemConfig.workers` was used to determine if the plugin works in the multi-worker.
It seems to me that this complicated specification could be the cause of bugs in the future..., anyway, it is not directly related to this issue.)

I will see #2651!

daipom:
It is hard to understand the purpose of code changes in #2651...

I'm concerned about the following points.

* About #2292
  * If `conf.for_this_worker?` is `true`, then `conf.target_worker_ids` appears to have at least one id. Why was the condition `if conf.target_worker_ids && !conf.target_worker_ids.empty?` needed?
    * https://github.com/fluent/fluentd/pull/2292/files#diff-601c8bc645b7008fa857b253a5295ef4b796c6236d0428ed8d72cecd572be6edR243-R245
    * https://github.com/fluent/fluentd/pull/2292/files#diff-a011588d74f931c0c2154891220af14f9181f28eb70b978d2e017866dfba3081R54-R59
  * It seems to me that the following code is enough at this point.
  ```rb
  if conf.for_this_worker?
    system_config_override(workers: conf.target_worker_ids.size)
  end
  ```
* About #2651
  * Why does `Fluent::Engine::run_configure()` run before launching the supervisor process even when `dry_run` is `false`?
    * Shouldn't it be run only when `dry_run` is `true`?
    * https://github.com/fluent/fluentd/pull/2651/files#diff-6042a07157b33b2d729ab3d10ed6bba89ad5f71aaed2be1aeb22a179a45ebbaeR499-R508
  * When `Fluent::Engine.supervisor_mode` is `true`, what is the expected value of `workers`?
    * I think we have to validate the config in one process in this case, but how many workers each plugin will work with will change depending on the setting.
      For example, plugin-A works with worker-1 and worker-2, and plugin-B works with only worker-3.
      With the current structure, would it be difficult to properly reflect and validate the number of workers?

I don't think all the points should be solved now.
I just noted points of concern as I work on this issue.
If you have noticed anything in the process of making this fix, I'd like to know.

The current fix may be enough to solve the issue #4051, but it may be better to leave some code comments for the future.

abetomo:
Thanks for your comment!
I'll look into it.

daipom:
Thanks! 

> When `Fluent::Engine.supervisor_mode` is `true`, what is the expected value of `workers`?
>     * I think we have to validate the config in one process in this case, but how many workers each plugin will work with will change depending on the setting.
>       For example, plugin-A works with worker-1 and worker-2, and plugin-B works with only worker-3.
>       With the current structure, would it be difficult to properly reflect and validate the number of workers?

I have checked this, and I understand this fix allows us to validate the config with the correct `workers` value!

I used this config.

```xml
<system>
  workers 8
</system>

<worker 0>
  <source>
    @id A
    @type sample
    tag test
  </source>
</worker>

<worker 1-2>
  <source>
    @id B
    @type sample
    tag test
  </source>
</worker>

<source>
  @id C
  @type sample
  tag test
</source>

<match test.**>
  @type stdout
</match>
```

I added a log for debugging to `in_sample`.

```diff
--- a/lib/fluent/plugin/in_sample.rb
+++ b/lib/fluent/plugin/in_sample.rb
@@ -63,6 +63,7 @@ module Fluent::Plugin
 
     def configure(conf)
       super
+      log.info id: @id, workers: system_config.workers
       @sample_index = 0
       config = conf.elements.select{|e| e.name == 'storage' }.first
```

The result of master branch:

```
2023-02-24 00:20:14 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2023-02-24 00:20:14 +0900 [info]: parsing config file is succeeded path="../work/fluent.conf"
...
2023-02-24 00:20:14 +0900 [info]: [C]  id="C" workers=1
2023-02-24 00:20:14 +0900 [info]: [A]  id="A" workers=1
2023-02-24 00:20:14 +0900 [info]: [B]  id="B" workers=2
...
2023-02-24 00:20:14 +0900 [info]: starting fluentd-1.15.3 pid=1286518 ruby="3.2.0"
...
2023-02-24 00:20:15 +0900 [info]: #7 [C]  id="C" workers=8
...
2023-02-24 00:20:15 +0900 [info]: #1 [B]  id="B" workers=2
...
2023-02-24 00:20:15 +0900 [info]: #0 [A]  id="A" workers=1
...
```

Before starting Fluentd, `workers` value of `C` is wrong.

The expected `workers` values before stating Fluentd are as follows.

```
[C]  id="C" workers=8
[A]  id="A" workers=1
[B]  id="B" workers=2
```

When applying this fix, these values are as expected!

daipom:
Summarize what we've learned so far.

> > * About [Support multi workers assign syntax on `<worker>` section. Fix #2289 #2292](https://github.com/fluent/fluentd/pull/2292)
> 
> As for #2292 condition, I too think @daipom's code is enough.

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/element.rb#L256-L258

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/plugin/base.rb#L56-L61

Originally, there was no need to overwrite `1`.

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/engine.rb#L213-L216

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/plugin/base.rb#L56

We need `Fluent::Engine.supervisor_mode` condition to validate the config before starting Fluentd.

However, with `supervisor_mode == true`,  the current logic overwrites the `workers` of all plugins that are not inside the `worker` directive.
This is the bug we have to fix.

After starting Fluentd, the `for_this_worker?` condition prevented this bug, so it was hard to notice.

This PR completely fixes this problem.

lib/fluent/plugin/base.rb
```diff
@@ -54,12 +54,9 @@ def fluentd_worker_id
 
       def configure(conf)
         if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
-          workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
-                      conf.target_worker_ids.size
-                    else
-                      1
-                    end
-          system_config_override(workers: workers)
+          if !conf&.target_worker_ids&.empty?
```
daipom, line 57:
```suggestion
          if !conf.target_worker_ids&.empty?
```

Do we need to consider the possibility that `conf` is `nil`?

daipom: We may fix this as follows. I'd like to hear others' opinions on this.

```rb
if Fluent::Engine.supervisor_mode || conf.for_this_worker?
  unless conf.target_worker_ids.empty?
```

abetomo:
`conf` of `Fluent::TextParser.configure(conf)` is `Hash`.
Thus, I added a `conf.is_a?(Fluent::Config::Element)` check.

`!conf.target_worker_ids.empty?` check is only necessary for `Fluent::Engine.supervisor_mode == true`.
So I changed to `(Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?)`.

daipom:
Thanks for the fix and adding tests! I will see this soon.

lib/fluent/plugin/base.rb
```diff
@@ -52,14 +52,12 @@ def fluentd_worker_id
...
+      # @param conf [Fluent::Config::Element, Hash] configuration
       def configure(conf)
-        if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
-          workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
-                      conf.target_worker_ids.size
-                    else
-                      1
-                    end
-          system_config_override(workers: workers)
+        if conf.is_a?(Fluent::Config::Element)
```
daipom, line 57: > `conf` of `Fluent::TextParser.configure(conf)` is `Hash`.
> Thus, I added a `conf.is_a?(Fluent::Config::Element)` check.

Thanks for the information.
This made me notice that API compatibility features for v0.12 may be involved.

* https://docs.fluentd.org/plugin-development#overview
* https://www.shigemk2.com/entry/memo_Fluentd_v0.14_Plugin_API_Updates

I have looked into this, but I am not sure whether we should really assume that the type of `conf` is not `Fluent::Config::Element`.

Some test codes look like passing `Hash` to `Fluent::Plugin::Base::configure()`.
(We need to be careful not to confuse it with `Fluent::Test::Driver::Base::configure()`)
Perhaps these tests are related to the v0.12 compatible layer.

* https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/test/test_formatter.rb
  * https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/test/test_formatter.rb#L317-L321
  * https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/test/test_formatter.rb#L332-L334
  * https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/test/test_formatter.rb#L339-L341
* https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/test/compat/test_parser.rb
  * Entirely.

However, it appears that `Hash` is not passed unless we explicitly pass it in tests.
(If `Hash` can be passed during normal operation, please let me know.)

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/supervisor.rb#L808-L814
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config.rb#L68-L79
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/v1_parser.rb#L43-L54
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/parser.rb#L25-L28
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/dsl.rb#L58-L62
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/element.rb#L67-L72
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/config/element.rb#L93-L97
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/root_agent.rb#L155-L163
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/plugin_helper/parser.rb#L40-L52
https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/plugin_helper/compat_parameters.rb#L304-L305

If so, I thought we could modify the tests and limit the type to `Element`.
But we should consider whether it is acceptable to break backward compatibility, even though this is a test-only impact.
Or it might be possible to fix it without breaking backward compatibility at all. (I haven't checked it yet.)

In the first place, I'm also beginning to wonder whether it's correct to overwrite `workers` in `Fluent::Plugin::Base::configure()`.
There may be a more appropriate place and timing, such as `Supervisor` or `RootAgent`.

I'll think about it a little more.

I would like to figure out what we should do about these problems in the future and then figure out how far we should go in this PR.

abetomo:
I fixed the use of `Hash` in the test with another PR.
#4074

lib/fluent/plugin/base.rb
ashie: Thanks to your explanation I'm able to understand the problem.

Now I have a question in your explanation.

> Or it might be possible to fix it without breaking backward compatibility at all. (I haven't checked it yet.)

What do you mean `breaking backward compatibility`?
Both this change and #4074 doesn't seem break backward compatibility.

daipom: > Now I have a question in your explanation.
> 
> > Or it might be possible to fix it without breaking backward compatibility at all. (I haven't checked it yet.)
> 
> What do you mean `breaking backward compatibility`? Both this change and #4074 doesn't seem break backward compatibility.

This is for the following.

> If so, I thought we could modify the tests and limit the type to `Element`.

If we limit the type of `conf` argument of `configure()`, it will break the backward compatibility.
Some tests, including 3rd-party plugins, may stop working.

So I said

> But we should consider whether it is acceptable to break backward compatibility, even though this is a test-only impact.

daipom: > If so, I thought we could modify the tests and limit the type to `Element`.

Oh, I understand the point. This sentence was hard to understand. Sorry.
This means the following 2 points.

1. We could modify the tests that passing `Hash` to `Fluent::Plugin::Base::configure()`
1. We could limit tye type of `conf` argument of the `Fluent::Plugin::Base::configure()`

`2.` can break the backward compatibility of some tests including 3rd-party plugins.

ashie: You mean the code completely rejecting non `Fluent::Config::Element` like the following breaks compatibility?
```ruby
def configure(conf)
  raise SomeError if conf.is_a?(Fluent::Config::Element)
```

daipom: Yes. I thought we could remove `if conf.is_a?(Fluent::Config::Element)` check in this PR.
This will raise `NoMethodError` when it is passed `Hash`.

```diff
def configure(conf)
-  if conf.is_a?(Fluent::Config::Element)
    if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
```

ashie: Yup, I think breaking compatibility doesn't good idea.
Even if we merge #4074, 3rd party plugins might use obsolete way.

daipom: Thanks for your advice!
OK! We shouldn't break compatibility even if the impact is only in tests!

> Or it might be possible to fix it without breaking backward compatibility at all.

About this, I think fixing these logic  may allow us to limit the type without breaking compatibility.

https://github.com/fluent/fluentd/blob/6c649d189b5cf65ccfce58f6952a31b24c775e72/lib/fluent/compat/parser.rb#L124-L140

However, perhaps we should consider this separately from this PR.

daipom: > I would like to figure out what we should do about these problems in the future and then figure out how far we should go in this PR.

The reason why I was thinking about limiting the type is that `workers` would not be overwritten unless the type is `Element`.
I think this is a bug of #1507.
When `Hash` is passed to `configure()`, this logic can't overwrite `workers`, and this can cause problems depending on the design of the plugin.

The currently intended specification is a bit unclear, but I think the correct specification is as follows.

* Only the `Fluent::Config::Element` should be passed to `Fluent::Plugin::Base::configure()`.
* For the v0.12 compatible feature, `Hash` can be passed to plugins in the `Fluent::Compat` layer, but it should be converted to the `Element` before being passed to `Fluent::Plugin::Base::configure()`.
* The test codes that pass `Hash` to `configure` of a plugin outside of the `Fluent::Compat` layer are incorrect, and should fail.

Or

* Even if `Hash` is passed to `Fluent::Plugin::Base::configure()`, `workers` is overwritten correctly.

I don't know much about the v0.12 compatible feature, so  I can miss something...
If so, please let me know.

Anyway, I hope we do the following in this PR.

* Clarify the correct specification.
* Leave the appropriate code comments or something for future development.

daipom: I have looked into the history of past revisions.

It seems that this problem is **NOT** related to the v0.12 compatible feature.

`Fluent::Config::Element` was implemented at the beginning of Fluentd and I can't find any specification that says we can use a `Hash` for `configure()`.

I think the trigger is this fix.

* #103

At this point, there was no `TestDriver` for `Parser`, and it seems that `Hash` was unintentionally passed to `configure()` .

I guess they didn't notice the problem because the other test passed `Hash` to `TestDriver` and it didn't look wrong at first glance.

`ParserTestDriver` is implemented in the following PR, but somehow some tests remained unfixed.

* #702

The same phenomenon is observed for `Formatter`.

* #712

I think these tests misled us and made us think that `Hash` could be passed on.

daipom: So, I think the correct specification is as follows.

* Only the `Fluent::Config::Element` should be passed to `Fluent::Plugin::Base::configure()`.
* All tests passing `Hash` directly to `Fluent::Plugin::Base::configure()` are wrong and must use `TestDriver`.

daipom: So I want to limit the type to `Fluent::Config::Element`, and fix some tests in #4074.

On the other hand, even if this was not the intended specification, I think there could be a decision to prioritize maintaining compatibility.
We need to discuss this point carefully.

What do you think? @ashie

ashie: Thank you for digging the history, it's very helpful to understand the issue.
Now I've found one more interesting thing.
test/test_buffer.rb had also `configure({})` in v0.14.0.pre.1 but it's replaced with `configure(config_element(...))` in test/plugin/test_buffer.rb of v0.14.0.rc1. test/plugin/test_buf_file.rb has also similar history.

Now I think these facts implies that your claim is reasonable.
O.K, we should pass only `Fluent::Config::Element` to `Fluent::Plugin::Base::configure()`.

@abetomo Could you modify your patch according to this policy?

daipom: Thanks, @ashie !!

abetomo: Thank you for the detailed information.
I will make the changes.

abetomo: Am I correct in my understanding that compatibility will not be maintained?

* Argument of `Fluent::Plugin::Base::configure()` is only `Fluent::Config::Element`.
* Passing `Hash` to `Fluent::Plugin::Base::configure()` makes an error.

daipom: Yes. Passing Hash is not intented specification. We can assume this argument is always Config::Element.

daipom: For example, I think simply removing the check of `if conf.is_a?(Fluent::Config::Element)` is fine.

abetomo: I thought it was difficult to understand the cause of the error with `NoMethodError`, so I changed it to throw `ArgumentError`.

lib/fluent/plugin/base.rb
```diff
@@ -52,15 +52,14 @@ def fluentd_worker_id
         @_fluentd_worker_id
       end
 
+      # @param conf [Fluent::Config::Element] configuration
```
daipom, line 55: I think we can remove this comment since many other methods don't specify the type with such a comment.
In addition, we can see the type of `conf` when we see the `raise ArgumentError` line.

lib/fluent/plugin/base.rb
```diff
@@ -52,15 +52,14 @@ def fluentd_worker_id
...
       def configure(conf)
-        if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
-          workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
-                      conf.target_worker_ids.size
-                    else
-                      1
-                    end
-          system_config_override(workers: workers)
+        raise ArgumentError, 'type of conf is Fluent::Config::Element' unless conf.is_a?(Fluent::Config::Element)
```
daipom, line 57:
```suggestion
        raise ArgumentError, 'BUG: type of conf must be Fluent::Config::Element, but #{conf.class} is passed.' unless conf.is_a?(Fluent::Config::Element)
```

From the other code, might a message like this be better?

https://github.com/fluent/fluentd/blob/fb641737d6bacace19be11cb676fec3fc20f0cf9/lib/fluent/engine.rb#L123-L125

lib/fluent/plugin/base.rb
daipom: (If `Hash` was also possible, we would need such a comment to understand that difficult specification.)

lib/fluent/plugin/base.rb
daipom: Nice idea!

lib/fluent/plugin/base.rb
abetomo: I see! I changed it that way.

test/plugin/test_base.rb
```diff
@@ -146,4 +146,134 @@ class FluentPluginBaseTest::DummyPlugin2 < Fluent::Plugin::TestBase
...
+          stub(Fluent::Engine).supervisor_mode { true }
+          stub(Fluent::Engine).worker_id { -1 }
+        end
+
+        test 'without <worker> directive' do
+          conf = config_element()
+          conf.set_target_worker_ids([])
+          @p.configure(conf)
+          assert{ @p.system_config.workers == 3 }
+        end
```
daipom, line 175: Could you summarize these new tests by using `data`?
It looks like it could be put together into one test.

Example:

https://github.com/fluent/fluentd/blob/fb641737d6bacace19be11cb676fec3fc20f0cf9/test/plugin/test_output.rb#L1077-L1105

abetomo: Changed to test using `data`.

daipom: Thanks! It is now easier to see!
Certainly, it would be better not to force them all into one test. This looks good.

I'm sorry I said we should put the tests together, but it looks like we need to split the tests up a bit more.
It is generally not desirable to assert different cases in a single test run.

Sorry for writing raw diffs, but how about something like this?
We can use `keep: true` to use the same data in the next test.

```diff
diff --git a/test/plugin/test_base.rb b/test/plugin/test_base.rb
index 7816f1a6..3d4404ba 100644
--- a/test/plugin/test_base.rb
+++ b/test/plugin/test_base.rb
@@ -161,16 +161,9 @@ class BaseTest < Test::Unit::TestCase
       assert{ @p.system_config.workers == data[:expected] }
     end
 
-    def when_supervisor_mode_is_true(data)
+    def stub_supervisor_mode
       stub(Fluent::Engine).supervisor_mode { true }
       stub(Fluent::Engine).worker_id { -1 }
-      assert_system_config_workers_value(data)
-    end
-
-    def when_supervisor_mode_is_false(data)
-      stub(Fluent::Engine).supervisor_mode { false }
-      stub(Fluent::Engine).worker_id { 0 }
-      assert_system_config_workers_value(data)
     end
 
     sub_test_case 'with <system> workers 3 </system>' do
@@ -185,32 +178,40 @@ class BaseTest < Test::Unit::TestCase
         {
           target_worker_ids: [],
           expected: 3
-        }
+        },
+        keep: true
       )
       data(
         'with <worker 0>',
         {
           target_worker_ids: [0],
           expected: 1
-        }
+        },
+        keep: true
       )
       data(
         'with <worker 0-1>',
         {
           target_worker_ids: [0, 1],
           expected: 2
-        }
+        },
+        keep: true
       )
       data(
         'with <worker 0-2>',
         {
           target_worker_ids: [0, 1, 2],
           expected: 3
-        }
+        },
+        keep: true
       )
       test 'system_config.workers value after configure' do
-        when_supervisor_mode_is_true(data)
-        when_supervisor_mode_is_false(data)
+        assert_system_config_workers_value(data)
+      end
+
+      test 'system_config.workers value after configure with supervisor_mode' do
+        stub_supervisor_mode
+        assert_system_config_workers_value(data)
       end
     end
 
@@ -220,18 +221,24 @@ class BaseTest < Test::Unit::TestCase
         {
           target_worker_ids: [],
           expected: 1
-        }
+        },
+        keep: true
       )
       data(
         'with <worker 0>',
         {
           target_worker_ids: [0],
           expected: 1
-        }
+        },
+        keep: true
       )
       test 'system_config.workers value after configure' do
-        when_supervisor_mode_is_true(data)
-        when_supervisor_mode_is_false(data)
+        assert_system_config_workers_value(data)
+      end
+
+      test 'system_config.workers value after configure with supervisor_mode' do
+        stub_supervisor_mode
+        assert_system_config_workers_value(data)
       end
     end
   end
```

abetomo: Thanks for the info.
I was looking for `keep: true`!

daipom: Thanks for fixing it!

lib/fluent/plugin/base.rb
```diff
@@ -53,14 +53,12 @@ def fluentd_worker_id
...
-        if Fluent::Engine.supervisor_mode || (conf.respond_to?(:for_this_worker?) && conf.for_this_worker?)
-          workers = if conf.target_worker_ids && !conf.target_worker_ids.empty?
-                      conf.target_worker_ids.size
-                    else
-                      1
-                    end
-          system_config_override(workers: workers)
+        raise ArgumentError, "BUG: type of conf must be Fluent::Config::Element, but #{conf.class} is passed." unless conf.is_a?(Fluent::Config::Element)
+
+        if (Fluent::Engine.supervisor_mode && !conf.target_worker_ids.empty?) || conf.for_this_worker?
```
daipom, line 58:
```suggestion
        if conf.for_this_worker? || (Fluent::Engine.supervisor_mode && !conf.for_every_workers?)
```

How about this?

* It is easier to understand if the `conf.for_this_worker?` condition is first.
* We can use `for_every_workers?` instead of `target_worker_ids.empty?`. It is easier to understand the intention.

abetomo: I changed it because it is easier to understand.

daipom (approved):
LGTM.
Thanks for fixing this difficult issue!

test/plugin/test_base.rb
ashie: I didn't know `keep: true`:
https://www.clear-code.com/blog/2018/12/26.html

daipom: I know it in that blog too :smile:

ashie (approved):
LGTM

ashie:
Thanks!

PR #4074, opened 2023-02-28
test: use `Fluent::Config::Element` in test

abetomo:

( https://github.com/fluent/fluentd/pull/4066#discussion_r1118388855 )

There is a case where `Hash` is used in the test of `configure` method.
I fixed it because `configure` method should use `Fluent::Config::Element`.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/test/test_formatter.rb
+++ b/test/test_formatter.rb
@@ -13,12 +13,16 @@ module FormatterTest
     {'message' => 'awesome', 'greeting' => 'hello'}
   end
 
+  def config_element_in_params(params)
+    config_element('test', '', params)
+  end
+
   class BaseFormatterTest < ::Test::Unit::TestCase
     include FormatterTest
 
     def test_call
       formatter = Formatter.new
-      formatter.configure({})
+      formatter.configure(config_element())
       assert_raise NotImplementedError do
         formatter.format('tag', Engine.now, {})
       end
@@ -135,7 +139,7 @@ module FormatterTest
     end
 
     def test_format
-      @formatter.configure({})
+      @formatter.configure(config_element())
       formatted = @formatter.format(tag, @time, record)
 
       assert_equal(record.to_msgpack, formatted)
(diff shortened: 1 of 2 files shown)


ashie:
Whether we should merge this or not is depends on the discussion in https://github.com/fluent/fluentd/pull/4066#discussion_r1118388855

daipom:
Sorry for the delay, yes I think so too.

I'm considering #4066's goal.
Please give me more time.

test/test_formatter.rb
```diff
@@ -13,12 +13,16 @@ def record
...
+    config_element('test', '', params)
+  end
+
   class BaseFormatterTest < ::Test::Unit::TestCase
     include FormatterTest
 
     def test_call
       formatter = Formatter.new
-      formatter.configure({})
+      formatter.configure(config_element())
```
daipom, line 25: I think we should use `Fluent::Test::FormatterTestDriver` along with other formatter tests.
Then we can pass `Hash` as before.

ashie: I think it doesn't use `Fluent::Test::FormatterTestDriver` intentionally.
Because `test_call` in `BaseFormatterTestWithTestDriver` does the test you intend.

ashie:
> Whether we should merge this or not depends on the discussion in [#4066 (comment)](https://github.com/fluent/fluentd/pull/4066#discussion_r1118388855)

Our conclusion in the discussion is https://github.com/fluent/fluentd/pull/4066#discussion_r1126481104
> So, I think the correct specification is as follows.
> 
>     * Only the `Fluent::Config::Element` should be passed to `Fluent::Plugin::Base::configure()`.
> 
>     * All tests passing `Hash` directly to `Fluent::Plugin::Base::configure()` are wrong and must use `TestDriver`.

So we'll merge.

test/test_formatter.rb
daipom: Oh, sorry. You are right. This code should be as is. Thanks!

ashie:
Thanks!

test/test_formatter.rb
```diff
@@ -135,7 +139,7 @@ def setup
     end
 
     def test_format
-      @formatter.configure({})
+      @formatter.configure(config_element())
```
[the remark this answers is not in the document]
ashie: Ah, sorry I've already merged this although you are right...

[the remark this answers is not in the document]
daipom: No problem! I'm making another PR for this now!

PR #4081, opened 2023-03-03
Fix race condition of out_secondary_file

### LINKED ISSUE #4080: `SecondaryFileOutput` sometimes appends logs even when `append false` in multiple workers
### Describe the bug

This logic causes race condition issues in multiple workers when `append false`.

https://github.com/fluent/fluentd/blob/e58fd66d71bf3df046f2fd29980521e59f6fcb83/lib/fluent/plugin/out_secondary_file.rb#L118-L129

### To Reproduce

Run the fluentd with the following config for about 1 minute.

```console
bundle exec fluentd -c fluent.conf
```

Then we can find some files where logs with different tags are mixed in the same file although the chunk-key was `tag` and `append` was `false`.

```
2023-03-03T19:53:28+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:29+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:30+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:31+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:32+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:33+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:34+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:35+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:36+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:37+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:38+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:28+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:29+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:30+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:31+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:32+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:33+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:34+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:35+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:36+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:37+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:38+09:00	aaa.fuga.bbb	{"message":"fuga"}
```

### Expected behavior

Each chunk is separated into a file.

File-A

```
2023-03-03T19:53:28+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:29+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:30+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:31+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:32+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:33+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:34+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:35+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:36+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:37+09:00	aaa.hoge.bbb	{"message":"hoge"}
2023-03-03T19:53:38+09:00	aaa.hoge.bbb	{"message":"hoge"}
```

File-B

```
2023-03-03T19:53:28+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:29+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:30+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:31+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:32+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:33+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:34+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:35+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:36+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:37+09:00	aaa.fuga.bbb	{"message":"fuga"}
2023-03-03T19:53:38+09:00	aaa.fuga.bbb	{"message":"fuga"}
```

### Your Environment

```markdown
- Fluentd version: fluentd 1.15.3
- Operating system: Ubuntu 20.04.5 LTS (Focal Fossa)
- Kernel version: 5.15.0-60-generic
```

### Your Configuration

```apache
<system>
  workers 4
</system>

<source>
  @type sample
  tag aaa.hoge.bbb
  sample {"message": "hoge"}
</source>

<source>
  @type sample
  tag aaa.fuga.bbb
  sample {"message": "fuga"}
</source>

<source>
  @type sample
  tag aaa.foo.bbb
  sample {"message": "foo"}
</source>

<match aaa.**>
  @type file
  path /test/fluentd/log/${tag}/fluentd.log
  append true
  <buffer tag>
    @type file
    path /test/fluentd/buffer
    flush_mode interval
    flush_interval 10s
    retry_max_times 1
  </buffer>
  <secondary>
    @type secondary_file
    directory /test/fluentd/error/
    append false
  </secondary>
</match>
```

### Your Error Log

```shell
Error logs are not related to this issue.
```

### Additional context

_No response_

daipom:
Fixes #4080

For solving race condition issues of `out_secondary_file`.

* Note: Similar problem in `out_file`: #3808.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -77,7 +77,7 @@ module Fluent::Plugin
 
     def write(chunk)
       path_without_suffix = extract_placeholders(@path_without_suffix, chunk)
-      return generate_path(path_without_suffix) do |path|
+      generate_path(path_without_suffix) do |path|
         FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
         case @compress
@@ -147,7 +147,7 @@ module Fluent::Plugin
       rescue FileAlreadyExist
         retry
       end
-      return path
+      path
     end
 
     class FileAlreadyExist < StandardError
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -61,18 +61,10 @@ module Fluent::Plugin
 
       @dir_perm = system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION
       @file_perm = system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION
-      @need_worker_lock = system_config.workers > 1
-      @need_thread_lock = @primary_instance.buffer_config.flush_thread_count > 1
     end
 
     def multi_workers_ready?
       true
-    end
-
-    def start
-      super
-      extend WriteLocker
-      @write_mutex = Mutex.new
     end
 
     def write(chunk)
@@ -151,36 +143,6 @@ module Fluent::Plugin
     end
 
     class FileAlreadyExist < StandardError
-    end
-
-    module WriteLocker
-      def lock_if_need(path)
-        get_worker_lock_if_need(path) do
-          get_thread_lock_if_need do
-            yield
-          end
-        end
-      end
-
-      def get_worker_lock_if_need(path)
-        unless @need_worker_lock
-          yield
-          return
-        end
-        acquire_worker_lock(path) do
-          yield
-        end
-      end
-
-      def get_thread_lock_if_need
-        unless @need_thread_lock
-          yield
-          return
-        end
-        @write_mutex.synchronize do
-          yield
-        end
-      end
     end
   end
 end
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -61,34 +61,40 @@ module Fluent::Plugin
 
       @dir_perm = system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION
       @file_perm = system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION
+      @need_worker_lock = system_config.workers > 1
+      @need_thread_lock = @primary_instance.buffer_config.flush_thread_count > 1
     end
 
     def multi_workers_ready?
-      ### TODO: add hack to synchronize for multi workers
       true
+    end
+
+    def start
+      super
+      extend WriteLocker
+      @write_mutex = Mutex.new
     end
 
     def write(chunk)
       path_without_suffix = extract_placeholders(@path_without_suffix, chunk)
-      path = generate_path(path_without_suffix)
-      FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
+      return generate_path(path_without_suffix) do |path|
+        FileUtils.mkdir_p File.dirname(path), mode: @dir_perm
 
-      case @compress
-      when :text
-        File.open(path, "ab", @file_perm) {|f|
-          f.flock(File::LOCK_EX)
-          chunk.write_to(f)
-        }
-      when :gzip
-        File.open(path, "ab", @file_perm) {|f|
-          f.flock(File::LOCK_EX)
-          gz = Zlib::GzipWriter.new(f)
-          chunk.write_to(gz)
-          gz.close
-        }
+        case @compress
+        when :text
+          File.open(path, "ab", @file_perm) {|f|
+            f.flock(File::LOCK_EX)
+            chunk.write_to(f)
+          }
+        when :gzip
+          File.open(path, "ab", @file_perm) {|f|
+            f.flock(File::LOCK_EX)
+            gz = Zlib::GzipWriter.new(f)
+            chunk.write_to(gz)
+            gz.close
+          }
+        end
       end
-
-      path
     end
 
     private
@@ -117,13 +123,62 @@ module Fluent::Plugin
 
     def generate_path(path_without_suffix)
       if @append
-        "#{path_without_suffix}#{@suffix}"
-      else
+        path = "#{path_without_suffix}#{@suffix}"
+        lock_if_need(path) do
+          yield path
+        end
+        return path
+      end
+
+      begin
         i = 0
         loop do
           path = "#{path_without_suffix}.#{i}#{@suffix}"
-          return path unless File.exist?(path)
+          break unless File.exist?(path)
           i += 1
+        end
+        lock_if_need(path) do
+          # If multiple processes or threads select the same path and another
+          # one entered this locking block first, the file should already
+          # exist and this one should retry to find new path.
+          raise FileAlreadyExist if File.exist?(path)
+          yield path
+        end
+      rescue FileAlreadyExist
+        retry
+      end
+      return path
+    end
+
+    class FileAlreadyExist < StandardError
+    end
+
+    module WriteLocker
+      def lock_if_need(path)
+        get_worker_lock_if_need(path) do
+          get_thread_lock_if_need do
+            yield
+          end
+        end
+      end
+
+      def get_worker_lock_if_need(path)
+        unless @need_worker_lock
+          yield
+          return
+        end
+        acquire_worker_lock(path) do
+          yield
+        end
+      end
+
+      def get_thread_lock_if_need
+        unless @need_thread_lock
+          yield
+          return
+        end
+        @write_mutex.synchronize do
+          yield
         end
       end
     end
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -116,7 +116,7 @@ module Fluent::Plugin
     def generate_path(path_without_suffix)
       if @append
         path = "#{path_without_suffix}#{@suffix}"
-        lock_if_need(path) do
+        acquire_lock_if_need(path) do
           yield path
         end
         return path
@@ -129,7 +129,7 @@ module Fluent::Plugin
           break unless File.exist?(path)
           i += 1
         end
-        lock_if_need(path) do
+        acquire_lock_if_need(path) do
           # If multiple processes or threads select the same path and another
           # one entered this locking block first, the file should already
           # exist and this one should retry to find new path.
--- a/lib/fluent/plugin/out_secondary_file.rb
+++ b/lib/fluent/plugin/out_secondary_file.rb
@@ -116,7 +116,7 @@ module Fluent::Plugin
     def generate_path(path_without_suffix)
       if @append
         path = "#{path_without_suffix}#{@suffix}"
-        acquire_lock_if_need(path) do
+        synchronize_path(path) do
           yield path
         end
         return path
@@ -129,7 +129,7 @@ module Fluent::Plugin
           break unless File.exist?(path)
           i += 1
         end
-        acquire_lock_if_need(path) do
+        synchronize_path(path) do
           # If multiple processes or threads select the same path and another
           # one entered this locking block first, the file should already
           # exist and this one should retry to find new path.
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -609,15 +609,15 @@ module Fluent
       # (For multiple workers, the lock is shared if `worker_lock_name` is the same value).
       # For multiple threads, `worker_lock_name` is not used, and the lock is shared by all
       # threads in the same process.
-      def lock_if_need(worker_lock_name)
-        get_worker_lock_if_need(worker_lock_name) do
-          get_flush_thread_lock_if_need do
+      def acquire_lock_if_need(worker_lock_name)
+        acquire_worker_lock_if_need(worker_lock_name) do
+          acquire_flush_thread_lock_if_need do
             yield
           end
         end
       end
 
-      def get_worker_lock_if_need(name)
+      def acquire_worker_lock_if_need(name)
         need_worker_lock = system_config.workers > 1
         if need_worker_lock
           acquire_worker_lock(name) { yield }
@@ -626,7 +626,7 @@ module Fluent
         end
       end
 
-      def get_flush_thread_lock_if_need
+      def acquire_flush_thread_lock_if_need
         need_thread_lock = actual_flush_thread_count > 1
         if need_thread_lock
           @flush_thread_mutex.synchronize { yield }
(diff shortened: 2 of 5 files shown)


daipom:
I have confirmed this solves #4080.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -117,15 +116,45 @@ def has_time_format?(str)
...
+          break unless File.exist?(path)
           i += 1
         end
+        lock_if_needed(path) do
+          # If multiple processes select the same path and another process
+          # entered this locking block first, the file should already exist
+          # and this process should retry to find new path.
+          raise FileAlreadyExist if File.exist?(path)
+          yield path
+        end
```
daipom, line 139: I just noticed that this logic perhaps must consider working on multiple threads, not only multiple processes.
I'll do additional tests and make the fix if necessary.

daipom: Test for fe65b4f37730e718ec23528142429f42d0961515.

Use the same way and the same environment as

* #4080#issuecomment-1455695378

Result: 

| workers | flush_thread_count |   percentage  |
|---------|--------------------|---------------|
|       1 |                 16 | 45.5% (10/22) |
|       3 |                  1 | 0.0% (0/51)   |

So the current fix is not enough.

daipom:
Test for  c95b52a.

Use the same way and the same environment as

* https://github.com/fluent/fluentd/issues/4080#issuecomment-1455695378

Result: 

| workers | flush_thread_count | percentage (1st) | percentage (2nd) |
|---------|--------------------|------------------|------------------|
|       1 |                 16 | 0.0% (0/111)     | 0.0% (0/42)      |
|       3 |                  1 | 0.0% (0/27)      | 0.0% (0/27)      |
|       3 |                  8 | 0.0% (0/267)     | 0.0% (0/29)      |

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -61,34 +61,40 @@ def configure(conf)
...
     end
 
     def multi_workers_ready?
-      ### TODO: add hack to synchronize for multi workers
       true
     end
 
+    def start
+      super
+      extend WriteLocker
```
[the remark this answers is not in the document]
daipom: I'm not sure how I should handle these Ruby features...
What I thought was as follows.

* I just wanted to clean up the methods by combining some specific function methods into a module.
* The `WriteLocker` module depends on the instance members of `SecondaryFileOutput`.
* So I felt uncomfortable with the form in which `SecondaryFileOutput` inherits `WriteLocker`.
  * I thought the dependency was reversed.

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -61,34 +61,40 @@ def configure(conf)
...
+      @write_mutex = Mutex.new
```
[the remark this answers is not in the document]
daipom: Understood! I will fix this!

test/command/test_cat.rb
```diff
@@ -83,6 +83,7 @@ def test_cat_json
   sub_test_case "msgpack" do
     def test_cat_secondary_file
       d = create_secondary_driver
+      d.instance_start
```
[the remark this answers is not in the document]
daipom: Hmm, even if we don't overwrite `start` method in `SecondaryFileOutput`, I wonder if it's really OK not to call `instance_start`.

Wouldn't it be correct to always call `instance_start` before calling the `instance` operation directly?

lib/fluent/plugin/out_secondary_file.rb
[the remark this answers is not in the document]
daipom: Hmm, should I just write the methods directly under the `SecondaryFileOutput`, not using `module`?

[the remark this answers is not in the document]
ashie: In this case I prefer not using module since there aren't so many code in this plugin for now.

BTW methods in `WriteLocker` might be useful for other output plugins, so place them just under `Fluent::Plugin::Output` might be better.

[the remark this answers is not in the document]
daipom: Indeed! I will try it!

lib/fluent/plugin/out_secondary_file.rb
```diff
@@ -61,34 +61,40 @@ def configure(conf)
 
       @dir_perm = system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION
       @file_perm = system_config.file_permission || Fluent::DEFAULT_FILE_PERMISSION
+      @need_worker_lock = system_config.workers > 1
+      @need_thread_lock = @primary_instance.buffer_config.flush_thread_count > 1
```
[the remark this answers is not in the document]
daipom: It seems that we don't need to consider it.
`Secondary`'s primary must have its `buffer_config`.

https://github.com/fluent/fluentd/blob/v1.15.3/lib/fluent/plugin/output.rb#L285-L301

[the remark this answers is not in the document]
daipom: In addition, even if `buffer` section is not defined in the primary, `buffer_config` will be initialized by its default values.

lib/fluent/plugin/out_secondary_file.rb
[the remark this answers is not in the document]
daipom: Fixed at ce32ac57a07121d69adf30d3b543ea11041ce613.

lib/fluent/plugin/out_secondary_file.rb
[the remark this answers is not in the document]
daipom: Fixed at ce32ac57a07121d69adf30d3b543ea11041ce613.

daipom:
Test for ce32ac57a07121d69adf30d3b543ea11041ce613.

Use the same way (`append false`) and the same environment as

* https://github.com/fluent/fluentd/issues/4080#issuecomment-1455695378

Result: 

| workers | flush_thread_count | percentage|
|---------|--------------------|------------------|
|       3 |                  8 | 0.0% (0/21)    |

I plan to do more tests at the end of this PR.

daipom:
I will add some unit tests for the lock feature, although it is hard to test race condition strictly.

daipom:
@ashie Could you please check if the direction of this fix is OK?

lib/fluent/plugin/output.rb
```diff
@@ -597,6 +598,47 @@ def terminate
...
+
+      def get_worker_lock_if_need(name)
+        need_worker_lock = system_config.workers > 1
+        unless need_worker_lock
+          yield
+          return
+        end
+        acquire_worker_lock(name) do
+          yield
+        end
```
ashie, line 628:
```suggestion
        if need_worker_lock
          acquire_worker_lock(name) { yield }
        else
          yield
        end
```

lib/fluent/plugin/output.rb
```diff
@@ -597,6 +598,47 @@ def terminate
...
+
+      def get_flush_thread_lock_if_need
+        need_thread_lock = actual_flush_thread_count > 1
+        unless need_thread_lock
+          yield
+          return
+        end
+        @flush_thread_mutex.synchronize do
+          yield
+        end
```
ashie, line 639:
```suggestion
        if need_thread_lock
          @flush_thread_mutex.synchronize { yeild }
        else
          yield
        end
```

lib/fluent/plugin/output.rb
```diff
@@ -597,6 +598,47 @@ def terminate
...
+      # threads in the same process.
+      def lock_if_need(worker_lock_name)
+        get_worker_lock_if_need(worker_lock_name) do
+          get_flush_thread_lock_if_need do
+            yield
+          end
+        end
+      end
+
+      def get_worker_lock_if_need(name)
```
ashie, line 620: The prefix of the method name seems inconsistent (`get` vs `acquire`).

lib/fluent/plugin/output.rb
daipom: Cool! Thanks for your advice!

test/command/test_cat.rb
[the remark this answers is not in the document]
daipom: Indeed! I will remove this diff from this PR for now.

lib/fluent/plugin/output.rb
daipom: Indeed. I will use `acquire`.

daipom: Fixed!

test/command/test_cat.rb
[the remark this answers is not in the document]
daipom: Removed!

daipom:
I will test this issue again, and add some tests for this new lock feature.

daipom:
Test for edcbd4b6a9cf2e83c2c88e34e7a103f3a4786314.

Use the same way (`append false`) and the same environment as

* https://github.com/fluent/fluentd/issues/4080#issuecomment-1455695378

Result: 

| workers | flush_thread_count | percentage (1st) | percentage (2nd) |
|---------|--------------------|------------------|------------------|
|       1 |                 16 | 0.0% (0/24)      | 0.0% (0/39)      |
|       3 |                  1 | 0.0% (0/27)      | 0.0% (0/21)      |
|       3 |                  8 | 0.0% (0/36)      | 0.0% (0/33)      |

daipom:
> I will test this issue again, and add some tests for this new lock feature.

Done.

lib/fluent/plugin/output.rb
```diff
@@ -597,6 +598,43 @@ def terminate
...
+        return @buffer_config.flush_thread_count unless @as_secondary
+        @primary_instance.buffer_config.flush_thread_count
+      end
+
+      # Run the passed block in the appropriate lock condition for multiple threads and workers.
+      # The lock between workers is made for every `worker_lock_name`.
+      # (For multiple workers, the lock is shared if `worker_lock_name` is the same value).
+      # For multiple threads, `worker_lock_name` is not used, and the lock is shared by all
+      # threads in the same process.
+      def acquire_lock_if_need(worker_lock_name)
```
ashie, line 612: I feel it's not Rubyish naming (original `acquire_worker_lock` is so too though).
`synchronize_foobar` might be better but I don't have good concrete idea for `foobar` to express it.
(`worker_and_flush_thread` is too redundant.)

@kou Do you have any idea for it?

kou: `synchronize_path(path)`?
Or `guard_path(path)`?

daipom: Thanks for your advice!

I have one point of concern.
Since the argument `path` is not used in the case of multiple threads, isn't `ooxx_path` misleadable?

daipom: How about simply `synchronize(path)`?

kou: > Since the argument `path` is not used in the case of multiple threads, isn't `ooxx_path` misleadable?

Sorry but I can't understand the point. `synchronize_path(path) {...}` ensures `path` processable only by the current thread in the current process. `synchronize_path` does or does not something against `path` is details of `synchronize_path`. `synchronize_path` users should not care about it.

BTW, another candidate `use_path(path)`.

> How about simply `synchronize(worker_lock_path)`?

I object it. Because `Output`'s main responsibility isn't synchronization. So only `synchronize` is ambitious. For example, `Output` has two `Mutex`s. It means that there are at least two synchronization units in `Output`.

`Mutex#synchronize` is OK because `Mutex`'s main responsibility is one synchronization.

daipom: Thanks for your detailed explanation! I see! I understand!

Although this does NOT provide exclusive processing for each `path` (but for all threads), it after all means that this ensures the `path` processable only be the current thread.
So we don't need to concern that the `path` is not used in the case of multiple threads.

https://github.com/fluent/fluentd/blob/f23e7e689c813a3040610e8f6b5958be3a369c28/lib/fluent/plugin/output.rb#L629-L636

daipom: Fixed! Thanks for your advice! @ashie @kou !

daipom:
By the way, might `Squash merge` be better when merging this?

test/plugin/test_output.rb
```diff
@@ -1072,4 +1072,263 @@ def invoke_slow_flush_log_threshold_test(i)
...
+      }
+    )
+    test "actual_flush_thread_count" do |data|
+      o = create_output(data[:output_type])
+      o.configure(data[:config])
+      assert_equal data[:expected], o.actual_flush_thread_count
+    end
+
+    data(
+      "Buffered with singile thread",
```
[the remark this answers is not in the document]
daipom: Thanks! Fixed!

ashie:
Thanks!

daipom:
Thanks for your review! @ashie @kou !

PR #4082, opened 2023-03-07
Use driver for formatter tests

daipom:
None. This is the following fix to

* #4074 

* #4074

In this PR, some tests passing `Hash` directly to `Fluent::Plugin::Base::configure()` are fixed so that passing `Fluent::Config::Element`.

However, a part of those tests should use the test-driver, and then we can passing `Hash` to the driver safely.

**Docs Changes**:
None.

**Release Note**: 
None.

### DIFF
--- a/test/test_formatter.rb
+++ b/test/test_formatter.rb
@@ -13,10 +13,6 @@ module FormatterTest
     {'message' => 'awesome', 'greeting' => 'hello'}
   end
 
-  def config_element_in_params(params)
-    config_element('test', '', params)
-  end
-
   class BaseFormatterTest < ::Test::Unit::TestCase
     include FormatterTest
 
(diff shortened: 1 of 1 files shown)


daipom:
@abetomo Could you please check this?

abetomo:
I should have used `Fluent::Test::FormatterTestDriver`.
I was not aware of the existence of `Fluent::Test::FormatterTestDriver`.

abetomo (approved):
Thanks!

daipom:
@abetomo Thanks for your review!

> I should have used `Fluent::Test::FormatterTestDriver`. I was not aware of the existence of `Fluent::Test::FormatterTestDriver`.

It's no problem!

FYI: The document of `TestDriver` is https://docs.fluentd.org/plugin-development/plugin-test-code

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4085, opened 2023-03-09
Use driver for RegexpParser tests

abetomo:

( https://github.com/fluent/fluentd/pull/4066#discussion_r1118388855 )

I didn't know why it was commented out, but it looked good on `Fluent::Test::Driver::Parser.new(...)`.
https://github.com/fluent/fluentd/pull/1286

**Docs Changes**:

None

**Release Note**: 

None


daipom (approved):
Thanks! We should fix tests like this.

Certainly, the intent of the #1286 fix is unclear.

`Fluent::Test::Driver::Parser::configure()` calls `instance.configure()`, so I think the old code is meaningless.

I have confirmed the test succeeds even on #1286 with this fix.

It looks good to me.

ashie:
Thanks!

abetomo:
Thanks for the merge!

PR #4086, opened 2023-03-09
Refactor Superviser.load_config()

daipom:
This PR is the following to

* #4065 

It does not load the config anymore since

* #2709

So it should simply return the config for `ServerEngine`.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed. (This fixes only internal specifications)

### DIFF
--- a/lib/fluent/daemon.rb
+++ b/lib/fluent/daemon.rb
@@ -9,7 +9,5 @@ require 'fluent/supervisor'
 
 server_module = Fluent.const_get(ARGV[0])
 worker_module = Fluent.const_get(ARGV[1])
-# it doesn't call ARGV in block because when reloading config, params will be initialized and then it can't use previous config.
-config_path = ARGV[2]
-params = JSON.parse(ARGV[3])
-ServerEngine::Daemon.run_server(server_module, worker_module) { Fluent::Supervisor.load_config(config_path, params) }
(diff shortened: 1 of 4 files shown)


lib/fluent/daemon.rb
```diff
@@ -9,7 +9,5 @@
 
 server_module = Fluent.const_get(ARGV[0])
 worker_module = Fluent.const_get(ARGV[1])
-# it doesn't call ARGV in block because when reloading config, params will be initialized and then it can't use previous config.
-config_path = ARGV[2]
-params = JSON.parse(ARGV[3])
-ServerEngine::Daemon.run_server(server_module, worker_module) { Fluent::Supervisor.load_config(config_path, params) }
+params = JSON.parse(ARGV[2])
```
ashie, line 12: Memo: daemon.rb is used here and original `ARGV[2]`(`config_path`) is removed in this commit:
https://github.com/fluent/fluentd/blob/041af39f401b26002c07a299ad7c85346909ce7c/lib/fluent/supervisor.rb#L454-L458

daipom: Yes, sorry for the lack of explanation.

ashie (approved):

daipom:
Thanks for your review!

PR #4090, opened 2023-03-12
Fix command option `--quiet`

daipom:
This fixes a bug related to #4037

Sorry, there was a bug in my following fix. 

* #4064
* 8e9f46ae84e6fd603cf46c63cdef6f3b2334cca3

This PR fixes it.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -166,7 +166,7 @@ op.on('-v', '--verbose', "increase verbose level (-v: debug, -vv: trace)", TrueC
 op.on('-q', '--quiet', "decrease verbose level (-q: warn, -qq: error)", TrueClass) {|b|
   return unless b
   cur_level = cmd_opts.fetch(:log_level, default_opts[:log_level])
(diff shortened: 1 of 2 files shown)


ashie:
After #4064 is landed, CI on Windows have apparently failed more often and rarely succeeded.
Especially, tests after sub test case `log_level by command line option` often failed.
e.g.)
* https://github.com/fluent/fluentd/actions/runs/4400546161/jobs/7705942975
* https://github.com/fluent/fluentd/actions/runs/4400546161/jobs/7705943318

Although I don't think #4064 itself doesn't have problem, it seems that it triggers something.
When I remove the sub test case, CI becomes stable again:

* https://github.com/ashie/fluentd/actions/runs/4401985736/jobs/7708692528

ashie:
CI on Windows are failed but it's not related with this PR as I described above.
So I merge this.

We need to investigate the CI failure later.

daipom:
Thank for your review!

PR #4092, opened 2023-03-12
SocketManager: Use new feature to select path automatically

daipom:
Use ServerEngine's new feature:

https://github.com/treasure-data/serverengine/pull/143

On Windows, this prevents SocketManager from wrongly selecting an unavailable port, such as a port in excluded port range.

Fixes https://github.com/treasure-data/serverengine/issues/140
(Sorry, I made the issue in ServerEngine, and not made it here.)

On Windows, `ServerEngine::SocketManager::generate_path` sometimes generates an unavailable port.
This makes Fluentd fail to start.

To fix this, I made the fix for ServerEngine:

* https://github.com/treasure-data/serverengine/pull/143

This is a fix on Fluentd's side.

**Docs Changes**:
Not needed.

**Release Note**: 
Fix bug that Fluentd sometimes tries to use an unavailable port and fails to start on Windows.

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("msgpack", [">= 1.3.1", "< 2.0.0"])
   gem.add_runtime_dependency("yajl-ruby", ["~> 1.0"])
   gem.add_runtime_dependency("cool.io", [">= 1.4.5", "< 2.0.0"])
-  gem.add_runtime_dependency("serverengine", [">= 2.3.0", "< 3.0.0"])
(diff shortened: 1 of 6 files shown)


daipom:
Waiting

* https://github.com/treasure-data/serverengine/pull/143

daipom:
https://github.com/treasure-data/serverengine/issues/140#issuecomment-1423448918

> Fluentd shouldn't open shared port if it's not configured using multi workers.

We should fix this too, but it seems difficult to do it before the next release...
I'll make the issue later.

daipom:
The currently failing tests are not related to this issue.

daipom:
> Rebased.

Thanks!

ashie (approved):
LGTM

daipom:
Thanks for your review!

PR #4096, opened 2023-03-13
Make `TestFluentdCommand` tests stable on Windows

daipom:
**Which issue(s) this PR fixes**: 
Fixes #4095

**What this PR does / why we need it**: 
Make`TestFluentdCommand` tests stable on Windows.
(Please see #4095 for details.)

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

PR #4097, opened 2023-03-14
CI: Reduce log level on Windows

daipom:
None

For some reason, the log level is Debug only on Windows.
This makes it harder to find failing tests.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

### DIFF
--- a/.github/workflows/windows-test.yaml
+++ b/.github/workflows/windows-test.yaml
@@ -46,4 +46,4 @@ jobs:
       - name: Install dependencies
         run: ridk exec bundle install
       - name: Run tests
-        run: bundle exec rake test TESTOPTS=-v ${{ matrix.ruby-lib-opt }}
+        run: bundle exec rake test ${{ matrix.ruby-lib-opt }}


ashie:
IMHO I want to leave it as is for a while.

On Windows some tests often become unstable due to other test's state although it shouldn't be.
When I investigate them, I often check which tests are executed before failed tests.
In fact I recently noticed a problem which is caused by a past unrelated pull request.

* https://github.com/fluent/fluentd/pull/4090#issuecomment-1465643573
* https://github.com/fluent/fluentd/pull/4096

It might be hard to find if `-v` isn't specified:

daipom:
I see! Thanks for the explanation! 
Now I understand the benefits of this setting.
I close this.

ashie:
BTW for finding failed test I search ` Failure: ` or `  Error: ` (including former & later spaces) with case sensitive.
It makes easy to find failed tests even if `-v` is added or not.

daipom:
Thanks!!
Do you use `View raw logs` and then search the words?

ashie:
I use both Web UI and raw logs.

daipom:
I see! Thanks!

PR #4106, opened 2023-03-21
Make sure JSON parser returns Hash

### LINKED ISSUE #4100: JSON parser plugin can sometimes output records as strings rather than `Hash`
### Describe the bug

### To Reproduce

```
    <filter gearset.*>
      @type parser
      <parse>
        @type json
        json_parser oj
      </parse>
      replace_invalid_sequence true
      key_name message
    </filter>
```

### Expected behavior

### Your Environment

```markdown
```

### Your Configuration

```apache
<source>
      @type tail
      read_from_head true
      tag gearset.x
      <parse>
        @type cri
      </parse>
    </source>

    <filter gearset.*>
      @type parser
      <parse>
        @type json
        json_parser oj
      </parse>
      replace_invalid_sequence true
      key_name message
    </filter>

    <filter gearset.*>
      @type prometheus
      <metric>
        name na
        type counter
        desc NA
        <labels>
          tag ${tag}
        </labels>
      </metric>
    </filter>
```

### Your Error Log

```shell
```

### Additional context

_No response_

daipom:
Fixes #4100

Make sure JSON parser returns `Hash`.
It is wrong for Parser to return a record that is not Hash.
Subsequent processing may result in errors.

The result of parsing JSON can be a value other than Hash.
Originally, the data represented by JSON was not limited to Hash.

* https://www.rfc-editor.org/rfc/rfc8259#section-2

The current implementation of `JSONParser` does not appear to have sufficiently considered this.

We need to consider carefully the specification of the case where raw data can not be parsed as `Hash`.

The currently intended usecases appears to be `Hash` or `Array` of `Hash`.
We should consider only these 2 patterns as parsable for now.

`in_http` assumed that Parser could return Array.
Therefore, it needed to be fixed along with this.

`MessagePackParser` appears to have the same issue. (Not fixed in this PR.)

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

# Difference

## Case: No subsequent processing

Config

```xml
<source>
  @type tcp
  tag test.tcp
  <parse>
    @type json
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

Operation

```console
$ netcat 0.0.0.0 5170
[{"k":"v"}, {"k2":"v2"}]

```

Result before this fix

```
2023-03-21 23:30:30.623063804 +0900 test.tcp: [{"k":"v"},{"k2":"v2"}]
```

Result after this fix

```
2023-03-21 23:29:20.147728222 +0900 test.tcp: {"k":"v"}
2023-03-21 23:29:20.147754433 +0900 test.tcp: {"k2":"v2"}
```

## Case: With subsequent filter processing

Config

```xml
<source>
  @type tcp
  tag test.tcp
  <parse>
    @type json
  </parse>
</source>

<filter test.**>
  @type record_transformer
  enable_ruby true
  <record>
    class ${record.class}
  </record>
</filter>

<match test.**>
  @type stdout
</match>
```

Operation

```console
$ netcat 0.0.0.0 5170
[{"k":"v"}, {"k2":"v2"}]

```

Result before this fix

```
2023-03-21 23:30:51 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for [{\"k\"=>\"v\"}, {\"k2\"=>\"v2\"}]:Array" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.tcp" time=2023-03-21 23:30:51.040692881 +0900 record=[{"k"=>"v"}, {"k2"=>"v2"}]
```

Result after this fix

```
2023-03-21 23:29:55.783183244 +0900 test.tcp: {"k":"v","class":"Hash"}
2023-03-21 23:29:55.783215076 +0900 test.tcp: {"k2":"v2","class":"Hash"}
```

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -203,54 +203,24 @@ module Fluent::Plugin
       begin
         path = path_info[1..-1]  # remove /
         tag = path.split('/').join('.')
-        record_time, record = parse_params(params)
 
-        # Skip nil record
-        if record.nil?
-          log.debug { "incoming event is invalid: path=#{path_info} params=#{params.to_json}" }
-          if @respond_with_empty_img
-            return RESPONSE_IMG
-          else
-            if @use_204_response
-              return RESPONSE_204
-            else
-              return RESPONSE_200
-            end
+        mes = Fluent::MultiEventStream.new
+        parse_params(params) do |record_time, record|
+          if record.nil?
+            log.debug { "incoming event is invalid: path=#{path_info} params=#{params.to_json}" }
+            next
           end
-        end
 
-        mes = nil
-        # Support batched requests
-        if record.is_a?(Array)
-          mes = Fluent::MultiEventStream.new
-          record.each do |single_record|
-            add_params_to_record(single_record, params)
-
-            if param_time = params['time']
-              param_time = param_time.to_f
-              single_time = param_time.zero? ? Fluent::EventTime.now : @float_time_parser.parse(param_time)
-            elsif @custom_parser
-              single_time = @custom_parser.parse_time(single_record)
-              single_time, single_record = @custom_parser.convert_values(single_time, single_record)
-            else
-              single_time = convert_time_field(single_record)
-            end
-
-            mes.add(single_time, single_record)
-          end
-        else
           add_params_to_record(record, params)
 
           time = if param_time = params['time']
                    param_time = param_time.to_f
                    param_time.zero? ? Fluent::EventTime.now : @float_time_parser.parse(param_time)
                  else
-                   if record_time.nil?
-                     convert_time_field(record)
-                   else
-                     record_time
-                   end
+                   record_time.nil? ? convert_time_field(record) : record_time
                  end
+
+          mes.add(time, record)
         end
       rescue => e
         if @dump_error_log
@@ -261,11 +231,7 @@ module Fluent::Plugin
 
       # TODO server error
       begin
-        if mes
-          router.emit_stream(tag, mes)
-        else
-          router.emit(tag, time, record)
-        end
+        router.emit_stream(tag, mes) unless mes.empty?
       rescue => e
         if @dump_error_log
           log.error "failed to emit data", error: e
@@ -308,20 +274,27 @@ module Fluent::Plugin
     def parse_params_default(params)
       if msgpack = params['msgpack']
         @parser_msgpack.parse(msgpack) do |_time, record|
-          return nil, record
+          if record.is_a?(Array)
+            # TODO: Temporarily supporting this case for compatibility.
+            # We should not consider this case here.
+            # We should fix MessagePackParser so that it doesn't return Array.
+            record.each do |single_record|
+              yield nil, single_record
+            end
+          else
+            yield nil, record
+          end
         end
       elsif js = params['json']
         @parser_json.parse(js) do |_time, record|
-          return nil, record
+          yield nil, record
         end
       elsif ndjson = params['ndjson']
-        events = []
         ndjson.split(/\r?\n/).each do |js|
           @parser_json.parse(js) do |_time, record|
-            events.push(record)
+            yield nil, record
           end
         end
-        return nil, events
       else
         raise "'json', 'ndjson' or 'msgpack' parameter is required"
       end
@@ -329,10 +302,9 @@ module Fluent::Plugin
 
     def parse_params_with_parser(params)
       if content = params[EVENT_RECORD_PARAMETER]
-        @custom_parser.parse(content) { |time, record|
-          raise "Received event is not #{@format_name}: #{content}" if record.nil?
-          return time, record
-        }
+        @custom_parser.parse(content) do |time, record|
+          yield time, record
+        end
       else
         raise "'#{EVENT_RECORD_PARAMETER}' parameter is required"
       end
--- a/test/plugin/test_parser_json.rb
+++ b/test/plugin/test_parser_json.rb
@@ -135,4 +135,60 @@ class JsonParserTest < ::Test::Unit::TestCase
       end
     end
   end
+
+  sub_test_case "various record pattern" do
+    data("Only string", { record: '"message"', expected: [nil] }, keep: true)
+    data("Only string without quotation", { record: "message", expected: [nil] }, keep: true)
+    data("Only number", { record: "0", expected: [nil] }, keep: true)
+    data(
+      "Array of Hash",
+      {
+        record: '[{"k1": 1}, {"k2": 2}]',
+        expected: [{"k1" => 1}, {"k2" => 2}]
+      },
+      keep: true,
+    )
+    data(
+      "Array of both Hash and invalid",
+      {
+        record: '[{"k1": 1}, "string", {"k2": 2}, 0]',
+        expected: [{"k1" => 1}, nil, {"k2" => 2}, nil]
+      },
+      keep: true,
+    )
+    data(
+      "Array of all invalid",
+      {
+        record: '["string", 0, [{"k": 0}]]',
+        expected: [nil, nil, nil]
+      },
+      keep: true,
+    )
+    def test_oj(data)
+      i = 0
+      @parser.configure('json_parser' => "oj")
+      @parser.instance.parse(data[:record]) { |time, record|
+        assert_equal(data[:expected][i], record)
+        i += 1
+      }
+    end
+
+    def test_yajl(data)
+      i = 0
+      @parser.configure('json_parser' => "yajl")
+      @parser.instance.parse(data[:record]) { |time, record|
+        assert_equal(data[:expected][i], record)
+        i += 1
+      }
+    end
+
+    def test_json(json)
+      i = 0
+      @parser.configure('json_parser' => "json")
+      @parser.instance.parse(data[:record]) { |time, record|
+        assert_equal(data[:expected][i], record)
+        i += 1
+      }
+    end
+  end
 end
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -79,6 +79,10 @@ module Fluent::Plugin
                 end
             @accessor.delete(record) if @remove_key_name_field
             r = handle_parsed(tag, record, t, values)
+            # Note: If the parser returns multiple records from one raw_value,
+            # this returns only the first one record.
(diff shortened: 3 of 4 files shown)


daipom:
Example

```xml
<source>
  @type sample
  tag test.hoge
  sample {"message": "{\"k\":\"v\"}"}
</source>

<source>
  @type sample
  tag test.fuga
  sample {"message": "\"HELLO\""}
</source>

<filter test.**>
  @type parser
  key_name message
  <parse>
    @type json
  </parse>
</filter>

<filter test.**>
  @type record_transformer
  enable_ruby true
  <record>
    class ${record.class}
  </record>
</filter>

<match test.**>
  @type stdout
</match>
```

Before

```
2023-03-17 11:18:09 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for \"HELLO\":String" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.fuga" time=2023-03-17 11:18:09.063985763 +0900 record="HELLO"
2023-03-17 11:18:09.064692451 +0900 test.hoge: {"k":"v","class":"Hash"}
2023-03-17 11:18:10 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for \"HELLO\":String" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.fuga" time=2023-03-17 11:18:10.066951425 +0900 record="HELLO"
2023-03-17 11:18:10.067684144 +0900 test.hoge: {"k":"v","class":"Hash"}
```

After

```
2023-03-21 11:08:52 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data '\"HELLO\"'" location=nil tag="test.fuga" time=2023-03-21 11:08:52.029211794 +0900 record={"message"=>"\"HELLO\""}
2023-03-21 11:08:52.030661373 +0900 test.hoge: {"k":"v"}
2023-03-21 11:08:53 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data '\"HELLO\"'" location=nil tag="test.fuga" time=2023-03-21 11:08:53.031778958 +0900 record={"message"=>"\"HELLO\""}
2023-03-21 11:08:53.032463736 +0900 test.hoge: {"k":"v"}
2023-03-21 11:08:54 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data '\"HELLO\"'" location=nil tag="test.fuga" time=2023-03-21 11:08:54.033762073 +0900 record={"message"=>"\"HELLO\""}
2023-03-21 11:08:54.034318296 +0900 test.hoge: {"k":"v"}
^C2023-03-21 11:08:54 +0900 [info]: Received graceful stop
2023-03-21 11:08:55 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data '\"HELLO\"'" location=nil tag="test.fuga" time=2023-03-21 11:08:55.035631987 +0900 record={"message"=>"\"HELLO\""}
2023-03-21 11:08:55.036183397 +0900 test.hoge: {"k":"v"}
```

daipom:
Additional confirming of the difference of the behavior,

```xml
<source>
  @type sample
  tag test.hash
  sample {"message": "{\"k\":\"v\"}"}
</source>

<source>
  @type sample
  tag test.string
  sample {"message": "\"HELLO\""}
</source>

<source>
  @type sample
  tag test.invalid
  sample {"message": "HELLO"}
</source>

<source>
  @type sample
  tag test.array
</source>

<filter test.**>
  @type parser
  key_name message
  <parse>
    @type json
  </parse>
</filter>

<filter test.**>
  @type record_transformer
  enable_ruby true
  <record>
    class ${record.class}
  </record>
</filter>

<match test.**>
  @type stdout
</match>
```

# Before this fix

Some patterns result in errors in subsequent processing.

```
2023-03-21 23:50:24.068859931 +0900 test.hash: {"k":"v","class":"Hash"}
2023-03-21 23:50:25 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for [{\"k\"=>\"v\"}, {\"k2\"=>\"v2\"}]:Array" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.array" time=2023-03-21 23:50:25.070141860 +0900 record=[{"k"=>"v"}, {"k2"=>"v2"}]
2023-03-21 23:50:25 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data 'HELLO'" location=nil tag="test.invalid" time=2023-03-21 23:50:25.071323231 +0900 record={"message"=>"HELLO"}
2023-03-21 23:50:25 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for \"HELLO\":String" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.string" time=2023-03-21 23:50:25.072303009 +0900 record="HELLO"
```

# After this fix

Correctly, `ParserError` occurs in `filter_parser`.

I left a comment in the code about `filter_parser` not being able to parse multiple records from one raw record.
(This is a limitation of the current specification.)

```
2023-03-21 23:43:25.061681273 +0900 test.array: {"k":"v","class":"Hash"}
2023-03-21 23:43:25 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data 'HELLO'" location=nil tag="test.invalid" time=2023-03-21 23:43:25.062907066 +0900 record={"message"=>"HELLO"}
2023-03-21 23:43:25.063493027 +0900 test.hash: {"k":"v","class":"Hash"}
2023-03-21 23:43:25 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="pattern not matched with data '\"HELLO\"'" location=nil tag="test.string" time=2023-03-21 23:43:25.063894687 +0900 record={"message"=>"\"HELLO\""}

```

lib/fluent/plugin/in_http.rb
```diff
@@ -203,54 +203,24 @@ def on_request(path_info, params)
...
-          mes = Fluent::MultiEventStream.new
-          record.each do |single_record|
-            add_params_to_record(single_record, params)
-
-            if param_time = params['time']
-              param_time = param_time.to_f
-              single_time = param_time.zero? ? Fluent::EventTime.now : @float_time_parser.parse(param_time)
-            elsif @custom_parser
-              single_time = @custom_parser.parse_time(single_record)
-              single_time, single_record = @custom_parser.convert_values(single_time, single_record)
```
daipom, line 234: We don't need this anymore.

Why we needed this before:

* When the parser returns `Array`, the parser can't parse the time because the data is `Array`.
* So we needed that parsing process again to separate records here.

After this fix, the following logic can always work.
So we don't need this logic anymore.

https://github.com/fluent/fluentd/pull/4106/files#diff-73e8d1657dbede482feb06b87014a3656632e7e507cd48fd7be2440aa848309eR94-R95

lib/fluent/plugin/in_http.rb
```diff
@@ -308,31 +274,37 @@ def on_server_connect(conn)
...
-        return nil, events
       else
         raise "'json', 'ndjson' or 'msgpack' parameter is required"
       end
     end
 
     def parse_params_with_parser(params)
       if content = params[EVENT_RECORD_PARAMETER]
-        @custom_parser.parse(content) { |time, record|
-          raise "Received event is not #{@format_name}: #{content}" if record.nil?
```
daipom, line 333: There did not seem to be any particular reason for `raise` only here.
We should align with `parse_params_default`.

test/plugin/test_parser_json.rb
```diff
@@ -135,4 +135,60 @@ def test_yajl_parse_io_with_buffer_smaller_than_input
       end
     end
   end
+
+  sub_test_case "various record pattern" do
+    data("Only string", { record: '"message"', expected: [nil] }, keep: true)
+    data("Only string without quaot", { record: "message", expected: [nil] }, keep: true)
```
abetomo, line 141: `quaot`
Is it a typo for "quote"?

daipom: Thanks! This is a typo! Now I fixed this!

daipom:
I confirmed the failing tests. They seem to be unstable tests and not related to this fix.

lib/fluent/plugin/in_http.rb
```diff
@@ -308,31 +274,37 @@ def on_server_connect(conn)
     def parse_params_default(params)
       if msgpack = params['msgpack']
         @parser_msgpack.parse(msgpack) do |_time, record|
-          return nil, record
+          if record.is_a?(Array)
+            # TODO: Temporarily supporting this case for compatibility.
+            # We should not consider this case here.
+            # We should fix MessagePackParser so that it doesn't return Array.
```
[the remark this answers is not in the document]
daipom: I thought it should be summarized as fixing the JSON parser.
Since there is no time before the next release, I also wanted to keep the scope of impact as small as possible.
About `in_http`, I had no choice but to fix it in this PR.

If it looks like I should fix it in this PR, I will try today or tomorrow.
Or I can try to fix it in another PR after this PR is merged.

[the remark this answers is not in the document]
ashie: It should be done in this PR instead of adding such workaround, since it's considered as a same problem and sometimes such workaround lives long unexpectedly.

I’ll postpone merging this to v1.17.0.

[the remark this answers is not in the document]
daipom: Thanks for your review. I Understood.
My understanding was that the existing workaround was minimized here, not adding a new workaround.
But certainly, if it is not fixed entirely, it may lead to unnecessary problems.
Especially for `MessagePackParser` as `@custom_parser`.

daipom:
TODO: Fix other parsers.

EnginnerDog:
Uhhh hellooo? This PR should unexist.

daipom:
Sorry for the delay. I'm going to fix this for the next release.

daipom:
I have add new PR for this

* #4474

PR #4116, opened 2023-03-29
Revive issue auto closer

ashie:
N/A

Although issue auto closer is disabled at d005002272e7e5b166d3150a5e544274e102c742 without discussion, I think we still need it at least for `waiting-for-users` and `stable` labels.

We need to revive it and tune which issues should be closed automatically.

**Docs Changes**:
None.

**Release Note**: 
None.


ashie:
Untriaged issues should be closed automatically.
We should tune it in this PR.

daipom:
I basically agree.

I have a question about automatically stale or closed items that are not labeled at all, but I think it is a good idea to automatically stale and close items that are labeled with `waiting-for-users`, for example.

Even if we don't decide in detail now, we can tune these things later.

ashie:
I introduced `waiting-for-triage` to fluent-plugin-kafka repository for this purpose:
https://github.com/fluent/fluent-plugin-kafka/commit/85538f6f92a13ab787d27075bcf80465338fc93c

daipom:
In `fluent-plugin-remote_syslog`, automatically staling and closing issues with `answered` label as follows.

https://github.com/fluent-plugins-nursery/fluent-plugin-remote_syslog/blob/master/.github/workflows/close-answered-issues.yml

daipom:
> I introduced `waiting-for-triage` to fluent-plugin-kafka repository for this purpose: [fluent/fluent-plugin-kafka@85538f6](https://github.com/fluent/fluent-plugin-kafka/commit/85538f6f92a13ab787d27075bcf80465338fc93c)

Oh, automatically adding `waiting-for-triage` at opening a new issue, I see! This looks nice idea!

ashie:
I've added same settings for fluent-plugin-kafka, and shorten `days-before-stale` to `30` because `90` is too long for answered issues.

.github/workflows/stale-actions.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v8
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-stale: 30
+        days-before-close: 30
```
daipom, line 14: I think `days-before-close` can be shorter, such as `7`.

daipom (approved):
LGTM. I agree.

.github/workflows/stale-actions.yml
ashie: I agree with you, I've fixed it.

PR #4123, opened 2023-04-03
Update security policy

ashie:
N/A

We've enabled `Private vulnerability reporting` feature of GitHub: https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability

Using it is desired instead of sending email to each developers.

**Docs Changes**:
N/A

**Release Note**: 
N/A

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,15 +4,11 @@
 
 | Version   | Supported          |
 | -------   | ------------------ |
-| 1.14.x    | :white_check_mark: |
-| <= 1.13.x | :x:                |
+| 1.16.x    | :white_check_mark: |
+| 1.15.x    | :white_check_mark: |
+| <= 1.14.x | :x:                |
 
 ## Reporting a Vulnerability
 
-Please contact to current active maintainers. (in alphabetical order)
-
-* ashie@clear-code.com
-* fujimoto@clear-code.com
-* hatake@calyptia.com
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -4,15 +4,11 @@
 
 | Version   | Supported          |
 | -------   | ------------------ |
-| 1.14.x    | :white_check_mark: |
-| <= 1.13.x | :x:                |
+| 1.16.x    | :white_check_mark: |
+| 1.15.x    | :white_check_mark: |
```
ashie, line 8: I'm planning to release fluentd v1.15.4 & td-agent v4.4.3 with some backport patches, so 1.15.x is still supported.

cosmo0920: Can we add documentation for supporting strategies/periods for multiple versions?
To stabilize maintenance process is needed, I guess.

e.g.)
When the new version v1.16 is released, the previous series should be supported after 6 months later or something like that.

kenhys: Yes, it should be clarified later. Anyway, main scope of this PR is updating contact address.

kenhys (approved):
LGTM.

SECURITY.md
ashie: We'll clarify it after LTS release is on rail.

PR #4117, opened 2023-03-29
in_udp: add test of message_length_limit

daipom:
**Which issue(s) this PR fixes**: 
None.

**What this PR does / why we need it**: 
Add test of `message_length_limit` for `in_udp` because there was no test for this option.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

Issue #3739, opened 2022-05-11
Hope the in_tail's option "max_line_size" can support retain the head part of the log


wangchaoforever:
### Is your feature request related to a problem? Please describe.

When I collect logs which has a big  single line log. It cause OOM KILL fluentd in in_tail stage. So I upgrade my fluentd to v1.14.4 to use the option "max_line_size". It works nice to skip the big log.But now, I have a need to retain the head 50000 length of the big single log.

### Describe the solution you'd like

I hope there is a option to control the option "max_line_size" to retain  the head length of the big single log.

### Describe alternatives you've considered

I don't want to skip the large log and hope to keep 50000 length. I have no idea to deal with it now.

PR #4131, opened 2023-04-07
Server helper: Suppress error of UDPServer over max_bytes on Windows

daipom:
None.
This fixes the problem revealed in the following PR.

* #4117

This PR just suppresses error message of UDPServer over max_bytes on Windows.

Suppressed error message on Windows:
(Please see https://github.com/fluent/fluentd/pull/4117#issuecomment-1488716289 for details)

> unexpected error in processing UDP data error_class=Errno::EMSGSIZE error="A message sent on a datagram socket was larger than the internal message buffer or some other network limit, or the buffer used to receive a datagram into was smaller than the datagram itself. - recvfrom(2)"

New info message on Windows:

> A received data was ignored since it was too large.

On Windows, `UDPSocket::recv` and `UDPSocket::recvfrom` can raise `Errno::EMSGSIZE` error when the receiving data size is larger than the `maxlen`, the value specified as the method's argument.
(On non-Windows, this error doesn't happen, and we can receive data that is cut off to the specified size.)

On Windows, we can't cut data off the same way as Non-Windows.
So the data is simply ignored on Windows.
It is not a big problem since cutting data off usually causes a failure of parsing and the data is eventually ignored.

We want information about what data is ignored, but it is difficult to know it.
That is an issue for the future.

* #3739

Ref

* `winsock` specification
  * https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recvfrom
  * https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2
* Ruby implementation
  * https://github.com/ruby/ruby/blob/e51014f9c05aa65cbf203442d37fef7c12390015/win32/win32.c#L3540-L3589
  * https://github.com/ruby/ruby/blob/e51014f9c05aa65cbf203442d37fef7c12390015/win32/win32.c#L3730-L3782

**Docs Changes**:
TODO. We should add a document about this Windows specification.

**Release Note**: 
Same as the title.

### DIFF
--- a/test/plugin_helper/test_server.rb
+++ b/test/plugin_helper/test_server.rb
@@ -29,6 +29,7 @@ class ServerPluginHelperTest < Test::Unit::TestCase
     ENV['SERVERENGINE_SOCKETMANAGER_PATH'] = @socket_manager_path.to_s
 
     @d = Dummy.new
+    @d.under_plugin_development = true
     @d.start
     @d.after_start
   end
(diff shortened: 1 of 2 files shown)


test/plugin_helper/test_server.rb
```diff
@@ -796,6 +797,51 @@ class Dummy < Fluent::Plugin::TestBase
     end
   end
 
+  sub_test_case "UDPServer" do
```
ashie, line 800: Since this `sub_test_case` focuses on `max_bytes`, it should be reflected in the `sub_test_case` name.
Otherwise it might be hard to recognize the difference with `sub_test_case '#server_create_udp'`.

daipom: Sorry for confusing you, but please let me explain this since perhaps you don't get the intent of this code.

Actually, this `sub_test_case` is for `EventHandler.UDPServer`.
I wanted to create a test that focused on it.
(I ended up using `server_create()` for initialization, though,,,)

https://github.com/fluent/fluentd/blob/206b46b91560177366371c1f05bbcf09ac006d86/lib/fluent/plugin_helper/server.rb#L518

So the test cases are `With sock` and `Without sock`.

> Since this `sub_test_case` focuses on `max_bytes`, it should be reflected in the `sub_test_case` name.

So this test is not focused on `max_bytes`, actually.
The data `Normal` is for the tests simply for `UDPServer.on_readable_without_sock` and `UDPServer.on_readable_with_sock`.
(`max_bytes` is specified just because it is a mandatory option.)

My intention is something like this.
But, certainly, it may be confusing.

Now, what do you think?

Would changing the name of the subtest like this improve this problem?

```
sub_test_case "UDPEventHandler::Server"
```

ashie: The design of `Fluent::PluginHelper::Server` doesn't seem intend to expose `EventHandler.UDPServer`, it just only intend to use via `server_create` or `server_create_udp`. In actual you don't use  `EventHandler.UDPServer` directly, you use it via `server_create` in this test. In my opinion users don't need to aware the implementation detail `EventHandler.UDPServer`, so that test don't need to be aware it too.

It seems that these tests should be placed under `sub_test_case '#server_create_udp'`.

daipom: Oh, I see. Thanks.

It means that there is not much benefit in trying to make unit tests of `EventHandler` right?

I will try to fix it that way.

ashie: > It means that there is not much benefit in trying to make unit tests of `EventHandler` right?

That's right.

daipom: Thanks for your advice!
Fixed!

ashie (approved):
Waiting CI...

daipom:
Thank for your review!

PR #4137, opened 2023-04-10
`in_tcp`: Add `message_length_limit` to drop large incoming data

daipom:
None.

Add the same feature to `in_tcp` as `message_length_limit` in `in_udp`, to drop too large incoming data.

**Docs Changes**:
TODO

**Release Note**: 
Same as the title.

**TODO**:

- [x] Add tests
- [ ] Add document

### DIFF
--- a/test/plugin/test_in_tcp.rb
+++ b/test/plugin/test_in_tcp.rb
@@ -255,4 +255,76 @@ class TcpInputTest < Test::Unit::TestCase
       assert_equal 'hello', event[2]['msg']
     end
   end
+
+  sub_test_case "message_length_limit" do
+    data("batch_emit", { extract: "" }, keep: true)
+    data("single_emit", { extract: "<extract>\ntag_key tag\n</extract>\n" }, keep: true)
+    test "drop records exceeding limit" do |data|
+      message_length_limit = 10
+      d = create_driver(base_config + %!
+        message_length_limit #{message_length_limit}
+        <parse>
+          @type none
+        </parse>
+        #{data[:extract]}
+      !)
+      d.run(expect_records: 2, timeout: 10) do
+        create_tcp_socket('127.0.0.1', @port) do |sock|
+          sock.send("a" * message_length_limit + "\n", 0)
+          sock.send("b" * (message_length_limit + 1) + "\n", 0)
+          sock.send("c" * (message_length_limit - 1) + "\n", 0)
+        end
+      end
+
+      expected_records = [
+        "a" * message_length_limit,
+        "c" * (message_length_limit - 1)
+      ]
+      actual_records = d.events.collect do |event|
+        event[2]["message"]
+      end
+
+      assert_equal expected_records, actual_records
+    end
+
+    test "clear buffer and discard the subsequent data until the next delimiter" do |data|
+      message_length_limit = 12
+      d = create_driver(base_config + %!
+        message_length_limit #{message_length_limit}
+        delimiter ";"
+        <parse>
+          @type json
+        </parse>
+        #{data[:extract]}
+      !)
+      d.run(expect_records: 1, timeout: 10) do
+        create_tcp_socket('127.0.0.1', @port) do |sock|
+          sock.send('{"message":', 0)
+          sock.send('"hello', 0)
+          sleep 1 # To make the server read data and clear the buffer here.
+          sock.send('world!"};', 0) # This subsequent data must be discarded so that a parsing failure doesn't occur.
+          sock.send('{"k":"v"};', 0) # This will succeed to parse.
+        end
+      end
+
+      logs = d.logs.collect do |log|
+        log.gsub(/\A\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2} [-+]\d{4} /, "")
+      end
+      actual_records = d.events.collect do |event|
+        event[2]
+      end
+
+      assert_equal(
+        {
+          # Asserting that '[warn]: pattern not matched message="world!\"}"' warning does not occur.
+          logs: ['[info]: The buffer size exceeds \'message_length_limit\', cleared: limit=12 size=17 head="{\"message\":\"hello"' + "\n"],
+          records: [{"k" => "v"}],
+        },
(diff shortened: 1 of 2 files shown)


daipom:
# Behavior of `message_length_limit`

* When the byte-size of the buffer exceeds the `message_length_limit`, the buffer will be cleared.
* If there is subsequent data, that part may fail to parse, and a warning log is output.

Run Fluentd with the following setting.

```xml
<source>
  @type tcp
  tag test
  message_length_limit 12
  delimiter ";"
  <parse>
    @type json
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

Send data as follows.

```console
$ netcat 0.0.0.0 5170
{"message":
"hello
world!};
```

Result:

```
[info]: #0 The receiving data is larger than 'message_length_limit', dropped: limit=12 size=19 head="{\"message\":\n\"hello\n"
[warn]: #0 pattern not matched message="world!}"
```

# Normal behavior

Run Fluentd with the following setting.

```xml
<source>
  @type tcp
  tag test
  delimiter ";"
  <parse>
    @type json
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

Send data as follows.

```console
$ netcat 0.0.0.0 5170
{"message":
"hello
world!};
```

Result:

```
2023-04-10 16:35:50.918992843 +0900 test: {"message":"hello\nworld!"}
```

daipom:
Writing to a socket at high speed can cause multiple writes to be stored in the buffer before the server reads them.
The new tests are exactly those.

It is undesirable that high-speed writing causes the total buffer size to exceed the limit and drop the records even if each record size does not exceed the limit.

So I fixed to check the size at 2 points: delimited one message size and total remaining buffer size.

One concern is whether this fix will affect the performance improvement of the next fix.

* #2567

test/plugin/test_in_tcp.rb
```diff
@@ -255,4 +255,78 @@ def create_tcp_socket(host, port, &block)
...
+      end
+      actual_records = d.events.collect do |event|
+        event[2]
+      end
+
+      assert_equal(
+        {
+          logs: [
+            '[info]: The buffer size exceeds \'message_length_limit\', cleared: limit=12 size=17 head="{\"message\":\"hello"' + "\n",
+            '[warn]: pattern not matched message="world!\"}"' + "\n"
```
ashie, line 321: Trailing data of dropped record should be obviously dropped as well.
How about like this:

```diff
diff --git a/lib/fluent/plugin/in_tcp.rb b/lib/fluent/plugin/in_tcp.rb
index 03e30d30..347458dc 100644
--- a/lib/fluent/plugin/in_tcp.rb
+++ b/lib/fluent/plugin/in_tcp.rb
@@ -106,6 +106,8 @@ module Fluent::Plugin
 
       log.info "listening tcp socket", bind: @bind, port: @port
       del_size = @delimiter.length
+      discard_till_next_delimiter = false
+
       if @_extract_enabled && @_extract_tag_key
         server_create(:in_tcp_server_single_emit, @port, bind: @bind, resolve_name: !!@source_hostname_key, send_keepalive_packet: @send_keepalive_packet) do |data, conn|
           unless check_client(conn)
@@ -120,6 +122,11 @@ module Fluent::Plugin
             msg = buf[pos...i]
             pos = i + del_size
 
+            if discard_till_next_delimiter
+              discard_till_next_delimiter = false
+              next
+            end
+
             if !@message_length_limit.nil? && @message_length_limit < msg.bytesize
               log.info "The received data is larger than 'message_length_limit', dropped:", limit: @message_length_limit, size: msg.bytesize, head: msg[...32]
               next
@@ -146,6 +153,7 @@ module Fluent::Plugin
           if !@message_length_limit.nil? && @message_length_limit < buf.bytesize
             log.info "The buffer size exceeds 'message_length_limit', cleared:", limit: @message_length_limit, size: buf.bytesize, head: buf[...32]
             buf.clear
+            discard_till_next_delimiter = true
             next
           end
         end
@@ -164,6 +172,11 @@ module Fluent::Plugin
             msg = buf[pos...i]
             pos = i + del_size
 
+            if discard_till_next_delimiter
+              discard_till_next_delimiter = false
+              next
+            end
+
             if !@message_length_limit.nil? && @message_length_limit < msg.bytesize
               log.info "The received data is larger than 'message_length_limit', dropped:", limit: @message_length_limit, size: msg.bytesize, head: msg[...32]
               next
@@ -189,6 +202,7 @@ module Fluent::Plugin
           if !@message_length_limit.nil? && @message_length_limit < buf.bytesize
             log.info "The buffer size exceeds 'message_length_limit', cleared:", limit: @message_length_limit, size: buf.bytesize, head: buf[...32]
             buf.clear
+            discard_till_next_delimiter = true
             next
           end
         end
```

daipom: I see! I will try to fix this in that direction!
`in_udp` on Non-Windows has this specification too, but it does not seem as easy as this fix.
So, certainly, it would be best to fix it for `in_tcp` so far now.

daipom: Fixed as https://github.com/fluent/fluentd/compare/283a3b3db39a625782d311671412d81b8d7e7aa0..0d9c3209de4bf1feaff23be048a0d6818b4e7b98

daipom: I will re-test and update https://github.com/fluent/fluentd/pull/4137#issuecomment-1501517920

ashie (approved):

test/plugin/test_in_tcp.rb
ashie: Please let me know if you O.K. to merge.

daipom: > I will re-test and update [#4137 (comment)](https://github.com/fluent/fluentd/pull/4137#issuecomment-1501517920)

Done!

> Please let me know if you are O.K. to merge.

I'm OK to merge!

ashie: Thanks!

daipom:
Thanks for your review!

PR #4138, opened 2023-04-10
Test `in_tcp`: Fix undesirable way to assert logs

### LINKED ISSUE #4136: CI: Use `Driver::logs` for asserting log messages
### Describe the bug

Some tests are asserting log messages as follows.
We should fix these tests.

* Use `shutdown: false` and manually do `d.instance_shutdown`
* Take logs from `Driver::instance`

https://github.com/fluent/fluentd/blob/206b46b91560177366371c1f05bbcf09ac006d86/test/plugin/test_in_forward.rb#L186-L203

We should not use this approach.

The log messages of the instance will be cleared after the shutdown, but we can still take log messages from `Driver::logs`.
So we should use `Driver::logs` and should not use `shutdown: false` for this purpose.

Note: `Driver::logs` is the same instance as `instance.log.out.logs` (= `instance.log.logs`), but `reset` will not clear it, just re-assign the new list to `DummyLogDevice`. So we can use `Driver::logs` after the shutdown.

* https://github.com/fluent/fluentd/blob/206b46b91560177366371c1f05bbcf09ac006d86/lib/fluent/test/driver/base_owner.rb#L30-L31
*  https://github.com/fluent/fluentd/blob/206b46b91560177366371c1f05bbcf09ac006d86/lib/fluent/test/log.rb#L32-L34

### To Reproduce

None.

### Expected behavior

None.

### Your Environment

```markdown
Not related.
```

### Your Configuration

```apache
Not related.
```

### Your Error Log

```shell
Not related.
```

### Additional context

The document also says we should use `Driver::logs`.

https://docs.fluentd.org/plugin-development/plugin-test-code#tests-for-logs

daipom:
* Partially fixes #4136

We should use `Driver::logs` so that we don't have to use `shutdown: false`.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed (or the same as the title).

### DIFF
--- a/test/plugin/test_in_tcp.rb
+++ b/test/plugin/test_in_tcp.rb
@@ -218,16 +218,14 @@ class TcpInputTest < Test::Unit::TestCase
           </client>
         </security>
       !)
-      d.run(shutdown: false, expect_records: 1, timeout: 2) do
+      d.run(expect_records: 1, timeout: 2) do
         create_tcp_socket('127.0.0.1', @port) do |sock|
           sock.send("hello\n", 0)
         end
       end
 
-      assert_equal 1, d.instance.log.logs.count { |l| l =~ /anonymous client/ }
+      assert_equal 1, d.logs.count { |l| l =~ /anonymous client/ }
       assert_equal 0, d.events.size
-    ensure
-      d.instance_shutdown if d&.instance
     end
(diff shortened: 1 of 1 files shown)


ashie:
Thanks!

PR #4145, opened 2023-04-11
Fix a link for the repository of td-agent

ashie:
N/A

omnibus-td-agent is obsolete, we should use fluent-package-builder for tracking issues of td-agent instead.

**Docs Changes**:
N/A

**Release Note**: 
N/A


daipom (approved):
LGTM! Thanks!

ashie:
Thanks!

PR #4159, opened 2023-04-28
Fix broken argument `unpacker` of `ChunkMessagePackEventStreamer.each`

daipom:
None.

The argument `unpacker` of [ChunkMessagePackEventStreamer.each](https://github.com/fluent/fluentd/blob/v1.16.1/lib/fluent/event.rb#L313) seems to have been added in order to match the feature with `EventStream` at

* #2559

However, that previous implementation at that point does not work as expected.
It has never caused any issues just because the argument was not used at all.

When `unpacker` is passed, the following will be executed.

```ruby
unpacker.each($block)
```

This just iterates the internal buffer of the unpacker without reading `io` of the chunk.
Thus it seems to me that this doesn't work as expected.

Note: It could be implemented so that this argument is used, but given that it has not been used so far, there is little need for it.

**Docs Changes**:
Not needed since it is an internal fix.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/event.rb
+++ b/lib/fluent/event.rb
@@ -308,11 +308,15 @@ module Fluent
   end
 
   module ChunkMessagePackEventStreamer
-    # chunk.extend(ChunkEventStreamer)
+    # chunk.extend(ChunkMessagePackEventStreamer)
     #  => chunk.each{|time, record| ... }
     def each(unpacker: nil, &block)
+      # Note: If need to use `unpacker`, then implement it,
+      # e.g., `unpacker.feed_each(io.read, &block)` (Not tested)
+      raise NotImplementedError, "'unpacker' argument is not implemented." if unpacker
+
       open do |io|
-        (unpacker || Fluent::MessagePackFactory.msgpack_unpacker(io)).each(&block)
+        Fluent::MessagePackFactory.msgpack_unpacker(io).each(&block)
       end
       nil
     end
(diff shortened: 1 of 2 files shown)


lib/fluent/event.rb
```diff
@@ -308,11 +308,13 @@ def ensure_decompressed!
   end
 
   module ChunkMessagePackEventStreamer
-    # chunk.extend(ChunkEventStreamer)
+    # chunk.extend(ChunkMessagePackEventStreamer)
     #  => chunk.each{|time, record| ... }
     def each(unpacker: nil, &block)
       open do |io|
-        (unpacker || Fluent::MessagePackFactory.msgpack_unpacker(io)).each(&block)
+        # Note: If need to use `unpacker`, then implement it,
```
ashie, line 315: It seems that we should raise `NotImplementedError` for this case to clarify it.

daipom: Thanks for your advice!
I see! I will fix it!

daipom:
Rebased to the latest master.

lib/fluent/event.rb
daipom: Done!

daipom:
I'm concerned about the many failures of CI.
I'll check it.

daipom:
I think the failures have nothing to do with this PR.
Rerunning...

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4160, opened 2023-05-01
Project governance: Fix voting rule

### LINKED ISSUE #4151: Project governance: tweak voting majority threshold
### Describe the bug

Currently, the voting requires 2/3 majority which is described in [Voting](https://github.com/fluent/fluentd/blob/master/GOVERNANCE.md#voting)

More contributors involved, it tends to be hard to satisfy with 2/3 majority because of #4150.

Also mentioned https://github.com/fluent/fluentd/issues/4150#issuecomment-1507874478

### To Reproduce

N/A

### Expected behavior

Introduce the new threshold in a practical manner.

Instead of 2/3 majority, changes the threshold like this:

* No objection ballot
* At least effective 3-organization affirmative vote

Both of the above conditions are met, the voting agenda will be approved.

### Your Environment

```markdown
N/A
```

### Your Configuration

```apache
N/A
```

### Your Error Log

```shell
N/A
```

### Additional context

* If these change was introduced, [the previous organization's vote](https://github.com/fluent/fluentd/pull/4069) was easy to make things forward.
* Before call for voting #4150, this issue should be processed. (it makes easy to process #4150 afterwards)

daipom:
Fixes #4151

The more contributors there are, the harder it will be to satisfy the current condition (2/3 majority organization vote).

Therefore, this updates the conditions based on the discussion in 

* #4151

These new conditions refer to Apache Voting Process:

* https://www.apache.org/foundation/voting#apache-voting-process

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

**Voting**:
I would like to propose taking a vote on this fix according to [our current governance policy](https://github.com/fluent/fluentd/blob/d6c34cbad241de8d2609f00666a32eb34cb68594/GOVERNANCE.md).

The voting rule for this change is not specified, but how about  2/3 organization's votes are sufficient for approval, as usual?

The total number of organizations is 8 and we need at least 6 votes to satisfy the majority of 2/3.

* Organizations
  * @sonots (ZOZO)
  * @cosmo0920, @edsiper (Calyptia)
  * @toru-takahashi (Treasure Data)
  * @ashie, @kenhys, @daipom (ClearCode)
* Individual developers
  * @okkez
  * @repeatedly
  * @tagomoris
  * @fujimotos

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -11,7 +11,26 @@ The Fluentd community adheres to the following principles:
 
 ## Voting
 
-The Fluentd project employs "organization voting" to ensure no single organization can dominate the project.
+The Fluentd project employs "Organization Voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, we follow these steps.
+
+1. Have enough period to discuss the topic so that all discussion members agree that there are no more points to be discussed.
+2. Add a specific statement of what is being voted on to the description of the relevant GitHub issue or pull request.
+3. Declare the start of voting.
+4. Ask maintainers to indicate their yes/no votes on the issue or PR.
+5. Tally the votes and note the outcomes after a suitable period.
+
+All the following conditions must be satisfied for the vote to be approved.
+
+- No effective objection ballot: See Vetoes, below.
+- At least effective __2-organization__ affirmative vote: See Organization Vote, below.
+- At least effective __3-maintainers__ affirmative vote.
+- At least __2-week__ for voting.
+
+Please note that the period for voting should depend on the topic. __2-week__ is merely the minimum period to ensure time for all organizations to say, "Wait! We need more discussion!". The more significant the decision's impact is, the longer the period should be.
+
+### Organization Vote
 
 Individuals not associated with or employed by a company or organization are allowed one organization vote. Each company or organization (regardless of the number of maintainers associated with or employed by that company/organization) receives one organization vote.
 
@@ -19,13 +38,19 @@ In other words, if two maintainers are employed by Company X, two by Company Y, 
 
 Any maintainer from an organization may cast the vote for that organization.
 
-For formal votes, a specific statement of what is being voted on should be added to the relevant github issue or PR, and a link to that issue or PR added to the maintainers meeting agenda document. Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
+### Vetoes
+
+The proposal is not approved as long as any maintainer votes an effective objection ballot.
+
+The maintainer who votes an objection ballot must explain the reason for the objection. Without reasonable justification, the objection ballot is not considered effective.
+
+The ballot can be changed during the voting period. For example, if the reason for the objection is solved by discussion or additional fixes, the objection ballot will be withdrawn and changed to an affirmative vote.
 
 ## Changes in Maintainership
 
-New maintainers are proposed by an existing maintainer and are elected by a 2/3 majority organization vote.
+New maintainers are proposed by an existing maintainer and are elected by the formal voting process: See Voting, above.
 
-Maintainers can be removed by a 2/3 majority organization vote.
+Maintainers can be removed by the formal voting process: See Voting, above.
 
 ## Github Project Administration
 
@@ -44,7 +69,7 @@ The fluent organization is open to receive new sub-projects under it umbrella. T
   - Data collection
   - Log management
   - Metering
-- Be supported by 2/3 majority of organization
+- Be supported by the formal voting process: See Voting, above.
 
 The submission process starts as a Pull Request on Fluentd repository with the required information mentioned above. Once a project is accepted, it's considered a __CNCF sub-project under the umbrella of Fluentd__
 


GOVERNANCE.md
```diff
@@ -11,21 +11,36 @@ The Fluentd community adheres to the following principles:
...
+
+For formal votes, a specific statement of what is being voted on should be added to the relevant GitHub issue or PR. Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
+
+All the following conditions must be satisfied to approve the formal votes.
+
+- No effective objection ballot: See Vetoes, below.
+- At least effective __2-organization__ affirmative vote: See Organization Vote, below.
+- At least effective __3-maintainers__ affirmative vote.
+- At least __2-week__ for voting.
+
```
kenhys, line 24: Maybe better to note that ballot can be changed during the voting period.
That rule is expected with further discussion, opinion may be changed from objection to affirmative onr.

GOVERNANCE.md
```diff
@@ -11,21 +11,36 @@ The Fluentd community adheres to the following principles:
...
 In other words, if two maintainers are employed by Company X, two by Company Y, two by Company Z, and one maintainer is an un-affiliated individual, a total of four "organization votes" are possible; one for X, one for Y, one for Z, and one for the un-affiliated individual.
 
 Any maintainer from an organization may cast the vote for that organization.
 
-For formal votes, a specific statement of what is being voted on should be added to the relevant github issue or PR, and a link to that issue or PR added to the maintainers meeting agenda document. Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
+### Vetoes
+
+The proposal is not approved as long as any maintainer votes an effective objection ballot.
+
+The maintainer who votes an objection ballot must explain the reason for the objection. Without reasonable justification, the objection ballot is not considered effective.
```
kenhys, line 37: It may be too sensitive, I have a concern about who judge "reasonable justification"?
It depends on which side you stand.

Can we make an agreement about "what is reasonable justification"?

GOVERNANCE.md
daipom: Thanks for the review!

Would it be good to add it to the next section?

https://github.com/fluent/fluentd/blob/050fe7a83b32565b4a49b59db5502aa8bbe78847/GOVERNANCE.md?plain=1#L33-L37

kenhys: Yes, it is reasonable to do so.

GOVERNANCE.md
daipom: Thanks!

There is some concern about not stating anything in this regard, but I think it would be difficult to set rules in too much detail for this.

I'm referring to Apache Voting Process here as well.

* https://www.apache.org/foundation/voting#Veto

kenhys: Hmm, then how about phrasing **reasonable "technical" justification** ?

daipom: I am concerned about whether it is appropriate to limit it to "technical" matters since I think that votes could be taken on non-technical matters as well.

For example, in a vote on a rule change such as this one, technical knowledge is not relevant.

kenhys: Ah, that's right.

GOVERNANCE.md
```diff
@@ -11,21 +11,36 @@ The Fluentd community adheres to the following principles:
...
+The Fluentd project employs "organization voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, a specific statement of what is being voted on should be added to the relevant GitHub issue or PR. Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
+
+All the following conditions must be satisfied to approve the formal votes.
+
+- No effective objection ballot: See Vetoes, below.
+- At least effective __2-organization__ affirmative vote: See Organization Vote, below.
+- At least effective __3-maintainers__ affirmative vote.
+- At least __2-week__ for voting.
```
cosmo0920, line 23: To change the voting system, we must follow the CNCF's election policy:
https://github.com/cncf/foundation/blob/main/maintainers-election-policy.md

>    Each eligible project maintainer will receive one vote. In order to participate in the election process, maintainers will need to add themselves to the list of active voters. This will occur at the same time as the nomination process and last at least two weeks.

We must follow the Voting System from **CNCF** not Apache.
But, I agree with your suggestion. It's acceptable change. :+1:

daipom: > We must follow the Voting System from CNCF not Apache.

Oh! I see! Sorry for missing that point.

> But, I agree with your suggestion. It's acceptable change. 👍

Thanks for your review from CNCF's point of view!

cosmo0920: However, CNCF's guideline declares project maintainers as follows:

> These project maintainer lists will be determined by the public lists at maintainers.cncf.io.

Have you registered and received an email from maintainers.cncf.io, @daipom?
If not, would you mind if you registered the mailing list?

ashie: Ah, we should post a pull request to https://github.com/cncf/foundation like this: https://github.com/cncf/foundation/pull/504
Thanks for notifying it.

daipom: Oh, I didn't know it! Thanks!

GOVERNANCE.md
daipom: Fixed! Thanks!

https://github.com/fluent/fluentd/compare/050fe7a83b32565b4a49b59db5502aa8bbe78847..5c00a9df86f0ac73f6a22a98505efd849cf9ddbf

GOVERNANCE.md
```diff
@@ -11,21 +11,38 @@ The Fluentd community adheres to the following principles:
 
 ## Voting
 
-The Fluentd project employs "organization voting" to ensure no single organization can dominate the project.
+The Fluentd project employs "organization voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, a specific statement of what is being voted on should be added to the relevant GitHub issue or PR. Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
```
daipom, line 16:
```suggestion
For formal votes, a specific statement of what is being voted on should be added to the relevant GitHub issue or PR. Maintainers should indicate their yes/no vote on that issue or PR. After a suitable period of time, the votes will be tallied.
```

It seems to me that this sentence is hard to understand.

I think it would be better to keep it as simple as the following, but I also think it would be better not to mix too many irrelevant modifications.

What do you all think?

cosmo0920: At least, `and the outcome noted.` shouldn't be removed. This is because the result should be _recorded_.

GOVERNANCE.md
ashie: https://github.com/cncf/foundation/pull/553

ashie: Added: https://maintainers.cncf.io

GOVERNANCE.md
daipom: Sorry for my late reply. Thanks for your feedback!

I see! Now I understand what the sentence `the outcome noted` wants to say. It wants to say that `the result should be recorded`.
(It seemed obvious to me that we can see whether the proposal is approved or not. I didn't read the intent, which says we should keep records of the voting results. )

There may be something we can improve, but it would be better to separate it from this fix.

GOVERNANCE.md
daipom: Thanks!

daipom:
Thanks for the reviews!

Now I would like to propose taking a vote on this fix according to [our current governance policy](https://github.com/fluent/fluentd/blob/d6c34cbad241de8d2609f00666a32eb34cb68594/GOVERNANCE.md).

The voting rule for this change is not specified, but how about  2/3 organization's votes are sufficient for approval, as usual?

tagomoris:
I'm still unsure what type of problems will be picked for voting. But "at least 2 weeks" should be shorter than my expectation.
For example, If someone proposed an idea to release Fluentd 2.0 in mid-December, it could get low intention from members in 2 weeks.

daipom:
@tagomoris Thanks for your feedback!

> I'm still unsure what type of problems will be picked for voting.

Same as before on this.

> "at least 2 weeks" should be shorter than my expectation.
> For example, If someone proposed an idea to release Fluentd 2.0 in mid-December, it could get low intention from members in 2 weeks.

As you say, the appropriate period should vary depending on the topic.

Basically, we will decide on an appropriate period for the topic.

The reason I am suggesting 2 weeks is that this is the minimum period.
This is designed to prevent people from completing their votes too quickly (no time for saying "Wait!").
This doesn't say a decision should be made in two weeks.

ashie:
> For example, If someone proposed an idea to release Fluentd 2.0 in mid-December, it could get low intention from members in 2 weeks.

IMHO we won't start voting until the discussion is over (as we are doing in this PR).
In your example, we never start voting until we get opinions of when Fluentd 2.0 should be released from the each active developers.
Still do you think 2 weeks is too short? If so, how long do you think it's appropriate?

daipom:
> IMHO we won't start voting until the discussion is over (as we are doing in this PR).

I have the same assumption.

tagomoris:
If you all have the same assumption (voting will start when ...), the voting rule should contain it.

ashie:
> If you all have the same assumption (voting will start when ...), the voting rule should contain it.

I see, thanks! We'll consider to add such sentence.

daipom:
@tagomoris Thanks for your review!

How about this? (Line breaks are temporarily inserted for readability.)

```diff
 ## Voting

 The Fluentd project employs "organization voting" and "Vetoes" to ensure no single organization can dominate the project.

 For formal votes, a specific statement of what is being voted on should be added to the relevant GitHub issue or PR.
+The start of voting must be clearly declared.
 Maintainers should indicate their yes/no vote on that issue or PR, and after a suitable period of time,
 the votes will be tallied and the outcome noted.

 All the following conditions must be satisfied to approve the formal votes.

 - No effective objection ballot: See Vetoes, below.
 - At least effective __2-organization__ affirmative vote: See Organization Vote, below.
 - At least effective __3-maintainers__ affirmative vote.
 - At least __2-week__ for voting.

+Please note that __2-week__ is merely the minimum period to ensure time for all organizations to say,
+"Wait! We need more discussion!".
+We should have enough period to discuss and vote depending on a topic. We should discuss enough on a topic
+before starting voting.
+
```

tagomoris:
@daipom Not bad, but it doesn't look like the definition of a rule.
I would add items to the list, for example:

```diff
  - No effective objection ballot: See Vetoes, below.
  - At least effective __2-organization__ affirmative vote: See Organization Vote, below.
  - At least effective __3-maintainers__ affirmative vote.
+ - An explicitly declared beginning of the term of voting.
  - At least __2-week__ for voting.

+Voting period should begin at a time when all discussion members agree that there are no more points to be discussed. And voting should have a longer period for the decisions with larger impacts.
```

daipom:
Hmm, I think the list consists of the conditions to approve the proposal.
It seems to me that `An explicitly declared beginning of the term of voting.` is a voting procedure, not a condition.
So, I feel something strange adding it to this list.

What do you all think? I would like to hear a wide range of opinions.

ashie:
> It seems to me that `An explicitly declared beginning of the term of voting.` is a voting procedure, not a condition.

I feel same thing a bit but not so negative for including this.
Because the next item is also about voting procedure:
```
  - At least __2-week__ for voting.
```

daipom:
> > It seems to me that `An explicitly declared beginning of the term of voting.` is a voting procedure, not a condition.
> 
> I feel same thing a bit but not so negative for including this. Because the next item is also about voting procedure:
> 
> ```
>   - At least __2-week__ for voting.
> ```

Certainly, we can assume this is also about the voting procedure.
(I think it can also be viewed as a condition that must be satisfied by elapsed time during voting, though)

Sorry, my explanation may be inaccurate.
Perhaps the point I felt strange with was the following.

If `An explicitly declared beginning of the term of voting.`  were not satisfied in advance, the voting would not have begun in the first place.
It seems I was thinking of `conditions` as conditions that had to be satisfied after starting voting.

I don't think there is any necessity for that (my concept of the list).
I want to prioritize a form that many people will find easy to understand.

daipom:
Given the points discussed so far, how about rewriting the entire chapter like this?

```markdown
## Voting

The Fluentd project employs "organization voting" and "Vetoes" to ensure no single organization can dominate the project.

For formal votes, we follow these steps.

1. Have enough period to discuss the topic so that all discussion members agree that there are no more points to be discussed.
2. Add a specific statement of what is being voted on to the relevant GitHub issue or PR.
3. Declare the start of voting.
4. Maintainers indicate their yes/no vote on that issue or PR.
5. Tally the votes and note the outcome after a suitable period.

All the following conditions must be satisfied for the vote to be approved.

- No effective objection ballot: See Vetoes, below.
- At least effective __2-organization__ affirmative vote: See Organization Vote, below.
- At least effective __3-maintainers__ affirmative vote.
- At least __2-week__ for voting.

Please note that __2-week__ is merely the minimum period to ensure time for all organizations to say, "Wait! We need more discussion!".
The more significant the decision's impact is, the longer the period should be.
```

daipom:
https://github.com/fluent/fluentd/pull/4160#issuecomment-1553988066

Slightly modified.

daipom:
I have updated the content of the "Voting" section to https://github.com/fluent/fluentd/pull/4160#issuecomment-1553988066.

daipom:
@cosmo0920 @tagomoris Could you please confirm the updated contents?

GOVERNANCE.md
```diff
@@ -11,21 +11,46 @@ The Fluentd community adheres to the following principles:
...
-The Fluentd project employs "organization voting" to ensure no single organization can dominate the project.
+The Fluentd project employs "Organization Voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, we follow these steps.
+
+1. Have enough period to discuss the topic so that all discussion members agree that there are no more points to be discussed.
+2. Add a specific statement of what is being voted on to the relevant GitHub issue or PR.
+3. Declare the start of voting.
+4. Ask maintainers to indicate their yes/no vote on the issue or PR.
+5. Tally the votes and note the outcome after a suitable period.
```
tagomoris, line 22: Who does this?

GOVERNANCE.md
```diff
@@ -11,21 +11,46 @@ The Fluentd community adheres to the following principles:
...
 
-The Fluentd project employs "organization voting" to ensure no single organization can dominate the project.
+The Fluentd project employs "Organization Voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, we follow these steps.
+
+1. Have enough period to discuss the topic so that all discussion members agree that there are no more points to be discussed.
+2. Add a specific statement of what is being voted on to the relevant GitHub issue or PR.
+3. Declare the start of voting.
+4. Ask maintainers to indicate their yes/no vote on the issue or PR.
```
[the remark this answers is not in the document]
daipom: > Most of these texts follow the current ones.

> https://github.com/fluent/fluentd/blob/dc7e4b295270d6a521f43331321628dba2692ca4/GOVERNANCE.md#L22

This too.

```
Maintainers should indicate their yes/no vote on that issue or PR
```

Do you mean we should fix it on this occasion?

I thought it didn't need to be the focus of this PR.

GOVERNANCE.md
daipom: > Most of these texts follow the current ones.

> https://github.com/fluent/fluentd/blob/dc7e4b295270d6a521f43331321628dba2692ca4/GOVERNANCE.md#L22

This too.

```
after a suitable period of time, the votes will be tallied and the outcome noted.
```

I think the subject is the person who proposes "Voting".
Currently, the subject is not specified, and I don't see a problem with that.

Do you want to specify it on this occasion?

GOVERNANCE.md
cosmo0920, line 21: nits: their vote**s**?

GOVERNANCE.md
cosmo0920, line 22: Maybe _outcome_ should be plural: outcome**s**.

GOVERNANCE.md
```diff
@@ -11,21 +11,46 @@ The Fluentd community adheres to the following principles:
 
 ## Voting
 
-The Fluentd project employs "organization voting" to ensure no single organization can dominate the project.
+The Fluentd project employs "Organization Voting" and "Vetoes" to ensure no single organization can dominate the project.
+
+For formal votes, we follow these steps.
+
+1. Have enough period to discuss the topic so that all discussion members agree that there are no more points to be discussed.
+2. Add a specific statement of what is being voted on to the relevant GitHub issue or PR.
```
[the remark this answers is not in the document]
cosmo0920: I also insist that we should use non-abbreviated words such as PR.
If you use PR in this document, you should add the description for "PR".

GOVERNANCE.md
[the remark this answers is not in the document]
cosmo0920: Maybe we should decide the voting style later in the dedicated issue. 
In my opinion, we have to write down the guideline to show yes/no voting. 
`:+1:` and `:-1:` signs should be enough in the most cases.

So, my suggestions is:

Maintainers should indicate their yes/no votes with formally defined signs. The signs must be defined when the voting is started. 

---

Or, we should define something like similar sentences.

GOVERNANCE.md
daipom: I will fix it! Thanks!

GOVERNANCE.md
daipom: I will fix it! Thanks!

GOVERNANCE.md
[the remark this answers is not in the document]
daipom: Fixed.

GOVERNANCE.md
daipom: Fixed.

GOVERNANCE.md
daipom: Fixed.

GOVERNANCE.md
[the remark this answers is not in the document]
daipom: I agree about making such improvements, as @tagomoris and @cosmo0920 say.

I think it should be made separately from this PR since this is a major rule revision.

GOVERNANCE.md
tagomoris: In my idea, this kind of document should define those things explicitly. We can easily imagine the situation:

1. a maintainer (A) started a voting
2. there are minimum "yes" 2 weeks later, but A is really busy
3. A is still busy and does not respond to the voting issue

What will happen in this situation? Another maintainer can close this voting or not?
My idea is, we should make rules as explicit as possible if we have them.

daipom: I see. Thanks for the explanation.
I understand that we can improve the document from that perspective,
but I think it should be made separately from this PR, as well as https://github.com/fluent/fluentd/pull/4160#discussion_r1206271345.

In this PR, I want to solve #4151.
Other than the threshold of the conditions required to approve the proposal, I keep the essential contents as unchanged as possible. 

I think the problem with that situation comes from the current document, not from this revision.
We should discuss it separately from this revision.

PR #4178, opened 2023-05-15
MessagePackFactory: Make sure to reset local unpacker to prevent received broken data from affecting other receiving data

daipom:
None.

@tkoba66 reported this bug. Thanks!

When `in_forward` abnormally receives incomplete (broken) data, it can't correctly receive the next sent data.

One of the main possible cases is as follows.

* There are multiple senders (`out_forward`) to one receiver (`in_forward`).
* An anomaly such as power failure occurs in one sender, and it sends broken data to the receiver.
* It can break the next data sent from another healthy sender.

Fixes bug in c6c6c038c5cacab6fcb9aa89a714d9366ac4902e (#2559).

Received incomplete data must not affect data from other senders.

In #2559, the local unpacker started to be cached to improve the performance, but this caused the problem where corrupted data was unintentionally left in the cached unpacker, which resulted in the next data being broken too.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/msgpack_factory.rb
+++ b/lib/fluent/msgpack_factory.rb
@@ -100,7 +100,12 @@ module Fluent
     end
 
     def self.thread_local_msgpack_unpacker
-      Thread.current[:local_msgpack_unpacker] ||= MessagePackFactory.engine_factory.unpacker
+      unpacker = Thread.current[:local_msgpack_unpacker]
+      if unpacker.nil?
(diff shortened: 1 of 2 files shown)


ashie (approved):
LGTM

Waiting for CI

ashie:
@tkoba66 Thanks for your report!

PR #4185, opened 2023-05-26
in_tail: Use inode for key of TailWatcher when follow_inodes

### LINKED ISSUE #3614: Fluentd in_tail "unreadable" file causes "following tail of <file>" to stop and no logs pushed

kattz-kawa:
Partially fixes #3614 (follow_inode true case)

Thank you for reading to the great developer's!
This PR is fixed for unexpected file close after logs rotate in fluentd v1.16.1.

I found the #3614 's reproducer.
Before applying this patch, fluentd cause unexpected file close after logs rotate every hour on my reproducer.
After applying this patch, fluentd does not cause it on my reproducer.
I have been running testing a long time.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -352,13 +352,13 @@ module Fluent::Plugin
 
     def existence_path
       hash = {}
-      @tails.each {|path, tw|
+      @tails.each_value do |tw|
         if @follow_inodes
           hash[tw.ino] = TargetInfo.new(tw.path, tw.ino)
         else
           hash[tw.path] = TargetInfo.new(tw.path, tw.ino)
         end
-      }
+      end
       hash
     end
 
@@ -443,7 +443,12 @@ module Fluent::Plugin
         return
       end
 
-      @tails[path] = tw
+      if @follow_inodes
+        @tails[target_info.ino] = tw
+      else
+        @tails[path] = tw
+      end
+
       tw.on_notify
     end
 
@@ -459,9 +464,17 @@ module Fluent::Plugin
         remove_path_from_group_watcher(target_info.path)
 
         if remove_watcher
-          tw = @tails.delete(target_info.path)
+          if @follow_inodes
+            tw = @tails.delete(target_info.ino)
+          else
+            tw = @tails.delete(target_info.path)
+          end
         else
-          tw = @tails[target_info.path]
+          if @follow_inodes
+            tw = @tails[target_info.ino]
+          else
+            tw = @tails[target_info.path]
+          end
         end
         if tw
           tw.unwatched = unwatched
@@ -475,8 +488,8 @@ module Fluent::Plugin
     end
 
     def close_watcher_handles
-      @tails.keys.each do |path|
-        tw = @tails.delete(path)
+      @tails.keys.each do |key|
+        tw = @tails.delete(key)
         if tw
           tw.close
         end
@@ -502,16 +515,27 @@ module Fluent::Plugin
       new_target_info = TargetInfo.new(path, new_inode)
 
       if @follow_inodes
-        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-        # So it should be unwatched here explicitly.
+        @tails.delete(tail_watcher.ino)
+
+        # TODO: This can cause log duplication. We need to fix this.
+        # (This problem exists from the start of implementation of follow_inodes)
+        #
+        # The old inode can still exist.
+        # In that case, we can't unwatch it because it causes log duplication.
+        # (`refresh_watcher` will find the inode as a new inode and read them again.)
+        #
+        # However, if the old inode already has disappeared, it must be unwatched here.
+        # It is because the old inode is removed from `@tails`, so `refresh_watcher` can't 
+        # recognize that it has disappeared and need to be unwatched.
+        # So simply removing this line causes leak of unwatch.
         tail_watcher.unwatched = true
 
         new_position_entry = @pf[new_target_info]
         # If `refresh_watcher` find the new file before, this will not be zero.
-        # In this case, only we have to do is detaching the current tail_watcher.
+        # In this case, we don't need to create a new TailWatcher.
         if new_position_entry.read_inode == 0
-          @tails[path] = setup_watcher(new_target_info, new_position_entry)
-          @tails[path].on_notify
+          @tails[new_inode] = setup_watcher(new_target_info, new_position_entry)
+          @tails[new_inode].on_notify
         end
       else
         @tails[path] = setup_watcher(new_target_info, pe)
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2156,13 +2156,13 @@ class TailInputTest < Test::Unit::TestCase
       target_info = create_target_info("#{@tmp_dir}/tail.txt")
       mock.proxy(Fluent::Plugin::TailInput::TailWatcher).new(target_info, anything, anything, true, true, anything, nil, anything, anything).once
       d.run(shutdown: false)
-      assert d.instance.instance_variable_get(:@tails)[target_info.path]
+      assert d.instance.instance_variable_get(:@tails)[target_info.ino]
 
       Timecop.travel(now + 10) do
         d.instance.instance_eval do
-          sleep 0.1 until @tails[target_info.path] == nil
+          sleep 0.1 until @tails[target_info.ino] == nil
         end
-        assert_nil d.instance.instance_variable_get(:@tails)[target_info.path]
+        assert_nil d.instance.instance_variable_get(:@tails)[target_info.ino]
       end
       d.instance_shutdown
     end
@@ -2679,12 +2679,20 @@ class TailInputTest < Test::Unit::TestCase
 
         # `watch_timer` calls `TailWatcher::on_notify`, and then `update_watcher` updates the TailWatcher:
         #     TailWatcher(path: "tail.txt", inode: inode_0) => TailWatcher(path: "tail.txt", inode: inode_1)
-        # The old TailWathcer is detached here since `rotate_wait` is just `1s`.
+        # @tails:
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #   }
+        # The old TailWatcher(path: "tail.txt", inode: inode_0) is detached here since `rotate_wait` is just `1s`.
         sleep 3
 
         # This reproduces the following situation:
         #     Rotation => update_watcher => refresh_watchers
         # This adds a new TailWatcher: TailWatcher(path: "tail.txt1", inode: inode_0)
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #     inode_0: TailWatcher(path: "tail.txt1", inode: inode_0),
+        #   }
         d.instance.refresh_watchers
 
         # Append to the new current log file.
@@ -2769,15 +2777,23 @@ class TailInputTest < Test::Unit::TestCase
         Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") {|f| f.puts "file2 log1"}
 
         # `watch_timer` calls `TailWatcher::on_notify`, and then `update_watcher` updates the TailWatcher:
-        #     TailWatcher(path: "tail.txt", inode: inode_0) => TailWatcher(path: "tail.txt", inode: inode_1)
+        #   TailWatcher(path: "tail.txt", inode: inode_0) => TailWatcher(path: "tail.txt", inode: inode_1)
+        # @tails:
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #   }
         sleep 2
 
         # This reproduces the following situation:
-        #     Rotation => update_watcher => refresh_watchers
+        #   Rotation => update_watcher => refresh_watchers
         # This adds a new TailWatcher: TailWatcher(path: "tail.txt1", inode: inode_0)
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #     inode_0: TailWatcher(path: "tail.txt1", inode: inode_0),
+        #   }
         d.instance.refresh_watchers
 
-        # The old TailWathcer is detached here since `rotate_wait` is `4s`.
+        # The old TailWatcher(path: "tail.txt", inode: inode_0) is detached here since `rotate_wait` is `4s`.
         sleep 3
 
         # Append to the new current log file.
@@ -2856,17 +2872,28 @@ class TailInputTest < Test::Unit::TestCase
         Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") {|f| f.puts "file2 log1"}
 
         # This reproduces the following situation:
-        #     Rotation => refresh_watchers => update_watcher
+        #   Rotation => refresh_watchers => update_watcher
         # This add a new TailWatcher: TailWatcher(path: "tail.txt", inode: inode_1)
-        #     This overwrites `@tails["tail.txt"]`
+        #   @tails => {
+        #     inode_0: TailWatcher(path: "tail.txt", inode: inode_0),
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #   }
         d.instance.refresh_watchers
 
-        # `watch_timer` calls `TailWatcher::on_notify`, and then `update_watcher` updates the TailWatcher:
+        # `watch_timer` calls `TailWatcher::on_notify`, and then `update_watcher` trys to update the TailWatcher:
         #     TailWatcher(path: "tail.txt", inode: inode_0) => TailWatcher(path: "tail.txt", inode: inode_1)
-        # The old TailWathcer is detached here since `rotate_wait` is just `1s`.
+        # However, it is already added in `refresh_watcher`, so this only remove and detach the old TailWatcher.
+        # It is detached here since `rotate_wait` is just `1s`.
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #   }
         sleep 3
 
         # This adds a new TailWatcher: TailWatcher(path: "tail.txt1", inode: inode_0)
+        #   @tails => {
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #     inode_0: TailWatcher(path: "tail.txt1", inode: inode_0),
+        #   }
         d.instance.refresh_watchers
 
         # Append to the new current log file.


ashie:
Thanks for your contribution!

DCO check is failed: https://github.com/fluent/fluentd/pull/4185/checks?check_run_id=13780778605

> Commit sha: [8f44700](https://github.com/fluent/fluentd/pull/4185/commits/8f4470087eb3dc24e0b8b3d1fb4db7f36f4ebb5c), Author: KatuyaKawakami, Committer: KatuyaKawakami; Can not find "KatuyaKawakami [kkawakami-je@nec.com](mailto:kkawakami-je@nec.com)", in ["Katuya Kawakami [kkawakami-je@nec.com](mailto:kkawakami-je@nec.com)", "Masaki Hatada [mas-hatada@nec.com](mailto:mas-hatada@nec.com)"].

Could you fix your name in the `Author` in your commit message?
```diff
- KatuyaKawakami
+ Katuya Kawakami
```

ashie:
> Could you fix your name in the Author in your commit message?

It's resolved, thanks!

ashie:
> Fix #3614

Could you add `partially` or something similar to this line in your commit message & the first comment of this PR?
As I described in #3614, it doesn't seem fix the issue on `follow_inode false` case.

ashie:
It seems that CI stalls on all platforms.

masaki-hatada:
> > Fix #3614
> 
> Could you add `partially` or something similar to this line in your commit message & the first comment of this PR? As I described in #3614, it doesn't seem fix the issue on `follow_inode false` case.

Thank you for commenting.
Katsuya will handle your comment next week.

By the way, please let me explain why we didn't take care of `follow_inode false` case.

To fix this unexpected file close issue, we have to modify the tail plugin to create a tail list with hash values **generated by inode numbers**.
But users who use `follow_inode false` might want the plugin to create a tail list with hash values **generated by file names**.
That's why we hesitated to take care of `follow_inode false` case in this commit.

In our understanding, `follow_inode false` is just for keeping the compatibility with old verions of the tail plugin.
https://docs.fluentd.org/input/tail#follow_inodes explains that `follow_inode false` can cause "read rotated files duplicately" problem.
So every user should use `follow_inode true` unless there is some particular reason.

Please let me know if our understanding is wrong.
We can change our code if needed.

ashie:
> It seems that CI stalls on all platforms.

```console
$ bundle install
$ bundle exec rake test TEST=test/plugin/test_in_tail.rb TESTOPTS="-v"
...
TailInputTest: 
  test_EACCES:																.: (0.513357)
...
  group rules line limit resolution: 
    test: valid:															.: (0.003195)
  inode_processing: 
    test_should_close_watcher_after_rotate_wait:											
```

The cause is `test_should_close_watcher_after_rotate_wait` doesn't follow the in_tail's modification.
In addition, your patch seems to be missing a fix.

```diff
diff --git a/lib/fluent/plugin/in_tail.rb b/lib/fluent/plugin/in_tail.rb
index 71d1be9b..27ee48da 100644
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -443,7 +443,11 @@ module Fluent::Plugin
         return
       end
 
-      @tails[path] = tw
+      if @follow_inodes
+        @tails[target_info.ino] = tw
+      else
+        @tails[path] = tw
+      end
       tw.on_notify
     end
 
diff --git a/test/plugin/test_in_tail.rb b/test/plugin/test_in_tail.rb
index b1e28fc9..37fbf950 100644
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2156,13 +2156,13 @@ class TailInputTest < Test::Unit::TestCase
       target_info = create_target_info("#{@tmp_dir}/tail.txt")
       mock.proxy(Fluent::Plugin::TailInput::TailWatcher).new(target_info, anything, anything, true, true, anything, nil, anything, anything).once
       d.run(shutdown: false)
-      assert d.instance.instance_variable_get(:@tails)[target_info.path]
+      assert d.instance.instance_variable_get(:@tails)[target_info.ino]
 
       Timecop.travel(now + 10) do
         d.instance.instance_eval do
-          sleep 0.1 until @tails[target_info.path] == nil
+          sleep 0.1 until @tails[target_info.ino] == nil
         end
-        assert_nil d.instance.instance_variable_get(:@tails)[target_info.path]
+        assert_nil d.instance.instance_variable_get(:@tails)[target_info.ino]
       end
       d.instance_shutdown
     end
```

Even though with above fix, some more other tests are failed.
We need to fix these tests too.

kattz-kawa:
@ashie 

Thank you for your comment and analysis.
I'm digging it.

ashie:
> In our understanding, follow_inode false is just for keeping the compatibility with old verions of the tail plugin.

Basically your recognition is reasonable, we should make `follow_inode true` default when it's proofed enough. But I don't think it's proofed enough. It's introduced at v1.12 and we experienced some fatal issues in v1.12 - v1.15.

> By the way, please let me explain why we didn't take care of `follow_inode false` case.

Even though in `follow_inode case` unexpected stall isn't acceptable.
Although `follow_inode true` is intended to resolve log duplication issue originally, modification for `follow_inode` feature seems introduce stall issue as in #3614. Such reports are increased after we merge this feature.

kattz-kawa:
@ashie 

> Could you add partially or something similar to this line in your commit message & the first comment of this PR?

Sure, I rewrite commit message and comment of this PR.

> Even though with above fix, some more other tests are failed.
> We need to fix these tests too.

I revised and rewrite the codes. 
All tests has passed locally.
Could you review this if all test passed in the github workflow?

daipom:
Thanks for this fix! I'm seeing this too.

daipom:
To me, this looks like a reasonable fix to address the problem of one scenario (https://github.com/fluent/fluentd/issues/3614#issuecomment-1578977365).

Though I think @ashie has a better grasp of the concerns of this fix, my concern is this code.

https://github.com/fluent/fluentd/blob/84b1b49a654121a57f16ab4f02cfac7656fc79e6/lib/fluent/plugin/in_tail.rb#L525

Why do we need this `deletion`?

daipom:
> https://github.com/fluent/fluentd/blob/84b1b49a654121a57f16ab4f02cfac7656fc79e6/lib/fluent/plugin/in_tail.rb#L525
> 
> Why do we need this `deletion`?

Hmm? Perhaps I'm confusing something...

ashie:
> Why do we need this deletion?

Whey the key for `@tails` is path, old TailWatcher can be dropped just by `@tails[path] = tw`.
But when the key is inode, old one cannot by `@tails[inode] = true`, need to drop it explicitly.
`rotated_target_info.ino` != `new_ino` here.

masaki-hatada:
> > Why do we need this deletion?
> 
> Whey the key for `@tails` is path, old TailWatcher can be dropped just by `@tails[path] = tw`. But when the key is inode, old one cannot be dropped by `@tails[inode] = true`, need to drop it explicitly. `rotated_target_info.ino` != `new_ino` here.

Yes, ashie-san is correct.

When using path as a key of "@tail", the key would be the same between a rotated log file and a new log file.
When using inode as a key, the key would be different between two files. So we need to delete old "@tail" entry here.

daipom:
> > Why do we need this deletion?
> 
> Whey the key for `@tails` is path, old TailWatcher can be dropped just by `@tails[path] = tw`. But when the key is inode, old one cannot be dropped by `@tails[inode] = true`, need to drop it explicitly. `rotated_target_info.ino` != `new_ino` here.

Thanks. I can imagine that, but something seems inconsistent to me...

The current implementation of managing `TailWatcher` by the paths may not be wrong even if `follow_inodes` is enabled.

For `TailWatcher`, the member `@path` and `@ino` are both constant.
So we have to recreate it when one of those needs to be updated.
Currently, `TailWatcher`s are managed by linking the paths.
Basically, it has nothing to do with `follow_inodes`.

It should be symmetrical whether we use the paths or the inodes as the keys for `@tails`.
So, I think something is wrong here with the need for deletion.

daipom:
If we use the inodes for the keys of `@tails`, shouldn't we recreate the watchers by the update of the paths for each inode?

Otherwise, it's not symmetrical.
(I don't think it has to be symmetrical if it doesn't seem to have any unintended effects, though)

daipom:
Sorry, maybe I'm saying something wrong...
I will reconsider this asymmetry.

daipom:
Now I understand the cause of this asymmetry.
It is caused by the watcher getting a new inode from the watcher's (old) path to determine the update.

daipom:
I have still some concerns about changing the key...

> Now I understand the cause of this asymmetry. It is caused by the watcher getting a new inode from the watcher's (old) path to determine the update.

Given this, it does not seem wrong to manage watchers by using the paths as the keys, regardless of `follow_inodes`.
The current update process of `update_watcher` seems not wrong to me.

It seems to me that the problem exists in `refresh_watchers`.

ashie:
> Given this, it does not seem wrong to manage watchers by using the paths as the keys, regardless of `follow_inodes`.

You are right, collecting TailWatchers is done by path by refresh_watcher (according to `path` setting in `<source> section`), so managing by `path` isn't wrong regardless of `follow_inodes`.

I think changing the key is just for comprehensibility of the code.

masaki-hatada:
Sorry, I couldn't understand what is "asymmetry".

Please let me explain my idea.

For example, let's consider the case of that fluentd watches "/var/log/0.log".
Now the log file was rotated. An old "/var/log/0.log" uses inodeA and a new "/var/log/0.log" uses inodeB.

When using a path as a key of `@tails`,  both of old and new `TailWatcher` objects would be stored to the same `@tails["/var/log/0.log"]`.
On the other hand, when using an inode as a key of `@tails`, the old one would be stored to `@tails[inodeA]` but the new one would be stored to `@tails[inodeB]`.
So `@tails[inodeA]` would be no longer used after "/var/log/0.log" got rotated. That's why we remove `@tails[inodeA]` entry.
It doesn't mean that we dispose the old `TailWatcher` object here. It would be disposed by the following line.

```
      detach_watcher_after_rotate_wait(rotated_tw, pe.read_inode) if rotated_tw
```

The old `TailWatcher` would be disposed in the same line even in `follow_inodes false`.
So I think that it's symmetry.

> Given this, it does not seem wrong to manage watchers by using the paths as the keys, regardless of follow_inodes.
> The current update process of update_watcher seems not wrong to me.

To change the key from path to inode is the most important point of this fix.

We believe that the issue is caused by using the same `@tails` entry even after the rotation.
As what #4190 explains, `@tails["/var/log/0.log"]` might have been already updated to the new "/var/log/0.log" by `refresh_watchers()`. In that case, `update_watcher()` would close the new one wrongly.

The path is the same even after the rotation, but the inode is changed.
So a different `@tails` entry would be used after the rotation if we use inode as a key. We can prevent wrongly closing by that.

> I think changing the key is just for comprehensibility of the code.

Our idea is to use inode as a key of all functions of in_tail.rb when `follow_inodes true`.
The plugin would be no longer work if we changed the key only in some of functions.

daipom:
Thanks, everyone.

I am wondering if the key of `@tails` is not a fundamental part of this problem.

I'm still not sure, but I commented in https://github.com/fluent/fluentd/issues/3614#issuecomment-1580130292 on what I thought was the possible alternative solution.

ashie:
> To change the key from path to inode is the most important point of this fix.
> 
> We believe that the issue is caused by using the same `@tails` entry even after the rotation. As what #4190 explains, `@tails["/var/log/0.log"]` might have been already updated to the new "/var/log/0.log" by `refresh_watchers()`. In that case, `update_watcher()` would close the new one wrongly.

> > I think changing the key is just for comprehensibility of the code.
> 
> Our idea is to use inode as a key of all functions of in_tail.rb when `follow_inodes true`. The plugin would be no longer work if we changed the key only in some of functions.

Ah, sorry, `just for comprehensibility of the code` is misleading.
We understand what's wrong and why this PR can fix the issue.
We are talking about whether there are more ideal solution or not.

Because tail watchers are collected according to `path` setting in `<source>` section, managing by path isn't wrong ideally. 
(It causes this issue in actual though).
We think there are some other ideas to resolve it without changing the key, #4191 is one of such idea.

daipom:
> Sorry, I couldn't understand what is "asymmetry".

Sorry, I just wanted to say that `path` as the key is reasonable for the current structure, regardless of `follow_inodes`.

A: The current implementation:

* Key of `@tails`: `path`
* Update of `TailWatcher` is processed based on the `path`
  * Based on the `path`, it sees the change of `ino`.

B: This PR's implementation:

* Key of `@tails`: `ino`
* Update of `TailWatcher` is processed based on the `path`
  * Based on the `path`, it sees the change of `ino`.

C:  State that I feel is symmetrical to A. (In this case, we don't need to worry about the deletion.)

* Key of `@tails`: `ino`
* Update of `TailWatcher` is processed based on the `ino`
  * Based on the `ino`, it sees the change of `path`.

The important point is that the current `TailWatcher` is updated based on its `path` by `TimerTrigger` or `StateWatcher`, and we don't want to change it.

B will certainly solve #3614 problem partially, but if we only consider the update process of `TailWatcher` (`update_watcher`), the current A is reasonable.

daipom:
Thanks to this PR and #4191, I am starting to see problems that I didn't understand before.
Thank you very much. I would like to think about it some more.

masaki-hatada:
Hi @ashie san, @daipom san,

Thank you for the explanations.
Maybe I got your idea.

> The important point is that the current TailWatcher is updated based on its path by TimerTrigger or StateWatcher, and we don't want to change it.

I couldn't find a code that fluentd passes `path` to `TimerTrigger`.
But at least fluentd passes `path` to StatWatcher.
`StatWatcher` is the child class of `Coolio::StatWatcher`. We cannot modify `Coolio::StatWatcher` to use inode instead of path.
I imagine that's why you guys want to keep path as the key of `@tails`.

But I feel that it's not weird even if we use inode as the key even though `StatWatcher` still uses path.
In my understanding:
* `StatWatcher` is used just for detecting the update of a log file specified by `path`.
* `tw.on_notify()` will be called if `StatWatcher` detected the update. It calls `@rotate_handler.on_notify(stat)` then check if inode got changed.
* `@on_rotate` would be called if inode change was detected.

To change the key of `@tails` would have no impact for the above.
`on_rotate()` cares inode. So I feel that it's natural that `@tails` is also managed by inode.

> We think there are some other ideas to resolve it without changing the key, #4191 is one of such idea.

Currently I haven't found any negative side-effect of #4191.
I agree with that #4191 is good if fluentd community wants a small fix.
(I also think that #4191's `if @follow_inodes && tw.ino != ino` might be able to be changed just to `if tw.ino != ino`, even though I haven't chekced it deeply)

But #4191 is a last catch fix as @garyzjq san said.

In our opinion, to use path as the key of `@tails` even in `follow_inodes true` case can reduce the maintainability of this plugin.
The difference of the behavior of `follow_inodes true` and `follow_inodes false` is big, I think.
For example:
- The same entry of pos file is updated even after rotation in `follow_inodes false`
- But a new entry is created in `follow_inodes true` case.

In `follow_inodes false` case:
```
/var/log/pods/aaaa/bbbb/0.log      000000000236aab7        000000000860571f <= the same entry is updated even after rotation
```

In `follow_inodes true` case:
```
/var/log/pods/aaaa/bbbb/0.log      ffffffffffffffff        000000000860571f
/var/log/pods/aaaa/bbbb/0.log      ffffffffffffffff        0000000008605732
/var/log/pods/aaaa/bbbb/0.log      000000000236aab7        000000000860571f <= A new entry is added after rotation
```

To match the behavior of `@tails` to the above is good to avoid confusion, I think.

Of course I can understand that a big code change isn't preferred to avoid regressions.
So It's still ok if the community doesn't choose our fix. It's a difficult decision...

daipom:
@masaki-hatada Thanks for considering my concern!

>> The important point is that the current TailWatcher is updated based on its path by TimerTrigger or StateWatcher, and we don't want to change it.
> 
> I couldn't find a code that fluentd passes `path` to `TimerTrigger`.
> But at least fluentd passes `path` to StatWatcher.

This is about here (`TailWatcher::on_notify`).

https://github.com/fluent/fluentd/blob/dcc05e59e038ffef16ee2b510d3d9e34b23bbdfe/lib/fluent/plugin/in_tail.rb#L807-L815

`TailWatcher` uses `path` as a constant value and takes a new inode of the path.

daipom:
Maybe @ashie isn't as hesitant to change the key as I am. :smile:

masaki-hatada:
@daipom san,

Thank you for the reply.

> This is about here (`TailWatcher::on_notify`).
> 
> https://github.com/fluent/fluentd/blob/dcc05e59e038ffef16ee2b510d3d9e34b23bbdfe/lib/fluent/plugin/in_tail.rb#L807-L815
> 
> `TailWatcher` uses `path` as a constant value and takes a new inode of the path.

I feel that to use path here is good in any case, because there is no os systemcall which requires inode instead of path to check the stat.
Even if such a systemcall existed, we would become unable to detect the rotation by that. (The old inode would still point to the old log file even after the file got renamed)

It's ideal that to get the latest stat of a file by path then check whether its inode got updated or not. 

> Maybe @ashie isn't as hesitant to change the key as I am. 

Ok!
I will wait the decision of him.

I or @kattz-kawa can update our code if needed.
(Adding more comments to better understanding, etc)

masaki-hatada:
> Yes. I don't want to change it, as I said before. (Sorry for my poor explanation.)

No, your explanation was good. ^^;

> Based on this fact (path is the constant value for the watcher), it seems natural to manage the watchers by its' paths as the keys.

What I want to say is that we don't need to match the key of `@tails` and the above.
For example:
- `expand_paths()` uses inode as a key of `hash` when `follow_inode true` case.
https://github.com/fluent/fluentd/blob/fbc3c2d2e2875cfea71fb36e4648dd2e7e5d33b9/lib/fluent/plugin/in_tail.rb#L342
- But `expand_paths()` still uses path for detecting new files.
https://github.com/fluent/fluentd/blob/fbc3c2d2e2875cfea71fb36e4648dd2e7e5d33b9/lib/fluent/plugin/in_tail.rb#L289

Can we change `expand_paths()` as to always use inode instead of path?
It's impossible, because to detect new files can be done only with path.

We can say the same thing for `Fluent::FileWrapper.stat(@path)` and `@tails`.
`Fluent::FileWrapper.stat(@path)` cannot be changed to use inode. But there is no relation between it and the key of `@tails`.

We can also say that inode should be used for `@tails` since `expand_paths()` uses inode for `hash`.
Although it depends on the way of thinking.

daipom:
As @kattz-kawa says in https://github.com/fluent/fluentd/issues/3614#issuecomment-1573422980, I am concerned this fix may cause a leak.

I will share more details later.

I feel the need to make the related test codes...

daipom:
# Use inodes as the key of `@tails`

Common condition

* `follow_inodes` is enabled.
* Files: `file.log (inode-A)`
  * `@tails[inode-A]` => `Watcher(file.log, inode-A)`
* Rotation!
  * file.log is moved to file.log.1.
  * A new file is created as file.log.
* Files: `file.log (inode-B)`, `file.log.1 (inode-A)`
* `refresh_watcher` and `update_watcher` can be processed at the same time...

If `refresh_watcher` is proceessed before `update_watcher`

1. `refresh_watcher`
   * Find `file.log (inode-B)` as a new target.
   * `@tails[inode-A]` => `Watcher(file.log, inode-A)` **Not changed**
   * `@tails[inode-B]` => `Watcher(file.log.1, inode-A)` **Newly added**
1. `update_watcher`
   * `Watcher(file.log, inode-A)` will call it's `update_watcher`
     * This will update the inode of file.log to inode-B.
     * Detach and delete `@tails[inode-A]` (`Watcher(file.log, inode-A)`)
       * It will read the remained data of `file.log.1 (inode-A)` since the `io_handler` is not reopened.
         (really?? What about `open_on_every_update`?)
     * Overwrite `@tails[inode-B]` by `Watcher(file.log, inode-B)` **WRONG**
       * `@tails[inode-B]` already has a watcher. We can't overwrite it here.
   * Result:
     * `@tails[inode-B]` => `Watcher(file.log, inode-B)`
     * `Watcher(file.log.1, inode-A)` remains, but it does **not exist** in `@tails`. **LEAK**
 
Certainly, we can trace the new current file, but there seems to be a leak of the watcher.

daipom:
For comparison.

# Use paths as the key of `@tails`

Common condition

* `follow_inodes` is enabled.
* Files: `file.log (inode-A)`
  * `@tails[file.log]` => `Watcher(file.log, inode-A)`
* Rotation!
  * file.log is moved to file.log.1.
  * A new file is created as file.log.
* Files: `file.log (inode-B)`, `file.log.1 (inode-A)`
* `refresh_watcher` and `update_watcher` can be processed at the same time...

If `refresh_watcher` is processed before `update_watcher`
1. `refresh_watcher`
   * Find `file.log (inode-B)` as a new target.
   * Overwrite `@tails[file.log]` by `Watcher(file.log, inode-B)`
     * **WRONG** `@tails[file.log]` already has a watcher. We can't overwrite it here.
   * Result:
     * `@tails[file.log]` => `Watcher(file.log, inode-B)`
     * `Watcher(file.log, inode-A)` remains, but it does **not exist** in `@tails`. **!!LEAK!!**
1. `update_watcher`
   * `Watcher(file.log, inode-A)` will call it's `update_watcher`
     * This will update the inode of file.log to inode-B.
     * Detach and delete `@tails[file.log]` (`Watcher(file.log, inode-B)`) **!!WRONG!!**
       * This stops tracking the new file!!
     * Overwrite `@tails[file.log]` by `Watcher(file.log, inode-B)` **!!Meaningless!!**
       * `@tails[file.log]` already has the completely same watcher.
     * Result:
       * `@tails[file.log]` => `Watcher(file.log, inode-B)`
         * Once detached, so it is no longer read?
       * `Watcher(file.log, inode-A)` still remains without detached.

daipom:
I thought of those simulations on the table.
I want to check the actual behavior later. (In addition, I want to create the test codes...)

daipom:
> Watcher(file.log.1, inode-A) remains, but it does not exist in @tails. !!LEAK!!

I think the watcher will be an orphan.
I still don't know if this will lead to some kind of resource leak or not.

In https://github.com/fluent/fluentd/issues/3614#issuecomment-1578261598, @ashie says 

> they will be scheduled to free by rotate_watcher in this case.

masaki-hatada:
@daipom san,

Thank you for the explanation.

> # Use inodes as the key of `@tails`
> Common condition
> 
> * `follow_inodes` is enabled.
> * Files: `file.log (inode-A)`
>   
>   * `@tails[inode-A]` => `Watcher(file.log, inode-A)`
> * Rotation!
>   
>   * file.log is moved to file.log.1.
>   * A new file is created as file.log.
> * Files: `file.log (inode-B)`, `file.log.1 (inode-A)`
> * `refresh_watcher` and `update_watcher` can be processed at the same time...
> 
> If `refresh_watcher` is processed before `update_watcher`
> 
> 1. `refresh_watcher`
>    
>    * Find `file.log (inode-B)` as a new target.
>    * `@tails[inode-A]` => `Watcher(file.log, inode-A)` **Not changed**

In this case, the watcher would be detached by `stop_watchers()` called from `refresh_workers()`, I think.

>    * `@tails[inode-B]` => `Watcher(file.log.1, inode-A)` **Newly added**
>
> 2. `update_watcher`
>    
>    * `Watcher(file.log, inode-A)` will call it's `update_watcher`
>      
>      * This will update the inode of file.log to inode-B.
>      * Detach and delete `@tails[inode-A]` (`Watcher(file.log, inode-A)`)
>        
>        * It will read the remained data of `file.log.1 (inode-A)` since the `io_handler` is not reopened.
>          (really?? What about `open_on_every_update`?)
>      * Overwrite `@tails[inode-B]` by `Watcher(file.log, inode-B)` **!!WRONG!!**
>        
>        * `@tails[inode-B]` already has a watcher. We can't overwrite it here.

Please correct me if I was wrong, but maybe the following "if" statement is for avoiding this.
https://github.com/fluent/fluentd/blob/84b1b49a654121a57f16ab4f02cfac7656fc79e6/lib/fluent/plugin/in_tail.rb#L521

In my understanding, "new_position_entry.read_inode == 0" means that the new log file hasn't been found by `refresh_watchers()` yet.
`update_watcher()` updates `@tails[inode-B]` only in this case.

I also had though that the current implementation of `update_watcher()` is very complicated, but I has hesitated to modify this since it would be a very big change...

daipom:
@masaki-hatada

Sorry for my late response.

> In my understanding, "new_position_entry.read_inode == 0" means that the new log file hasn't been found by `refresh_watchers()` yet.

Thanks! You are right!
I will fix the comments!

daipom:
Sorry for my incorrect description.
I updated the following 2 comments.

* https://github.com/fluent/fluentd/pull/4185#issuecomment-1580767216
* https://github.com/fluent/fluentd/pull/4185#issuecomment-1580772174

daipom:
I ran #4207 `test_updateTW_after_refreshTW` test with this fix.
The test fails because the expected value does not match this modification, but that is not the problem with this PR.
This PR certainly solves one of in_tail's problems.

```rb
<{:position_entries=>
  [["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
    "ffffffffffffffff",
    45351307],
   ["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
    "0000000000000016",
    45351310]],
 :record_values=>["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
 :tail_watcher_inodes=>[45351307, 45351310, 45351307],
 :tail_watcher_io_handler_opened_statuses=>[false, false, false],
 :tail_watcher_paths=>
  ["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
   "/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
   "/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt1"]}> expected but was
<{:position_entries=>
  [["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
    "0000000000000016",
    45351307],
   ["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
    "0000000000000016",
    45351310]],
 :record_values=>["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
 :tail_watcher_inodes=>[45351307, 45351310],
 :tail_watcher_io_handler_opened_statuses=>[false, false],
 :tail_watcher_paths=>
  ["/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt",
   "/home/daipom/work/fluentd/fluentd/test/plugin/../tmp/tail/876881f23793c42b295c/tail.txt"]}>
```

daipom:
Based on the information I have learned from this PR and #4191, I have created a PR.

* #4208

If there is no problem with #4208, I think #4208 would be a more direct solution.

masaki-hatada:
> Sorry for my incorrect description. I updated the following 2 comments.

Thank you for the update!
 
> * [in_tail: Fix for unexpected file close after logs rotate in update_wa… #4185 (comment)](https://github.com/fluent/fluentd/pull/4185#issuecomment-1580767216)

I have a comment about this.

> 2. `update_watcher`
>    
>    * `Watcher(file.log, inode-A)` will call it's `update_watcher`
>      
>      * This will try to update the inode of file.log to inode-B and try to delete `@tails[inode-A]` (`Watcher(file.log, inode-A)`).
>      * However, the update and deletion is not executed because inode-B is already found (`new_position_entry.read_inode != 0`)
>      * Only detach
>      * Overwrite `@tails[inode-B]` by `Watcher(file.log, inode-B)` **!!WRONG!!**
>        
>        * `@tails[inode-B]` already has a watcher. We can't overwrite it here.

If my understanding is correct, "Overwrite `@tails[inode-B]` by `Watcher(file.log, inode-B)`" won't happen in "`new_position_entry.read_inode != 0`" case.
(Maybe you just forgot to remove this line, I think)

https://github.com/fluent/fluentd/blob/84b1b49a654121a57f16ab4f02cfac7656fc79e6/lib/fluent/plugin/in_tail.rb#L515-L529

daipom:
@masaki-hatada Thanks! You are right! I'm sorry. I forgot to remove it.
I have updated it again!

daipom (commented):
Thanks again for creating this PR!
Thanks to this PR and #4191, I can create the following PR, and it is merged now.

* #4208

The problem this PR wants to fix is fixed by that PR.

Now, we need to figure out if we should change the key of `@tails` in terms of maintainability.

This PR fixes overwriting TailWatcher by `refresh_watcher`.
After #4208, overwritten TailWatchers will be correctly detached by notifying, but it is not in a good state in terms of maintainability.

However, I have concerns about the following.
For the current entire logic that manages TailWatcher based on path, it makes sense that the key of `@tails` is `path`.

I think it might be best to leave the key unchanged and fix `refresh_watcher` so that when it overwrites TailWatcher in `@tails`,  the TailWatcher should be detached at `refresh_watcher`'s responsibility.

I will leave this PR open for a while for the above considerations.

daipom:
@kattz-kawa @masaki-hatada I have pushed the commit to solve the conflicts.
After all, I think we can change the key to solve some problems in #4221.

test/plugin/test_in_tail.rb
```diff
@@ -2856,17 +2872,26 @@ def test_updateTW_after_refreshTW
...
+        #   @tails => {
+        #     inode_0: TailWatcher(path: "tail.txt", inode: inode_0) **DETACHED**,
+        #     inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
+        #   }
         sleep 3
 
-        # This adds a new TailWatcher: TailWatcher(path: "tail.txt1", inode: inode_0)
+        # This does nothing because inode_0 still exists in @tails. (The TailWatcher is detached, but it is not removed.)
+        # TODO: Check if this is expected.
         d.instance.refresh_watchers
```
daipom, line 2895: I think we need to figure out if this is an expected behavior.

test/plugin/test_in_tail.rb
```diff
@@ -2886,22 +2911,20 @@ def test_updateTW_after_refreshTW
...
-          # TODO: This is BUG!! We need to fix it and replace this with the next.
+          record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
+
+          # TODO: Check if this is expected. 
+          tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt"],
+          tail_watcher_inodes: [inode_0, inode_1],
+          tail_watcher_io_handler_opened_statuses: [false, false],
+          # tail_watcher_inodes: [inode_0, inode_1, inode_0],
+          # tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt1"],
+          # tail_watcher_io_handler_opened_statuses: [false, false, false],
```
daipom, line 2922: I think we need to figure out if this is an expected behavior.

ashie:
Hmm, there is a failure case that is concerned with this :thinking: 

https://github.com/fluent/fluentd/actions/runs/5408475806/jobs/9827580198?pr=4185
```
Failure: test_updateTW_before_refreshTW_and_detach_after_refreshTW(TailInputTest::Update watchers for rotation with follow_inodes)
/home/runner/work/fluentd/fluentd/test/plugin/test_in_tail.rb:2814:in `test_updateTW_before_refreshTW_and_detach_after_refreshTW'
     2811:         end
     2812:       end
     2813: 
  => 2814:       assert_equal(
     2815:         {
     2816:           record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
     2817:           tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt1"],
<{:position_entries=>
  [["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
    "ffffffffffffffff",
    1114610],
   ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
    "0000000000000016",
    1114614]],
 :record_values=>["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
 :tail_watcher_inodes=>[1114610, 1114614, 1114610],
 :tail_watcher_io_handler_opened_statuses=>[false, false, false],
 :tail_watcher_paths=>
  ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
   "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
   "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt1"]}> expected but was
<{:position_entries=>
  [["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
    "0000000000000016",
    1114610],
   ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
    "0000000000000016",
    1114614]],
 :record_values=>["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
 :tail_watcher_inodes=>[1114610, 1114614, 1114610],
 :tail_watcher_io_handler_opened_statuses=>[true, false, false],
 :tail_watcher_paths=>
  ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
   "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
   "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt1"]}>

diff:
  {:position_entries=>
    [["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
?     "ffffffffffffffff",
?      0000000000000016  
?      ????????????????  
      1114610],
     ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
      "0000000000000016",
      1114614]],
   :record_values=>["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
   :tail_watcher_inodes=>[1114610, 1114614, 1114610],
?  :tail_watcher_io_handler_opened_statuses=>[true, false, false, false],
   :tail_watcher_paths=>
    ["/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
     "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt",
     "/home/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/94e63360cbdd8f447f72/tail.txt1"]}
```

test/plugin/test_in_tail.rb
daipom: # Master

* `update_watcher` overwrites the path of `@tails`, so the old inode disappears from `@tails`.
* So, `refresh_watcher` find the old inode as a new target.
  * It should not cause log duplication. PositionFile should remember the pos.
  * Related to #4221

```rb
@tails => {
  tail.txt: TailWatcher(path: "tail.txt", inode: inode_1),
  tail.txt1: TailWatcher(path: "tail.txt1", inode: inode_0) **Newly added**,
}
```

# This PR

* `update_watcher` doesn't overwrites `@tails`, so the old inode still exists in `@tails`.
* So, `refresh_watcher` can't find the old inode as a new target.

```rb
@tails => {
  inode_0: TailWatcher(path: "tail.txt", inode: inode_0) **DETACHED**,
  inode_1: TailWatcher(path: "tail.txt", inode: inode_1),
}
```

# Question

I'm confusing what the expected behavior is...

daipom:
> Hmm, there is a failure case that is concerned with this 🤔
> 
> https://github.com/fluent/fluentd/actions/runs/5408475806/jobs/9827580198?pr=4185

Oh, somehow, the first TailWatcher is not detached and not unwatched.
I've tried several times locally and still can't reproduce it.

We should check if there is a possibility that the results change depending on the timing...
Anyway, some handlers not being closed is a problem...

masaki-hatada:
The following "path" should be changed to "key" or something for avoiding confusion. (We didn't notice this when posting the patch in past...)
https://github.com/fluent/fluentd/blob/719cdee2fe1235c7ef7bf69f3e708f3bf25c12ff/lib/fluent/plugin/in_tail.rb#L491-L492

Thank you for rebasing our patch. We will recheck this and write here if we found any issue.

daipom:
> The following "path" should be changed to "key" or something for avoiding confusion. (We didn't notice this when posting the patch in past...)
> 
> https://github.com/fluent/fluentd/blob/719cdee2fe1235c7ef7bf69f3e708f3bf25c12ff/lib/fluent/plugin/in_tail.rb#L491-L492

Thanks!
Fixed: https://github.com/fluent/fluentd/compare/719cdee2fe1235c7ef7bf69f3e708f3bf25c12ff..0915daf405ca75d9695fbd0094643b6365bfb365

> Thank you for rebasing our patch. We will recheck this and write here if we found any issue.

Thanks! If you find any issue, feel free to fix it!

masaki-hatada:
Hi @daipom san, @ashie san,

Just FYI.
Currently @kattz-kawa is also running https://github.com/fluent/fluentd/pull/4185#issuecomment-1612437497 in our local env, but we haven't faced the issue.

daipom:
Thanks for your check! I want to conclude by the end of this week.
I will consider this point: https://github.com/fluent/fluentd/pull/4185#discussion_r1246115818

masaki-hatada:
I'm also confusing why this commit can solve #4221 .
I leave a memo.

# Concern1: Really is it ok if `update_watcher()` detaches rotated log files?

I had thought only the case that fluentd monitors only current log files.
(Because this is the configuration we use for our environment)

```
  path /test/fluentd/input4/test.log
```

Rotated log files can be detached in `update_watcher()` in this case.

But how about the case that fluentd monitors also rotated log files?
```
```

Rotated log files should continue to be monitored even after they got renamed. (e.g., test.log => test.log.1)
So my question is "Is it ok if `update_watcher()` detaches rotated log files in that case?".

If the answer is "No", we need to leave the detaching task to `stop_watchers()` called in `refresh_watchers()` instead of `update_watcher()`.

# Concern2: Why cannot `stop_watchers()` work when follow_inodes is true?

The following line mentions "When follow_inodes is true, it's not cleaned up by refresh_watcher".
https://github.com/fluent/fluentd/blob/155bdcabfc61180c0880a47725952e1db5fea49a/lib/fluent/plugin/in_tail.rb#L518

Maybe it means that `stop_watchers()` isn't called by `refresh_watchers()` when follow_inodes is true.
I considered why it isn't called.

1. `expand_paths()` creates a hash list. The key will be inode when follow_inodes is true.
   https://github.com/fluent/fluentd/blob/155bdcabfc61180c0880a47725952e1db5fea49a/lib/fluent/plugin/in_tail.rb#L342
   => If inode-1 of a deleted file(log-A) was reused with another new log file(log-B), log-B's TailWatcher will be stored in to its `hash[inode-1]`.
2. `existence_path()` also creates a hash list. The key will be inode when follow_inodes is true.
   https://github.com/fluent/fluentd/blob/155bdcabfc61180c0880a47725952e1db5fea49a/lib/fluent/plugin/in_tail.rb#L357
   => log-A's TailWatcher will be stored into its `hash[inode-1]` 
3. Compare existence_path's hash list and expand_paths's hash list to create `unwatched_hash`
   https://github.com/fluent/fluentd/blob/155bdcabfc61180c0880a47725952e1db5fea49a/lib/fluent/plugin/in_tail.rb#L380
   => log-A's TailWatcher won't be added to `unwatched_hash` since we can find the same inode-1 key from expand_paths's hash list. But the entry of expand_paths's hash list is for log-B. This is not for log-A.
   ==> Cannot call `stop_watchers()` for log-A.
   ==> Cannot call `start_watchers()` for log-B due to the same reason.

So we need to detach log-A's TailWatcher in `update_watcher()` instead of `stop_watchers()`.
(`refresh_watchers()` becomes able to call `start_watchers()` for log-B after the detaching)

# How can we solve the conflict between Concern1 and 2?

We can ignore Concern1 if the expected behavior of the tail plugin is that TailWatcher object should be recreated every time the log file path got changed.
However, to avoid reading the same log twice, we should tell the previous position to the new TailWatcher object.
But where is the code to do that?

Isn't the plan B of https://github.com/fluent/fluentd/pull/4221#discussion_r1246007375 still needed even after we use inode as a key?

daipom:
Thanks for your considering it!

> Concern1: Really is it ok if `update_watcher()` detaches rotated log files?

I think It is OK.
Newly created TailWatchers will tail it.

* TailWatcher-A (test.log (current))
* Rotation
  * test.log -> test.log.1
  * new test.log
* TailWatcher-A detaches test.log and newly creates TailWatcher-A' (test.log)
* test.log.1 is newly found in `refresh_watcher` and a new TailWatcher-B (test.log.1) is created, but it doesn't tail it due to the pos file.
* Rotation again
  * test.log.1 -> test.log.2
  * test.log -> test.log.1
  * new test.log
* TailWatcher-A' detaches test.log and newly creates TailWatcher-A'' (test.log)
* TailWatcher-B detaches test.log.1 and newly creates TailWatcher-B' (test.log.1)
* test.log.2 is newly found in `refresh_watcher` ...
* ...

> Concern2: Why cannot `stop_watchers()` work when follow_inodes is true?

It is mostly true.

This is because `refresh_watcher` cannot recognize that the inode is disappeared.

* As you say,  `refresh_watcher` recognize the disappeared targets by comparing `@tails` and the existing file lists.
* However, `update_watcher` overwrites `@tails[path]`, so the information about the old inode of the path is lost from `@tails`.
* So, `refresh_watcher` can't find the old inode from `@tails`, and can't recognize that the inode is disappeared.

This happens because the key of `@tails` is path.
If it is inode, then that overwriting doesn't occur, and `refresh_watcher` can unwatch the disappeared inode by rotation.

> However, to avoid reading the same log twice, we should tell the previous position to the new TailWatcher object.
But where is the code to do that?

These lines load the position from the pos file.
If the previous position is recorded, it is loaded here and passed to the new TailWatcher. 

https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail.rb#L435

https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail.rb#L509

masaki-hatada:
@daipom san,

Thank you for your explanation!

> > Concern2: Why cannot `stop_watchers()` work when follow_inodes is true?
> 
> Your assumptions seem mostly correct.
> 
> This is because `refresh_watcher` cannot recognize that the inode is disappeared.
> 
> * As you say, `refresh_watcher` recognize the disappeared targets by comparing `@tails` and the existing file lists.
> * However, `update_watcher` overwrites `@tails[path]`, so the information about the old inode of the path is lost from `@tails`.
> * So, `refresh_watcher` can't find the old inode from `@tails`, and can't recognize that the inode is disappeared.
> 
> This happens because the key of `@tails` is path. If it is inode, then that overwriting doesn't occur, and `refresh_watcher` can unwatch the disappeared inode by rotation.

I was worried if `@tails[inode]` can be still overwritten by another log file due to inode re-use while fluentd still has a TailWatcher of the old log file.
But I noticed that it cannot happen, since inode won't be freed until fluentd closes the old file by update_watcher()^^;
So my concern was cleared.

> > However, to avoid reading the same log twice, we should tell the previous position to the new TailWatcher object.
> > But where is the code to do that?
> 
> These lines load the position from the pos file. If the previous position is recorded, it is loaded here and passed to the new TailWatcher.
> 
> https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail.rb#L435
> 
> https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail.rb#L509

Ok. We can refer to the previous position by using inode because we set follow_inodes to true.

My current concern is that the previous position might have been cleared by the following line.
https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail/position_file.rb#L58

The `unwatch()` function will be called when `detach_watcher_after_rotate_wait()` is called with `tail_watcher.unwatched = true`.
The current code sets `tail_watcher.unwatched = true` when follow_inodes is true.
https://github.com/fluent/fluentd/blob/e20697e12003e2c993ccf6f1c0bac190010b121b/lib/fluent/plugin/in_tail.rb#L504-L521

> - TailWatcher-A detaches test.log and newly creates TailWatcher-A' (test.log)
> - test.log.1 is newly found in refresh_watcher and a new TailWatcher-B (test.log.1) is created, but it doesn't tail it due to the pos file.

Isn't the entry of the pos file cleared before test.log.1 is newly found?

daipom:
> My current concern is that the previous position might have been cleared by the following line.
> 
> The `unwatch()` function will be called when `detach_watcher_after_rotate_wait()` is called with `tail_watcher.unwatched = true`. The current code sets `tail_watcher.unwatched = true` when follow_inodes is true.
> 
> > * TailWatcher-A detaches test.log and newly creates TailWatcher-A' (test.log)
> > * test.log.1 is newly found in refresh_watcher and a new TailWatcher-B (test.log.1) is created, but it doesn't tail it due to the pos file.
> 
> Isn't the entry of the pos file cleared before test.log.1 is newly found?

Yes!
The PR #4221 is to fix this problem!

daipom:
Instead of #4221, I would like to fix it as follows:

* Change the key of `@tails` to inode (This PR)
* Stop unwatching in `update_watcher` (Future PR)

daipom:
Thank you too for all your help!
If you have any concerns or anything you've noticed, please let me know!

daipom:
https://github.com/fluent/fluentd/pull/4185#discussion_r1246115818

I'm about the correct specifications.
I think the detached TailWathcer must be surely removed from `@tails`.
If not removing it from `@tails`, then it must not be detached.

We need to add the fix.

test/plugin/test_in_tail.rb
daipom: I think the following must be ensured when an inode is rotated.
We need to add some fixes for this.

* Remove the TailWatcher tied to the old path
  * Remove it from `@tails` and detach it
  * Unwatching has nothing to do with the rotation. (When `refresh_watcher` can't find the inode, then it should be unwatched)
* Create a new TailWathcer tied to the new path.

daipom: What makes this complicated is that we want to think about rotation based on one inode, but `update_watcher` is called based on one path...

daipom: I think we need to do `@tails.delete(tail_watcher.ino)` even if `refresh_watcher` find the new file before.

https://github.com/fluent/fluentd/blob/155bdcabfc61180c0880a47725952e1db5fea49a/lib/fluent/plugin/in_tail.rb#L523-L529

daipom:
> > Instead of #4221, I would like to fix it as follows:
> > 
> > * Change the key of `@tails` to inode (This PR)
> > * Stop unwatching in `update_watcher` (Future PR)
> 
> Ok. I got that Both of PRs are required for fixing #4221 . Thank you for your explanation!

Sorry, I said I wanted to stop watching in `update_wathcer` in the future PR, but now I think we should do it in this PR.
It is not good to separate these fixes since using inodes for the key directly makes it unnecessary.

test/plugin/test_in_tail.rb
daipom: Done as https://github.com/fluent/fluentd/compare/155bdcabfc61180c0880a47725952e1db5fea49a..bc130634eebc03c45d8ab1170f8e5a162110d311

test/plugin/test_in_tail.rb
daipom: Done as https://github.com/fluent/fluentd/compare/155bdcabfc61180c0880a47725952e1db5fea49a..bc130634eebc03c45d8ab1170f8e5a162110d311

daipom:
I fixed the concerns. I think this is ideal behavior.
I will check the actual behavior more.

daipom:
Sorry, there was an easy mistake in the testcases.
Fixed: https://github.com/fluent/fluentd/compare/bc130634eebc03c45d8ab1170f8e5a162110d311..ce898dd572d5745663abd456486169d772daa8aa

daipom:
Sorry, one test is still failing...
I will fix it.

daipom:
> This is because `refresh_watcher` cannot recognize that the inode is disappeared.
> 
>     * As you say,  `refresh_watcher` recognize the disappeared targets by comparing `@tails` and the existing file lists.
> 
>     * However, `update_watcher` overwrites `@tails[path]`, so the information about the old inode of the path is lost from `@tails`.
> 
>     * So, `refresh_watcher` can't find the old inode from `@tails`, and can't recognize that the inode is disappeared.
> 
> 
> This happens because the key of `@tails` is path.
> If it is inode, then that overwriting doesn't occur, and `refresh_watcher` can unwatch the disappeared inode by rotation.

I'm sorry again 😢 
By using inode for the key, the overwriting of `@tails` does not occur anymore.
However, we remove TailWatcher from `@tails`, so the information of old inode is lost from `@tails` and `refresh_watcher` can not recognize the inode is disappeared after all... 😢 

https://github.com/fluent/fluentd/blob/ce898dd572d5745663abd456486169d772daa8aa/lib/fluent/plugin/in_tail.rb#L518

daipom:
I believe the current logic is the ideal behavior for `update_watcher` if using inode for the key.
However, the problem of unwatching still remains.

lib/fluent/plugin/in_tail.rb
```diff
@@ -502,16 +515,14 @@ def update_watcher(tail_watcher, pe, new_inode)
       new_target_info = TargetInfo.new(path, new_inode)
 
       if @follow_inodes
-        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-        # So it should be unwatched here explicitly.
-        tail_watcher.unwatched = true
```
masaki-hatada, line 507: If possible, could you let me know who will call `detach_watcher()` if we removed this line?

I agree that `update_watcher()` shouldn't do `tail_watcher.unwatched = true` when fluentd still needs to watch the rotated file.
But how about after fluentd stopped watching the rotated file?

In my understanding, `refresh_watchers()` will never call stop_watchers() for the rotated file since `update_watcher()` has already removed its TailWatcher from `@tails`.
Fluentd will never reset the previous position to zero in pos file if nobody detachs.
So I'm worried if fluentd reads the previous position of the old rotated file by mistake when start monitoring a new file which has the same inode.
https://github.com/fluent/fluentd/blob/ce898dd572d5745663abd456486169d772daa8aa/lib/fluent/plugin/in_tail.rb#L435

daipom: > If possible, could you let me know who will call `detach_watcher()` if we removed this line?

`update_watcher` always detach the old watcher in this line:

https://github.com/fluent/fluentd/blob/ce898dd572d5745663abd456486169d772daa8aa/lib/fluent/plugin/in_tail.rb#L532

masaki-hatada:
The following is my idea, although there might be still some overlooks...
* Use inode as a key of `@tails` if follow_inodes = true.
* Always use path as a key of `hash` in `existence_path()` and `expand_paths()` regardless of follow_inodes is true or not.
* Don't make `update_watcher()` to call `detach_watcher_after_rotate_wait()` if follow_inodes is true. Instead, update the path of its TailWatcher.
* Make `start_watchers()` to create a new TailWatcher only when if the same inode doesn't exist in `@tails`  if follow_inodes is true.

The base idea of this is "Don't recreate TailWatcher even if log files got rotated. Keep using the same TailWatchers until fluentd stopped monitoring them".

lib/fluent/plugin/in_tail.rb
daipom: Oh maybe I misunderstood!
You say if we stop unwatching in `update_watcher`, then nobody unwatches its inode and the position remains.
It would be dangerous because a new file coincidently having the same inode would use that position wrongly.

That's right!
The problem of unwatching still remain, so now I think we can't stop unwatching in `update_watcher` in this PR.
(I'm sorry for this. I though we could stop it in this PR, but it was wrong.)

Perhaps I will revert that modification, Let me think about it some more.

daipom:
> @kattz-kawa @masaki-hatada I have pushed the commit to solve the conflicts.
After all, I think we can change the key to solve some problems in #4221.

I thought this PR could nicely solve the current problem of unwatching (#4221), but it was wrong.
The problem of unwatching should be fixed in another PR.

daipom:
I'm very sorry for rebasing this branch again and again.
Now this PR is for simply changing the key.
I want to consider whether to include this PR in the next version while considering the remaining in_tail issues.

After all, the only change I made is to ensure that the old TailWatcher is removed from `@tails`. (https://github.com/fluent/fluentd/pull/4185#discussion_r1246115818)

https://github.com/fluent/fluentd/blob/45b193fbdb264d8ff41832e56aa7227e010ac17c/lib/fluent/plugin/in_tail.rb#L518

(It was done in `if new_position_entry.read_inode == 0` block originally, but I moved it outside the if block)

daipom:
@masaki-hatada @kattz-kawa I'm very sorry for troubling you all.

Thanks for your suggestion to solve the remaining problem of unwatching: https://github.com/fluent/fluentd/pull/4185#issuecomment-1630218546

Maybe that problem should be handled separately from this PR.
I'll take this idea into account and see what I can do about this problem.

ashie:
For now we don't have plan changing the key from path to inode, so I close this.
In the future we may consider it again, we'll refer this pull request again at that time.
Thanks for your contribution!

PR #4207, opened 2023-06-20
Draft: test: in_tail: Add tests for rotation with follow_inodes

daipom:
None.
Related to #3614.

This PR adds tests to reproduce the problems with follow_inodes, especially talked in #3614.
At this point, several fail.
We need to fix `in_tail` so that these tests succeed.

Perhaps, we don't need to merge this PR.
I created this PR to share these tests and discuss them.
We can add these tests to the future PR that actually fixes `in_tail`.

**Docs Changes**:
No need.

**Release Note**: 
No need.

### DIFF
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2638,4 +2638,262 @@ class TailInputTest < Test::Unit::TestCase
       end
     end
   end
+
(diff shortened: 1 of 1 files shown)


daipom:
These tests are added in the following PR:

* #4208

Issue #4190, opened 2023-05-31
[in_tail][follow_inode enabled] found a racing bug that new created watcher be unwatched unexpectedly and thus log stuck


garyzjq:
### Describe the bug

**Env & Setting**
We are using fluentd in_tail plugin to parse log from pods running on k8s. 

k8s log rotation policy: latest log will always write log file with same name, let's say 'container.log'. When file reach to certain size and need rotate, k8s will rename the original one to 'container.backup.log' as backup (inode keep the same) and create a new file still named 'container.log' (different inode)

In our env, follow_inode is enabled.

**Issue Behavior**
When log file rotated, fluentd in_tail sometimes no longer track new file change, no offset change in pos file although new entrypoint is created with no inode number.

**RC Analysis**
After update the in_tail.rb to print more troubleshooting logs, we locate the potential RC, which is a racing issue.

When a file is rotated, expectedly only old watcher for rotated file should be detach. But in some case, both old and newly created watcher are detached.

Here's the racing description:
1. File rotated: previous log file 'container.log' is renamed to ''container.backup.log'' (inode_old), new 'container.log' created (inode_new)
2. refresh_watchers regularly triggered, mark **old watcher for inode_old to be detach** and create a new watcher for inode_new (**@tail[path] now point to new watcher**)
3. old watcher on_rotate triggers update_watcher as watcher_needs_update = true. 
4. in update_watcher, "rotated_tw = @tails[**path**]" will **return the new watcher to be detached** because tw with same **path** has been updated to step 2.
5. After rotate_wait, both old and new watchers are detached in step 2 and 4, thus new watcher stuck

The main problem is in #4, when try to detach a watcher, need to determine whether the rotated_tw is still the old inode, which can be detach, or already refresh to new inode, which need skip detach (detached by other event).

### To Reproduce

Here's the detail trace log to repro the **racing issue**
2023-05-31 03:09:08 +0000 [info]: #0 stop_watchers path="<log_path>" ino=<old_ino>
2023-05-31 03:09:08 +0000 [info]: #0 detach_watcher_after_rotate_wait path="<log_path>" ino=<old_ino>
2023-05-31 03:09:08 +0000 [info]: #0 construct_watcher path="<log_path>"
2023-05-31 03:09:08 +0000 [info]: #0 construct_watcher2 path="<log_path>" ino=<new_ino>
[...]

PR #4208, opened 2023-06-20
in_tail: Ensure to detach correct watcher on rotation with follow_inodes

### LINKED ISSUE #3614: Fluentd in_tail "unreadable" file causes "following tail of <file>" to stop and no logs pushed

daipom:
* Partially fixes #3614

If `refresh_watchers` run before `update_watcher`, the old implementation of `update_watcher` detach wrongly the new TailWatcher which is added in `refresh_watcher`. This causes the problem of stopping tailing log and handle leak.

The test case `test_updateTW_after_refreshTW` reproduces this problem.
This PR solves it.

NOTE: There is another BUG about unwatching. (I adjusted some expected values of the tests for this BUG.)
When `refresh_watcher` find the rotated old file AFTER unwatching it (`rotate_wait`), then the logs will be collected in duplicate.
If `refresh_watcher` find it BEFORE unwatching it (`rotate_wait`), this problem doesn't occur because the position entry is still alive.
We need to fix this and fix the adjusted expected values.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

**OTHERS**:
This fix is based on the content and discussion of the following issue and PR:

* #3614
* #4185
* #4191

Thanks to those who have given me much insight, especially @kattz-kawa, @masaki-hatada, and @garyzjq!

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -484,8 +484,8 @@ module Fluent::Plugin
     end
 
     # refresh_watchers calls @tails.keys so we don't use stop_watcher -> start_watcher sequence for safety.
-    def update_watcher(target_info, pe)
-      path = target_info.path
+    def update_watcher(tail_watcher, pe, new_inode)
+      path = tail_watcher.path
 
       log.info("detected rotation of #{path}; waiting #{@rotate_wait} seconds")
 
@@ -499,23 +499,26 @@ module Fluent::Plugin
         end
       end
 
-      rotated_tw = @tails[path]
+      new_target_info = TargetInfo.new(path, new_inode)
 
       if @follow_inodes
-        new_position_entry = @pf[target_info]
+        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
+        # So it should be unwatched here explicitly.
+        tail_watcher.unwatched = true
 
+        new_position_entry = @pf[new_target_info]
+        # If `refresh_watcher` find the new file before, this will not be zero.
+        # In this case, only we have to do is detaching the current tail_watcher.
         if new_position_entry.read_inode == 0
-          # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-          # So it should be unwatched here explicitly.
-          rotated_tw.unwatched = true if rotated_tw
-          @tails[path] = setup_watcher(target_info, new_position_entry)
+          @tails[path] = setup_watcher(new_target_info, new_position_entry)
           @tails[path].on_notify
         end
       else
-        @tails[path] = setup_watcher(target_info, pe)
+        @tails[path] = setup_watcher(new_target_info, pe)
         @tails[path].on_notify
       end
-      detach_watcher_after_rotate_wait(rotated_tw, pe.read_inode) if rotated_tw
+
+      detach_watcher_after_rotate_wait(tail_watcher, pe.read_inode)
     end
 
     # TailWatcher#close is called by another thread at shutdown phase.
@@ -877,18 +880,14 @@ module Fluent::Plugin
               # No need to update a watcher if stat is nil (file not present), because moving to inodes will create
               # new watcher, and old watcher will be closed by stop_watcher in refresh_watchers method
               # don't want to swap state because we need latest read offset in pos file even after rotate_wait
-              if stat
-                target_info = TargetInfo.new(@path, stat.ino)
-                @update_watcher.call(target_info, @pe)
-              end
+              @update_watcher.call(self, @pe, stat.ino) if stat
             else
               # Permit to handle if stat is nil (file not present).
               # If a file is mv-ed and a new file is created during
               # calling `#refresh_watchers`s, and `#refresh_watchers` won't run `#start_watchers`
               # and `#stop_watchers()` for the path because `target_paths_hash`
               # always contains the path.
-              target_info = TargetInfo.new(@path, stat ? stat.ino : nil)
-              @update_watcher.call(target_info, swap_state(@pe))
+              @update_watcher.call(self, swap_state(@pe), stat&.ino)
             end
           else
             @log.info "detected rotation of #{@path}"
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2638,4 +2638,279 @@ class TailInputTest < Test::Unit::TestCase
       end
     end
   end
+
+  sub_test_case "Update watchers for rotation with follow_inodes" do
+    def test_updateTW_before_refreshTW_and_detach_before_refreshTW
+      config = config_element(
+        "ROOT",
+        "",
+        {
+          "path" => "#{@tmp_dir}/tail.txt*",
+          "pos_file" => "#{@tmp_dir}/tail.pos",
+          "tag" => "t1",
+          "format" => "none",
+          "read_from_head" => "true",
+          "follow_inodes" => "true",
+          # In order to detach the old watcher quickly.
+          "rotate_wait" => "1s",
+          # In order to reproduce the same condition stably, ensure that `refresh_watchers` is not
+          # called by a timer.
+          "refresh_interval" => "1h",
+          # stat_watcher often calls `TailWatcher::on_notify` faster than creating a new log file,
+          # so disable it in order to reproduce the same condition stably.
+          "enable_stat_watcher" => "false",
+        }
+      )
+      d = create_driver(config, false)
+
+      tail_watchers = []
+      stub.proxy(d.instance).setup_watcher do |tw|
+        tail_watchers.append(tw)
+        tw
+      end
+
+      Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") {|f| f.puts "file1 log1"}
+
+      d.run(expect_records: 4, timeout: 10) do
+        # Rotate (If the timing is bad, `TailWatcher::on_notify` might be called between mv and new-file-creation)
+        Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "ab") {|f| f.puts "file1 log2"}
+        FileUtils.move("#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt" + "1")
+        Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") {|f| f.puts "file2 log1"}
+
+        # `watch_timer` calls `TailWatcher::on_notify`, and then `update_watcher` updates the TailWatcher:
+        #     TailWatcher(path: "tail.txt", inode: inode_0) => TailWatcher(path: "tail.txt", inode: inode_1)
+        # The old TailWathcer is detached here since `rotate_wait` is just `1s`.
+        sleep 3
+
+        # This reproduces the following situation:
+        #     Rotation => update_watcher => refresh_watchers
+        # This adds a new TailWatcher: TailWatcher(path: "tail.txt1", inode: inode_0)
+        d.instance.refresh_watchers
+
+        # Append to the new current log file.
+        Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "ab") {|f| f.puts "file2 log2"}
+      end
+
+      inode_0 = tail_watchers[0].ino
+      inode_1 = tail_watchers[1].ino
+      record_values = d.events.collect { |event| event[2]["message"] }.sort
+      position_entries = []
+      Fluent::FileWrapper.open("#{@tmp_dir}/tail.pos", "r") do |f|
+        f.readlines(chomp: true).each do |line|
+          values = line.split("\t")
+          position_entries.append([values[0], values[1], values[2].to_i(16)])
+        end
+      end
+
+      assert_equal(
+        {
+          # TODO: This is BUG!! We need to fix it and replace this with the next.
+          record_values: ["file1 log1", "file1 log1", "file1 log2", "file1 log2", "file2 log1", "file2 log2"],
+          # record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
+          tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt1"],
+          tail_watcher_inodes: [inode_0, inode_1, inode_0],
+          tail_watcher_io_handler_opened_statuses: [false, false, false],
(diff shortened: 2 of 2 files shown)


daipom:
The scenario of this problem (basically from https://github.com/fluent/fluentd/pull/4185#issuecomment-1580772174)
Please refer to the test case `test_updateTW_after_refreshTW`.

When `follow_inodes` is enabled.

* Files: `file.log (inode-A)`
  * `@tails[file.log]` => `Watcher(file.log, inode-A)`
* Rotation!
  * file.log is moved to file.log.1.
  * A new file is created as file.log.
* Files: `file.log (inode-B)`, `file.log.1 (inode-A)`
* `refresh_watcher` and `update_watcher` can be processed at the same time...

If `refresh_watcher` is processed before `update_watcher`
1. `refresh_watcher`
   * Find `file.log (inode-B)` as a new target.
   * **OVERWRITE** `@tails[file.log]` by `Watcher(file.log, inode-B)`
     * `@tails[file.log]` already has a watcher. We can't overwrite it here.
   * Result:
     * `@tails[file.log]` => `Watcher(file.log, inode-B)`
     * `Watcher(file.log, inode-A)` remains, but it does **NOT EXIST** in `@tails`.
       * This was causing handle leaks.
1. `update_watcher`
   * `Watcher(file.log, inode-A)` will call it's `update_watcher`
     * This will update the inode of file.log to inode-B.
     * Old implementation
       * Detach `@tails[file.log]` (`Watcher(file.log, inode-B)`) **WRONG**
         * This stops tailing the new file!!
     * This PR
       * Detach self (`Watcher(file.log, inode-A)`). **CORRECT**
       * This solves the problem both of stopping tailing and handle leaks.

lib/fluent/plugin/in_tail.rb
```diff
@@ -499,23 +499,26 @@ def update_watcher(target_info, pe)
         end
       end
 
-      rotated_tw = @tails[path]
```
daipom, line 502: NOTE: `@tails[path]` should point to the caller of this function ( = self-watcher).
So I pass self-watcher directly to this function.

This way, we can avoid the impact of `refresh_watcher` being executed first and `@tails` being overwritten.

masaki-hatada: > NOTE: @tails[path] should point to the caller of this function ( = self-watcher).
> So I pass self-watcher directly to this function.

Maybe this is a good idea to prevent closing a new watcher wrongly.
Currently I'm checking this patch as code level and @kattz-kawa is evaluating this patch on our env.

lib/fluent/plugin/in_tail.rb
```diff
@@ -484,8 +484,8 @@ def close_watcher_handles
     end
 
     # refresh_watchers calls @tails.keys so we don't use stop_watcher -> start_watcher sequence for safety.
-    def update_watcher(target_info, pe)
-      path = target_info.path
+    def update_watcher(watcher, pe, new_inode)
```
ashie, line 487: nit: I prefer tail_watcher for the first argument.
```suggestion
    def update_watcher(tail_watcher, pe, new_inode)
```
In this plugin various type of `watcher`s are appeared so that `watcher`  is confusing.

Probably there are 3 local variable names to represent it in this plugin:

1. `tw`
2. `watcher`
3. `tail_watcher`

1. is also acceptable for me.
Although it's too short but not so confusing in this context, and it's appeared most often in this code.

daipom: Thanks for your review! I see!
Should we change the method name too?

```diff
- update_watcher(watcher, pe, new_inode)
+ update_tail_watcher(tail_watcher, pe, new_inode)
```

daipom: Fixed!
I used `tail_watcher`!

kattz-kawa:
Hi @daipom

Thank you for the disuccsing regarding #3614. 
And I appreciate the adding co-authored sign on this PR's commit message.

@masaki-hatada and Me have checked and tested heavily last week.
It looks good.
There is no concern for us.
I hope that this PR merged in the future release for fixing incorrect file handle issue.

masaki-hatada (approved):
Looks good for me.

lib/fluent/plugin/in_tail.rb
ashie: I agree with your idea too.

test/plugin/test_in_tail.rb
```diff
@@ -2638,4 +2638,278 @@ def test_lines_collected_with_no_throttling(data)
...
+      Fluent::FileWrapper.open("#{@tmp_dir}/tail.pos", "r") do |f|
+        f.readlines(chomp: true).each do |line|
+          values = line.split("\t")
+          position_entries.append([values[0], values[1], values[2].to_i(16)])
+        end
+      end
+
+      assert_equal(
+        {
+          # TODO: This is BUG!! We need to fix it and replace this with the next.
```
ashie, line 2708: :thinking:

test/plugin/test_in_tail.rb
```diff
@@ -2638,4 +2638,278 @@ def test_lines_collected_with_no_throttling(data)
       end
     end
   end
+
+  # https://github.com/fluent/fluentd/issues/3614
```
ashie, line 2642: It prefer #4190 instead of #3614 here to clarify the mechanism of this bug.

test/plugin/test_in_tail.rb
```diff
@@ -2638,4 +2638,278 @@ def test_lines_collected_with_no_throttling(data)
...
+          record_values: record_values,
+          tail_watcher_paths: tail_watchers.collect { |tw| tw.path },
+          tail_watcher_inodes: tail_watchers.collect { |tw| tw.ino },
+          tail_watcher_io_handler_opened_statuses: tail_watchers.collect { |tw| tw.instance_variable_get(:@io_handler)&.opened? || false },
+          position_entries: position_entries
+        },
+      )
+    end
+
+    def test_updateTW_after_refreshTW
```
ashie, line 2820: This is the case described in #4190, isn't it?
Can you add something a mark to distinguish it?
It's okay for me that Just moving the comment at the head of `sub_test_case` (`# https://github.com/fluent/fluentd/issues/3614`) to here.

test/plugin/test_in_tail.rb
ashie: It's not intended usage, rotated files should be excluded by `path` parameter although they are included in this test (`#{@tmp_dir}/tail.txt*`).

daipom:
@kattz-kawa @masaki-hatada Thanks for your review!

daipom:
@ashie Thanks for your review! I'll check it.

test/plugin/test_in_tail.rb
ashie: It's hard to detect whether tail.txt1 is a completely new file or rotated file, since same inode with a different path is possible especially on just after deleting a file (since a newly created file is often given a same inode with the file delete just before it).
So I think we don't need to care rotated files by our code, they should be excluded by the `path` pattern.

So the test code should exclude the rotated path pattern.

ashie: ... :thinking:（That should not be.)
Sorry, probably it's my misunderstanding.
Please ignore my comment.

daipom: Since the position is erased during [unwatch](https://github.com/fluent/fluentd/blob/b71ab139b25bfcb8cb3df87385a7f94c8f6fd220/lib/fluent/plugin/in_tail/position_file.rb#L56-L60), there is a possibility of collecting duplicate files after the rotation depending on timing.

This is for reducing memory consumption:

* #2803

However, it does not seem to work correctly with `follow_inodes`, which is implemented after this fix.

I want to fix it following this PR.

ashie (approved):
LGTM (above my comments are nitpicking)

test/plugin/test_in_tail.rb
daipom: Thanks! I see!
I moved that comment here and fix it to #4190.

test/plugin/test_in_tail.rb
daipom: Thanks! I see!
Fixed in https://github.com/fluent/fluentd/pull/4208#discussion_r1241854041.

ashie:
Thanks to all contributors who help investigating this issue!

ashie:
TODO:
* Apply it to v1.16 branch
* How about `follow_inode false`?

daipom:
Thanks to all contributors who have given me much insight, especially @kattz-kawa, @masaki-hatada, and @garyzjq!
And thanks for your review!

PR #4221, opened 2023-06-29
in_tail: prevent wrongly unwatching with follow_inodes

daipom:
* Following up to #4208

We must not unwatch targets that still exist.
If unwatching an existing target, it causes log duplication.
This problem exists from v1.13.3 (the following fix).

* #3466

This can occur when enabling `follow_inodes`.

The mechanism is as follows.

The TailWatcher for rotated old file can be newly created because the path is changed (the inode is the same). (file.log -> file.log.1)
If unwatching the inode before creating that new TailWatcher, it can't refer to the previous position and it reads logs in duplicate.

**Docs Changes**:
No need.

**Release Note**: 
Same as the title.

**How to Reproduce**

Config:

```xml
<source>
  @type tail
  tag test
  read_from_head true
  follow_inodes true
  refresh_interval 15s
  enable_stat_watcher false
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

* note about `enable_stat_watcher`
  * `enable_stat_watcher false` is for making sure to detect rotation.
    * I think this is problematic too... If stat_watcher notifies too quickly before the new current file is created, [TailWatcher sets `nil` to it's `rotate_handler`](https://github.com/fluent/fluentd/blob/77d7229f75a15b1d817cebb43044bbeaf6f24eb8/lib/fluent/plugin/in_tail.rb#L874).  This causes some problems. (I will create an issue or PR about this later.)
    * This may have prevented this problem from occurring in a real environment. If `in_tail` can't detect the rotation of the current file because of this, then this problem does not occur. The focus is on whether the rotation of the current file can be detected when `enable_stat_watcher` is enabled. If not, then this can occur only when disabling `enable_stat_watcher`.

Files:

```console
$ ls /test/fluentd/input4
test.log

$ cat /test/fluentd/input4/test.log
foooooaaa
Barrrrraaa
```

Script to rotate:

```sh
mv /test/fluentd/input4/test.log /test/fluentd/input4/test.log.1
echo "aaa" > /test/fluentd/input4/test.log
```

Run:

* Run Fluentd
* Run the script to rotate

Result:

```
2023-06-29 09:55:07 +0900 [info]: #0 starting fluentd worker pid=686256 ppid=686236 worker=0
2023-06-29 09:55:07 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log
2023-06-29 09:55:07.232909021 +0900 test: {"message":"foooooaaa"}
2023-06-29 09:55:07.232911541 +0900 test: {"message":"Barrrrraaa"}
2023-06-29 09:55:07 +0900 [info]: #0 fluentd worker is now running worker=0
2023-06-29 09:55:12 +0900 [info]: #0 detected rotation of /test/fluentd/input4/test.log; waiting 5 seconds
2023-06-29 09:55:12 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log
2023-06-29 09:55:12.234238467 +0900 test: {"message":"aaa"}
2023-06-29 09:55:22 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log.1
2023-06-29 09:55:22.234924294 +0900 test: {"message":"foooooaaa"}
2023-06-29 09:55:22.234935622 +0900 test: {"message":"Barrrrraaa"}
```

* The old rotated file was read in duplicate.
* This occurs if detach is done before `refresh_watcher`.
  * It is because detach of `update_watcher` wrongly unwatches the existing inode.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -502,9 +502,13 @@ module Fluent::Plugin
       new_target_info = TargetInfo.new(path, new_inode)
 
       if @follow_inodes
-        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-        # So it should be unwatched here explicitly.
-        tail_watcher.unwatched = true
+        # If the old inode still exists, we can't unwatch it.
+        # If the old inode disappears, we should unwatch it here because
+        # `refresh_watcher` can't detect that the inode disappears.
+        # (It is because the path of `@tails` will be overwritten here.)
+        # (It is not a big problem if we don't unwatch here, as long as we make
+        # sure to detach it. The position will be cleared on restart eventually.)
+        tail_watcher.unwatched = true unless expand_paths.key?(tail_watcher.ino)
 
         new_position_entry = @pf[new_target_info]
         # If `refresh_watcher` find the new file before, this will not be zero.
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2704,22 +2704,14 @@ class TailInputTest < Test::Unit::TestCase
 
       assert_equal(
         {
-          # TODO: This is BUG!! We need to fix it and replace this with the next.
-          record_values: ["file1 log1", "file1 log1", "file1 log2", "file1 log2", "file2 log1", "file2 log2"],
-          # record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
+          record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
           tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt1"],
           tail_watcher_inodes: [inode_0, inode_1, inode_0],
           tail_watcher_io_handler_opened_statuses: [false, false, false],
-          # TODO: This is BUG!! We need to fix it and replace this with the next.
           position_entries: [
-            ["#{@tmp_dir}/tail.txt", "ffffffffffffffff", inode_0],
+            ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_0],
             ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_1],
-            ["#{@tmp_dir}/tail.txt1", "0000000000000016", inode_0],
           ],
-          # position_entries: [
-          #   ["#{@tmp_dir}/tail.txt", "ffffffffffffffff", inode_0],
-          #   ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_1],
-          # ],
         },
         {
           record_values: record_values,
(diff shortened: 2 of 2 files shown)


lib/fluent/plugin/in_tail.rb
```diff
@@ -502,9 +502,13 @@ def update_watcher(tail_watcher, pe, new_inode)
...
-        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-        # So it should be unwatched here explicitly.
-        tail_watcher.unwatched = true
+        # If the old inode still exists, we can't unwatch it.
+        # If the old inode disappears, we should unwatch it here because
+        # `refresh_watcher` can't detect that the inode disappears.
+        # (It is because the path of `@tails` will be overwritten here.)
+        # (It is not a big problem if we don't unwatch here, as long as we make
+        # sure to detach it. The position will be cleared on restart eventually.)
+        tail_watcher.unwatched = true unless expand_paths.key?(tail_watcher.ino)
```
ashie, line 511: I'm worry calling `expand_paths` here, because it's relatively high cost.
It's might to too high cost on heavy rotated environment.
In addition, the file might be disappear soon just after calling expand_paths, as described in https://github.com/fluent/fluentd/blob/691b2e0c39d7d93a7d648dc5308ff961a4d1bda1/lib/fluent/plugin/in_tail.rb#L337-L348
(Although it might not so big problem as you described in the comment).

daipom: Thanks for your review!

> I'm worry calling expand_paths here, because it's relatively high cost.
> It might be too high cost on heavy rotated environment.

I see.
I thought it would be okay since it was only called on rotation, but certainly, this is worrisome considering the case of so many files being rotated so often...

It may be better to cache the list in `refresh_watcher` and use it here...
However, I'm unsure if I should go this far for unwatching here.

> (Although it might not so big problem as you described in the comment).

What we should be concerned about is that, without unwatching here, the memory and file for PositionFile would be larger.
(It will be cleared only when restarting. We can't use `pos_file_compaction_interval` to clear this because it doesn't check the target existence.)

daipom: Possible fix:

* A: Don't unwatch here.
* B: Fix `refresh_watcher` to refresh PositionFile as well.
* C: Cache the list in `refresh_watcher` and use it in `update_watcher`
* D: Change the key of `@tails` to inode when enabling `follow_inodes`

daipom: >     * A: Don't unwatch here.

We want to avoid this if possible since the memory and file for PositionFile would be larger.

>     * B: Fix `refresh_watcher` to refresh PositionFile as well.

It could be possible.

>     * C: Cache the list in `refresh_watcher` and use it in `update_watcher`

This is not good. It depends on the interval of `refresh_watcher`.

>     * D: Change the key of `@tails` to inode when enabling `follow_inodes`

It could be possible.
We should consider the following PR for this.

* #4185

ashie: I close this PR since it should be fixed by B or D.
Are you OK for closing it?

daipom: Yes! I close this! Thanks!

daipom:
We should do this in another way.

PR #4225, opened 2023-07-04
Update sigdump to 0.2.5

daipom:
None.

In 0.2.5, the bug of wrong value of `hash pairs` and `array elements` fixed.

That version will be installed without this fix, but I think we should explicitly update the version.
(Is it correct?)

We can confirm it as follows on non-Windows. (On Windows, we can use `fluent-ctl`.)

```console
$ sudo kill -CONT {pid}
$ less /tmp/sigdump-{pid}.log
```

Before: Too small Array and Hash sizes

```
...
   Array 61 elements
    Hash 8 pairs
```

After: Correct size

```
...
   Array 41,555 elements
    Hash 15,752 pairs
```

**Docs Changes**:
No need.

**Release Note**: 
Fix wrong Object Array and Hash sizes of SIGDUMP.

**ETC**:
Should we merge this into v1.16 too?

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("cool.io", [">= 1.4.5", "< 2.0.0"])
   gem.add_runtime_dependency("serverengine", [">= 2.3.2", "< 3.0.0"])
   gem.add_runtime_dependency("http_parser.rb", [">= 0.5.1", "< 0.9.0"])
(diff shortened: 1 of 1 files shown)


ashie (approved):

ashie:
> Should we merge this into v1.16 too?

Not always needed.
Even if we don't backport it, the latest version will be installed.
And we'll bundle latest version of sigdump explicitly for td-agent (fluent-package).

daipom:
Thanks for your review!

PR #4229, opened 2023-07-06
logger: Fix cache for `ignore_same_log_interval` growing unlimitedly

### LINKED ISSUE #4174: Large Memory Consumption when ignore_same_log_interval
### Describe the bug

My colleague Lester Lu and I found the issue related to https://github.com/fluent/fluentd/issues/3401

The abnormal increasing memory usage is because of the ignore repeated log feature. 
The current implementation uses dictionary to store the message as key. However, depends on the plugin implementation, the message could be the actual log that being send out. For example in azure-loganalytics, the log message is like `log.fatal "Exception occured in posting to DataCollector API: " + "'#{ex}', data=>" + Yajl.dump(records)`. The records/log could be large like 1k bytes per log and never repeat.

It leads to the dictionary `cached_log` keep growing and the memory will never be released, therefore can observe significant memory usage for some plugins.

https://github.com/fluent/fluentd/blob/0a6d706a9cee5882d751b2cc6169696709df0134/lib/fluent/log.rb#L480

### To Reproduce

Use plugin azure-loganalytics, config with an invalid account so that the log will not be send out, with large traffic 3k log/s, with ignore_same_log_interval 60s, observed memory usage keep increasing.

### Expected behavior

Memory should not increase infinitely. The `cached_log` need to be freed once a while.

### Your Environment

```markdown
- Fluentd version: 1.15.2
- TD Agent version: 4.4.1
- Operating system: Ubuntu
- Kernel version:
```

### Your Configuration

```apache
<system>
  log_level      info
  ignore_repeated_log_interval   60s
  ignore_same_log_interval       60s
</system>

<match **>
    @type azure-loganalytics
    customer_id invalidid
    shared_key anykeyshouldbefine
    log_type ApacheAccessLog
<buffer tag>
	@type                      memory
	chunk_limit_size           10M
	total_limit_size           50M
	flush_at_shutdown          true
	overflow_action            block
	retry_forever              true
	disable_chunk_backup       true
</buffer>
</match>
```

### Your Error Log

```shell
No error, abnormal memory usage.
```

### Additional context

_No response_

daipom:
* Fixes #4174

With `ignore_same_log_interval`, the cache can grow unlimitedly.
This limits the size to 1000.

Specification:

* When the size is over 1000, clear the unneeded caches.
* If the size is still over, clear all the caches.

If needed, we can make the max size configurable, but for now, I don't think we need it.

**Docs Changes**:
Adding an explanation to https://docs.fluentd.org/deployment/logging#ignore_same_log_interval is desirable.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/log.rb
+++ b/lib/fluent/log.rb
@@ -51,6 +51,8 @@ module Fluent
     LOG_TYPES = [LOG_TYPE_SUPERVISOR, LOG_TYPE_WORKER0, LOG_TYPE_DEFAULT].freeze
     LOG_ROTATE_AGE = %w(daily weekly monthly)
 
+    IGNORE_SAME_LOG_MAX_CACHE_SIZE = 1000 # If need, make this an option of system config.
+
     def self.str_to_level(log_level_str)
       case log_level_str.downcase
       when "trace" then LEVEL_TRACE
@@ -477,6 +479,13 @@ module Fluent
           false
         end
       else
+        if cached_log.size >= IGNORE_SAME_LOG_MAX_CACHE_SIZE
+          cached_log.reject! do |_, cached_time|
+            (time - cached_time) > @ignore_same_log_interval
+          end
+        end
+        # If the size is still over, we have no choice but to clear it.
+        cached_log.clear if cached_log.size >= IGNORE_SAME_LOG_MAX_CACHE_SIZE
         cached_log[message] = time
         false
       end
--- a/test/test_log.rb
+++ b/test/test_log.rb
@@ -472,6 +472,43 @@ class LogTest < Test::Unit::TestCase
       ]
       assert_equal(expected, log.out.logs)
     end
+
+    def test_reject_on_max_size
+      ignore_same_log_interval = 10
+
+      logger = Fluent::Log.new(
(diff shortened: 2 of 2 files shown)


lib/fluent/log.rb
```diff
@@ -51,6 +51,8 @@ module TTYColor
     LOG_TYPES = [LOG_TYPE_SUPERVISOR, LOG_TYPE_WORKER0, LOG_TYPE_DEFAULT].freeze
     LOG_ROTATE_AGE = %w(daily weekly monthly)
 
+    IGNORE_SAME_LOG_MAX_CACHE_SIZE = 100 # If need, make this an option of system config.
```
ashie, line 54: It might be better to make it larger value for safety, e.g. `1000`.
I think there is no impact to memory & CPU usage at this level.

daipom: Thanks for your review!
I see! I will change it.

daipom: DONE!

ashie (approved):
LGTM
Waiting CI...

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4237, opened 2023-07-12
in_tail: prevent wrongly unwatching with follow_inodes

daipom:
* Following up to #4208
* Another idea of #4221

We must not unwatch targets that still exist.
If unwatching an existing target, it causes log duplication.

This problem exists from v1.13.3 (the following fix).

* #3466

This can occur when enabling `follow_inodes`.

In the existing implementation, `update_watcher` needs to unwatch the old TailWatcher since `refresh_watcher` can't unwatch it when `update_watcher` is called first (when rotation happens).
It is because `update_watcher` discards the old TailWatcher and `refresh_watcher` can't recognize the old inode is disappeared.

https://github.com/fluent/fluentd/blob/e12069351a431b30bd84996d60ff9c4a7e78adce/lib/fluent/plugin/in_tail.rb#L504-L507

However, it can wrongly unwatch an existing inode because the old inode may still exist.
(See the diff of test cases.)

Thus, we need a new mechanism to unwatch targets correctly.

This fix is based on the idea that we should unwatch based on directly on PositionFile's data.

**Docs Changes**:
No need.

**Release Note**: 
Same as the title.

**How to Reproduce**

Config:

```xml
<source>
  @type tail
  tag test
  read_from_head true
  follow_inodes true
  refresh_interval 15s
  enable_stat_watcher false
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

* note about `enable_stat_watcher`
  * `enable_stat_watcher false` is for making sure to detect rotation.
    * I think this is problematic too... If stat_watcher notifies too quickly before the new current file is created, [TailWatcher sets `nil` to it's `rotate_handler`](https://github.com/fluent/fluentd/blob/77d7229f75a15b1d817cebb43044bbeaf6f24eb8/lib/fluent/plugin/in_tail.rb#L874).  This causes some problems. (I will create an issue or PR about this later.)
    * This may have prevented this problem from occurring in a real environment. If `in_tail` can't detect the rotation of the current file because of this, then this problem does not occur. The focus is on whether the rotation of the current file can be detected when `enable_stat_watcher` is enabled. If not, then this can occur only when disabling `enable_stat_watcher`.

Files:

```console
$ ls /test/fluentd/input4
test.log

$ cat /test/fluentd/input4/test.log
foooooaaa
Barrrrraaa
```

Script to rotate:

```sh
mv /test/fluentd/input4/test.log /test/fluentd/input4/test.log.1
echo "aaa" > /test/fluentd/input4/test.log
```

Run:

* Run Fluentd
* Run the script to rotate

Result:

```
2023-06-29 09:55:07 +0900 [info]: #0 starting fluentd worker pid=686256 ppid=686236 worker=0
2023-06-29 09:55:07 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log
2023-06-29 09:55:07.232909021 +0900 test: {"message":"foooooaaa"}
2023-06-29 09:55:07.232911541 +0900 test: {"message":"Barrrrraaa"}
2023-06-29 09:55:07 +0900 [info]: #0 fluentd worker is now running worker=0
2023-06-29 09:55:12 +0900 [info]: #0 detected rotation of /test/fluentd/input4/test.log; waiting 5 seconds
2023-06-29 09:55:12 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log
2023-06-29 09:55:12.234238467 +0900 test: {"message":"aaa"}
2023-06-29 09:55:22 +0900 [info]: #0 following tail of /test/fluentd/input4/test.log.1
2023-06-29 09:55:22.234924294 +0900 test: {"message":"foooooaaa"}
2023-06-29 09:55:22.234935622 +0900 test: {"message":"Barrrrraaa"}
```

* The old rotated file was read in duplicate.
* This occurs if detach is done before `refresh_watcher`.
  * It is because detach of `update_watcher` wrongly unwatches the existing inode.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -370,17 +370,30 @@ module Fluent::Plugin
     def refresh_watchers
       target_paths_hash = expand_paths
       existence_paths_hash = existence_path
-      
+
       log.debug {
         target_paths_str = target_paths_hash.collect { |key, target_info| target_info.path }.join(",")
         existence_paths_str = existence_paths_hash.collect { |key, target_info| target_info.path }.join(",")
         "tailing paths: target = #{target_paths_str} | existing = #{existence_paths_str}"
       }
 
-      unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
+      if !@follow_inodes
+        need_unwatch_in_stop_watchers = true
+      else
+        # When using @follow_inodes, need this to unwatch the rotated old inode when it disappears.
+        # After `update_watcher` detaches an old TailWatcher, the inode is lost from the `@tails`.
+        # So that inode can't be contained in `removed_hash`, and can't be unwatched by `stop_watchers`.
+        #
+        # This logic may work for `@follow_inodes false` too.
+        # Just limiting the case to supress the impact to existing logics.
+        @pf&.unwatch_removed_targets(target_paths_hash)
+        need_unwatch_in_stop_watchers = false
+      end
+
+      removed_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
       added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
 
-      stop_watchers(unwatched_hash, immediate: false, unwatched: true) unless unwatched_hash.empty?
+      stop_watchers(removed_hash, unwatched: need_unwatch_in_stop_watchers) unless removed_hash.empty?
       start_watchers(added_hash) unless added_hash.empty?
       @startup = false if @startup
     end
@@ -502,10 +515,6 @@ module Fluent::Plugin
       new_target_info = TargetInfo.new(path, new_inode)
 
       if @follow_inodes
-        # When follow_inodes is true, it's not cleaned up by refresh_watcher.
-        # So it should be unwatched here explicitly.
-        tail_watcher.unwatched = true
-
         new_position_entry = @pf[new_target_info]
         # If `refresh_watcher` find the new file before, this will not be zero.
         # In this case, only we have to do is detaching the current tail_watcher.
--- a/lib/fluent/plugin/in_tail/position_file.rb
+++ b/lib/fluent/plugin/in_tail/position_file.rb
@@ -53,10 +53,16 @@ module Fluent::Plugin
         }
       end
 
+      def unwatch_removed_targets(existing_targets)
+        @map.reject { |key, entry|
+          existing_targets.key?(key)
+        }.each_key { |key|
+          unwatch_key(key)
+        }
+      end
+
       def unwatch(target_info)
-        if (entry = @map.delete(@follow_inodes ? target_info.ino : target_info.path))
-          entry.update_pos(UNWATCHED_POSITION)
-        end
+        unwatch_key(@follow_inodes ? target_info.ino : target_info.path)
       end
 
       def load(existing_targets = nil)
@@ -117,6 +123,12 @@ module Fluent::Plugin
       end
 
       private
+
+      def unwatch_key(key)
+        if (entry = @map.delete(key))
+          entry.update_pos(UNWATCHED_POSITION)
+        end
+      end
 
       def compact(existing_targets = nil)
         @file_mutex.synchronize do
--- a/test/plugin/in_tail/test_position_file.rb
+++ b/test/plugin/in_tail/test_position_file.rb
@@ -26,6 +26,10 @@ class IntailPositionFileTest < Test::Unit::TestCase
     "valid_path" => Fluent::Plugin::TailInput::TargetInfo.new("valid_path", 1),
     "inode23bit" => Fluent::Plugin::TailInput::TargetInfo.new("inode23bit", 0),
   }
+  TEST_CONTENT_INODES = {
+    1 => Fluent::Plugin::TailInput::TargetInfo.new("valid_path", 1),
+    0 => Fluent::Plugin::TailInput::TargetInfo.new("inode23bit", 0),
+  }
 
   def write_data(f, content)
     f.write(content)
@@ -221,7 +225,7 @@ class IntailPositionFileTest < Test::Unit::TestCase
   end
 
   sub_test_case '#unwatch' do
-    test 'deletes entry by path' do
+    test 'unwatch entry by path' do
       write_data(@file, TEST_CONTENT)
       pf = Fluent::Plugin::TailInput::PositionFile.load(@file, false, {}, logger: $log)
       inode1 = File.stat(@file).ino
@@ -238,6 +242,32 @@ class IntailPositionFileTest < Test::Unit::TestCase
       assert_equal Fluent::Plugin::TailInput::FilePositionEntry, p2.class
 
       assert_not_equal p1, p2
+    end
+
+    test 'unwatch entries by inode' do
+      write_data(@file, TEST_CONTENT)
+      pf = Fluent::Plugin::TailInput::PositionFile.load(@file, true, TEST_CONTENT_INODES, logger: $log)
+
+      existing_targets = TEST_CONTENT_INODES.select do |inode, target_info|
+        inode == 1
+      end
+      pe_to_unwatch = pf[TEST_CONTENT_INODES[0]]
+
+      pf.unwatch_removed_targets(existing_targets)
+
+      assert_equal(
+        {
+          map_keys: [TEST_CONTENT_INODES[1].ino],
+          unwatched_pe_pos: Fluent::Plugin::TailInput::PositionFile::UNWATCHED_POSITION,
+        },
+        {
+          map_keys: pf.instance_variable_get(:@map).keys,
+          unwatched_pe_pos: pe_to_unwatch.read_pos,
+        }
+      )
+
+      unwatched_pe_retaken = pf[TEST_CONTENT_INODES[0]]
+      assert_not_equal pe_to_unwatch, unwatched_pe_retaken
     end
   end
 
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2704,22 +2704,14 @@ class TailInputTest < Test::Unit::TestCase
 
       assert_equal(
         {
-          # TODO: This is BUG!! We need to fix it and replace this with the next.
-          record_values: ["file1 log1", "file1 log1", "file1 log2", "file1 log2", "file2 log1", "file2 log2"],
-          # record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
+          record_values: ["file1 log1", "file1 log2", "file2 log1", "file2 log2"],
           tail_watcher_paths: ["#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt1"],
           tail_watcher_inodes: [inode_0, inode_1, inode_0],
           tail_watcher_io_handler_opened_statuses: [false, false, false],
-          # TODO: This is BUG!! We need to fix it and replace this with the next.
           position_entries: [
-            ["#{@tmp_dir}/tail.txt", "ffffffffffffffff", inode_0],
+            ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_0],
             ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_1],
-            ["#{@tmp_dir}/tail.txt1", "0000000000000016", inode_0],
           ],
-          # position_entries: [
-          #   ["#{@tmp_dir}/tail.txt", "ffffffffffffffff", inode_0],
-          #   ["#{@tmp_dir}/tail.txt", "0000000000000016", inode_1],
-          # ],
         },
         {
           record_values: record_values,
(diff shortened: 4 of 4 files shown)


lib/fluent/plugin/in_tail/position_file.rb
```diff
@@ -53,10 +53,12 @@ def [](target_info)
...
-      def unwatch(target_info)
-        if (entry = @map.delete(@follow_inodes ? target_info.ino : target_info.path))
-          entry.update_pos(UNWATCHED_POSITION)
-        end
+      def unwatch_removed_targets(existing_targets)
+        @map.reject { |key, entry|
+          existing_targets.key?(key)
+        }.each_key { |key|
+          unwatch(key)
+        }
```
daipom, line 61: Note: I think we don't need entire exclusive lock here.

Each `unwatch` process has the lock and `try_compact` safely updates the seek positions in one lock.

https://github.com/fluent/fluentd/blob/e12069351a431b30bd84996d60ff9c4a7e78adce/lib/fluent/plugin/in_tail/position_file.rb#L101-L116

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,17 +370,19 @@ def existence_path
...
       }
 
-      unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
+      @pf&.unwatch_removed_targets(target_paths_hash)
+
+      removed_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
       added_hash = target_paths_hash.reject {|key, value| existence_paths_hash.key?(key)}
 
-      stop_watchers(unwatched_hash, immediate: false, unwatched: true) unless unwatched_hash.empty?
+      stop_watchers(removed_hash, immediate: false) unless removed_hash.empty?
```
ashie, line 385: I'm not sure whether we can safely remove `unwatched` for `follow_inodes false` case.

daipom:
TODO: I'm checking if it is OK to unwatch the position tied to TailWatcher that is not detached yet.

daipom:
https://github.com/fluent/fluentd/actions/runs/5526720226/jobs/10081747103?pr=4237

The test `test_move_truncate_move_back` looks unstable on macOS. I will check if it is related to this fix.

lib/fluent/plugin/in_tail.rb
ashie: In `follow_inodes false` case it's safer not to change the behaviour for now.
The issue we currently want to solve is only the log duplication issue on `follow_inodes false` case.

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,17 +370,19 @@ def existence_path
...
-      
+
       log.debug {
         target_paths_str = target_paths_hash.collect { |key, target_info| target_info.path }.join(",")
         existence_paths_str = existence_paths_hash.collect { |key, target_info| target_info.path }.join(",")
         "tailing paths: target = #{target_paths_str} | existing = #{existence_paths_str}"
       }
 
-      unwatched_hash = existence_paths_hash.reject {|key, value| target_paths_hash.key?(key)}
+      @pf&.unwatch_removed_targets(target_paths_hash)
```
masaki-hatada, line 380: Although I might overlook something, but could you let me ask the following?

I considered the pattern that /var/log/test1.log was removed then its inode(inode-1) was reused by /var/log/test2.log in the setting `path = /var/log/*.log*`.

- target_paths_hash will include inode-1, since /var/log/test2.log has the same inode as the removed /var/log/test1.log.
- So `@pf&.unwatch_removed_targets(target_paths_hash)` will skip inode-1 so the position info of inode-1 won't be reset.
- `removed_hash` also won't include inode-1 due to the same reason.

In this case, won't fluentd misrecognize the previous position for /var/log/test2.log?

lib/fluent/plugin/in_tail.rb
daipom: Thanks for your review!
I see!
I have limited the impact only to `follow_inodes true`!

daipom:
I fixed this so that this logic works only for `follow_inodes true`.
Now this does not change the behavior of `follow_indoes false`.

daipom:
> TODO: I'm checking if it is OK to unwatch the position tied to TailWatcher that is not detached yet.

This fix changes that unwatching is done immediately (without waiting `rotate_wait`). 
I think this can change some behaviors...
To limit the impact, we should ensure that unwatching is done at the same time as detaching, just as before.
However, ensuring this will be a high-impact fix. So I gave up. :cry: 

Even if the existing TailWatcher's position is unwatched, the target file is already removed.
So I think we don't need to wait `rotate_wait` for unwatching...
but I worry that it might result in unexpected changes in behavior...

daipom:
> but I worry that it might result in unexpected changes in behavior...

The following test failure is one example of this.
This test is for `follow_inodes false`, so now it will succeed, but I think the same phenomenon can occur for `follow_inodes true` as well.

> https://github.com/fluent/fluentd/actions/runs/5526720226/jobs/10081747103?pr=4237
> 
> The test `test_move_truncate_move_back` looks unstable on macOS. I will check if it is related to this fix.
> 
> Windows too. It may have something to do with this fix.
> 
> https://github.com/fluent/fluentd/actions/runs/5526720222/jobs/10081747201?pr=4237

It is difficult to explain why this test became unstable before rebasing this branch.
We need to understand the following scenario and mechanism first.

* While the existing TailWatcher is detaching and waiting for the `rotate_wait` interval, new TailWatcher can be created that collects the same target.
* In this case, there are multiple TailWatchers for the same target, so this risks log duplication.
* However, log duplication does not occur because those multiple TailWatchers refer to the same PositionEntry.

This fix changes this behavior.
If referring to the inode that is already unwatched, then PositionFile returns a new PositionEntry.
So, if the old TailWatcher is unwatched immediately without waiting `rotate_wait`, the new TailWatcher for the same target will have another PositionEntry for the same target.

Because of the following logic, TailWathcer can update the position to zero even after being unwatched.

https://github.com/fluent/fluentd/blob/e12069351a431b30bd84996d60ff9c4a7e78adce/lib/fluent/plugin/in_tail.rb#L836-L843

Thus, this fix can cause the log duplication in this scenario.

This happens only when the target having the same inode resurrects (disappears and appears again) in the interval `rotate_wait`.

daipom:
About https://github.com/fluent/fluentd/pull/4237#issuecomment-1632176184, I think we should fix the entire logic so that a new TailWatcher for the same target should be added after the detach of the old TailWatcher is completed.

But it doesn't look like we'll have time to get there before the next release of 1.16... :cry:

lib/fluent/plugin/in_tail.rb
masaki-hatada: > That scenario can happen in the current master branch, right?

I wanted to say that https://github.com/fluent/fluentd/pull/4237#discussion_r1260712488 is the side effect introduced by adding `@pf&.unwatch_removed_targets(target_paths_hash)` in your patch.

But maybe I was wrong, because `added_hash` also won't include inode-1 in my suggested pattern.
I rewrote the scenario a bit as follows.

In the pattern that `/var/log/test1.log` was removed then its inode(inode-1) was reused by `/var/log/test2.log` in the setting `path = /var/log/*.log*`:
- `target_paths_hash` will include inode-1, since `/var/log/test2.log` has the same inode as the removed `/var/log/test1.log`.
  - It means that `@pf&.unwatch_removed_targets(target_paths_hash)` will skip inode-1 so the position info of inode-1 won't be reset.
  - `removed_hash` also won't include inode-1 due to the same reason.
  - `added_hash` also won't include inode-1 due to the same reason.
- So `refresh_watchers()` will call neither `start_watchers()` nor `stop_watchers()` for both of `/var/log/test1.log` and `/var/log/test2.log`.

Yes, the above scenario can happen even in the current master branch.

The situation will be resolved after a new `/var/log/test1.log` was created and `@tails["/var/log/test1.log"]` got updated with a new inode.
So it won't be a big problem in real environment, I believe...

daipom: Is it rotation? (Is the `test1.log` the same file as `test2.log`?)

If so, it is correct that `refresh_watcher` does nothing for `inode-1` because it means the target `inode-1` still exists.
In this case, `update_watcher` update the TailWatcher.

However, I wonder the old TailWathcer can collect duplicate log in `rotate_wait` interval...
I'll check it.

daipom: > However, I wonder the old TailWathcer can collect duplicate log in rotate_wait interval...
I'll check it.

This may not be the point that you were originally concerned about, but I found this can cause log duplication.
This seems to be a problem that has existed for a long time.

This is a similar problem to https://github.com/fluent/fluentd/pull/4237#issuecomment-1632176184
It seems that in_tail has a log duplication problem since a long time ago with `rotate_wait`.

I will explain this in another comment.

daipom:
I have commented on some cases where multiple TailWatcher exists due to `rotate_wait`.
I have checked the behavior of v1.16.1, and I found out that in_tail has this problem since a long time ago.

So now I think we should not be concerned about that point in this PR.
(It's been wrong since a long time ago)

https://github.com/fluent/fluentd/pull/4237#discussion_r1261830930
https://github.com/fluent/fluentd/pull/4237#issuecomment-1632176184

The scenario: rotate -> update_watcher -> refresh_watcher -> `rotate_wait` elapsed

* Initial state

```
path-1: inode-b, TailWatcher-1 
path-2: inode-a, TailWathcer-2
```

* Rotate happens

```
path-1: inode-c, TailWatcher-1-new
path-2: inode-b, TailWathcer-2-new, TailWatcher-1-detaching
path-3: inode-a, TailWatcher-2-detaching
```

* refresh_watcher

```
path-1: inode-c, TailWatcher-1-new
path-2: inode-b, TailWathcer-2-new, TailWatcher-1-detaching
path-3: inode-a, TailWatcher-2-detaching, TailWatcher-3
```

* `rotate_wait` elapsed

```
path-1: inode-c, TailWatcher-1-new
path-2: inode-b, TailWathcer-2-new
path-3: inode-a, TailWatcher-3
```

As above, `rotate_wait` causes the situation that multiple TailWatchers exist for the same target temporarily.
This causes log duplication.
I have confirmed this on v1.16.1.

```xml
<source>
  @type tail
  tag test
  path /path/to/test.log*
  read_from_head true
  follow_inodes true
  refresh_interval 5s
  enable_stat_watcher false # To ensure that TailWathcer recognizes rotation
  rotate_wait 30s
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

* Start to tail the following file

```
test.log
```

* Rotate

```
test.log
test.log.1
```

* Append a log to `test.log.1`.
  * When `refresh_watchers` NOT done and `rotate_wait` NOT elapsed
    * No duplication
  * When `refresh_watchers` done and `rotate_wait` NOT elapsed
    * DUPLICATION
  * When `refresh_watchers` done and `rotate_wait` elapsed
    * No duplication

lib/fluent/plugin/in_tail.rb
daipom: I commented on this in https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632

daipom:
In some cases, as in the case `test_move_truncate_move_back `, it seems that referring to the same PositionEntry can prevent duplication, but in other cases, this has not been possible before.
(We can reproduce it easily like https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632)

This problem should be solved fundamentally.
(in_tail should wait for the old TailWatcher being completely detached)

ashie:
> The test test_move_truncate_move_back looks unstable on macOS. I will check if it is related to this fix.

The situation that this test cares seems rare, and it doesn't necessarily have to be resolved on `follow_inodes` case, because files are assumed to be rotated normally without such odd behaviour in this case.

> This happens only when the target having the same `inode` resurrects (disappears and appears again) with the smaller file size in the interval rotate_wait.

There are 2 possible reasons for same `inode` resurrection:

1. File is moved and moved back like `test test_move_truncate_move_back`
2. The `inode` is reused for a new file

For 1, we don't always need to resolve it (at least in this PR) as I described above.

For 2, it's not possible in this case because the `TailWatcher` keeps the file opened while `rotate_wait`.
(Although there is an exception `open_on_every_update`, `rotate_wait` doesn't happen in this case.)
When someone still open the file, the `inode` won't be reused.
Here is a simplified verification for it:

test.rb:
```ruby
#!/usr/bin/env ruby                                                                                                                                           

require 'fileutils'

FileUtils.rm_rf("hoge.txt")

File.open("hoge.txt", "wb") { |f| f.puts("hoge") }

s1 = File.stat("hoge.txt")
puts "ino: #{s1.ino}"

if ARGV[0] == "keep-opened"
  f = File.open("hoge.txt", "wb")
end

FileUtils.rm_rf("hoge.txt")

File.open("hoge.txt", "wb") { |f| f.puts("hage") }

s2 = File.stat("hoge.txt")
puts "ino: #{s2.ino}, same-inode?: #{s1.ino == s2.ino}"
```

```console
$ ./test.rb
ino: 20192736
ino: 20192736, same-inode?: true
$ ./test.rb keep-opened
ino: 20192736
ino: 20197162, same-inode?: false
```

ashie (approved):
> I have commented on some cases where multiple TailWatcher exists due to rotate_wait.
> I have checked the behavior of v1.16.1, and I found out that in_tail has this problem since a long time ago.

Not needed to resolve all of this kine of problems by this PR.
Since we need more time to resolve it fundamentally, I think that's enough for now as it solves part of them.

daipom:
Thanks for your review! I agree.
I will create an issue for https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 later

ashie:
Thanks!

daipom:
Thanks! @ashie @masaki-hatada

PR #4239, opened 2023-07-13
in_tail: Ensure to discard TailWatcher with missing target when follow_inodes

daipom:
Not related to a particular issue, but it can have something to do with reported log duplication or resource leak phenomenon about `in_tail`.

Ensure to discard TailWatcher with missing target when follow_inodes.

For example, when a rotation process is slow, there is a small time lag between moving and adding files.
There is a possibility that StatWatcher notifies too quickly to the TailWatcher before the new file is moved to that target path. From the TailWatcher, it appears as if the file is resurrected once it disappeared.

In this case, `refresh_watcher` can't recognize it, so TailWatcher needs to discard self correctly.
In the previous implementation, it was not done.
So it caused the handle leak and log duplication.
(Please check the added test case)

Because of the problem https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632, log duplication can still occur during `rotate_wait`.
We need to set `rotate_wait 0` for a workaround.
(Duplication will occur if `refresh_watcher` is called during the `rotate_wait`. We need the mechanism not to add duplicated TailWathcer with detaching TailWatcher.)

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

**How to Reproduce**:

```xml
<source>
  @type tail
  tag test
  path /path/to/test.log*
  read_from_head true
  follow_inodes true
  refresh_interval 15s
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

* A: initial start

```
test.log
```

* B: move the file

```
test.log1
```

* C: add the new current file

```
test.log
test.log1
```

* D: wait `rerfresh_watcher` TWICE
  * first: add TailWatcher for `test.log` (new inode)
    * this overwrites `@tails[test.log]` so old inode information is lost from `@tails` at this timing
  * second: add TailWatcher for `test.log1` (old inode)
* E:  append logs to `test.log1`
  * Because now there are 2 TailWatchers for `test.log1`, log duplication occurs.
    * `TailWatcher(test.log, old inode)` (orphan) (This actually opens the file handle of `test.log1`)
    * `TailWatcher(test.log1, old inode)`

When StatWatcher (or TimeTrigger) notifies between B and C, then the TailWatcher becomes an orphan, and a handle leak happens.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -498,6 +498,24 @@ module Fluent::Plugin
 
     # refresh_watchers calls @tails.keys so we don't use stop_watcher -> start_watcher sequence for safety.
     def update_watcher(tail_watcher, pe, new_inode)
+      # TODO we should use another callback for this.
+      # To supress impact to existing logics, limit the case to `@follow_inodes`.
+      # We may not need `@follow_inodes` condition.
+      if @follow_inodes && new_inode.nil?
+        # nil inode means the file disappeared, so we only need to stop it.
+        @tails.delete(tail_watcher.path)
+        # https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 
+        # Because of this problem, log duplication can occur during `rotate_wait`.
+        # Need to set `rotate_wait 0` for a workaround.
+        # Duplication will occur if `refresh_watcher` is called during the `rotate_wait`.
+        # In that case, `refresh_watcher` will add the new TailWatcher to tail the same target,
+        # and it causes the log duplication.
+        # (Other `detach_watcher_after_rotate_wait` may have the same problem.
+        #  We need the mechanism not to add duplicated TailWathcer with detaching TailWatcher.)
+        detach_watcher_after_rotate_wait(tail_watcher, pe.read_inode)
+        return
+      end
+
       path = tail_watcher.path
 
       log.info("detected rotation of #{path}; waiting #{@rotate_wait} seconds")
@@ -890,10 +908,14 @@ module Fluent::Plugin
 
           if watcher_needs_update
             if @follow_inodes
-              # No need to update a watcher if stat is nil (file not present), because moving to inodes will create
-              # new watcher, and old watcher will be closed by stop_watcher in refresh_watchers method
-              # don't want to swap state because we need latest read offset in pos file even after rotate_wait
-              @update_watcher.call(self, @pe, stat.ino) if stat
+              # If stat is nil (file not present), NEED to stop and discard this watcher.
+              #   When the file is disappeared but is resurrected soon, then `#refresh_watcher`
+              #   can't recognize this TailWatcher needs to be stopped.
+              #   This can happens when the file is rotated.
+              #   If a notify comes before the new file for the path is created during rotation,
+              #   then it appears as if the file was resurrected once it disappeared.
+              # Don't want to swap state because we need latest read offset in pos file even after rotate_wait
+              @update_watcher.call(self, @pe, stat&.ino)
             else
               # Permit to handle if stat is nil (file not present).
               # If a file is mv-ed and a new file is created during
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2899,5 +2899,122 @@ class TailInputTest < Test::Unit::TestCase
         },
       )
     end
+
+    def test_path_resurrection
+      config = config_element(
+        "ROOT",
+        "",
+        {
+          "path" => "#{@tmp_dir}/tail.txt*",
+          "pos_file" => "#{@tmp_dir}/tail.pos",
+          "tag" => "t1",
+          "format" => "none",
+          "read_from_head" => "true",
+          "follow_inodes" => "true",
+          # In order to reproduce the same condition stably, ensure that `refresh_watchers` is not
+          # called by a timer.
+          "refresh_interval" => "1h",
+          # https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632
+          # Because of this problem, log duplication can occur during `rotate_wait`.
+          # Need to set `rotate_wait 0` for a workaround.
+          "rotate_wait" => "0s",
+        }
+      )
+      d = create_driver(config, false)
+
+      tail_watchers = []
+      stub.proxy(d.instance).setup_watcher do |tw|
+        tail_watchers.append(tw)
+        tw
+      end
+
+      Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") {|f| f.puts "file1 log1"}
+
+      d.run(expect_records: 5, timeout: 10) do
+        # Rotate
+        Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "ab") {|f| f.puts "file1 log2"}
+        FileUtils.move("#{@tmp_dir}/tail.txt", "#{@tmp_dir}/tail.txt" + "1")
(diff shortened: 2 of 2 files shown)


daipom:
I will check this

* https://github.com/fluent/fluentd/actions/runs/5539804706/jobs/10111142203?pr=4239

daipom:
> I will check this
> 
>     * https://github.com/fluent/fluentd/actions/runs/5539804706/jobs/10111142203?pr=4239

I think this is just a list order problem. I will improve the test.

daipom:
It seems that log duplication and handle leak still can occur on macOS...
I will check it 😢 

https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141849?pr=4239
https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141988?pr=4239

daipom:
> It seems that log duplication and handle leak still can occur on macOS... I will check it 😢
> 
> https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141849?pr=4239 https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141988?pr=4239

It seems that this can happen only on macOS.
I can reproduce this on my macOS (Ventura, Apple M1) too, but I can't reproduce it on my Windows 10.

daipom:
> > It seems that log duplication and handle leak still can occur on macOS... I will check it 😢
> > https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141849?pr=4239 https://github.com/fluent/fluentd/actions/runs/5539804708/jobs/10111141988?pr=4239
> 
> It seems that this can happen only on macOS. I can reproduce this on my macOS (Ventura, Apple M1) too, but I can't reproduce it on my Windows 10.

It can happen on Windows and macOS because StatWatcher does not work on those platforms (specification).
So this is merely a matter of the interval time between moving and adding files.
I will change the sleep time a bit longer for those platform to make the test valid for those platforms.

daipom:
The above problems are fixed.

daipom:
I'm sorry. There was an easy mistake.

https://github.com/fluent/fluentd/actions/runs/5539804706/jobs/10111142203?pr=4239

daipom:
Rebased to the latest master.

lib/fluent/plugin/in_tail.rb
```diff
@@ -498,6 +498,19 @@ def close_watcher_handles
...
+      # TODO we should use another callback for this.
+      # To supress impact to existing logics, limit the case to `@follow_inodes`.
+      # We may not need `@follow_inodes` condition.
+      if @follow_inodes && new_inode.nil?
+        # nil inode means the file disappeared, so we only need to stop it.
+        @tails.delete(tail_watcher.path)
+        # Detach should be done immediately to avoid log duplication.
+        # (Other `detach_watcher_after_rotate_wait` may have the same problem.
+        #  We need another mechanism not to add duplicated TailWathcer with detaching TailWatcher.)
+        detach_watcher(tail_watcher, pe.read_inode)
```
daipom, line 510: I fixed the comment.
This is related to https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632.
If we use `detach_watcher_after_rotate_wait` here, we can't avoid the problem that this PR wants to fix.

ashie: Now I understand the bug, but cannot understand why we cannot `detach_watcher_after_rotate_wait` here yet.

daipom: Because this fix mainly focuses on rotation cases, not cases where the file is actually removed.
In rotation cases, the new file is added very soon,
If waiting `rotate_wait`, the problem of https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 definitely happens and causes log duplication.

daipom: > * A: initial start
> 
> ```
> test.log
> ```
> 
>     * B: move the file
> 
> 
> ```
> test.log1
> ```
> 
>     * C: add the new current file
> 
> 
> ```
> test.log
> test.log1
> ```
> 
> When StatWatcher (or TimeTrigger) notifies between B and C, then the TailWatcher becomes an orphan, and a handle leak happens.

I consider the case where the notifying happens between B and C.
After C, the new file `test.log` will be added very soon.

It's an issue of which is faster, `StatWatcher`(or `TimeTrigger`) or rotation process, in intervals of less than a second.

daipom: > If waiting rotate_wait, the problem of https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 definitely happens and causes log duplication.

Oh, sorry. Not `definitely`.
This is depending on the timing of `refresh_watcher`.

If `refresh_watcher` add the new TailWatcher during `rotate_wait`, then log duplication occurs.

daipom: You can see this problem by changing this line to `detach_watcher_after_rotate_wait ` and running `$ bundle exec rake test TESTOPTS='-t"TailInputTest::Update watchers for rotation with follow_inodes"'`.

ashie: > If waiting `rotate_wait`, the problem of https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 definitely happens and causes log duplication.

Ah I see, thanks.

It's a tough issue. It should be `detach_watcher_after_rotate_wait` when we resolve duplicate TailWatcher issue to support both rotation & removing.
One of solution or work around is setting `rotate_wait` to `0`. When all rotated files are expected to be tracked by `follow_inode`, `rotate_wait` isn't mostly needed. But whether or not it actually tails all rotated files is depend on user's setting, so we can't assume that all rotated files are tracked.

daipom: > It should be detach_watcher_after_rotate_wait when we resolve duplicate TailWatcher issue to support both rotation & removing.

Yes! I thinks so too!

daipom: > One of solution or work around is setting rotate_wait to 0. When all rotated files are expected to be tracked by follow_inode, rotate_wait isn't mostly needed. But whether or not it actually tails all rotated files is depend on user's setting, so we can't assume that all rotated files are tracked.

Yes. I think so too.

daipom: Changing this to `detach_watcher_after_rotate_wait ` is possible for now.
At least, it will solve the handle leak problem, and duplicate logs are less likely to occur.

ashie: > If waiting `rotate_wait`, the problem of https://github.com/fluent/fluentd/pull/4237#issuecomment-1633358632 definitely happens and causes log duplication.

Could you add this URL to the comment?

> Changing this to `detach_watcher_after_rotate_wait` is possible for now.

Hmm, it's hard to judge but I think it's better to change for consistency of the current code.
Users can work around it by `rotate_wait 0`.

daipom: I see! I will change those points!

daipom: Done as https://github.com/fluent/fluentd/compare/f87ab7b96cf5618730a781bd1dd4dbdbea6c1ab2..781e31319649e0a229dedd7abca8bc8d20a76034

daipom: I have updated the issue comment as well.

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4282, opened 2023-08-29
plugin/in_http: recognize CSP reports as JSON data

GiedriusS:
To make it easy to use fluentd for CSP reporting, let's parse retrieved data as JSON if the content type is `application/csp-report`. It is guaranteed to be a well-formatted JSON according to the specification: https://www.w3.org/TR/CSP2/#violation-reports.

### DIFF
--- a/test/plugin/test_in_http.rb
+++ b/test/plugin/test_in_http.rb
@@ -517,6 +517,28 @@ class HttpInputTest < Test::Unit::TestCase
     assert_equal_event_time time, d.events[1][1]
   end
 
+  def test_csp_report
+    d = create_driver
+    time = event_time("2011-01-02 13:14:15 UTC")
+    time_i = time.to_i
+    events = [
+      ["tag1", time, {"a"=>1}],
+      ["tag2", time, {"a"=>2}],
+    ]
+    res_codes = []
+
+    d.run(expect_records: 2) do
+      events.each do |tag, t, record|
+        res = post("/#{tag}?time=#{time_i.to_s}", record.to_json, {"Content-Type"=>"application/csp-report; charset=utf-8"})
+        res_codes << res.code
+      end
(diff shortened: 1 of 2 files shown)


daipom (commented):
Thanks for your contribution!

I'm not familiar with [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), but I understand it has a feature to send reports.

* https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#enabling_reporting

This fix is to allow the reports to be received by `in_http` plugin, right?

test/plugin/test_in_http.rb
```diff
@@ -517,6 +517,28 @@ def test_application_json
...
   end
 
+  def test_csp_report
+    d = create_driver
+    time = event_time("2011-01-02 13:14:15 UTC")
+    time_i = time.to_i
+    events = [
+      ["tag1", time, {"a"=>1}],
+      ["tag2", time, {"a"=>2}],
+    ]
```
ashie, line 527: It would be better that the sample contents contains actual syntax of CSP report:

* [report syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only#violation_report_syntax)
* [Sample violation report](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only#sample_violation_report)

ashie (changes requested):

test/plugin/test_in_http.rb
ashie: It's a trivial thing, no need to block merging.

ashie:
Thanks for your contribution!
I'm sorry for delaying merge.

PR #4288, opened 2023-09-05
Raise minimum required ruby version

kenhys:

N/A

The service discovery plugin helper use Array#prepend, it means that the ruby version must be 2.5 or later precisely.

   fluentd.work/lib/fluent/plugin_helper/service_discovery.rb:71:in
   `service_discovery_configure': undefined method `prepend' for
   []:Array (NoMethodError) from
   /work/fluent/fluentd/fluentd.work/lib/fluent/plugin/out_

**Docs Changes**:

https://github.com/fluent/fluentd-docs-gitbook/pull/471

**Release Note**: 

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Fluentd: Open-Source Log Collector
 
 ### Prerequisites
 
-- Ruby 2.4 or later
+- Ruby 2.7 or later
 - git
 
 `git` should be in `PATH`. On Windows, you can use `Github for Windows` and `GitShell` for easy setup.
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.license = "Apache-2.0"
 
-  gem.required_ruby_version = '>= 2.4'
+  gem.required_ruby_version = '>= 2.7'
 
   gem.add_runtime_dependency("bundler")
   gem.add_runtime_dependency("msgpack", [">= 1.3.1", "< 2.0.0"])


cosmo0920 (approved):
Good catch! :+1:

daipom:
Thanks!
Shouldn't we update the minimum version to 2.7 since the minimum version tested in CI is 2.7?

kenhys:
It depends on perspective:

* Precisely set minimum: 
  * Pros: can run a bit older even though already reached EOL.
  * Cons: allow non-supported version to run with.
* Only set 
  * Pros: can recommend verified version with CI.
  * Cons: cut off ruby version actually work.

Created PR based on the former, but reconsidered it should be the latter.

Thanks.

kenhys:
Fixed.

daipom (approved):
> It depends on perspective:
> 
>     * Precisely set minimum:
>       
>       * Pros: can run a bit older even though already reached EOL.
>       * Cons: allow non-supported version to run with.
> 
>     * Only set CI certified version:
>       
>       * Pros: can recommend verified version with CI.
>       * Cons: cut off ruby version actually work.
> 
> 
> Created PR based on the former, but reconsidered it should be the latter.
> 
> Thanksm, I'll fix it.

I see!
At least, we need this fix.

We can think about whether we should update the version to 2.7 or not later.

daipom:
> Fixed.

Thanks!

kenhys:
> We can think about whether we should update the version to 2.7 or not later.

When td-agent reached EOL (Dec, 2023), we will drop 2.7 from CI.

PR #4294, opened 2023-09-08
Minor performance improvements

isimluk:
Dunno. Minor perf improvements never hurt badly.

### DIFF
--- a/lib/fluent/registry.rb
+++ b/lib/fluent/registry.rb
@@ -60,10 +60,10 @@ module Fluent
       # search from additional plugin directories
       if @dir_search_prefix
         path = "#{@dir_search_prefix}#{type}"
-        files = @paths.map { |lp|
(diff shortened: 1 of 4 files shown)


lib/fluent/registry.rb
```diff
@@ -60,31 +60,31 @@ def search(type)
       # search from additional plugin directories
       if @dir_search_prefix
         path = "#{@dir_search_prefix}#{type}"
-        files = @paths.map { |lp|
+        files = @paths.filter_map { |lp|
```
ashie, line 63: Note: `filter_map` is available as of Ruby 2.7.
Recently we changed required Ruby version to 2.7, so it's no problem to use it: #4288

ashie (approved):

ashie:
Thanks!

PR #4299, opened 2023-09-12
remove unnecessary self-assignment

isimluk:
Previously this code was assigning instance variable like

    @log_event_router = log_event_router

later, during the refactoring of Fluent Engine it was changed to

    log_event_router = log_event_router

which can be now removed. For reference please review changeset at 3a3aee0227b621b61d1d7f258147f7ff661316e0


ashie (approved):

ashie:
Thanks!

PR #4303, opened 2023-09-19
Buffer: Fix NoMethodError with empty unstaged chunk arrays

amdoolittle:
When the write function encounters a chunk that is defined as a Nil object, the call to synchronize fails. This failure causes an unexpected error that is caught and handled like a failure to connect, with a 30 second retry delay. 

In an environment with a large amount of data, and this Nil scenario occurring frequently, it introduces significant processing delays since all processing activity stops during the 30 second window. The topic being queried continues to fill but no agents parse the data, leading to an ever-increasing backlog.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -419,15 +419,19 @@ module Fluent
               enqueue_chunk(m)
               if unstaged_chunks[m]
                 u = unstaged_chunks[m].pop
-                u.synchronize do
-                  if u.unstaged? && !chunk_size_full?(u)
-                    # `u.metadata.seq` and `m.seq` can be different but Buffer#enqueue_chunk expect them to be the same value
-                    u.metadata.seq = 0
-                    synchronize {
-                      @stage[m] = u.staged!
-                      @stage_size_metrics.add(u.bytesize)
-                    }
+                if !u.nil?
+                  u.synchronize do
+                    if u.unstaged? && !chunk_size_full?(u)
+                      # `u.metadata.seq` and `m.seq` can be different but Buffer#enqueue_chunk expect them to be the same value
+                      u.metadata.seq = 0
+                      synchronize {
+                        @stage[m] = u.staged!
+                        @stage_size_metrics.add(u.bytesize)
+                      }
+                    end
                   end
+                else
+                  log.warn "encountered nil chunk, skipping"
                 end
               end
             elsif c.unstaged?
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -417,21 +417,17 @@ module Fluent
             if c.staged? && (enqueue || chunk_size_full?(c))
               m = c.metadata
               enqueue_chunk(m)
-              if unstaged_chunks[m]
+              if unstaged_chunks[m] && !unstaged_chunks[m].empty?
                 u = unstaged_chunks[m].pop
-                if !u.nil?
-                  u.synchronize do
-                    if u.unstaged? && !chunk_size_full?(u)
-                      # `u.metadata.seq` and `m.seq` can be different but Buffer#enqueue_chunk expect them to be the same value
-                      u.metadata.seq = 0
-                      synchronize {
-                        @stage[m] = u.staged!
-                        @stage_size_metrics.add(u.bytesize)
-                      }
-                    end
+                u.synchronize do
+                  if u.unstaged? && !chunk_size_full?(u)
+                    # `u.metadata.seq` and `m.seq` can be different but Buffer#enqueue_chunk expect them to be the same value
+                    u.metadata.seq = 0
+                    synchronize {
+                      @stage[m] = u.staged!
+                      @stage_size_metrics.add(u.bytesize)
+                    }
                   end
-                else
-                  log.warn "encountered nil chunk, skipping"
                 end
               end
             elsif c.unstaged?


daipom (commented):
Thanks for your contribution.

> When the write function encounters a chunk that is defined as a Nil object, the call to synchronize fails.

Is it reproducible?
Could you tell us how to reproduce it?

The error will occur if `unstaged_chunks[m]` is an empty list or contains a `nil` object.
However, it does not seem possible.

https://github.com/fluent/fluentd/blob/78c91468d65da7cce0999fd94dbe61f7a35b606b/lib/fluent/plugin/buffer.rb#L356-L359

amdoolittle:
>
> However, it does not seem possible.
> 
> https://github.com/fluent/fluentd/blob/78c91468d65da7cce0999fd94dbe61f7a35b606b/lib/fluent/plugin/buffer.rb#L356-L359

Tell that to my constant 100MB/sec data flow that throws this error frequently. :)

I'll see if I can forcefully reproduce it outside that environment.

Here's the stack trace, for clarification:

```
2023-09-15 17:33:32 +0000 [warn]: #0 emit transaction failed: error_class=NoMethodError error="undefined method `synchronize' for nil:NilClass" location="/opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/buffer.rb:422:in `block in write'" tag="[redacted]"
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/buffer.rb:422:in `block in write'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/buffer.rb:416:in `each'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/buffer.rb:416:in `write'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/output.rb:1095:in `block in handle_stream_simple'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/output.rb:977:in `write_guard'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/output.rb:1094:in `handle_stream_simple'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/output.rb:967:in `execute_chunking'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin/output.rb:897:in `emit_buffered'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/event_router.rb:115:in `emit_stream'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluent-plugin-kafka-0.19.0/lib/fluent/plugin/in_rdkafka_group.rb:287:in `emit_events'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluent-plugin-kafka-0.19.0/lib/fluent/plugin/in_rdkafka_group.rb:269:in `block in run'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluent-plugin-kafka-0.19.0/lib/fluent/plugin/in_rdkafka_group.rb:214:in `each_batch'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluent-plugin-kafka-0.19.0/lib/fluent/plugin/in_rdkafka_group.rb:229:in `run'
  2023-09-15 17:33:32 +0000 [warn]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/plugin_helper/thread.rb:78:in `block in thread_create'
```

daipom (commented):
> I'll see if I can forcefully reproduce it outside that environment.

Thanks!

> Once we've popped all the values off the array, it'll still evaluate truthy while empty (thanks Ruby), so we could end up attempting to pop another value off the end of an empty array, resulting in a nil return to "u".

I see!
Certainly, it seems possible to execute `pop` against an empty list because this process can be repeated.

https://github.com/fluent/fluentd/blob/78c91468d65da7cce0999fd94dbe61f7a35b606b/lib/fluent/plugin/buffer.rb#L416-L421

If so, this fix would be reasonable.
I want to confirm the reproduction conditions a little more.

daipom:
Sorry for the delay.
I think this fix would be reasonable, but I'm still checking the mechanism and impact of this bug.
I will merge this after it.

daipom:
> I think this fix would be reasonable, but I'm still checking the mechanism and impact of this bug.

I don't yet fully understand the role of `unstaged_chunks`...
At least, we can merge this to prevent the unexpected error.
It does not change the specification.
It only prevents this error.

So, I merge this now for the next release.

daipom:
Seeing #4342, I understand now some mechanics of this bug.
When retrying `write_step_by_step`, the first staged chunk is added to `operated_chunks` duplicatedly.
So, `unstaged_chunks[m].pop` can be called more than once for the same metadata.

I believe this duplication of `operated_chunks` is another bug and we should fix it.

Anyway, these unstaged chunks will be enqueued immediately in the `Buffer::write()`,
so, for now, it is sufficient to avoid the error by this fix.

PR #4320, opened 2023-10-10
Drop unused @_plugin_id_configured from all own-able plugins

isimluk:
This variable is unused since b32781d0f03e7984f3b4eeec63f7310fa340b562.

### DIFF
--- a/lib/fluent/plugin/owned_by_mixin.rb
+++ b/lib/fluent/plugin/owned_by_mixin.rb
@@ -21,7 +21,6 @@ module Fluent
         @_owner = plugin
 
         @_plugin_id = plugin.plugin_id
-        @_plugin_id_configured = plugin.plugin_id_configured?
 
         @log = plugin.log
       end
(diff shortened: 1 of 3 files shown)


daipom:
Sorry, it may be a week or two before I can review this.
If other maintainers can't make time to review this, please wait a moment.

ashie:
`@_plugin_id_configured` is introduced at `f601453903e1fa64d3fca5d77eb894c55ac2efa7` (v0.14.0.rc1) and removed at b32781d0f03e7984f3b4eeec63f7310fa340b562 (v0.14.11).
It seems be able to remove safely since it's used only v0.14 series and `@_` prefix (should be used only internally) is used.

ashie (approved):

daipom (approved):
Thanks.
Sorry for the delay in merging this.
Now, I merge this to the master branch for v1.17.0.

Note: It seems that we don't need `@_plugin_id` either.

PR #4327, opened 2023-10-18
in_tail: Fix a stall bug on !follow_inode case

ashie:
**Which issue(s) this PR fixes**: 
Fixes #3614
Closes #4264

**What this PR does / why we need it**: 

Although known stall issues of in_tail on `follow_inode` case are fixed
in v1.16.2, it has still a similar problem on `!follow_inode` case.

In this case, a tail watcher is possible to mark the position entry as
`unwatched` if it's tansitioned to `rotate_wait` state by
`refresh_watcher` even if another newer tail watcher is managing it.
It's hard to occur in usual because `stat_watcher` will be called
immediately after the file is changed  while `refresh_wather` is called
every 60 seconds in default. However, there is a rare possibility that
this order might be swapped especillay if in_tail is busy on processing
large amount of logs. Because in_tail is single threadied, event queues
such as timers or inotify will be stucked in this case.
    
There is no such problem on `follow_inode` case because position entries
are always marked as `unwatched` before entering `rotate_wait` state.

**Docs Changes**:
N/A

**Release Note**: 
Same with the title

PR #4334, opened 2023-10-31
in_tail: Manage tail watchers that are `rorate_wait` state too

ashie:
N/A (Follow up for https://github.com/fluent/fluentd/pull/4327#issuecomment-1774330298)

After a tail watcher transitions to `rotate_wait` state, the `rotate_wait` timer is no longer managed by in_tail, it might cause unexpected behaviour. e.g.)

* It's never unwatched when shutdown occurs before `rotate_wait` passed.
* Needless `rotate_wait` timers are executed when it detects more rotations.

This patch fixes such unexpected behaviour.

**Docs Changes**:
N/A

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -594,8 +594,6 @@ module Fluent::Plugin
         target_info = TargetInfo.new(tw.path, ino)
         @pf.unwatch(target_info)
       end
-
-      @tails_rotate_wait.delete(tw)
     end
 
     def throttling_is_enabled?(tw)
@@ -622,6 +620,7 @@ module Fluent::Plugin
           elapsed = Fluent::Clock.now - start_time_to_wait
           if tw.eof? && elapsed >= @rotate_wait
             timer.detach
+            @tails_rotate_wait.delete(tw)
             detach_watcher(tw, ino)
           end
         end
@@ -629,6 +628,7 @@ module Fluent::Plugin
       else
         # when the throttling feature isn't enabled, just wait @rotate_wait
         timer = timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
+          @tails_rotate_wait.delete(tw)
           detach_watcher(tw, ino)
         end
         @tails_rotate_wait[tw] = { ino: ino, timer: timer }
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -52,6 +52,7 @@ module Fluent::Plugin
       super
       @paths = []
       @tails = {}
+      @tails_rotate_wait = {}
       @pf_file = nil
       @pf = nil
       @ignore_list = []
@@ -267,6 +268,9 @@ module Fluent::Plugin
       @shutdown_start_time = Fluent::Clock.now
       # during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.
       stop_watchers(existence_path, immediate: true, remove_watcher: false)
+      @tails_rotate_wait.keys.each do |tw|
+        detach_watcher(tw, @tails_rotate_wait[tw][:ino])
+      end
       @pf_file.close if @pf_file
 
       super
@@ -590,6 +594,8 @@ module Fluent::Plugin
         target_info = TargetInfo.new(tw.path, ino)
         @pf.unwatch(target_info)
       end
+
+      @tails_rotate_wait.delete(tw)
     end
 
     def throttling_is_enabled?(tw)
@@ -604,7 +610,11 @@ module Fluent::Plugin
       if @open_on_every_update
         # Detach now because it's already closed, waiting it doesn't make sense.
         detach_watcher(tw, ino)
-      elsif throttling_is_enabled?(tw)
+      end
+
+      return if @tails_rotate_wait[tw]
+
+      if throttling_is_enabled?(tw)
         # When the throttling feature is enabled, it might not reach EOF yet.
         # Should ensure to read all contents before closing it, with keeping throttling.
         start_time_to_wait = Fluent::Clock.now
@@ -615,11 +625,13 @@ module Fluent::Plugin
             detach_watcher(tw, ino)
           end
         end
+        @tails_rotate_wait[tw] = { ino: ino, timer: timer }
       else
         # when the throttling feature isn't enabled, just wait @rotate_wait
-        timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
+        timer = timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
           detach_watcher(tw, ino)
         end
+        @tails_rotate_wait[tw] = { ino: ino, timer: timer }
       end
     end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -578,10 +578,6 @@ module Fluent::Plugin
       detach_watcher_after_rotate_wait(tail_watcher, pe.read_inode)
     end
 
-    # TailWatcher#close is called by another thread at shutdown phase.
-    # It causes 'can't modify string; temporarily locked' error in IOHandler
-    # so adding close_io argument to avoid this problem.
-    # At shutdown, IOHandler's io will be released automatically after detached the event loop
     def detach_watcher(tw, ino, close_io = true)
       if @follow_inodes && tw.ino != ino
         log.warn("detach_watcher could be detaching an unexpected tail_watcher with a different ino.",
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -269,7 +269,7 @@ module Fluent::Plugin
       # during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.
       stop_watchers(existence_path, immediate: true, remove_watcher: false)
       @tails_rotate_wait.keys.each do |tw|
-        detach_watcher(tw, @tails_rotate_wait[tw][:ino])
+        detach_watcher(tw, @tails_rotate_wait[tw][:ino], false)
       end
       @pf_file.close if @pf_file
 
@@ -279,6 +279,7 @@ module Fluent::Plugin
     def close
       super
       # close file handles after all threads stopped (in #close of thread plugin helper)
+      # It may be because we need to wait IOHanlder.ready_to_shutdown()
       close_watcher_handles
     end
 
@@ -519,6 +520,9 @@ module Fluent::Plugin
         if tw
           tw.close
         end
+      end
+      @tails_rotate_wait.keys.each do |tw|
+        tw.close
       end
     end
 
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -3049,13 +3049,13 @@ class TailInputTest < Test::Unit::TestCase
       d.run(expect_records: 6, timeout: 15) do
         Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt0", "ab") {|f| f.puts "file1 log2"}
 
-        sleep 1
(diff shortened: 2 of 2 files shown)


ashie:
>     * It's never unwatched when shutdown occurs before `rotate_wait` passed.
>     * Needless `rotate_wait` timers are executed when it detects more rotations.

Although they are unexpected behaviour, probably not harmful.
So we might pass merging this patch into the next release.
This PR is just for remind & discuss this issue.

ashie:
>     * It's never unwatched when shutdown occurs before `rotate_wait` passed.

BTW I'm suspecting that some unstable in_tail tests are caused by it.

ashie:
Sometimes we get an error regarding unmanaged rotated tail watchers on shutdown:

e.g.)
```
[error]: #0 [in_tail_fluentd] closed stream
[error]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.3/lib/fluent/plugin/in_tail/position_file.rb:228:in `pos='
[error]: #0 /opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.3/lib/fluent/plugin/in_tail/position_file.rb:228:in `block in update_pos'
...
```

We should merge this fix to v1.16.

daipom:
Thanks for this fix!

> It's never unwatched when shutdown occurs before rotate_wait passed.

At this point, this fix basically looks good to me.

> Needless rotate_wait timers are executed when it detects more rotations.

At this point, we need more modifications.
Maybe they should be done in other PRs, but I'd like to clarify this point at this time.

I think this occurs **with** `follow_inodes`, and it makes duplicate TailWatchers and handle-leak.

When a very fast rotation occurs, a TailWatcher in the state of `rotate_wait` can call `update_watcher()` again.
That is the issue we should be concerned about.

**Without** `follow_inodes`, this check in `update_watcher()` prevents the duplicate `update_watcher()` process.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L545-L553

It is because `@pf[target_info_from_position_entry].read_inode ` returns the inode of the current file **without** `follow_inodes`.
So, if the TailWatcher is old, those inodes do not match.

However, **with** `follow_inodes`, this check does not work, and executes `update_watcher()` process duplicately.

This fix will prevent duplicate timer execution for detaching.
However, it is not enough.
We have to prevent the overall duplicate `update_watcher()` process.
Maybe it should be done in another PR.

lib/fluent/plugin/in_tail.rb
```diff
@@ -615,11 +625,13 @@ def detach_watcher_after_rotate_wait(tw, ino)
...
           end
         end
+        @tails_rotate_wait[tw] = { ino: ino, timer: timer }
       else
         # when the throttling feature isn't enabled, just wait @rotate_wait
-        timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
+        timer = timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
           detach_watcher(tw, ino)
         end
+        @tails_rotate_wait[tw] = { ino: ino, timer: timer }
```
daipom, line 634: Why is it necessary to keep these `timer` objects?

ashie: In my understanding, the complete process of cancelling roate_wait needs `@tails_rotate_wait[tw][:timer].detach)`.
If we need to cancel rotaet_wait at certain timing, we need to keep the timer ID.
Ideally it's also needed in `shutdown` but this PR omits it since event_loop will be stopped after shutdown so that the timer never called.

daipom: I see! Thanks!

> If we need to cancel rotaet_wait at certain timing, we need to keep the timer ID (we never cancel it in the current implementation though).

Yes!

> Ideally it's also needed in shutdown but this PR omits it since event_loop will be stopped after shutdown so that the timer never called.

Yes! `event_loop` will detach them automatically, so we can omit it.

lib/fluent/plugin/in_tail.rb
```diff
@@ -267,6 +268,9 @@ def shutdown
       @shutdown_start_time = Fluent::Clock.now
       # during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.
       stop_watchers(existence_path, immediate: true, remove_watcher: false)
+      @tails_rotate_wait.keys.each do |tw|
+        detach_watcher(tw, @tails_rotate_wait[tw][:ino])
+      end
```
daipom, line 273: > during shutdown phase, don't close io. It should be done in close after all threads are stopped. See close.

I'm concerned about this point.

`stop_watchers()` with `immediate: true` calls `detach_watcher()` with `close_io: false`.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L503-L505

It seems that we should not close the io handler here.
We should do it in `close_watcher_handles()`, separately.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L275-L279

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L512-L519

I'm concerned about these comments as well, but I think they are old.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L573-L577

At that time, `in_tail` used `Coolio::Loop`, but it is not used now.

* #348

I'm not sure if we really need to pay attention to the timing of the IO handler close...
At this point, I think it would be better to unify the timing of the process.
(Closing IO hanlders should be done in `close_watcher_handles()` when stopping.)

daipom: > I'm not sure if we really need to pay attention to the timing of the IO handler close...

It may be because `io_handler` does not stop immediately.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L844-L850

daipom:
> > Needless rotate_wait timers are executed when it detects more rotations.
> 
> At this point, we need more modifications. Maybe they should be done in other PRs, but I'd like to clarify this point at this time.
> 
> I think this occurs **with** `follow_inodes`, and it makes duplicate TailWatchers and handle-leak.
> 
> ...
> 
> However, **with** `follow_inodes`, this check does not work, and executes `update_watcher()` process duplicately.
> 
> This fix will prevent duplicate timer execution for detaching. However, it is not enough. We have to prevent the overall duplicate `update_watcher()` process. Maybe it should be done in another PR.

Hmm, I may be wrong about this.
When using `follow_inodes`, the following logic prevents making duplicate TailWatcher.

https://github.com/fluent/fluentd/blob/175d8669b8444505bd5168524b07d6a3133dca7e/lib/fluent/plugin/in_tail.rb#L558-L561

daipom:
I have added tests for fast rotation.

lib/fluent/plugin/in_tail.rb
```diff
@@ -590,6 +594,8 @@ def detach_watcher(tw, ino, close_io = true)
         target_info = TargetInfo.new(tw.path, ino)
         @pf.unwatch(target_info)
       end
+
+      @tails_rotate_wait.delete(tw)
```
daipom, line 598: I believe this is not the responsibility of `detach_wacher()`.
Logic inside `detach_watcher()` should be independent of `rotate_wait`.

daipom:
@ashie Can I add a commit about the points I am commenting on?

ashie:
> @ashie Can I add a commit about the points I am commenting on?

Of course.

lib/fluent/plugin/in_tail.rb
daipom: Done @ 341d8b89eb7314ac72ad663907a8cc05904c21ea

daipom:
> > @ashie Can I add a commit about the points I am commenting on?
> 
> Of course.

Done.

The new tests on macOS seem to be unstable. I will check them.

daipom:
The new tests are unstable on Windows or macOS.
We need to check if we need to fix these tests or if in_tail on Windows or macOS can not collect files that are rotated very fast.

daipom:
I will merge this if it is confirmed that the newly added tests are stable.
Please let me know if anyone has concerns.

daipom:
The new tests are now stable.
However,  I have noticed that we need to consider `rotate_wait 0` case.

daipom:
> However, I have noticed that we need to consider rotate_wait 0 case.

I am concerned about the possibility of the key being deleted before the key is added.
Although I can't confirm such a behavior, I have added the countermeasure just in case.

https://github.com/fluent/fluentd/pull/4334/commits/8b1547803f503304ebe8b7beef95ba8a656c701b

daipom:
> > However, I have noticed that we need to consider rotate_wait 0 case.
> 
> I am concerned about the possibility of the key being deleted before the key is added. Although I can't confirm such a behavior, I have added the countermeasure just in case.
> 
> [8b15478](https://github.com/fluent/fluentd/commit/8b1547803f503304ebe8b7beef95ba8a656c701b)

@ashie 
Could you briefly check this?

lib/fluent/plugin/in_tail.rb
```diff
@@ -630,15 +632,17 @@ def detach_watcher_after_rotate_wait(tw, ino)
...
         timer = timer_execute(:in_tail_close_watcher, @rotate_wait, repeat: false) do
           @tails_rotate_wait.delete(tw)
           detach_watcher(tw, ino)
         end
-        @tails_rotate_wait[tw] = { ino: ino, timer: timer }
       end
+
+      # To avoid conflict with timer execution (It deletes the key concurrently)
+      v = @tails_rotate_wait[tw]
+      v[:timer] = timer if v
```
ashie, line 645: > > I am concerned about the possibility of the key being deleted before the key is added. Although I can't confirm such a behavior, I have added the countermeasure just in case.
> > [8b15478](https://github.com/fluent/fluentd/commit/8b1547803f503304ebe8b7beef95ba8a656c701b)
> 
> @ashie Could you briefly check this?

I can't understand what you worry about.
Do you mean that timers created in this method may be executed before reaching here (` v = @tails_rotate_wait[tw]`)?
If so, it's not possible. All timers in in_tail are executed in a same thread.
So created timers are always executed after exiting `detach_watcher_after_rotate_wait` method.

If my above guess for your concern is correct, I don't want this change, I want keep it simple as possible.

daipom: > Do you mean that timers created in this method may be executed before reaching here (` v = @tails_rotate_wait[tw]`)?
> If so, it's not possible. All timers in in_tail are executed in a same thread. It means that timers created in this method are always executed after exiting this `detach_watcher_after_rotate_wait` method.
> 
> If my above guess for your concern is correct, I don't want this change, I want keep it simple as possible.

Yes!! Thanks so much! I overlooked both `detach_watcher_after_rotate_wait` and the timer event were executed in the same thread for the event loop.

I will remove this fix, and check the following codes as well.

https://github.com/fluent/fluentd/pull/4334/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802R272-R275

daipom:
> I will remove this fix, and check the following codes as well.
> 
> https://github.com/fluent/fluentd/pull/4334/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802R272-R275

The callbacks of timers are not called after the `Stop` phase.
So we don't need these codes.
I will remove them too.

daipom:
The code for conflicts with timer execution is unnecessary, so I have removed them.
@ashie Thanks and sorry I overlooked this.

ashie (commented):
LGTM
(It seems that I can't change the status of this pull request, because it's originally mine 🤪)

daipom:
Thanks! I will merge this soon!

daipom (approved):

PR #4344, opened 2023-11-17
build: replace deprecated method

daipom:
None.

Because `Bundler.with_clean_env` has been deprecated, we need to replace it.

    $ bundle exec rake build:all
    [DEPRECATED] `Bundler.with_clean_env` has been deprecated in
    favor of `Bundler.with_unbundled_env`. If you instead want
    the environment before bundler was originally loaded, use
    `Bundler.with_original_env`

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed. (Not related to the behavior of the product)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,7 +13,7 @@ task test: [:base_test]
 namespace :build do
   desc 'Build gems for all platforms'
   task :all do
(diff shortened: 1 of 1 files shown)


kenhys (approved):
LGTM.

PR #4361, opened 2023-12-11
Add parameter to skip ignoring files

apilaskowski:
In our use-case we have a race condition.

We are creating files, then we are setting up privileges. In function 'extend_paths' we can see that if file is not readable once it is ignored, so it happened to us, that some logs were ignored forever. 

I think that introducing such parameter should be the easiest way to avoid similar problems in the future.

Fixes #
Rare race condition with updating privileges to files after executing 'in_tail.rb'

**Docs Changes**:
New parameter in 'in_tail'

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -113,6 +113,8 @@ module Fluent::Plugin
     config_param :follow_inodes, :bool, default: false
     desc 'Maximum length of line. The longer line is just skipped.'
     config_param :max_line_size, :size, default: nil
+    desc 'Skip ignoring files.'
+    config_param :skip_ignoring_files, :bool, defaults: false
 
     config_section :parse, required: false, multi: true, init: true, param_name: :parser_configs do
       config_argument :usage, :string, default: 'in_tail_parser'
@@ -303,7 +305,7 @@ module Fluent::Plugin
                   true
                 end
               else
-                if is_file
+                if is_file && !skip_ignoring_files?
                   unless @ignore_list.include?(p)
                     log.warn "#{p} unreadable. It is excluded and would be examined next time."
                     @ignore_list << p if @ignore_repeated_permission_error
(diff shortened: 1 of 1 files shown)


apilaskowski:
Why is this DCO failing? I added message required by it at the end of the commit. Can someone help?
Also, should I prepare documentation changes in this PR also, or should it be done separately?

ashie:
Thanks for your contribution!

> Why is this DCO failing? 

The answer is described in [the link](https://github.com/fluent/fluentd/pull/4361/checks?check_run_id=19536987782): 
> Expected "Artur Piotr Izaak Laskowski [apilaskowski@google.com](mailto:apilaskowski@google.com)", but got "Artur Laskowski [apilaskowski@google.com](mailto:apilaskowski@google.com)".

apilaskowski:
DCO is now passing

apilaskowski:
I will try to fix it later today, I assume this is some problem with new parameter not being properly created. I will take a look

apilaskowski:
@ashie PTAL, i made a misspelling before

daipom:
Thanks for requesting the new feature!

> that some logs were ignored forever.

Does this mean that the following warning log will be no longer shown?

https://github.com/fluent/fluentd/blob/95438b2eeb4bb5586a4031c1fbc35756f3c12565/lib/fluent/plugin/in_tail.rb#L306-L312

Or does that mean that the collection of the file will stop forever?

apilaskowski:
What I want to achieve is not ignoring any files, it happens that they are unreadable at the beginning, but then gets readable after some time.

I am wondering and probably missed that at the beginning, is `ignore_repeated_permission_error` doing just that?
I will investigate and get back to you.

apilaskowski:
I think that `ignore_repeated_permission_error` is working for me, thank you for your cooperation.
If I will need more I will return to this, however, for now it is working for us.

daipom:
@apilaskowski
Thanks for your feedback!

> What I want to achieve is not ignoring any files, it happens that they are unreadable at the beginning, but then gets readable after some time.

`@ignore_list` is just a list to control the warning message.
The list has nothing to do with whether the file is collected or not.
If a file is added to the list, the file should be collected after it gets readable.

If the file is not collected after it gets readable, there may be some other issues.
If you have any problems collecting files, please tell us again!

PR #4369, opened 2023-12-27
test: use IO#popen instead of Kernel#open

kenhys:

N/A

open("|command") spawns command and create pipes for processing pipeline.

This feature of Kernel#open is deprecated, replace it with IO#popen.

It fixes the following warning:

  warning: Calling Kernel#open with a leading '|' is deprecated and will
  be removed in Ruby 4.0; use IO.popen instead

**Docs Changes**:

N/A

**Release Note**: 

N/A


ashie (approved):
LGTM

PR #4399, opened 2024-02-09
Guard launching duplicated fluentd with same configuration

kenhys:

Fixes https://github.com/fluent/fluent-package-builder/issues/611

**This is just a PoC PR.**

Before:

  If you launch multiple Fluentd instance with same
  configuration file, it causes a disaster with inconsistency
  processed buffer or pos file.

After:

  Detect fluentd service's main process and fetch FLUENT_CONF.
  if configuration is same as spawned process, abort it.

  It can block the following conditions are met:

  * fluentd is launched via systemd (fluent-package) configuration file is specified via FLUENT_CONF.
  * manually try to launch fluentd  with same configuration file as fluentd user with -c option.

  Thus running fluentd service and manually try to launch normal
  user case can't be detected.

NOTE: Windows is out of scope in this PR.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -20,6 +20,7 @@ require 'fluent/supervisor'
 require 'fluent/log'
 require 'fluent/env'
 require 'fluent/version'
+require 'fluent/process'
 
 $fluentdargv = Marshal.load(Marshal.dump(ARGV))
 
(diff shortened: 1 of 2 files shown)


kenhys:
Reconsider, it is not appropriate to introduce package specific fix to fluentd internals. :thinking:

PR #4401, opened 2024-02-15
in_tail: Expand glob capability for square brackets and one character matcher

cosmo0920:
This PR will fix none of the issue but this was raised in community slack channel.

This feature was requested on the community slack's #fluentd channel.
For now, we didn't support for square brackets pattern and one character matcher in #expand_paths.

This shouldn't be fit for the following case.

The logs are named as follows:

```
.
├── 01.log
├── 02.log
├── 03.log
├── 12.log
├── 13.log
├── 20.log
```

Then, a user wants to tail 02.log, 03.log, 12.log, and 13.log, and excluding to tail 01.log and 20.log with the following configuration:

```
<source>
  @type tail
  path "[0-1][2-3].log"
  <parse>
    @type none
  </parse>
  tag log.test
  read_from_head
  use_extended_glob # To use this expansion, users need to enable this feature explicitly.
</source>
```

However, the current Fluentd doesn't handle the "[0-1][2-3].log" pattern as glob. It just handles as normal file name.
This shouldn't be effective for this use case.

It would be nice to handle this pattern on the in_tail plugin.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/488

**Release Note**: 
Same as title.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -301,6 +301,10 @@ module Fluent::Plugin
     # This should be collided for wildcard pattern for curly braces.
     def use_glob?(path)
       if @glob_policy == :always
+        # For future extensions, we decided to use `always' term to handle
+        # regular expressions as much as possible.
+        # This is because not using `true' as a returning value
+        # when choosing :always here.
         extended_glob_pattern(path) || /\{.*,.*\}/.match(path)
       elsif @glob_policy == :extended
         extended_glob_pattern(path)
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -64,8 +64,8 @@ module Fluent::Plugin
     config_param :path, :string
     desc 'path delimiter used for spliting path config'
     config_param :path_delimiter, :string, default: ','
-    desc 'Choose using glob patterns. Adding whether a capability to handle [] and ? or not.'
-    config_param :glob_policy, :enum, list: [:backward_compatible, :extended], default: :backward_compatible
+    desc 'Choose using glob patterns. Adding capabilities to handle [] and ?, and {}.'
+    config_param :glob_policy, :enum, list: [:backward_compatible, :extended, :always], default: :backward_compatible
     desc 'The tag of the event.'
     config_param :tag, :string
     desc 'The paths to exclude the files from watcher list.'
@@ -140,6 +140,10 @@ module Fluent::Plugin
 
       if !@enable_watch_timer && !@enable_stat_watcher
         raise Fluent::ConfigError, "either of enable_watch_timer or enable_stat_watcher must be true"
+      end
+
+      if @glob_policy == :always && @path_delimiter == ','
+        raise Fluent::ConfigError, "cannot use glob_policy as always with the default path_delimitor: `,\""
       end
 
       if RESERVED_CHARS.include?(@path_delimiter)
@@ -288,7 +292,9 @@ module Fluent::Plugin
     # Curly braces is not supported for now because the default delimiter of path is ",".
     # This should be collided for wildcard pattern for curly braces.
     def use_glob?(path)
-      if @glob_policy == :extended
+      if @glob_policy == :always
+        path.include?('*') || path.include?('?') || /\[.*\]/.match(path) || /\{.*,.*\}/.match(path)
+      elsif @glob_policy == :extended
         path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
       elsif @glob_policy == :backward_compatible
         path.include?('*')
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -283,6 +283,12 @@ module Fluent::Plugin
         @capability.have_capability?(:effective, :dac_override)
     end
 
+    # Curly braces is not supported for now because the default delimiter of path is ",".
+    # This should be collided for wildcard pattern for curly braces.
+    def use_glob?(path)
+      path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
+    end
+
     def expand_paths
       date = Fluent::EventTime.now
       paths = []
@@ -292,7 +298,7 @@ module Fluent::Plugin
                else
                  date.to_time.strftime(path)
                end
-        if path.include?('*') || /\[.*\]/.match(path)
+        if use_glob?(path)
           paths += Dir.glob(path).select { |p|
             begin
               is_file = !File.directory?(p)
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -65,7 +65,7 @@ module Fluent::Plugin
     desc 'path delimiter used for spliting path config'
     config_param :path_delimiter, :string, default: ','
     desc 'Choose using glob patterns. Adding whether a capability to handle [] and ? or not.'
-    config_param :enable_glob, :enum, list: [:with_wildcards, :always], default: :with_wildcards
+    config_param :glob_policy, :enum, list: [:backward_compatible, :extended], default: :backward_compatible
     desc 'The tag of the event.'
     config_param :tag, :string
     desc 'The paths to exclude the files from watcher list.'
@@ -288,9 +288,9 @@ module Fluent::Plugin
     # Curly braces is not supported for now because the default delimiter of path is ",".
     # This should be collided for wildcard pattern for curly braces.
     def use_glob?(path)
-      if @enable_glob == :always
+      if @glob_policy == :extended
         path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
-      elsif @enable_glob == :with_wildcards
+      elsif @glob_policy == :backward_compatible
         path.include?('*')
       end
     end
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -65,7 +65,7 @@ module Fluent::Plugin
     desc 'path delimiter used for spliting path config'
     config_param :path_delimiter, :string, default: ','
     desc 'Use extended glob patterns. Adding a capability to handle [] and ?.'
-    config_param :use_extended_glob, :bool, default: false
+    config_param :enable_glob, :enum, list: [:with_wildcards, :extended, :no], default: :with_wildcards
     desc 'The tag of the event.'
     config_param :tag, :string
     desc 'The paths to exclude the files from watcher list.'
@@ -288,10 +288,13 @@ module Fluent::Plugin
     # Curly braces is not supported for now because the default delimiter of path is ",".
     # This should be collided for wildcard pattern for curly braces.
     def use_glob?(path)
-      if @use_extended_glob
+      if @enable_glob == :extended
         path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
-      else
+      elsif @enable_glob == :with_wildcards
         path.include?('*')
+      elsif @enable_glob == :no
+        # Always return false when glob is turned off.
+        false
       end
     end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -342,7 +342,7 @@ module Fluent::Plugin
                else
                  date.to_time.strftime(path)
                end
-        path.include?('*') ? Dir.glob(path) : path
+        use_glob?(path) ? Dir.glob(path) : path
       }.flatten.uniq
       # filter out non existing files, so in case pattern is without '*' we don't do unnecessary work
       hash = {}
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -65,7 +65,7 @@ module Fluent::Plugin
     desc 'path delimiter used for spliting path config'
     config_param :path_delimiter, :string, default: ','
     desc 'Choose using glob patterns. Adding whether a capability to handle [] and ? or not.'
-    config_param :enable_glob, :enum, list: [:with_wildcards, :extended], default: :with_wildcards
+    config_param :enable_glob, :enum, list: [:with_wildcards, :always], default: :with_wildcards
     desc 'The tag of the event.'
     config_param :tag, :string
     desc 'The paths to exclude the files from watcher list.'
@@ -288,7 +288,7 @@ module Fluent::Plugin
     # Curly braces is not supported for now because the default delimiter of path is ",".
     # This should be collided for wildcard pattern for curly braces.
     def use_glob?(path)
-      if @enable_glob == :extended
+      if @enable_glob == :always
         path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
       elsif @enable_glob == :with_wildcards
         path.include?('*')
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -297,8 +297,10 @@ module Fluent::Plugin
       path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
     end
 
-    # Curly braces is not supported for now because the default delimiter of path is ",".
-    # This should be collided for wildcard pattern for curly braces.
+    # Curly braces is not supported with default path_delimiter
+    # because the default delimiter of path is ",".
+    # This should be collided for wildcard pattern for curly braces and
+    # be handled as an error on #configure.
     def use_glob?(path)
       if @glob_policy == :always
         # For future extensions, we decided to use `always' term to handle
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -64,8 +64,8 @@ module Fluent::Plugin
     config_param :path, :string
     desc 'path delimiter used for spliting path config'
     config_param :path_delimiter, :string, default: ','
-    desc 'Use extended glob patterns. Adding a capability to handle [] and ?.'
-    config_param :enable_glob, :enum, list: [:with_wildcards, :extended, :no], default: :with_wildcards
+    desc 'Choose using glob patterns. Adding whether a capability to handle [] and ? or not.'
+    config_param :enable_glob, :enum, list: [:with_wildcards, :extended], default: :with_wildcards
     desc 'The tag of the event.'
     config_param :tag, :string
     desc 'The paths to exclude the files from watcher list.'
@@ -292,9 +292,6 @@ module Fluent::Plugin
         path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
       elsif @enable_glob == :with_wildcards
         path.include?('*')
-      elsif @enable_glob == :no
-        # Always return false when glob is turned off.
-        false
       end
     end
 
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -292,7 +292,7 @@ module Fluent::Plugin
                else
                  date.to_time.strftime(path)
                end
-        if path.include?('*')
+        if path.include?('*') || /\[.*\]/.match(path)
           paths += Dir.glob(path).select { |p|
             begin
               is_file = !File.directory?(p)
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -146,6 +146,10 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "cannot use glob_policy as always with the default path_delimitor: `,\""
       end
 
+      if @glob_policy == :extended && /\{.*,.*\}/.match(@path) && extended_glob_pattern(@path)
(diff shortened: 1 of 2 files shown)


ashie:
From https://fluent-all.slack.com/archives/C0CTT63EE/p1707473140708389

daipom:
Thanks!
It's a Very useful enhancement!

I'm concerned about backward compatibility.
If there is already a setting where the `path` option contains the characters, the behavior will change, correct?

For example, if using the following configuration, is it necessary to change the setting?

```xml
<source>
  @type tail
  path "/path/to/foo[bar]/test.log"
  tag test
  <parse>
    @type none
  </parse>
</source>
```

I want to consider this point carefully.

cosmo0920:
> For example, if using the following configuration, is it necessary to change the setting?

You are right.
macOS, and also Windows permit to use square brackets in filename. So, we need to maintain backward compatibility for it.

cosmo0920:
Once I implemented this feature, I didn't notice the backward compatibility issue and also macOS and Windows permit to use square brackets in their filesystems. So, we need to provide a switch for enabling this fearture.

daipom (changes requested):
Sorry for the delay, I've been busy with the 1.16 release.

Now, compatibility is OK, thanks!

I have 2 concerns about the specifications.

1. Shouldn't we consider `exclude_paths` as well?
1. `use_extended_glob` option's name and feature are slightly inconsistent and confusing.
   * This is simply a feature about changing the condition to enable `Dir.glob`.
   * Whether this setting is `true` or `false`, `Dir.glob` itself remains the same.
     * For example, even if it is `false`, we can use `[]` for glob as long as `path` includes `*`.
   * Wouldn't it be more convenient and clearer to make a new option that just changes the condition to enable `Dir.glob`?
     * For example: the new option `enable_glob`.
       * Enum: `with_wildcard` (default), `always`, `no`
         * `with_wildcard`: Enabled if wildcards are included. (The current specification).
         * `always`: Always enabled.
           * We don't need to care about curly braces here. We should make it available to all who want to use it because we can avoid the collision by setting `path_delimiter`. We can explain it in the documentation or check it when we do `@path.split(@path_delimiter)` and give a warning message, if it is needed.
         * `no`: Always disabled.
           * It would be natural for this option to exist.

cosmo0920:
> Wouldn't it be more convenient and clearer to make a new option that just changes the condition to enable `Dir.glob`?
> 
>     * For example: the new option `enable_glob`.

I'd prefer to use enum approach. Because switching with `true` or `false` shouldn't be representing the conditions that we considered. Also, enum approach is more effective to prevent specifying malformed values.

So, my recommendation is:

`with_wildcard`: This is default setting for that option we'll provide
`extended`: This should be more concrete name for that option. Because we didn't provide full set of the glob patterns on Dir.blog. In this case, I'd prefer to use "extended" instead of "always".
However, I didn't fully follow your suggestion what "no" means here.  Perhaps, you would be interested to collect `*` contained files with in_tail?

cosmo0920:
Hey @daipom and @ashie, could you all take a look the applied changes again? This PR is ready to review.

daipom (commented):
@cosmo0920 Thanks for fixing!

> `extended`: This should be more concrete name for that option. Because we didn't provide full set of the glob patterns on Dir.glob.

Is this about not supporting curly braces?
In my opinion, as I commented below, it would not be a problem to provide the full set.

> We don't need to care about curly braces here. We should make it available to all who want to use it because we can avoid the collision by setting `path_delimiter`. We can explain it in the documentation or check it when we do `@path.split(@path_delimiter)` and give a warning message, if it is needed.

In addition, even if `path_delimiter` is the default value `,`, there seems to be no concern about causing the glob to fail because the path split is done first.
Do you have concerns about this?

> However, I didn't fully follow your suggestion what "no" means here. Perhaps, you would be interested to collect `*` contained files with in_tail?

I understand your concern.
This option is not necessarily needed.
If you think you don't need it, let's remove it!

This may rarely actually help.
I just thought it would be natural to have an option to disable the feature completely.
We can add the option if we need it later.

cosmo0920:
> > `extended`: This should be more concrete name for that option. Because we didn't provide full set of the glob patterns on Dir.glob.
> 
> Is this about not supporting curly braces? In my opinion, as I commented below, it would not be a problem to provide the full set.

Yes. I will not support curly braces here. Because this is intended to extend a capability to handle square brackets in glob patterns. This should be suitable to consider how extend the glob capabilities.



> > We don't need to care about curly braces here. We should make it available to all who want to use it because we can avoid the collision by setting `path_delimiter`. We can explain it in the documentation or check it when we do `@path.split(@path_delimiter)` and give a warning message, if it is needed.
> 
> In addition, even if `path_delimiter` is the default value `,`, there seems to be no concern about causing the glob to fail because the path split is done first. Do you have concerns about this?

No, I experienced the feature collisions to support curly braces. This is because curly braces are usually used with comma `,` like as `{a,b}`. So, it is not able to extend to handle curly braces. This should be handled as a future task or put it in ice box for now.




> > However, I didn't fully follow your suggestion what "no" means here. Perhaps, you would be interested to collect `*` contained files with in_tail?
> 
> I understand your concern. This option is not necessarily needed. If you think you don't need it, let's remove it!
> 
> This may rarely actually help. I just thought it would be natural to have an option to disable the feature completely. We can add the option if we need it later.

Sure. Let's remove on that choice from the `enable_glob` parameter.

daipom:
@cosmo0920 

> > Is this about not supporting curly braces? In my opinion, as I commented below, it would not be a problem to provide the full set.
> 
> Yes. I will not support curly braces here. Because this is intended to extend a capability to handle square brackets in glob patterns only. This should be suitable for the first step to consider how extend the glob capabilities.

I see.
I am concerned that once an option is added, it is difficult to change it later to keep compatibility.
If it's not a problem to provide the full set, I'd like `always`.

> > In addition, even if `path_delimiter` is the default value `,`, there seems to be no concern about causing the glob to fail because the path split is done first. Do you have concerns about this?
> 
> No, I experienced the feature collisions to support curly braces. This is because curly braces are usually used with comma `,` like as `{a,b}`. So, it is not able to extend to handle curly braces easily. This should be handled as a future task or put it in ice box for now.

Hmm, but it can be avoided by changing `path_delimiter`.
Users who want to use curly braces can use it by changing `path_delimiter`.
I don't see the benefit of prohibiting it.

Rather, the point is whether it would negatively impact users who **don't** use curly braces. (users who want to use `[]` only)
We must avoid it.
And, I wouldn't be concerned about that.
Supporting curly braces seems to have no negative impact on such users. 

In summary, from the following three points, it seems to me that there is no need to narrow the conditions now.

* Once an option is added, it is difficult to change it later to keep compatibility.
* Users who want to use curly braces can use it by changing `path_delimiter`.
* Supporting curly braces seems to have no negative impact on users who want to use `[]` only. 

What do you think?

cosmo0920:
> * Once an option is added, it is difficult to change it later to keep compatibility.
> * Users who want to use curly braces can use it by changing `path_delimiter`.
> * Supporting curly braces seems to have no negative impact on users who want to use `[]` only.

1. Got it. I'll change the choice name later.
2. I **don't** think so. _The most of the functionality should work well out of the box without any additional settings._ Your suggestion is definitely opposite my concerns and recommendations.
3. Same as the above. But curly braces shall introduce the functional glitches that I already pointed out.

cosmo0920:
Additionally, if we were supporting the curly braces in in_tail, users should confuse how to specify `path_delimiter` character. The current default character is `,`. This should introduce the complexity which user must choose the delimiter that won't collide the divide the path for monitoring.
I wish I could add this feature in this PR but adding a capability to handle curly braces introduces such complexity.
If we had a cycle to support user forums and community, it would be nice to add.
However, our supporting resource is _limited_ for OSS version, right? So, I wouldn't prefer to add this feature.

daipom (approved):
>     2. I **don't** think so. _The most of the functionality should work well out of the box without any additional settings._ Your suggestion is definitely opposite my concerns and recommendations.

> Additionally, if we were supporting the curly braces in in_tail, users should confuse how to specify `path_delimiter` character. The current default character is `,`. This should introduce the complexity which user must choose the delimiter that won't collide the divide the path for monitoring. I wish I could add this feature in this PR but adding a capability to handle curly braces introduces such complexity. If we had a cycle to support user forums and community, it would be nice to add. However, our supporting resource is _limited_ for OSS version, right? So, I wouldn't prefer to add this feature.

Thanks for your explanation!
I see! I understand the points!

I don't feel we need to go that far in this case, but I understand it's a very important idea.
I understand your idea, so if there are no other opinions, I agree with your direction!

Note: Just to be clear, let me explain why I don't think so in this case.

* Including curly braces in the condition itself is not harmful.
  * It only affects those who want to use curly braces.
* Simpler specifications are better for users.
  * It is less confusing for users to say, "You can use all of them, so use whatever you like, but be careful when using curly braces", rather than having users understand that some characters are exceptions to the specification,

cosmo0920:
> Note: Just to be clear, let me explain why I don't think so in this case.
> 
> * Including curly braces in the condition itself is not harmful.
>   
>   * It only affects those who want to use curly braces.
> * Simpler specifications are better for users.
>   
>   * It is less confusing for users to say, "You can use all of them, so use whatever you like, but be careful when using curly braces", rather than having users understand that some characters are exceptions to the specification,

I understand the opinions on your side. Yeah, I'm always confusing how should I do/implement this newly implementing feature?
Then, I assumed myself as a newbie who is not knowing well how to use/work this software.
With this assumption, I often realized that the option/direction which should be confusing easily is not good direction. If I were implementing a feature which is silently dependent another feature, I felt sick to use that feature or simply give up to use.
So, my responded concerns are based on the above assumptions.

Thanks for your suggestions. They are really helpful and well considered. You rock! :+1:

cosmo0920:
@ashie @kenhys Do you guys have any concerns?

ashie:
> @ashie @kenhys Do you guys have any concerns?

Still reading the discussion.

ashie:
For the discussion about braces, basically I agree with @cosmo0920.
I don't want to introduce implicit dependencies between parameters which may confuse end users.

On the other hand, the enum label `:always` doesn't seem suitable if we drop brace support.
I tried to consider more suitable name or other configuration way, but I haven't yet find the best one.

How about this one as a a compromise?

* Support brace and keep the label name `:always`
* Check `:enable_glob`, `path` and `path_delimiter` configurations at start up then show error or warning when `path` includes `{}` and `path_delimiter` is default value `,` (and `:enable_glob` is `:always`).

kenhys:
enhancement concept is reasonable, but I felt some concern.

* `wildcard`  accepts `*` in path, but it also expands other patterns. wildcard may be misreading in some extent
even though it is same behavior as before.

daipom:
@ashie 
> How about this one as a a compromise?
> 
>     * Support brace and keep the label name `:always`
> 
>     * Check `:enable_glob`, `path` and `path_delimiter` configurations at start up then show error or warning when `path` includes `{}` and `path_delimiter` is default value `,` (and `:enable_glob` is `:always`).

I agree.

@kenhys 
> enhancement concept is reasonable, but I felt some concern.
> 
>     * `wildcard`  accepts `*` in path, but it also expands other patterns. wildcard may be misreading in some extent
>       even though it is same behavior as before.

I understand your concern.
This is a difficult point in this enhancement.

This option manages only the condition to enable the glob feature.
It does not limit the glob feature itself.

This would be an option name issue.
I proposed `enable_glob` because I thought this was the condition to enable the blog feature.
If you have a more descriptive name, let's change it.

cosmo0920:
Got it. `glob_policy` may better to change from `enable_glob`. This is because `enable_glob` implies true or false.

cosmo0920:
> Do we need to care about the cycles?

Hmm, I mean cycles as the time to be able to spent for something.

cosmo0920:
> * `wildcard`  accepts `*` in path, but it also expands other patterns. wildcard may be misreading in some extent
>   even though it is same behavior as before.

I'm also considered which names should be better but I didn't find suitable and relatively shorter word. So, I just use `backward_compatible` for keeping backward compatiblities.

cosmo0920:
implementing `always` option should be done.

ashie (approved):
LGTM

Thanks for your work!
I don't have further objection.

daipom:
I understand the points!
The current proposal is to make the following specifications, right?

* `glob_policy always`:  We can use all glob features, but `@path_delimiter` must be changed.
* `glob_policy extended`: We can use glob features except curly braces. If using curly braces, it will be ConfigError.
* `glob_policy backward_compatible`: Glob features are enabled if `path` includes `*`. (The current spec).

Great!! It looks good to me.

I have one concern about the name `backward_compatible`.

> > * `wildcard`  accepts `*` in path, but it also expands other patterns. wildcard may be misreading in some extent
> >   even though it is same behavior as before.
> 
> I'd also considered which name should be better but I didn't find any suitable and relatively shorter word. So, I just use `backward_compatible` for keeping backward compatiblities.

Wouldn't users be more confused when they see this default value named `backward_compatible`?

If the priority is to make it easy for users to understand, it would be better to be able to assume the feature from the name.
If we think about it, the setting name could be something like this, for example,

* `glob_policy enable_with_wildcard`

What do you think about this? @cosmo0920 @kenhys

cosmo0920:
> * `glob_policy enable_with_wildcard`
> 
> What do you think about this? @cosmo0920 @kenhys

I think that `only_wildcard` would be better to use but I wanted to emphasize the default value is "barkward compatible".

daipom:
> > * `glob_policy enable_with_wildcard`
> > 
> > What do you think about this? @cosmo0920 @kenhys
> 
> I think that `only_wildcard` would be better to use 

Why? It looks to me to be misleading as @kenhys is concerned:

> wildcard accepts * in path, but it also expands other patterns.

---

> but I wanted to emphasize the default value is "backward compatible".

Why? It may not confuse existing users, but wouldn't it confuse new users?

cosmo0920:
> > > * `glob_policy enable_with_wildcard`
> > > 
> > > What do you think about this? @cosmo0920 @kenhys
> > 
> > 
> > I think that `only_wildcard` would be better to use
> 
> Why? It looks to me to be misleading as @kenhys is concerned:
> 
> > wildcard accepts * in path, but it also expands other patterns.

As @kenhys said that this was intended to represent for the extending regular expressions not only wildcard but for other expressions. But the current circumstances request to this option for only handling wildcard.



> > but I wanted to emphasize the default value is "backward compatible".
> 
> Why? It may not confuse existing users, but wouldn't it confuse new users?

No, it would confuse new users. Because the default value should be backward compatible and the name should be noun. Not including verbs. Maybe perfect sentence would be effective but sometimes it makes redundant for the users.

But, only using `why?` is just like representing that a parent is anger for their kids. It's too direct to represent negative emotions. So, it wouldn't be suitable behavior on the community. Did you follow what I mean?

daipom:
I see. I would also like to hear @kenhys's opinion.

> But, only using `why?` is just like representing that a parent is anger for their kids. It's too direct to represent negative emotions. So, it wouldn't be suitable behavior on the community. Did you follow what I mean?

I apologize. I didn't understand the nuance. :cry:
Thanks for letting me know.
Could you tell me how you normally ask for a reason?

cosmo0920:
> > But, only using `why?` is just like representing that a parent is anger for their kids. It's too direct to represent negative emotions. So, it wouldn't be suitable behavior on the community. Did you follow what I mean?
> 
> I apologize. I didn't understand the nuance. 😢 Thanks for letting me know. Could you tell me how you normally ask for a reason?

How come bra bra? or How come? should be better instead of using why?
Also, asking reason in text is sometimes difficult to represent in English via direct translation from Japanese.
As might you noticed that I often use would/should. Because these auxiliary verb should work for a room of choices for listeners. Making choices and take a distance should be represent politeness in English.

cosmo0920:
> > But, only using `why?` is just like representing that a parent is anger for their kids. It's too direct to represent negative emotions. So, it wouldn't be suitable behavior on the community. Did you follow what I mean?
> 
> I apologize. I didn't understand the nuance. 😢 Thanks for letting me know. Could you tell me how you normally ask for a reason?

And no worries. I'm not native English speaker, neither. But I've been in an "immersion" environment in these years.

daipom:
I see! Thanks!

I use English without understanding the nuances. That's difficult for me. 😢 

I would very appreciate it if you would let me know if my English is impolite.
Thanks so much!

kenhys (approved):
LGTM.

daipom (approved):
I commented on one question, but it's basically looks good to me.
Thanks so much for this enhancement!

lib/fluent/plugin/in_tail.rb
```diff
@@ -283,6 +293,22 @@ def have_read_capability?
...
 
+    def extended_glob_pattern(path)
+      path.include?('*') || path.include?('?') || /\[.*\]/.match(path)
+    end
+
+    # Curly braces is not supported for now because the default delimiter of path is ",".
+    # This should be collided for wildcard pattern for curly braces.
+    def use_glob?(path)
+      if @glob_policy == :always
+        extended_glob_pattern(path) || /\{.*,.*\}/.match(path)
```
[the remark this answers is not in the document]
cosmo0920: This is because I wanted to show the possibly handled regular expressions.
If always returning true, there is no hint to debug for erroneous results from `Dir.glob(..)`.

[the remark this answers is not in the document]
daipom: Hmm.
Doesn't this make it look as if there are cases where glob should not be done even when `glob_policy` is `always`?
If you want to leave information for debugging, how about using comments (instead of using codes)?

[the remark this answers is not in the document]
cosmo0920: Fair enough. So, I added comments to describe why we don't return `true` here.

cosmo0920:
Also, I found that the comment for curly braces is already outdated. I updated it, too.
This is why I didn't want to use comments too much for describing the reasons. 😅

lib/fluent/plugin/in_tail.rb
```diff
@@ -283,6 +293,28 @@ def have_read_capability?
...
+    # because the default delimiter of path is ",".
+    # This should be collided for wildcard pattern for curly braces and
+    # be handled as an error on #configure.
+    def use_glob?(path)
+      if @glob_policy == :always
+        # For future extensions, we decided to use `always' term to handle
+        # regular expressions as much as possible.
+        # This is because not using `true' as a returning value
+        # when choosing :always here.
+        extended_glob_pattern(path) || /\{.*,.*\}/.match(path)
```
daipom, line 310: Thanks! Now I can read the intent of the code.

This is fine, but if we reduce comments as much as possible, I would do the following.

```suggestion
        # Intended to be equivalent to `extended_glob_pattern(path) || /\{.*,.*\}/.match(path)`
        return true
```

Would you prefer to keep it as is? If so, I would not object anymore.
Let's merge this!

Thanks for helping us out so far!

daipom: If so, could we simplify the comments as follows?

```suggestion
        # To reduce the cost of calling `Dir.glob` every time
        extended_glob_pattern(path) || /\{.*,.*\}/.match(path)
```

cosmo0920: Your suggestion will completely remove the intent of the why we choose _always_ term in the glob_policy option.
Just simplifying is not good option to describe "simply" why we choose this.

daipom: OK!
We don't need to force simplification, but isn't the following information the most important?

> there is a discussion for how much the calling costs between Dir.glob and regular expressions. I'd say validating with regular expressions might be lower costs to process all the specified paths.

cosmo0920: Not sure. I don't have a benchmark result for it. 
The most important things should be described there:
* Why we choose _always_ term in the glob_policy option
* Display possible regular expressions there (as a code not in the comment)

cosmo0920: FYI: I can describe the reason why I wrote the above comment.
_I'd say_ implies that I'm going to say what is not certainly things but I wanted to tell
_might_ implies 50% percent or less possibility. Note that it's not equivaIent to "たぶん" in Japanese. That equivalent is _probably_ not _might_.

daipom: Thanks! I see.

With just the current comment, I misread the codes as https://github.com/fluent/fluentd/pull/4401#discussion_r1551357082.
So, I'd be glad if we could clarify the intent of the code.

Even if it's only 50%, it would be good to leave it as a comment.
For example, like this.

```suggestion
        # For future extensions, we decided to use `always' term to handle
        # regular expressions as much as possible.
        # It might be costly to call `Dir.glob` every time.
        # This is because not using `true' as a returning value
        # when choosing :always here.
        extended_glob_pattern(path) || /\{.*,.*\}/.match(path)
```

However, I understand the disadvantages of having more comments.
In addition, we can fix those comments later.
So, I think it's fine as it is!

cosmo0920: Thanks for continuously patient reviewing, @daipom. You rock! :muscle: 💎

daipom:
Merged!
Thanks so much for this enhancement @cosmo0920 !
Thanks, everyone!

PR #4380, opened 2024-01-15
WIP: Refine CI

ashie:
**Which issue(s) this PR fixes**: 
N/A
See also: https://github.com/fluent/fluentd/pull/4375#issuecomment-1886109420

**What this PR does / why we need it**: 

Refine CI settings:

* Unify CI settings for each platforms to one file
* Don't run CI for Ruby head on each `push`/`pull_request` event
  * Because Ruby head is continually changed and not released yet, running CI one shot isn't so meaningful.
  * Instead run it regularly on every Sunday.
* Add `TESTOPT=-v` to all platfroms
  * It's useful to investigate when a test is stalled.
* Remove a hack for Ruby 3.0 on Windows
  * It seems no longer needed.

**Docs Changes**:
N/A

**Release Note**: 
N/A

PR #4428, opened 2024-03-11
Backport (v1.16): Refine CI (#4380)

daipom:
* ##4380


kenhys:
Should we use .yaml instead of .yml for consistency?

.github/workflows/test-ruby-head.yaml
```diff
@@ -0,0 +1,31 @@
+name: Test
+
+on:
+  schedule:
+    - cron: '11 14 * * 0'
```
ashie, line 5: It would be better to choose a different time than master branch.
BTW I'm not sure whether it happens for this branch or master branch.

daipom: Oh! I see. I'll consider it.

daipom: > BTW I'm not sure whether it happens for this branch or master branch.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

It seems that it happens for the default branch.

I am very confused by the system that allows this can be set up per branch...

daipom: I'm not sure, but I'm starting to think it's correct to set `schedule` only on the default branch (master).

ashie: Just an idea is remove this for v1.16 branch, because we never support v1.16 against Ruby head.

daipom: Agree!

daipom:
> Should we use .yaml instead of .yml for consistency in fluent/fluentd?

Thanks for

* #4429

We can backport this later as well.

ashie (approved):

daipom:
Thanks!

daipom:
Need to backport some commits to fix Ruby 3.3 tests. I'm doing.

PR #4429, opened 2024-03-11
github: unify YAML file extension to .yml

kenhys:

Fixes N/A

It is not harmful because GitHub allows both of file extensions (.yaml and .yml), but it may be better to unify it for consistency.

.yml is used in https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions

**Docs Changes**:

N/A

**Release Note**: 

N/A


daipom:
Thanks!

> .yml is used in https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions

I agree with unifying `.yml` based on this reason.

Should we wait a bit to see if there are any other opinions?

daipom (approved):

ashie (approved):

PR #4437, opened 2024-03-14
Apply v1.16.4 to master

daipom:
* Version
* CHANGELOG

When releasing v1.16.3, we forgot to apply the version to the master.
So, the version on the master is still v1.16.2.
Currently, the master is the same minor version as the LTS branch v1.16, so we should update it along with the LTS branch, until releasing v1.17 on the master.


kenhys (approved):
LGTM.

daipom:
Thanks!

PR #4452, opened 2024-03-27
ci: do not show test detail immediately

kenhys:

N/A

TESTOPTS=-v shows running test and the result.
If something weird happen, it show error immediately.

As rake task executes many tests, so it may be better to delay showing test details later.

NOTE: --progress-style=fault-only helps you to focus on failure, but it is inconvenient when test case has stalled.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/.github/workflows/test-ruby-head.yml
+++ b/.github/workflows/test-ruby-head.yml
@@ -28,4 +28,4 @@ jobs:
       - name: Install dependencies
         run: bundle install
       - name: Run tests
-        run: bundle exec rake test TESTOPTS=-v
+        run: bundle exec rake test TESTOPTS="-v --no-show-detail-immediately"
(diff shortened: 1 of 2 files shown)


.github/workflows/test-ruby-head.yml
```diff
@@ -28,4 +28,4 @@ jobs:
       - name: Install dependencies
         run: bundle install
       - name: Run tests
-        run: bundle exec rake test TESTOPTS=-v
+        run: bundle exec rake test TESTOPTS="-v --no-show-detail-immediately"
```
ashie, line 31: Oh, I didn't know such option!

daipom (approved):
Great!! I want this!!

kenhys:
Error detail is shown before Omission:

image

daipom:
Perfect!! Thanks!!

PR #4459, opened 2024-04-02
Implements AWS SigV4 for the HTTP output plugin. 

### LINKED ISSUE #4444: Support AWS SigV4 in the http output plugin

We would like to enable Fluentd to write to [Amazon OpenSearch Ingestion](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/ingestion.html) (OSIS). The current `http` output plugin in Fluentd can write to open-source Data Prepper's `http` source plugin, which OSIS supports.

All OSIS pipelines require [AWS SigV4](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) authentication which the Fluentd `http` output plugin does not support.

Provide an AWS SigV4 option in Fluentd's `http` output plugin. Create a new `auth` option named `aws_sigv4`. Add options with the `aws_` prefix to support AWS configurations.

```
  <auth>
    method aws_sigv4
    aws_service osis
    aws_region us-east-1
    aws_role_arn arn:aws:iam::123456789012:role/MyRole
  </auth>
```

The `aws_role_arn` will be optional. If not provided, the plugin will use the default credential chain. If provided, the plugin will assume the role supplied by `aws_role_arn`.

One alternative considered is a new OSIS output plugin. However, this seems unnecessary as the `http` plugin is sufficient. Also, there may be other use-cases of writing `http` to other AWS services.

### Additional context

_No response_

dlvenable:
Fixes #4444

This PR adds a new authentication method to the HTTP output plugin for AWS sigV4. We need it so that customers can use Fluentd to send data to Amazon OpenSearch Ingestion which supports the Fluentd output plugin, but requires SigV4 authentication.

**Docs Changes**:

* `aws_sigv4` - When this option is specified, Fluentd will sign requests using AWS Signature Version 4.

* `aws_service` - The AWS service to authenticate against
* `aws_region` - The AWS region to use when authenticating
* `aws_role_arn` - The AWS role ARN to assume when authenticating

You can optionally specify an `aws_role_arn`. If you provide it, Fluentd will assume that AWS role and send requests signing from that role. Otherwise, Fluentd will use the credentials found by the [credential provider chain](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html) as defined in the AWS documentation.

**Release Note**: 

The http output plugin supports AWS Signature Version 4 authentication.

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -29,6 +29,9 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("tzinfo-data", ["~> 1.0"])
   gem.add_runtime_dependency("strptime", [">= 0.2.4", "< 1.0.0"])
   gem.add_runtime_dependency("webrick", ["~> 1.4"])
+  gem.add_runtime_dependency("aws-sigv4", ["~> 1.8"])
+  gem.add_runtime_dependency("aws-sdk-sts", ["~> 1.11"])
+  gem.add_runtime_dependency("rexml", ["~> 3.2"])
 
   # gems that aren't default gems as of Ruby 3.4
   gem.add_runtime_dependency("base64", ["~> 0.2"])
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -29,9 +29,6 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("tzinfo-data", ["~> 1.0"])
   gem.add_runtime_dependency("strptime", [">= 0.2.4", "< 1.0.0"])
   gem.add_runtime_dependency("webrick", ["~> 1.4"])
-  gem.add_runtime_dependency("aws-sigv4", ["~> 1.8"])
-  gem.add_runtime_dependency("aws-sdk-sts", ["~> 1.11"])
-  gem.add_runtime_dependency("rexml", ["~> 3.2"])
 
   # gems that aren't default gems as of Ruby 3.4
   gem.add_runtime_dependency("base64", ["~> 0.2"])
@@ -59,4 +56,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency("oj", [">= 2.14", "< 4"])
   gem.add_development_dependency("async", "~> 1.23")
   gem.add_development_dependency("async-http", ">= 0.50.0")
+  gem.add_development_dependency("aws-sigv4", ["~> 1.8"])
+  gem.add_development_dependency("aws-sdk-core", ["~> 3.191"])
+  gem.add_development_dependency("rexml", ["~> 3.2"])
 end
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -263,7 +263,6 @@ module Fluent::Plugin
         end
       end
       req['Content-Type'] = @content_type
-      req['Host'] = uri.host
     end
 
     def set_auth(req, uri)
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -20,8 +20,6 @@ require 'openssl'
 require 'fluent/tls'
 require 'fluent/plugin/output'
 require 'fluent/plugin_helper/socket'
-require 'aws-sigv4'
-require 'aws-sdk-core'
 
 # patch Net::HTTP to support extra_chain_cert which was added in Ruby feature #9758.
 # see: https://github.com/ruby/ruby/commit/31af0dafba6d3769d2a39617c0dddedb97883712
@@ -131,6 +129,12 @@ module Fluent::Plugin
       end
 
       if @auth and @auth.method == :aws_sigv4
+        begin
+          require 'aws-sigv4'
+          require 'aws-sdk-core'
+        rescue LoadError
+          raise Fluent::ConfigError, "The aws-sigv4 and aws-sdk-core gems are required for aws_sigv4 auth. Run: 'gem install aws-sigv4 -v 1.8.0' and 'gem install aws-sdk-core -v 3.191'."
+        end
 
         raise Fluent::ConfigError, "aws_service is required for aws_sigv4 auth" unless @auth.aws_service != nil
         raise Fluent::ConfigError, "aws_region is required for aws_sigv4 auth" unless @auth.aws_region != nil
@@ -262,16 +266,7 @@ module Fluent::Plugin
       req['Host'] = uri.host
     end
 
-    def create_request(chunk, uri)
-      req = case @http_method
-            when :post
-              Net::HTTP::Post.new(uri.request_uri)
-            when :put
-              Net::HTTP::Put.new(uri.request_uri)
-            end
-      set_headers(req, uri, chunk)
-      req.body = @json_array ? "[#{chunk.read.chop}]" : chunk.read
-
+    def set_auth(req, uri)
       if @auth
         if @auth.method == :basic
           req.basic_auth(@auth.username, @auth.password)
@@ -291,8 +286,23 @@ module Fluent::Plugin
           req.add_field('authorization', signature.headers['authorization'])
         end
       end
+    end
+
+    def create_request(chunk, uri)
+      req = case @http_method
+            when :post
+              Net::HTTP::Post.new(uri.request_uri)
+            when :put
+              Net::HTTP::Put.new(uri.request_uri)
+            end
+      set_headers(req, uri, chunk)
+      req.body = @json_array ? "[#{chunk.read.chop}]" : chunk.read
+
+      # At least one authentication method requires the body and other headers, so the order of this call matters
+      set_auth(req, uri)
       req
     end
+
 
     def send_request(uri, req)
       res = if @proxy_uri
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -20,6 +20,8 @@ require 'openssl'
 require 'fluent/tls'
 require 'fluent/plugin/output'
 require 'fluent/plugin_helper/socket'
+require 'aws-sigv4'
+require 'aws-sdk-core'
 
 # patch Net::HTTP to support extra_chain_cert which was added in Ruby feature #9758.
 # see: https://github.com/ruby/ruby/commit/31af0dafba6d3769d2a39617c0dddedb97883712
@@ -87,11 +89,17 @@ module Fluent::Plugin
 
     config_section :auth, required: false, multi: false do
       desc 'The method for HTTP authentication'
-      config_param :method, :enum, list: [:basic], default: :basic
+      config_param :method, :enum, list: [:basic, :aws_sigv4], default: :basic
       desc 'The username for basic authentication'
       config_param :username, :string, default: nil
       desc 'The password for basic authentication'
       config_param :password, :string, default: nil, secret: true
+      desc 'The AWS service to authenticate against'
+      config_param :aws_service, :string, default: nil
+      desc 'The AWS region to use when authenticating'
+      config_param :aws_region, :string, default: nil
+      desc 'The AWS role ARN to assume when authenticating'
+      config_param :aws_role_arn, :string, default: nil
     end
 
     def initialize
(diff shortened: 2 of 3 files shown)


daipom:
@dlvenable Thanks so much for this enhancement!
I am very happy that adding [AWS SigV4](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) authentication will allow Fluentd to work with AWS services such as Prepper!

fluentd.gemspec
```diff
@@ -29,6 +29,9 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency("tzinfo-data", ["~> 1.0"])
   gem.add_runtime_dependency("strptime", [">= 0.2.4", "< 1.0.0"])
   gem.add_runtime_dependency("webrick", ["~> 1.4"])
+  gem.add_runtime_dependency("aws-sigv4", ["~> 1.8"])
+  gem.add_runtime_dependency("aws-sdk-sts", ["~> 1.11"])
+  gem.add_runtime_dependency("rexml", ["~> 3.2"])
```
ashie, line 34: We want to minimize mandatory requirements for installing fluentd core.
So it would be better that make these gems optional.
For example:

* Remove these gems from fluentd.gemspec
* Do `require 'aws-sigv4'` and `require 'aws-sdk-core'` only when AWS related parameters are set.
* Show error log then prompt users to install these gems when `require` is failed.

ashie (changes requested):

lib/fluent/plugin/out_http.rb
```diff
@@ -20,6 +20,8 @@
 require 'fluent/tls'
 require 'fluent/plugin/output'
 require 'fluent/plugin_helper/socket'
+require 'aws-sigv4'
+require 'aws-sdk-core'
```
cosmo0920, line 24: These gems aren't always required. So, as @ashie mentioned, we need to verify whether `:aws_sigv4` method is used or not in #configure method.
When the aws related gems are not installed to attempt to use this feature, then the error should be displayed.

lib/fluent/plugin/out_http.rb
```diff
@@ -227,6 +259,7 @@ def set_headers(req, chunk)
         end
       end
       req['Content-Type'] = @content_type
+      req['Host'] = uri.host
```
daipom, line 262: Could you tell me why you changed the host header to be explicitly set?
(Sorry I'm not familiar with this, but I'm concerned that this change may affect the existing logic.)

lib/fluent/plugin/out_http.rb
```diff
@@ -236,11 +269,28 @@ def create_request(chunk, uri)
...
+              'Host' => uri.host
+            },
+            body: req.body
+          )
+          req.add_field('x-amz-date', signature.headers['x-amz-date'])
+          req.add_field('x-amz-security-token', signature.headers['x-amz-security-token'])
+          req.add_field('x-amz-content-sha256', signature.headers['x-amz-content-sha256'])
+          req.add_field('authorization', signature.headers['authorization'])
+        end
       end
```
daipom, line 293: Since there are more code lines for `auth`, it would be better to extract them into a new method, such as `set_auth`. (like `set_headers`)

lib/fluent/plugin/out_http.rb
cosmo0920: It seems that it's required for SignV4 requirements: https://github.com/aphillips19/aws-sdk-ruby/blob/master/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L173
https://github.com/aphillips19/aws-sdk-ruby/blob/master/gems/aws-sigv4/lib/aws-sigv4/signature.rb#L18-L22
And its specification: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

BTW, the aws-sdk's source documentation says that:

>       # @return [Hash<String,String>] A hash of headers that should
>      #   be applied to the HTTP request. Header keys are lower
>      #   cased strings and may include the following:
>      #
>      #   * 'host'
>      #   * 'x-amz-date'
>      #   * 'x-amz-security-token'
>      #   * 'x-amz-content-sha256'
>      #   * 'authorization'

Is `host' header always needed?

dlvenable: @daipom , Yes, the `Host` header needs to be signed. I found that Fluentd was not sending an `Host` header. And the documentation for SigV4 states that this must be signed header.

I can test to see if having an empty `Host` header works with SigV4. Another alternative is to only set the `Host` header when using SigV4.

I chose this route because the `Host` header can be helpful, especially when using a proxy. I'm happy to take one of the alternatives I mentioned above if you prefer.

fluentd.gemspec
dlvenable: @ashie , Thank you for the solid suggestions. I've made these changes accordingly. The one difference is that I moved these dependencies to development dependencies so that the tests run.

lib/fluent/plugin/out_http.rb
dlvenable: @daipom , I was able to remove the `Host` header and still authenticate. So I've remove this from this PR.

dlvenable:
Thank you all for the useful feedback. I've pushed changes to address these comments.

fluentd.gemspec
ashie: It seems nice. Thanks!

lib/fluent/plugin/out_http.rb
ashie: Resolved.

lib/fluent/plugin/out_http.rb
ashie: Resolved.

lib/fluent/plugin/out_http.rb
```diff
@@ -229,21 +265,44 @@ def set_headers(req, chunk)
...
+              'Host' => uri.host
+            },
+            body: req.body
+          )
+          req.add_field('x-amz-date', signature.headers['x-amz-date'])
+          req.add_field('x-amz-security-token', signature.headers['x-amz-security-token'])
+          req.add_field('x-amz-content-sha256', signature.headers['x-amz-content-sha256'])
+          req.add_field('authorization', signature.headers['authorization'])
+        end
+      end
```
daipom, line 287: Thanks for extracting these codes to the new method!

It's not very important, but how about inverting `if @auth` condition? (`return unless @auth`)

```suggestion
      return unless @auth
      if @auth.method == :basic
        req.basic_auth(@auth.username, @auth.password)
      elsif @auth.method == :aws_sigv4
        signature = @aws_signer.sign_request(
          http_method: req.method,
          url: uri.request_uri,
          headers: {
            'Content-Type' => @content_type,
            'Host' => uri.host
          },
          body: req.body
        )
        req.add_field('x-amz-date', signature.headers['x-amz-date'])
        req.add_field('x-amz-security-token', signature.headers['x-amz-security-token'])
        req.add_field('x-amz-content-sha256', signature.headers['x-amz-content-sha256'])
        req.add_field('authorization', signature.headers['authorization'])
      end
```

lib/fluent/plugin/out_http.rb
daipom: > I found that Fluentd was not sending an `Host` header.

Thanks for confirming it.
I am also concerned about this point.

`Host` header seems to be set automatically, but I'm not sure.
Is Fluentd really sending requests without a host header?
If so, we may need to put a host header like this.

daipom (approved):
I commented on my concerns, but it looks basically good to me.
Thanks so much for this enhancement!

lib/fluent/plugin/out_http.rb
```diff
@@ -121,6 +127,36 @@ def configure(conf)
...
         define_singleton_method(:format, method(:format_json_array))
       end
+
+      if @auth and @auth.method == :aws_sigv4
+        begin
+          require 'aws-sigv4'
+          require 'aws-sdk-core'
+        rescue LoadError
+          raise Fluent::ConfigError, "The aws-sigv4 and aws-sdk-core gems are required for aws_sigv4 auth. Run: 'gem install aws-sigv4 -v 1.8.0' and 'gem install aws-sdk-core -v 3.191'."
+        end
```
cosmo0920, line 137: I have some concern for installing XML parser. Because AWS requests are basically handled via XML responses. So, we need to indicate that XML parsers are also required to work for AWS SigV4 requests.
Should we add another require checks for XML parser gems such as nokogiri or rexml?

cosmo0920 (approved):
Basically, it looks good to me. I added a comment about XML parsers.
aws-sdk-ruby needs to work with XML parsers to handle XML responses from AWS services.

lib/fluent/plugin/out_http.rb
dlvenable: @cosmo0920,

I considered this, but the AWS SDK already does this check and includes information on the available parsers. I think we should rely on theirs because otherwise we need to check for the same list that they have.

https://github.com/aws/aws-sdk-ruby/blob/a703da0525ac26d2ac4a68253b3718b6d86461b8/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L74-L76

dlvenable:
@daipom , @cosmo0920 , I see that some tests are failing, but they don't appear to be related to http. Are these possibly flaky tests?

I'm very interested in getting this PR in to allow us to move forward. Thank you.

lib/fluent/plugin/out_http.rb
cosmo0920: Ah, sounds reasonable.

lib/fluent/plugin/out_http.rb
```diff
@@ -121,6 +127,36 @@ def configure(conf)
         end
         define_singleton_method(:format, method(:format_json_array))
       end
+
+      if @auth and @auth.method == :aws_sigv4
+        begin
+          require 'aws-sigv4'
+          require 'aws-sdk-core'
+        rescue LoadError
+          raise Fluent::ConfigError, "The aws-sigv4 and aws-sdk-core gems are required for aws_sigv4 auth. Run: 'gem install aws-sigv4 -v 1.8.0' and 'gem install aws-sdk-core -v 3.191'."
```
ashie, line 136: Sorry, one more thing.
Does it really need to stick to aws-sigv4  v1.8.0?
aws-sdk-core v3.191 depends on `aws-sigv4 ~> 1.8`, so that latest version of aws-sigv4 1.8 series will be installed automatically when user installs aws-sdk-core v3.191 series.
It would be happy if we can omit specifying strict version of aws-sigv4 to receive latest bug fixes (especially vulnerability fixes).

lib/fluent/plugin/out_http.rb
ashie: `Host` header must be sent in all HTTP/1.1 requests

* https://datatracker.ietf.org/doc/html/rfc7230#section-5.4
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

So it's done in standard library.

lib/fluent/plugin/out_http.rb
ashie: I also prefer early return style.

cosmo0920:
> @daipom , @cosmo0920 , I see that some tests are failing, but they don't appear to be related to http. Are these possibly flaky tests?
> 
> I'm very interested in getting this PR in to allow us to move forward. Thank you.

These failed tests are flaky tests on macOS and Windows. They should be able to ignored for now.

@daipom @ashie Any ETA for getting merged this PR? In Fluent Bit side, SigV4 signature is set up in the almost HTTP related plugins. So, I wanted to equip the similar functionality as a parity for Fluent Bit.
And this PR should be prioritized to process.

ashie:
> @daipom @ashie Any ETA for getting merged this PR? 

It's almost ready for merge.
We'll merge soon when we get @dlvenable's response about https://github.com/fluent/fluentd/pull/4459#discussion_r1552620586

daipom:
Yes. I think we can merge this soon, at least once the concerns pointed out by @ashie are resolved.

lib/fluent/plugin/out_http.rb
dlvenable: @daipom , @ashie , I made this change in the last commit I pushed.

lib/fluent/plugin/out_http.rb
dlvenable: @ashie , Correct, it doesn't need to be v1.8.0 strictly. I have updated the code in the last commit.

The error message now looks like:

```
2024-04-05 08:46:10 -0500 [error]: config error file="fluent.conf" error_class=Fluent::ConfigError error="The aws-sdk-core and aws-sigv4 gems are required for aws_sigv4 auth. Run: gem install aws-sdk-core -v '~> 3.191'"
```

lib/fluent/plugin/out_http.rb
dlvenable: I looked further into this and it does include the `Host` header. As you can you see in my most recent commits, I am signing the host using `uri.host`.

```
headers: {
            'Content-Type' => @content_type,
            'Host' => uri.host
          }
```

I think the `Host` header is not set until later in the process. But, this solution has worked in my testing.

lib/fluent/plugin/out_http.rb
ashie: Thanks!

lib/fluent/plugin/out_http.rb
ashie: Thanks!

ashie (approved):

ashie:
Thanks for your contribution!

daipom:
@dlvenable Thanks for your contribution!

PR #4474, opened 2024-04-26
Make sure parser returns hash

### LINKED ISSUE #4100: JSON parser plugin can sometimes output records as strings rather than `Hash`

daipom:
* Partially Fixes #4100 
* Remake #4106
  * Add fixes for `parser_msgpack` and some refactoring

* Make sure `parser_json` and `parser_msgpack` return Hash.
* Make `parser_json` and `parser_msgpack` accept only `Hash` or `Array` of `Hash`

It is wrong for Parser to return a record that is not Hash.
Subsequent processing may result in errors.

`parser_json` and `parser_msgpack` could return non-Hash objects, so they have been fixed in this PR.
In addition, `in_http` was designed based on this wrong behavior, so it has been fixed as well in this PR.

**However, we have a remaining problem with `filter_parser`.**
`filter_parser` could return Array based on this wrong behavior.
This PR disables it, so it can not return multiple parsed results anymore.
Even though it was wrong to begin with, it's possible that this change in specifications would be unacceptable.
We need to consider this change carefully.
I explain some examples in detail below.

**Docs Changes**:
TODO

**Release Note**: 
TODO

# Example: parser_json

## Case: No subsequent processing (without `convert_values`)

* Before: Array => Single record
* After: Array => Multiple records

Config

```xml
<source>
  @type tcp
  tag test.tcp
  <parse>
    @type json
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

Operation

```console
$ netcat 0.0.0.0 5170
[{"k":"v"}, {"k2":"v2"}]
```

Result BEFORE this fix

```
2023-03-21 23:30:30.623063804 +0900 test.tcp: [{"k":"v"},{"k2":"v2"}]
```

Result AFTER this fix

```
2023-03-21 23:29:20.147728222 +0900 test.tcp: {"k":"v"}
2023-03-21 23:29:20.147754433 +0900 test.tcp: {"k2":"v2"}
```

## Case: No subsequent processing (with `convert_values`)

* Before: Array => **NoMethodError**
* After: Array => Multiple records

Config

```xml
<source>
  @type tcp
  tag test.tcp
  <parse>
    @type json
    null_empty_string
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

Operation

```console
$ netcat 0.0.0.0 5170
[{"k":"v"}, {"k2":"v2"}]
```

Result BEFORE this fix

```
2024-04-26 18:48:43 +0900 [error]: #0 unexpected error on reading data host="127.0.0.1" port=33268 error_class=NoMethodError error="undefined method `each_key' for [{\"k\"=>\"v\"}, {\"k2\"=>\"v2\"}]:Array"
```

Result AFTER this fix

```
2024-04-26 18:51:02.763961229 +0900 test.tcp: {"k":"v"}
2024-04-26 18:51:02.763971914 +0900 test.tcp: {"k2":"v2"}
```

## Case: With subsequent filter processing

* Before: Array => **NoMethodError**
* After: Array => Multiple records

Config

```xml
<source>
  @type tcp
  tag test.tcp
  <parse>
    @type json
  </parse>
</source>

<filter test.**>
  @type record_transformer
  enable_ruby true
  <record>
    class ${record.class}
  </record>
</filter>

<match test.**>
  @type stdout
</match>
```

Operation

```console
$ netcat 0.0.0.0 5170
[{"k":"v"}, {"k2":"v2"}]

```

Result BEFORE this fix

```
2023-03-21 23:30:51 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for [{\"k\"=>\"v\"}, {\"k2\"=>\"v2\"}]:Array" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.tcp" time=2023-03-21 23:30:51.040692881 +0900 record=[{"k"=>"v"}, {"k2"=>"v2"}]
```

Result AFTER this fix

```
2023-03-21 23:29:55.783183244 +0900 test.tcp: {"k":"v","class":"Hash"}
2023-03-21 23:29:55.783215076 +0900 test.tcp: {"k2":"v2","class":"Hash"}
```

# Example: filter_parser

## Case: No subsequent processing (without any option)

* Before: Array => Single record
* After: Array => **Imcomleted single record that has only the first value** **!!Remaining Problem!!**

Config

```xml
<source>
  @type sample
  tag test.array
</source>

<filter test.**>
  @type parser
  key_name message
  <parse>
    @type json
  </parse>
</filter>

<match test.**>
  @type stdout
</match>
```

Result BEFORE this fix

```
2024-04-26 18:53:30.029602646 +0900 test.array: [{"k":"v"},{"k2":"v2"}]
```

Result AFTER this fix

```
2024-04-26 18:55:28.010836436 +0900 test.array: {"k":"v"}
```

## Case: No subsequent processing (with some options such as `reserve_data`)

* Before: Array => **ParserError and return the original record**
  * `parse failed no implicit conversion of Array into Hash`
* After: Array => **Imcomleted record that has only the first value** **!!Remaining Problem!!**

Config

```xml
<source>
  @type sample
  tag test.array
</source>

<filter test.**>
  @type parser
  key_name message
  reserve_data
  <parse>
    @type json
  </parse>
</filter>

<match test.**>
  @type stdout
</match>
```

Result BEFORE this fix

```
2024-04-26 18:58:25 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="parse failed no implicit conversion of Array into Hash" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_parser.rb:110:in `rescue in filter_with_time'" tag="test.array" time=2024-04-26 18:58:25.027466823 +0900 record={"message"=>"[{\"k\":\"v\"}, {\"k2\":\"v2\"}]"}
2024-04-26 18:58:25.027466823 +0900 test.array: {"message":"[{\"k\":\"v\"}, {\"k2\":\"v2\"}]"}
```

Result AFTER this fix

```
2024-04-26 18:58:08.091274876 +0900 test.array: {"message":"[{\"k\":\"v\"}, {\"k2\":\"v2\"}]","k":"v"}
```

## Case: With subsequent filter processing

* Before: Array => **dump an error event: error_class=NoMethodError**
* After: Array => **Imcomleted record that has only the first value** **!!Remaining Problem!!**

Config

```xml
<source>
  @type sample
  tag test.array
</source>

<filter test.**>
  @type parser
  key_name message
  <parse>
    @type json
  </parse>
</filter>

<filter test.**>
  @type record_transformer
  enable_ruby true
  <record>
    class ${record.class}
  </record>
</filter>

<match test.**>
  @type stdout
</match>
```

Result BEFORE this fix

```
2024-04-26 19:06:23 +0900 [warn]: #0 dump an error event: error_class=NoMethodError error="undefined method `merge!' for [{\"k\"=>\"v\"}, {\"k2\"=>\"v2\"}]:Array" location="/home/daipom/work/fluentd/fluentd/lib/fluent/plugin/filter_record_transformer.rb:135:in `reform'" tag="test.array" time=2024-04-26 19:06:23.032939419 +0900 record=[{"k"=>"v"}, {"k2"=>"v2"}]
```

Result AFTER this fix

```
2024-04-26 19:07:25.096004693 +0900 test.array: {"k":"v","class":"Hash"}
```

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -70,14 +70,31 @@ module Fluent
       end
 
       def parse(text)
-        record = @load_proc.call(text)
-        time = parse_time(record)
-        if @execute_convert_values
-          time, record = convert_values(time, record)
+        parsed_json = @load_proc.call(text)
+
+        if parsed_json.is_a?(Hash)
+          time, record = parse_one_record(parsed_json)
+          yield time, record
+        elsif parsed_json.is_a?(Array)
+          parsed_json.each do |record|
+            unless record.is_a?(Hash)
+              yield nil, nil
+              next
+            end
+            time, parsed_record = parse_one_record(record)
+            yield time, parsed_record
+          end
+        else
+          yield nil, nil
         end
-        yield time, record
+
       rescue @error_class, EncodingError # EncodingError is for oj 3.x or later
         yield nil, nil
+      end
+
+      def parse_one_record(record)
+        time = parse_time(record)
+        convert_values(time, record)
       end
 
       def parser_type
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -79,6 +79,10 @@ module Fluent::Plugin
                 end
             @accessor.delete(record) if @remove_key_name_field
             r = handle_parsed(tag, record, t, values)
+            # Note: https://github.com/fluent/fluentd/issues/4100
+            # If the parser returns multiple records from one raw_value,
+            # this returns only the first one record.
+            # This should be fixed in the future version.
             return t, r
           else
             if @emit_invalid_record_to_error
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -203,54 +203,24 @@ module Fluent::Plugin
       begin
         path = path_info[1..-1]  # remove /
         tag = path.split('/').join('.')
-        record_time, record = parse_params(params)
 
-        # Skip nil record
-        if record.nil?
-          log.debug { "incoming event is invalid: path=#{path_info} params=#{params.to_json}" }
-          if @respond_with_empty_img
-            return RESPONSE_IMG
-          else
-            if @use_204_response
-              return RESPONSE_204
-            else
-              return RESPONSE_200
-            end
+        mes = Fluent::MultiEventStream.new
+        parse_params(params) do |record_time, record|
+          if record.nil?
+            log.debug { "incoming event is invalid: path=#{path_info} params=#{params.to_json}" }
+            next
           end
-        end
 
-        mes = nil
-        # Support batched requests
-        if record.is_a?(Array)
-          mes = Fluent::MultiEventStream.new
-          record.each do |single_record|
-            add_params_to_record(single_record, params)
-
-            if param_time = params['time']
-              param_time = param_time.to_f
-              single_time = param_time.zero? ? Fluent::EventTime.now : @float_time_parser.parse(param_time)
-            elsif @custom_parser
-              single_time = @custom_parser.parse_time(single_record)
-              single_time, single_record = @custom_parser.convert_values(single_time, single_record)
-            else
-              single_time = convert_time_field(single_record)
-            end
-
-            mes.add(single_time, single_record)
-          end
-        else
           add_params_to_record(record, params)
 
           time = if param_time = params['time']
                    param_time = param_time.to_f
                    param_time.zero? ? Fluent::EventTime.now : @float_time_parser.parse(param_time)
                  else
-                   if record_time.nil?
-                     convert_time_field(record)
-                   else
-                     record_time
-                   end
+                   record_time.nil? ? convert_time_field(record) : record_time
                  end
+
+          mes.add(time, record)
         end
       rescue => e
         if @dump_error_log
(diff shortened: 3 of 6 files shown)


lib/fluent/plugin/parser_json.rb
```diff
@@ -70,16 +70,33 @@ def configure_json_parser(name)
       end
 
       def parse(text)
-        record = @load_proc.call(text)
-        time = parse_time(record)
-        if @execute_convert_values
```
ashie, line 75: Note: No need to check `@execute_convert_values` here because it's also checked in `convert_values`.

ashie (approved):
LGTM

daipom:
Thanks! I have added `Release Note` to https://github.com/fluent/fluentd/pull/4474#issue-2265457668.

PR #4478, opened 2024-04-30
filter_parser: add error event for multiple parsed results

daipom:
* Following to #4474
* Related to #4100

The following problem of `filter_parser` remains in #4474 :

> However, we have a remaining problem with filter_parser.
filter_parser could return Array based on this wrong behavior.
This PR disables it, so it can not return multiple parsed results anymore.
Even though it was wrong to begin with, it's possible that this change in specifications would be unacceptable.
We need to consider this change carefully.
I explain some examples in detail below.

At least, it would not be happy if some events were lost without any warnings or errors.
So, this PR adds an error handling for that.

**Docs Changes**:
TODO

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -70,6 +70,13 @@ module Fluent::Plugin
         end
       end
       begin
+        # Note: https://github.com/fluent/fluentd/issues/4100
+        # If the parser returns multiple records from one raw_value,
+        # this returns only the first one record.
+        # This should be fixed in the future version.
+        result_time = nil
+        result_record = nil
+
         @parser.parse(raw_value) do |t, values|
           if values
             t = if @reserve_time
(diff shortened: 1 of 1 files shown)


daipom:
Sample:

```xml
<source>
  @type sample
  tag test.array
</source>

<filter test.**>
  @type parser
  key_name message
  <parse>
    @type json
  </parse>
</filter>

<match test.**>
  @type stdout
</match>
```

```
2024-04-30 14:35:58 +0900 [warn]: #0 dump an error event: error_class=Fluent::Plugin::Parser::ParserError error="Could not emit the event. The parser returned multiple results, but currently filter_parser plugin only returns the first parsed result. Raw data: '[{\"k\":\"v\"}, {\"k2\":\"v2\"}]'" location=nil tag="test.array" time=2024-04-30 14:35:58.020965248 +0900 record={"k2"=>"v2"}
2024-04-30 14:35:58.020948041 +0900 test.array: {"k":"v"}
```

ashie (approved):
LGTM

PR #4482, opened 2024-05-01
config: yaml_parser: Handle $log_level element for special case

cosmo0920:
**Which issue(s) this PR fixes**: 

Closes https://github.com/fluent/fluentd/issues/4463.

**What this PR does / why we need it**: 

This is because `@log_level` is invalid for YAML.
Instead, we should interpret $log_level as `@log_level` on YAML parser.

**Docs Changes**:

Needed. Will handle on it.

**Release Note**: 

Same as title.

PR #4477, opened 2024-04-30
Fix a failed test of ConsoleAdapter

ashie:
**Which issue(s) this PR fixes**: 
N/A

**What this PR does / why we need it**: 
Follow the change of the log format as of console v1.24.

**Docs Changes**:
N/A

**Release Note**: 
N/A

PR #4486, opened 2024-05-07
Backport(v1.16): config: yaml_parser: Handle $log_level element for special case (#4482)

daipom:
Backported from 236d87d07fea8b3396f33985247c96ad59911f3e. (#4482)

### DIFF
--- a/lib/fluent/config/yaml_parser/parser.rb
+++ b/lib/fluent/config/yaml_parser/parser.rb
@@ -138,6 +138,10 @@ module Fluent
             sb.add_line('@id', v)
           end
 
+          if (v = config.delete('$log_level'))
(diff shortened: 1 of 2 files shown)


ashie:
Hmm, CI is failed due to a change of dependent gem. It seems that console gem is needed pinned.

daipom:
> Hmm, CI is failed due to a change of dependent gem. It seems that console gem is needed pinned.

I have made the issue for this:

* #4487

We should resolve this first.

daipom:
For `console` v1.25, we need to consider #4487

For `console` v1.24, we need to backport #4477.

ashie (approved):

Issue #4487, opened 2024-05-07
'console' gem v1.25 causes LoadError


daipom:
### Describe the bug

https://github.com/socketry/console v1.25 series was released on 2024-05-03.

* https://github.com/socketry/console/commit/39056038e819be133d07224f4cc5c776b4c8cbb5

It causes CI failure (Ruby ver >= 3.1): https://github.com/fluent/fluentd/actions/runs/8978822806/job/24659864778?pr=4486

```
cannot load such file -- console/terminal/logger (LoadError)`
...
from /.../lib/fluent/log/console_adapter.rb:17:in `<top (required)>'
```

https://github.com/fluent/fluentd/blob/236d87d07fea8b3396f33985247c96ad59911f3e/lib/fluent/log/console_adapter.rb#L17-L24

This is caused by the following path change (`lib/console/terminal/logger.rb` -> `lib/console/output/terminal.rb`):

* https://github.com/socketry/console/compare/v1.24.0...v1.25.0#diff-6e7355da8b6a8f794e74e462aebbe4b23f7752dab2a29967d17707eabb71ce12

We need to consider how to handle this new `console` gem ver.

### To Reproduce

```bash
git clone git@github.com:fluent/fluentd.git
cd fluentd
bundle
bundle exec rake test
```

### Expected behavior

Tests are executed without errors.

### Your Environment

```markdown
- Fluentd version: 1.17.0
- Operating system: Ubuntu 20.04.6 LTS (Focal Fossa)
- Kernel version: 5.15.0-105-generic
```

### Your Configuration

```apache
None
```

### Your Error Log

```shell
[...]

PR #4489, opened 2024-05-07
Backport(v1.16): Fix a failed test of ConsoleAdapter (#4477)

daipom:
Backported from a226e28b482c8ca97ee03cd8823e32875b22cc5f. (#4477)


daipom:
This fixes the CI fails for Ruby v3.0. (related to `console` gem v1.24)

For Ruby v3.1, 3.2, and 3.3, we need to consider #4487. (related to `console` gem v1.25)

PR #4491, opened 2024-05-10
Fix line skipping issue in receive_lines method

### LINKED ISSUE #4494: in_tail plugin can cause breakdowns in fluentd
### Describe the bug

A requirement for our work is to read a 1GB single-line text. This text is generated by PostgreSQL, which has no limit for the log. However, Fluentd breaks down when attempting to read this 1GB single-line text.

I believe this could be considered a bug and vulnerability.

### To Reproduce

Add below code in test_in_tail.rb
```
  def test_max_line_size_with_large_singleline
    label = "1k"
    size = 1000000000
    config = config_element("", "", {
                          "tag" => "max_line_size",
                          "path" => "#{@tmp_dir}/with_long_lines.txt",
                          "format" => "none",
                          "read_from_head" => true,
                          "max_line_size" => label,
                          "log_level" => "debug"
                        })
    Fluent::FileWrapper.open("#{@tmp_dir}/with_long_lines.txt", "w+") do |f|
      f.puts "foo"
      f.puts "x" * size # 'x' * size + \n > @max_line_size
      f.puts "bar"
    end
    d = create_driver(config, false)
    timestamp = Time.parse("Mon Nov 29 11:22:33 UTC 2021")
    Timecop.freeze(timestamp)
    d.run(expect_records: 2)
    assert_equal([
                    [{"message" => "foo"},{"message" => "bar"}],
                    [
                      "2021-11-29 11:22:33 +0000 [warn]: received line length is longer than #{size}\n",
                      "2021-11-29 11:22:33 +0000 [debug]: skipped line\n"
                    ]
                  ],
                  [
                    d.events.collect { |event| event.last },
                    d.logs[-2..]
                  ])
  end
```

Run below command
```
bundle exec rake test TEST=test/plugin/test_in_tail.rb TESTOPTS="--name=test_max_line_size_with_large_singleline"
```

Add below code in test_in_tail.rb
```
  def test_max_line_size_with_large_multiline
    Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") { |f| }

    config = config_element("", "", {
                              "format" => "multiline",
                              "format_firstline" => "/^s/",
                              "format1" => "/^s (?<message1>[^\\n]+)\\n?/",
                              "format2" => "/^f (?<message2>[^\\n]+)\\n?/",
                              "max_line_size" => "1k",
                              "multiline_flush_interval" => "1s",
                          })
    d = create_driver(config)
    msg = "t" * 1000000000
    d.run(expect_emits: 1) do
      Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "ab") { |f|
        f.puts "s #{msg}"
        f.puts "f #{msg}"
        f.puts "s test1"
        f.puts "f test2"
      }
    end

    events = d.events

    assert_equal({"message1" => "test1", "message2" => "test2"}, events[0][2])
    assert_equal(1, events.length)
  end
```

Run below command
```
bundle exec rake test TEST=test/plugin/test_in_tail.rb TESTOPTS="--name=test_max_line_size_with_large_multiline"
```

### Expected behavior

test_max_line_size_with_large_singleline

**AS-IS** : it passed. but it took much long time. Finished in 2365.724927 seconds.

**TO-BE** : it works. Finished in 0.915693 seconds.

test_max_line_size_with_large_multiline

**AS-IS** : it failed. Finished in 305.73504 seconds.

**TO-BE** : it works. Finished in 24.163525 seconds

### Your Environment

```markdown
- Fluentd version: master branch
- TD Agent version: 
- Operating system: Monterey 12.6.1
- Kernel version: 21.6.0
```

### Your Configuration

```apache
None
```

### Your Error Log

```shell
test_max_line_size_with_large_multiline

Error: test_max_line_size_with_large_multiline(TailInputTest): NoMethodError: undefined method `[]' for nil
```

### Additional context

_No response_

yugeeklab:
Fixes #4494

Before this patch, long lines could cause breakdowns in fluentd, potentially posing a vulnerability. With this patch, max_line_size will be integrated into the FIFO, enabling the system to skip lines exceeding the maximum size before executing receive_lines.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -50,23 +50,15 @@ module Fluent
       def configure_json_parser(name)
         case name
         when :oj
-          raise LoadError unless Fluent::OjOptions.available?
-          [Oj.method(:load), Oj::ParseError]
+          return [Oj.method(:load), Oj::ParseError] if Fluent::OjOptions.available?
+
+          log&.info "Oj is not installed, and failing back to Yajl for json parser"
+          configure_json_parser(:yajl)
         when :json then [JSON.method(:load), JSON::ParserError]
         when :yajl then [Yajl.method(:load), Yajl::ParseError]
         else
           raise "BUG: unknown json parser specified: #{name}"
         end
-      rescue LoadError => ex
-        name = :yajl
-        if log
-          if /\boj\z/.match?(ex.message)
-            log.info "Oj is not installed, and failing back to Yajl for json parser"
-          else
-            log.warn ex.message
-          end
-        end
-        retry
       end
 
       def parse(text)
--- a/test/plugin/test_parser_json.rb
+++ b/test/plugin/test_parser_json.rb
@@ -8,6 +8,37 @@ class JsonParserTest < ::Test::Unit::TestCase
     @parser = Fluent::Test::Driver::Parser.new(Fluent::Plugin::JSONParser)
   end
 
+  sub_test_case "configure_json_parser" do
+    data("oj", [:oj, [Oj.method(:load), Oj::ParseError]])
+    data("json", [:json, [JSON.method(:load), JSON::ParserError]])
+    data("yajl", [:yajl, [Yajl.method(:load), Yajl::ParseError]])
+    def test_return_each_loader((input, expected_return))
+      result = @parser.instance.configure_json_parser(input)
+      assert_equal expected_return, result
+    end
+
+    def test_raise_exception_for_unknown_input
+      assert_raise RuntimeError do
+        @parser.instance.configure_json_parser(:unknown)
+      end
+    end
+
+    def test_fall_back_oj_to_yajl_if_oj_not_available
+      stub(Fluent::OjOptions).available? { false }
+
+      result = @parser.instance.configure_json_parser(:oj)
+
+      assert_equal [Yajl.method(:load), Yajl::ParseError], result
+      logs = @parser.logs.collect do |log|
+        log.gsub(/\A\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2} [-+]\d{4} /, "")
+      end
+      assert_equal(
+        ["[info]: Oj is not installed, and failing back to Yajl for json parser\n"],
+        logs
+      )
+    end
+  end
+
   data('oj' => 'oj', 'yajl' => 'yajl')
   def test_parse(data)
     @parser.configure('json_parser' => data)


daipom:
@yugeeklab Thanks for this fix!
CI is currently unstable because of #4487. We will fix it. Sorry for the trouble.

I see the intent of this fix as follows.

* In the current implementation, large lines that would eventually be discarded in `receive_lines` are temporarily held in `IOHandler`'s `@lines`.
* This is a waste of memory.
* This fix resolves the waste. 

Surely, such a fix would allow us to limit memory consumption by the `max_line_size` setting to some extent!

This PR would be effective to some extent, however I believe the problem of memory consumption will remain.
It would be possible that `FIFO`'s `@buffer` becomes unlimitedly large if the `@eol` does not appear in the data.

Are these my understandings correct?

yugeeklab:
Hi, @daipom 

I published an issue #4491 for more information.

> This PR would be effective to some extent, however I believe the problem of memory consumption will remain.
> It would be possible that FIFO's @buffer becomes unlimitedly large if the @eol does not appear in the data.

When max_line_size isn't set, FIFO's @buffer can grow indefinitely. Or if max_line_size has large value, FIFO's buffer will be limited, but there's still a possibility of fluentd experiencing slowdowns.

Summary:

as-is: max_line_size helps you avoid buffer overflow configuring via buffer section.
to-be: max_line_size helps prevent buffer overflow by configuring the buffer section and also ensures FIFO's buffer size remains limited.

If you have any suggestions, such as the fifo_buffer_size parameter or any other ideas, please feel free to discuss them with me.

Thank you for your review!

daipom:
@yugeeklab

> I've just published an issue #4491 for more information.

Thanks so much!

> as-is: max_line_size helps you avoid buffer overflow configuring via buffer section. to-be: max_line_size helps prevent buffer overflow by configuring the buffer section and also ensures FIFO's buffer size remains limited.

Now I understand!
The following understanding was not correct.

> This PR would be effective to some extent, however I believe the problem of memory consumption will remain.
It would be possible that `FIFO`'s `@buffer` becomes unlimitedly large if the `@eol` does not appear in the data.

This fix clears `FIFO`'s `@buffer` when `read_lines`.
So, this fix ensures `FIFO`'s buffer size remains limited.

> If you have any suggestions, such as the fifo_buffer_size parameter or any other ideas, please feel free to discuss them with me.

Thanks! 
Basically, it seems to be a very good idea to limit the `FIFO`'s buffer.

yugeeklab:
> Basically, it seems to be a very good idea to limit the FIFO's buffer.

Thank you for your comment!! @daipom 

Please let me know if there's any feedback on my code or idea. I'll review and accept your feedback as soon as possible.

Thank you.

daipom:
About CI failures, although #4493 has been resolved, we still need to resolve #4487.

daipom:
@yugeeklab The CI issue has been resolved. Sorry for the trouble.
Could you please rebase this branch on the latest master?

yugeeklab:
Hi, @daipom 

Rebase is done!!

Thank you for your review!!

daipom:
Sorry for waiting.
I will review this soon.

test/plugin/in_tail/test_fifo.rb
```diff
@@ -86,11 +86,31 @@ class IntailFIFO < Test::Unit::TestCase
...
+
+    test 'return lines when line size is less than or equal to max_line_size' do
+      fifo = Fluent::Plugin::TailInput::TailWatcher::FIFO.new(Encoding::ASCII_8BIT, Encoding::ASCII_8BIT, $log, 5)
+      t = "test"
+      text = ("#{t}\n" * 2).force_encoding(Encoding::ASCII_8BIT)
+      fifo << text
+      lines = []
+      fifo.read_lines(lines)
+      assert_equal ["test\n", "test\n"], lines
+    end
```
daipom, line 108: I want to commonize these tests and add some test cases as follows.

```suggestion

    data('bigger than max_line_size', [
      ["test test test\n" * 3],
      [],
    ])
    data('less than or equal to max_line_size', [
      ["test\n" * 2],
      ["test\n", "test\n"],
    ])
    data('mix', [
      ["test test test\ntest\ntest test test\ntest\ntest test test\n"],
      ["test\n", "test\n"],
    ])
    data('mix and multiple', [
      [
        "test test test\ntest\n",
        "test",
        " test test\nt",
        "est\nt"
      ],
      ["test\n", "test\n"],
    ])
    data('remaining data bigger than max_line_size should be discarded', [
      [
        "test\nlong line still not having EOL",
        "following texts to the previous long line\ntest\n",
      ],
      ["test\n", "test\n"],
    ])
    test 'return lines only that size is less than or equal to max_line_size' do |(input_texts, expected)|
      max_line_size = 5
      fifo = Fluent::Plugin::TailInput::TailWatcher::FIFO.new(Encoding::ASCII_8BIT, Encoding::ASCII_8BIT, $log, max_line_size)
      lines = []

      input_texts.each do |text|
        fifo << text.force_encoding(Encoding::ASCII_8BIT)
        fifo.read_lines(lines)
        # The size of remaining buffer (i.e. a line still not having EOL) must not exceed max_line_size.
        assert { fifo.bytesize <= max_line_size }
      end

      assert_equal expected, lines
    end
```

test/plugin/in_tail/test_io_handler.rb
```diff
@@ -146,5 +146,36 @@ def create_watcher
       assert_equal 5, returned_lines[0].size
       assert_equal 3, returned_lines[1].size
     end
+
+    test 'call receive_lines some times when long line(more than 8192) and max_line_size is 4096' do
```
daipom, line 150: I would like this test to be more clear about what is being tested.

This test does not seem to be intended to test if `receive_lines` is called more than once (some times).
(Actually, it is called only once in this test.)

Perhaps it would be better to make this test independent of the `when limit is 5` sub test.

daipom (changes requested):
@yugeeklab Thanks for this fix!
This fix basically looks good to me!
I've commented on some minor details (about the following), please check!

* Keeping the same debug log as before
* Improving codes
* Improving tests

lib/fluent/plugin/in_tail.rb
```diff
@@ -1059,7 +1055,27 @@ def read_lines(lines)
...
+              @was_long_line = false
+            end
+          end
+
+          is_long_line = !@max_line_size.nil? && (
+            @buffer.bytesize > @max_line_size || @was_long_line
+          )
+
+          if is_long_line
+            @buffer = ''.force_encoding(@from_encoding)
```
daipom, line 1077:
```suggestion
            @buffer.clear
```

daipom:
@yugeeklab 
Thanks for updating.
I'm checking the CI failures.

daipom:
Current CI failures have nothing to do with this PR.
Sorry for the trouble again.

lib/fluent/plugin/in_tail.rb
yugeeklab: Hi @daipom

I discovered that the change in this suggestion(https://github.com/fluent/fluentd/pull/4491#discussion_r1615863064) causes the following test to fail:

```
  data(
    small: ["128", 8192],
  )
  test 'max_line_size_8192' do |(label, size)|
    config = config_element("", "", {
                              "tag" => "max_line_size",
                              "path" => "#{@tmp_dir}/with_long_lines.txt",
                              "format" => "none",
                              "read_from_head" => true,
                              "max_line_size" => label,
                              "log_level" => "debug"
                            })
    Fluent::FileWrapper.open("#{@tmp_dir}/with_long_lines.txt", "w+") do |f|
      f.puts "foo"
      f.puts "x" * size # 'x' * size + \n > @max_line_size
      f.puts "bar"
    end
    d = create_driver(config, false)
    timestamp = Time.parse("Mon Nov 29 11:22:33 UTC 2021")
    Timecop.freeze(timestamp)
    d.run(expect_records: 2)
    assert_equal([
                    [{"message" => "foo"},{"message" => "bar"}],
                    [
                      "2021-11-29 11:22:33 +0000 [warn]: received line length is longer than #{label}\n",
                      "2021-11-29 11:22:33 +0000 [debug]: skipped line: #{'x' * size}\n"
                    ]
                  ],
                  [
                    d.events.collect { |event| event.last },
                    d.logs[-3..]
                  ])
  end
```

When the line_size is greater than max_line_size and line_size is greater than or equal to 8192, the test fails.

To log the long line even if it is skipped, you can log the line above this code like this:

```
          if is_long_line
            @log.warn "received line length is longer than #{@max_line_size}"
            @log.debug("skipped line: ") { convert(@buffer).chomp }
            @buffer.clear
            @was_long_line = true
          end
```

If you have any idea, please let me know.

Thank you for your review!!

daipom: I see! Thanks for checking this behavior!

`@log.debug("skipped line: ") { convert(rbuf).chomp }` sometimes can not output the entire line that is expected.
It is because some parts of the line are already cleared if `@was_long_line` is `true`, 

I think this is simply a debug log issue, right?
It is less important, so it would be good if we could fix it within the possible range.

daipom:
The CI issue has been resolved.
So, could you please rebase this to the latest master?
Sorry for the trouble again.

test/plugin/in_tail/test_io_handler.rb
yugeeklab: Hi @daipom 

What about the test code below?

```
  test 'does not call receive_lines when line_size exceeds max_line_size' do
    t = 'x' * (8192)
    text = "#{t}\n"

    max_line_size = 8192

    @file.write(text)
    @file.close

    update_pos = 0

    watcher = create_watcher
    stub(watcher).pe do
      pe = 'position_file'
      stub(pe).read_pos {0}
      stub(pe).update_pos { |val| update_pos = val }
      pe
    end

    returned_lines = []
    r = Fluent::Plugin::TailInput::TailWatcher::IOHandler.new(watcher, path: @file.path, read_lines_limit: 1000, read_bytes_limit_per_second: -1, max_line_size: max_line_size, log: $log, open_on_every_update: false, metrics: @metrics) do |lines, _watcher|
      returned_lines << lines.dup
      true
    end

    r.on_notify
    assert_equal text.bytesize, update_pos
    assert_equal 0, returned_lines.size
  end
```

I put this test outside of the 'when limit is 5' subtest and changed its name and behavior.

But I discovered that this test failed because it doesn't not call update_pos.
So, we need to fix the code. I will push one more commit for it.

yugeeklab:
Hi, @daipom 

Rebase is done.
I also added a commit to resolve the following [issue](https://github.com/fluent/fluentd/pull/4491#discussion_r1632200754).
Please review it if you don't mind.

Thank you.

test/plugin/in_tail/test_io_handler.rb
daipom: @yugeeklab Thanks for considering the test code!
Basically, it looks good, but we need to be careful about how we handle `pos`.

> But I discovered that this test failed because it doesn't not call update_pos.
> So, we need to fix the code. I will push one more commit for it.

I saw bf73efeea722376f30bccac0f10095fb6601939c and realized that there is a problem that needs to be solved about the management of `pos`
I'll comment on it in the new thread.

daipom (changes requested):
Thanks for the fix!
The following point is a remaining concern.

https://github.com/fluent/fluentd/pull/4491#discussion_r1635832302

> I saw bf73efeea722376f30bccac0f10095fb6601939c and realized that there is a problem that needs to be solved about the management of `pos`

We should not update `pos` like [this commit (bf73efe)](bf73efeea722376f30bccac0f10095fb6601939c).
We should only update `pos` at points where recovery is possible.
This means that we should not update `pos` until we can be sure that `@lines` has been successfully handled by `@receive_lines`.
If updating `pos` like this commit, some data may be lost when `BufferOverflowError` occurs or when Fluentd is forced to stop.

So, we need to consider how to manage `pos` correctly for this feature.
It needs to be able to continue processing correctly even if Fluentd is forced to stop.
Repeating process to skip long lines would be acceptable.
Data loss or sending corrupted data would be unacceptable.

For this feature, we need to take care of `@was_long_line` in particular.
We need to make sure that the restart of Fluentd does not cause a subsequent incomplete log to be sent.

daipom:
> We should not update `pos` like [this commit (bf73efe)](https://github.com/fluent/fluentd/commit/bf73efeea722376f30bccac0f10095fb6601939c).
> ...
> If updating `pos` like this commit, some data may be lost when `BufferOverflowError` occurs or when Fluentd is forced to stop.

Oh, sorry,  it was wrong.
The `@lines.empty?` condition will prevent it.

https://github.com/fluent/fluentd/blob/bf73efeea722376f30bccac0f10095fb6601939c/lib/fluent/plugin/in_tail.rb#L1230-L1232

So, we need to consider only the following points.

> For this feature, we need to take care of `@was_long_line` in particular.
> We need to make sure that the restart of Fluentd does not cause a subsequent incomplete log to be sent.

lib/fluent/plugin/in_tail.rb
```diff
@@ -1011,15 +1004,19 @@ def swap_state(pe)
...
-        def initialize(from_encoding, encoding)
+        def initialize(from_encoding, encoding, log, max_line_size=nil)
           @from_encoding = from_encoding
           @encoding = encoding
           @need_enc = from_encoding != encoding
           @buffer = ''.force_encoding(from_encoding)
           @eol = "\n".encode(from_encoding).freeze
+          @max_line_size = max_line_size
+          @was_long_line = false
+          @has_skipped_line = false
```
daipom, line 1015: Is there any reason why keeping `@has_skipped_line` as an instance variable?
If it is not necessary to keep it, we should make it the local variable in `read_lines()`  and make `read_lines()` return it as a returning value.

daipom:
> So, we need to consider only the following points.
> 
> > For this feature, we need to take care of `@was_long_line` in particular.
> > We need to make sure that the restart of Fluentd does not cause a subsequent incomplete log to be sent.

I think we should change `FIFO#bytesize`.

https://github.com/fluent/fluentd/blob/bf73efeea722376f30bccac0f10095fb6601939c/lib/fluent/plugin/in_tail.rb#L1087-L1089

It is used in the following `pos` logic:

https://github.com/fluent/fluentd/blob/bf73efeea722376f30bccac0f10095fb6601939c/lib/fluent/plugin/in_tail.rb#L1230-L1241

https://github.com/fluent/fluentd/blob/bf73efeea722376f30bccac0f10095fb6601939c/lib/fluent/plugin/in_tail.rb#L1246-L1248

The `bytesize` should be the uncommitted byte size that `FIFO` is still handling.
It does not equal the size of the buffer of `FIFO` anymore because `FIFO` can clear the buffer to skip the long line.

In the following case (`max_line_size 12`), `very long line not finished yet` will be cleared from the buffer soon.

```
short line\n # To be committed to the pos
very long line not finished yet # Not to be committed to the pos until the `@eol` occurs.
```

However, that data size should be considered for `pos` handling.
Since the line is not finished yet, the `pos` update should be done up to the end of `short line\n`.
(When Fluentd restarts, Fluentd should continue the process from the end of `short line\n`.)
Also, the reopening pos should be from the end of `very long line not finished yet` (especially for the case `open_on_every_update`).

For this, `FIFO#bytesize` should be the uncommitted byte size that `FIFO` is still handling, not the real buffer size of `FIFO`.

daipom:
@yugeeklab I have fixed the remaining points and pushed them to my tmp branch (the following 3 commits).
Could you please check them?
If there is no problem, I will push these commits to this PR.
If you have any concerns or ideas, please let me know.

* https://github.com/daipom/fluentd/tree/in_tail-improve-max_line_size
  * [make has_skipped_line local variable](https://github.com/fluent/fluentd/commit/05e7caf493ab07f91b6f58468883518de8a30cd8)
  * [refactor some variable names](https://github.com/fluent/fluentd/commit/6c0371729811635b7aab5d9381be7b4a367d0d13)
  * [fix to commit the correct pos to continue processing correctly](https://github.com/fluent/fluentd/commit/fdf74e5d715b2a769026bb22b13f5e690ea67dcd)

The main point is to resolve the issue that is tested on the `discards a subsequent data in a long line even if restarting occurs between` test in [fix to commit the correct pos to continue processing correctly](https://github.com/fluent/fluentd/commit/fdf74e5d715b2a769026bb22b13f5e690ea67dcd).
This test would fail in the current branch.

lib/fluent/plugin/in_tail.rb
daipom: Please see https://github.com/fluent/fluentd/pull/4491#issuecomment-2162508391.

[make has_skipped_line local variable](https://github.com/fluent/fluentd/commit/05e7caf493ab07f91b6f58468883518de8a30cd8) is for this.

yugeeklab:
Hi @daipom 

So, Here is the summary of your code

AS-IS:
 When a restart occurs while reading a long line, because it doesn't record the position properly, a long line can be recognized as a short line.

T0-BE:
 When a restart occurs while reading a long line and the position is recorded properly, the long line is recognized as a long line.

It looks good to me!!

Thank you!!!

daipom:
@yugeeklab Yes! Thanks for checking it!
I will push them.

daipom:
@yugeeklab Sorry, I failed to push. Something wrong happens...
I'm fixing it. Please wait...

daipom:
aa

daipom:
@yugeeklab Sorry for the trouble.
Could you please run the following command to recover your `origin/master` branch?

```bash
git push -f
```

yugeeklab:
Hi @daipom 

I recover my origin/master!!

Should I also reopen Pull Request?

Thank you.

daipom:
> Should I also reopen Pull Request?

Of course! Thanks for reopening it!
Sorry for the trouble.

PR #4502, opened 2024-05-23
out_file:add warn message for symlink_path-setting

### LINKED ISSUE #4490: out_file:Inconsistent path settings and symlink_path settings.
### Describe the bug

If tag is set in the chunk key, We will get an error unless we specify the tag in the path setting of the out_file plugin.
However, there is no such check in the symlink_path setting. (i.e., no error occurs even if no tag is specified.)
Also, this check is a setting to make the pass unique, but if we use a part of tag such as tag[0] instead of tag, no error will occur.

Is this an intended spec? Or is there a check omitted?

### To Reproduce

For example, use the following input.
```apache
<source>
  @type sample
  sample {"message":"hoge"}
  tag test.hoge
</source>

<source>
  @type sample
  sample {"message":"fuga"}
  tag test.fuga
</source>
```
At this time, setting the out_file plugin as follows will result in an error.
```apache
<match test.**>
  @type file
  path /tmp/test
 symlink_path /tmp/test/link/${tag}
  append true

  <buffer time,tag>
    @type file
    path /tmp/test/buffer
    timekey 60 
    timekey_wait 3
  </buffer>
</match>
```
```
 $ bundle exec fluentd -c /Users/fluentd/configuration.txt
2024-04-26 10:20:51 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2024-04-26 10:20:51 +0900 [info]: parsing config file is succeeded path="/Users/fluentd/configuration.txt"
2024-04-26 10:20:51 +0900 [info]: gem 'fluentd' version '1.16.5'
2024-04-26 10:20:51 +0900 [error]: config error file="/Users/fluentd/configuration.txt" error_class=Fluent::ConfigError error="Parameter 'path: /tmp/test/test.%Y%m%d%H%M.log' doesn't have tag placeholder"
```
However, if the out_file plugin is set up as follows, there is no error and two outputs compete for a single path.
```apache
<match test.**>
  @type file
  path /tmp/test/${tag[0]}
 symlink_path /tmp/test/link/${tag}
  append true

  <buffer time,tag>
    @type file
    path /tmp/test/buffer/tag
    timekey 60 
    timekey_wait 3
  </buffer>
</match>
```
```
/tmp/test $ ls -R
buffer			test.202405091047.log	test.202405091049.log
link			test.202405091048.log

./buffer:
tag

./buffer/tag:
buffer.b617fba30cfcdcb64c554db644d9ef7c5.log
buffer.b617fba30cfcdcb64c554db644d9ef7c5.log.meta
buffer.b617fba30dd0f05a6f9a793e93f657172.log
buffer.b617fba30dd0f05a6f9a793e93f657172.log.meta

./link:
test		test.fuga	test.hoge
/tmp/test $ tail test.202405091048.log
2024-05-09T10:48:00+09:00	test.fuga	{"message":"fuga"}
2024-05-09T10:48:01+09:00	test.fuga	{"message":"fuga"}
2024-05-09T10:48:02+09:00	test.fuga	{"message":"fuga"}
...
2024-05-09T10:48:59+09:00	test.fuga	{"message":"fuga"}
2024-05-09T10:48:00+09:00	test.hoge	{"message":"hoge"}
2024-05-09T10:48:01+09:00	test.hoge	{"message":"hoge"}
...
2024-05-09T10:48:59+09:00	test.hoge	{"message":"hoge"}
```
Furthermore, if the out_file plugin is set up as follows, there will be no error and the two will compete for a single symlink_path.
```apache
<match test.**>
  @type file
  path /tmp/test/${tag[0]}
 symlink_path /tmp/test/link
  append true

  <buffer time,tag>
    @type file
    path /tmp/test/buffer
    timekey 60
    timekey_wait 3
  </buffer>
</match>
```
```
$ ls -R
buffer  link

./buffer:
buffer.b616f84e16b5e9575a8a61ca0cbf34536.log        buffer.b616f84e16b968e52f8d0c2bc37bdb9d9.log
buffer.b616f84e16b5e9575a8a61ca0cbf34536.log.meta   buffer.b616f84e16b968e52f8d0c2bc37bdb9d9.log.meta

./link:
test
/tmp/test $ tail link/test
2024-04-26T13:26:39+09:00   test.fuga   {"message":"fuga"}
2024-04-26T13:26:40+09:00   test.fuga   {"message":"fuga"}
2024-04-26T13:26:41+09:00   test.fuga   {"message":"fuga"}
...
/tmp/test $ tail link/test
2024-04-26T13:27:00+09:00   test.hoge   {"message":"hoge"}
2024-04-26T13:27:01+09:00   test.hoge   {"message":"hoge"}
2024-04-26T13:27:02+09:00   test.hoge   {"message":"hoge"}
...
```

### Expected behavior

In either case, an error should occur. (Or this is the intended specification.)

### Your Environment

```markdown
- Fluentd version:1.17.0
- Operating system:
ProductName:		macOS
ProductVersion:		13.0
BuildVersion:		22A380
- Kernel version:22.1.0
```

### Your Configuration

See `To Reproduce` section.

### Your Error Log

See `To Reproduce` section.

### Additional context

_No response_

Shingo-Nakayama:
* Partially fixes #4490

According to the issue https://github.com/fluent/fluentd/issues/4490 , if a `tag` or `keys` is set in the chunk key and those placeholders do not exist in the `symlink_path`, multiple chunks to compete for the same symlink and some logs can't be taken from it.
Since this is unintended behavior, I think it is necessary to inform the user that this is a bad setting by the warn message.

If a `tag` or `keys` is set in the chunk key and those placeholders do not exist in the `symlink_path`, the following warning log is displayed.

Ex:
```
<source>
  @type sample
  sample {"message":"hoge","key1":"A","key2":"B"}
  tag test.hoge
</source>

<source>
  @type sample
  sample {"message":"fuga","key1":"A","key2":"B"}
  tag test.fuga
</source>

<match test.**>
  @type file
  path /tmp/test/${tag}/${key1}/${key2}
 symlink_path /tmp/test/link
  append true

  <buffer time,tag,key1,key2>
    @type file
    path /tmp/test/buffer/tag
    timekey 60 
    timekey_wait 3
  </buffer>
</match>
```
```
[warn]: #0 Parameter 'symlink_path: /tmp/test/link' doesn't have tag placeholder. This means multiple chunks are competing for a single symlink_path, so some logs may not be taken from the symlink.
[warn]: #0 Parameter 'symlink_path: /tmp/test/link' doesn't have enough placeholders for keys key1,key2. This means multiple chunks are competing for a single symlink_path, so some logs may not be taken from the symlink.
```

**Docs Changes**:
None (may be necessary)

**Release Note**: 
Same with the title

### DIFF
--- a/test/plugin/test_out_file.rb
+++ b/test/plugin/test_out_file.rb
@@ -130,7 +130,7 @@ class FileOutputTest < Test::Unit::TestCase
           'path' => "#{TMP_DIR}/${tag}/${type}/conf_test.%Y%m%d.%H%M.log",
           'add_path_suffix' => 'false',
           'append' => "true",
-          'symlink_path' => "#{TMP_DIR}/conf_test.current.log",
+          'symlink_path' => "#{TMP_DIR}/${tag}/conf_test.current.log",
           'compress' => 'gzip',
           'recompress' => 'true',
         }, [
@@ -181,6 +181,25 @@ class FileOutputTest < Test::Unit::TestCase
       ])
       assert_nothing_raised do
         Fluent::Test::Driver::Output.new(Fluent::Plugin::NullOutput).configure(conf)
+      end
+    end
+
+    test 'symlink path has not tag placeholder or key placeholder' do
+      conf = config_element('match', '**', {
+          'path' => "#{TMP_DIR}/${tag}/${key1}/${key2}/conf_test.%Y%m%d.%H%M.log",
+          'symlink_path' => "#{TMP_DIR}/conf_test.current.log",
+        }, [
+          config_element('buffer', 'time,tag,key1,key2', {
+              '@type' => 'file',
+              'timekey' => '1d',
+              'path' => "#{TMP_DIR}/buf_conf_test",
+          }),
+      ])
+      assert_nothing_raised do
+        d = create_driver(conf)
+        assert do
+          d.logs.count { |log| log.include?("symlink_path:") } == 2
+        end
       end
     end
   end
--- a/test/plugin/test_out_file.rb
+++ b/test/plugin/test_out_file.rb
@@ -184,7 +184,8 @@ class FileOutputTest < Test::Unit::TestCase
       end
     end
 
-    test 'symlink path has not tag placeholder or key placeholder' do
+    test 'warning for symlink_path not including correct placeholders corresponding to chunk keys' do
+      omit "Windows doesn't support symlink" if Fluent.windows?
       conf = config_element('match', '**', {
           'path' => "#{TMP_DIR}/${tag}/${key1}/${key2}/conf_test.%Y%m%d.%H%M.log",
           'symlink_path' => "#{TMP_DIR}/conf_test.current.log",
@@ -198,7 +199,7 @@ class FileOutputTest < Test::Unit::TestCase
       assert_nothing_raised do
         d = create_driver(conf)
         assert do
-          d.logs.count { |log| log.include?("symlink_path:") } == 2
+          d.logs.count { |log| log.include?("multiple chunks are competing for a single symlink_path") } == 2
         end
       end
     end
--- a/lib/fluent/plugin/out_file.rb
+++ b/lib/fluent/plugin/out_file.rb
@@ -172,15 +172,14 @@ module Fluent::Plugin
           log.warn "symlink_path is unavailable on Windows platform. disabled."
           @symlink_path = nil
         else
-          #add check if symlink_path has a tag placeholder.
-            placeholder_validators(:symlink_path, @symlink_path).reject{|v| v.type == :time }.each do |v|
+          placeholder_validators(:symlink_path, @symlink_path).reject{ |v| v.type == :time }.each do |v|
             begin
               v.validate!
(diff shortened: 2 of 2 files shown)


test/plugin/test_out_file.rb
```diff
@@ -183,6 +183,25 @@ def create_driver(conf = CONFIG, opts = {})
...
+          config_element('buffer', 'time,tag,key1,key2', {
+              '@type' => 'file',
+              'timekey' => '1d',
+              'path' => "#{TMP_DIR}/buf_conf_test",
+          }),
+      ])
+      assert_nothing_raised do
+        d = create_driver(conf)
+        assert do
+          d.logs.count { |log| log.include?("symlink_path:") } == 2
```
daipom, line 201:
```suggestion
          d.logs.count { |log| log.include?("multiple chunks are competing for a single symlink_path") } == 2
```

It would be safer to look for a string that essentially represents the expected log.

daipom (changes requested):
Thanks for this fix!
Could you please fix the following points?

daipom (approved):
LGTM. Thanks!

PR #4513, opened 2024-05-29
github: stop using rendering for freestyle text area

daipom:
None.

Stop using rendering for freestyle text area.
Not happy about free style text area being enclosed in a code block automatically.

I think this is not intended, right?
If there is a reason, please let me know.

Q&A:
Screenshot from 2024-05-29 11-44-38

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.

### DIFF
--- a/.github/DISCUSSION_TEMPLATE/q-a-japanese.yml
+++ b/.github/DISCUSSION_TEMPLATE/q-a-japanese.yml
@@ -12,7 +12,6 @@ body:
       description: |
         何について困っているのかを書いてください。試したことや実際の結果を示してください。
         期待する挙動と実際の結果の違いがあればそれも書くのをおすすめします。
-      render: markdown
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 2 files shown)


daipom:
@kenhys Could you check this when you have time (this is not urgent)?

ashie (approved):

kenhys (approved):
As there is no preview feature when posting discussion topic, so it may be difficult  to
expect that discussion form will be rendered with markdown or not.

It was intentional to apply `render: markdown` itself, but if it is harmful in practical use case, it should be removed.

daipom:
Thanks for your review!

daipom:
> As there is no preview feature when posting discussion topic, so it may be difficult to expect that discussion form will be rendered with markdown or not.
> 
> It was intentional to apply `render: markdown` itself, but if it is harmful in practical use case, it should be removed.

I see! Thanks!

PR #4515, opened 2024-05-29
test: fix IOError about Tempfile closed when GC

daipom:
None.

`Tempfile#binmode` returns `File` object, not own `Tempfile` object.
So, GC will cause its finalizer and the file can be closed during the test.
This is the cause why these tests sometimes fail by `IOError: closed stream`.

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.


daipom:
@kenhys Could you please review this?

ashie (approved):
LGTM

PR #4522, opened 2024-06-06
parser_json: fix wrong LoadError warning

daipom:
None

If Oj is not installed, LoadError with the empty message is raised.

* https://github.com/fluent/fluentd/blob/b25e989daf1501e606a6980e64670cbd3a755209/lib/fluent/plugin/parser_json.rb#L53

So, the current condition `/\boj\z/.match?(ex.message)` does not work and the following meaningless warning is displayed.

    {datetime} [warn]: #x {id} LoadError

After this fix, the log message will be:

    {datetime} [info]: #x {id} Oj is not installed, and failing back to Yajl for json parser

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -50,23 +50,15 @@ module Fluent
       def configure_json_parser(name)
         case name
         when :oj
-          raise LoadError unless Fluent::OjOptions.available?
-          [Oj.method(:load), Oj::ParseError]
+          return [Oj.method(:load), Oj::ParseError] if Fluent::OjOptions.available?
+
+          log&.info "Oj is not installed, and failing back to Yajl for json parser"
+          configure_json_parser(:yajl)
         when :json then [JSON.method(:load), JSON::ParserError]
         when :yajl then [Yajl.method(:load), Yajl::ParseError]
         else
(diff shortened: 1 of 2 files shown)


ashie (changes requested):
It seems that this `rescue` no longer needed.
It would be better to do `retry` at `when :oj` instead of `rescue` block.

daipom:
> It seems that this `rescue` no longer needed. It would be better to do `retry` at `when :oj` instead of `rescue` block.

Thanks for your review!
I have fixed this as https://github.com/fluent/fluentd/compare/151e6314a371add407f602d50cb8be777109afd1..d515b3a6a277d854378a5961223b4666fb438c31

daipom:
Fixed typo: https://github.com/fluent/fluentd/compare/29ff3815c38bb3f36be0b64a9c27ecae3cdbf11a..891ce71120198fa168232a796d4267a3cdb2c91f

daipom:
It looks like `failing back` should be `falling back`.
For now, this PR does not handle it.
(Changing log messages should not be done casually. For this level, it might be better to leave the existing.)

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

PR #4528, opened 2024-06-13
out_http: add gzip option

### LINKED ISSUE #4410: Compression Support in out_http plugin

Similar to #2047 , I would like to be able to generate compressed gzip HTTP output that would able to be transparently ingested by the `in_http` plugin so I can accomodate larger message sizes in a http based fluentd to fluentd setup.

the http request should be sent with `Content-Encoding: gzip` headers set and the serialized message content should be gzip encoded

Unfortunately our enviroment makes using the `forward` plugin, which supports this type of setup much easier, hard for us to implement and HTTP suits us much better.

### Additional context

_No response_

rockliffelewis:
Fixes #4410

Adds a `compress gzip` option to `out_http` to enable compressing the http request before sending it to save on bandwidth

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/out_http.rb
+++ b/lib/fluent/plugin/out_http.rb
@@ -55,6 +55,8 @@ module Fluent::Plugin
     config_param :headers, :hash, default: nil
     desc 'Additional placeholder based headers for HTTP request'
     config_param :headers_from_placeholders, :hash, default: nil
+    desc 'Compress HTTP request body'
+    config_param :compress, :enum, list: [:text, :gzip], default: :text
 
     desc 'The connection open timeout in seconds'
     config_param :open_timeout, :integer, default: nil
@@ -290,6 +292,9 @@ module Fluent::Plugin
           req[k] = extract_placeholders(v, chunk)
         end
       end
+      if @compress == :gzip
+        req['Content-Encoding'] = "gzip"
+      end
       req['Content-Type'] = @content_type
     end
 
@@ -323,7 +328,14 @@ module Fluent::Plugin
               Net::HTTP::Put.new(uri.request_uri)
             end
       set_headers(req, uri, chunk)
+
       req.body = @json_array ? "[#{chunk.read.chop}]" : chunk.read
+
+      if @compress == :gzip
+        gz = Zlib::GzipWriter.new(StringIO.new)
+        gz << req.body
+        req.body = gz.close.string
+      end
 
       # At least one authentication method requires the body and other headers, so the order of this call matters
       set_auth(req, uri)
--- a/test/plugin/test_out_http.rb
+++ b/test/plugin/test_out_http.rb
@@ -75,21 +75,30 @@ class HTTPOutputTest < Test::Unit::TestCase
         @@result.headers[key] = value
       end
 
+      body = ""
       data = []
+
+      case req['content-encoding']
+      when 'gzip'
+        body = Zlib::GzipReader.new(StringIO.new(req.body)).read
+      else 
+        body = req.body
+      end
+
       case req.content_type
       when 'application/x-ndjson'
-        req.body.each_line { |l|
+        body.each_line { |l|
           data << JSON.parse(l)
         }
       when 'application/json'
-        data = JSON.parse(req.body)
+        data = JSON.parse(body)
       when 'text/plain'
         # Use single_value in this test
-        req.body.each_line { |line|
+        body.each_line { |line|
           data << line.chomp
         }
       else
-        data << req.body
+        data << body
       end
       @@result.data = data
 
@@ -515,6 +524,41 @@ class HTTPOutputTest < Test::Unit::TestCase
       assert_equal 'POST', result.method
       assert_equal 'application/x-ndjson', result.content_type
       assert_equal test_events, result.data
+      assert_not_empty result.headers
+    end
+  end
+
+  sub_test_case 'GZIP' do
+    def server_port
+      19882
+    end
+
+    data(:json_array, [false, true])
+    data(:buffer_compress, ["text", "gzip"])
+    def test_write_with_gzip
+      d = create_driver(%[
+        endpoint http://127.0.0.1:#{server_port}/test
+        compress gzip
+        json_array #{data[:json_array]}
+        <buffer>
+          @type memory
+          compress #{data[:buffer_compress]}
+        </buffer>
+      ])
+      d.run(default_tag: 'test.http') do
+        test_events.each { |event|
+          d.feed(event)
+        }
+      end
+
+      result = @@result
+      assert_equal 'POST', result.method
+      assert_equal(
+        data[:json_array] ? 'application/json' : 'application/x-ndjson',
+        result.content_type
+      )
+      assert_equal 'gzip', result.headers['content-encoding']
+      assert_equal test_events,  result.data
       assert_not_empty result.headers
     end
   end


lib/fluent/plugin/out_http.rb
```diff
@@ -325,6 +330,12 @@ def create_request(chunk, uri)
       set_headers(req, uri, chunk)
       req.body = @json_array ? "[#{chunk.read.chop}]" : chunk.read
 
+      if @compress == :gzip
+        gz = Zlib::GzipWriter.new(StringIO.new)
+        gz << req.body
+        req.body = gz.close.string
+      end
+
```
daipom, line 338: We need to consider the possibility that the chunk is already compressed ([compress option of buffer](https://docs.fluentd.org/configuration/buffer-section#buffering-parameters)).

Example: `out_file` changes the behavior depending on `@buffer.compress`.

https://github.com/fluent/fluentd/blob/c0cd1e6960290954b442ea24077a80cee9229d41/lib/fluent/plugin/out_file.rb#L214-L222

Example: `out_forward` forces `@buffer.compress` to `gzip` when own `compress` is `gzip`.

https://github.com/fluent/fluentd/blob/c0cd1e6960290954b442ea24077a80cee9229d41/lib/fluent/plugin/out_forward.rb#L254-L258

https://github.com/fluent/fluentd/blob/c0cd1e6960290954b442ea24077a80cee9229d41/lib/fluent/plugin/out_forward.rb#L664-L668

Like `out_forward`, we could force `@buffer.compress` to `gzip` when `out_http`'s `compress` is `gzip`.
(Then, it would be enough to read chunk as is like `chunk.read(compressed: @compress)`.)
On the other hand, we should note that the compression efficiency decreases when data is compressed frequently. (`out_file` has [recompress option](https://docs.fluentd.org/output/file#recompress) for this.)

Anyway, we need to take into account the compression of buffer to some extent.

test/plugin/test_out_http.rb
```diff
@@ -511,9 +520,35 @@ def test_write_with_https
         }
       end
 
+      result = @@result
+      p result
```
daipom, line 524:
```suggestion
```

lib/fluent/plugin/out_http.rb
```diff
@@ -55,6 +55,8 @@ class RetryableResponse < StandardError; end
     config_param :headers, :hash, default: nil
     desc 'Additional placeholder based headers for HTTP request'
     config_param :headers_from_placeholders, :hash, default: nil
+    desc 'Compress HTTP request body'
+    config_param :compress, :enum, list: [:none, :gzip], default: :none
```
daipom, line 59: It would be better to align with `out_forward` and `out_file`

```suggestion
    config_param :compress, :enum, list: [:text, :gzip], default: :text
```

daipom (changes requested):
Thanks for the PR!
I think we need to take into account the compression of buffer.
I commentted below on the details, so please check.

lib/fluent/plugin/out_http.rb
daipom: The current PR's specification is to always recompress data regardless of the Buffer's compression settings, right?
(`chunk.read` without arguments will decompress data if the chunk is compressed).

If this is the intended specification, the current implementation would be fine.

rockliffelewis: At current yes it will decompress the buffer, format the response, and recompress before sending.

I think trying to send already compressed chunks straight from the buffer might interfere with the json_array setting, or at least theres a case to be handled there.

I'll take a look and see what improvements can be made here

lib/fluent/plugin/out_http.rb
```diff
@@ -323,7 +328,26 @@ def create_request(chunk, uri)
...
+          req.body = chunk.read(compressed: @compress)
+          compress_body = false
+        end
+      end
+
+      if compress_body
+        gz = Zlib::GzipWriter.new(StringIO.new)
+        gz << req.body
+        req.body = gz.close.string
+      end
```
rockliffelewis, line 350: Ok, so the json_array setting throws a bit of a spanner in the works, but the gist here is that if both out_http and buffer are set to gzip and json_array isnt set, then send the gzipped chunk as is.

Otherwise decompress (if required) and recompress (if required).

lib/fluent/plugin/out_http.rb
daipom: Thanks!
Note: https://github.com/fluent/fluentd/pull/4528#discussion_r1639078250

lib/fluent/plugin/out_http.rb
daipom: Thanks!
I understand the specification!

rockliffelewis: Sorry @daipom, I just updated this section to tidy it up a little more and make it easier to read

daipom: I see! No problem!

daipom:
I'd like to add test patterns like this: https://github.com/fluent/fluentd/commit/c68450926bd51b2d6b2daa545e0b3ba4280835e7

However, the case of sending the gzipped chunk as is fails in my environment.
(Only one record decompressed on the test server.)
I'm checking the cause.

```rb
Failure: test_write_with_gzip[buffer_compress: "gzip", json_array: false](HTTPOutputTest::GZIP)
/home/daipom/work/fluentd/fluentd/test/plugin/test_out_http.rb:564:in `test_write_with_gzip'
     561:         result.content_type
     562:       )
     563:       assert_equal 'gzip', result.headers['content-encoding']
  => 564:       assert_equal test_events, result.data
     565:       assert_not_empty result.headers
     566:     end
     567:   end
<[{"bool"=>true, "message"=>"hello", "num"=>10},
 {"bool"=>false, "message"=>"hello", "num"=>11}]> expected but was
<[{"bool"=>true, "message"=>"hello", "num"=>10}]>

diff:
? [{"bool"=>true, "message"=>"hello", "num"=>10},
?                                               ]
?                                               ?
-  {"bool"=>false, "message"=>"hello", "num"=>11}]
```

rockliffelewis:
> I'd like to add test patterns like this: [c684509](https://github.com/fluent/fluentd/commit/c68450926bd51b2d6b2daa545e0b3ba4280835e7)
> 
> However, the case of sending the gzipped chunk as is fails in my environment. (Only one record decompressed on the test server.) I'm checking the cause.
> 
> ```ruby
> Failure: test_write_with_gzip[buffer_compress: "gzip", json_array: false](HTTPOutputTest::GZIP)
> /home/daipom/work/fluentd/fluentd/test/plugin/test_out_http.rb:564:in `test_write_with_gzip'
>      561:         result.content_type
>      562:       )
>      563:       assert_equal 'gzip', result.headers['content-encoding']
>   => 564:       assert_equal test_events, result.data
>      565:       assert_not_empty result.headers
>      566:     end
>      567:   end
> <[{"bool"=>true, "message"=>"hello", "num"=>10},
>  {"bool"=>false, "message"=>"hello", "num"=>11}]> expected but was
> <[{"bool"=>true, "message"=>"hello", "num"=>10}]>
> 
> diff:
> ? [{"bool"=>true, "message"=>"hello", "num"=>10},
> ?                                               ]
> ?                                               ?
> -  {"bool"=>false, "message"=>"hello", "num"=>11}]
> ```

I also noticed this in my testing, im assuming that the behavior is correct but not what the test is expecting.

Its delivering a single gzipped message rather than batching a few messages in the request.

I can add a few more tests to cover off on the different option combinations

rockliffelewis:
The issue here is that the test harness doesn't really handle getting multiple http requests for one test, and in the gzipped buffer case the messages will come one by one. Im not entirely sure how to fix this properly without doing something like this:

```
  assert_equal(
    data[:buffer_compress] == 'gzip' && !data[:json_array] ? [test_events[0]] : test_events,
    result.data
  )
```

daipom:
@rockliffelewis Thanks for considering it!
I found the cause!

The cause is that `zlib` does not support `gzip` concatenation.
The chunk of Fluentd handles it as follows.

https://github.com/fluent/fluentd/blob/c0cd1e6960290954b442ea24077a80cee9229d41/lib/fluent/plugin/compressable.rb#L57-L75

So, as you say, it's generally not a problem.
However, if we consider a use case where forwarding the gzip data to `in_http`, this could be a problem.
Currently, `in_http` does not support `gzip` concatenation.

https://github.com/fluent/fluentd/blob/c0cd1e6960290954b442ea24077a80cee9229d41/lib/fluent/plugin/in_http.rb#L509-L523

So, if we consider a use case where forwarding the gzip data to `in_http`, it would be safer to start with a simple specification that always recompresses the chunk as the first implementation.

What do you think about this?

rockliffelewis:
Ah i see! I didnt realise the buffer was doing something fancy like that. I wouldn't expect most HTTP receivers to implement that either so at best we would need a `recompress` flag.

In my usecase I want to use `in_http` as the receiver, and I want the maximum compression so I would end up buffering it uncompressed anyway.

daipom (approved):
Thanks! LGTM.
I will wait a bit to merge for the other maintainers.

ashie (approved):
LGTM
Thanks for your contribution!

PR #4578, opened 2024-08-05
Add throttling metrics

Athishpranav2003:
Fixes #
https://github.com/fluent/fluent-plugin-prometheus/issues/190

It adds metrics to the throttling of logs while taling logs using in_tail plugin as asked in the above issue.
Once this PR is accepted corresponding changes in prometheus plugin will be made to expose the newly added throttling metric
**Docs Changes**:
Need docs change in prometheus plugin for metrics description 
**Release Note**: 
NA

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -1197,6 +1197,7 @@ module Fluent::Plugin
           if limit_bytes_per_second_reached? || group_watcher&.limit_lines_reached?(@path)
             @metrics.throttled.inc
             return
+          end
 
           with_io do |io|
             begin
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -36,7 +36,7 @@ module Fluent::Plugin
     helpers :timer, :event_loop, :parser, :compat_parameters
 
     RESERVED_CHARS = ['/', '*', '%'].freeze
-    MetricsInfo = Struct.new(:opened, :closed, :rotated)
+    MetricsInfo = Struct.new(:opened, :closed, :rotated, :throttled)
 
     class WatcherSetupError < StandardError
       def initialize(msg)
@@ -208,7 +208,8 @@ module Fluent::Plugin
       opened_file_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_opened_total", help_text: "Total number of opened files")
       closed_file_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_closed_total", help_text: "Total number of closed files")
       rotated_file_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_rotated_total", help_text: "Total number of rotated files")
-      @metrics = MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics)
+      throttling_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_throttled_total", help_text: "Total number of times throttling occurs per file when throttling enabled")
+      @metrics = MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics, throttling_metrics)
     end
 
     def configure_tag
@@ -793,6 +794,7 @@ module Fluent::Plugin
           'opened_file_count' => @metrics.opened.get,
           'closed_file_count' => @metrics.closed.get,
           'rotated_file_count' => @metrics.rotated.get,
+          'throttled_log_count' => @metrics.throttled.get,
         })
       }
       stats
@@ -1192,8 +1194,9 @@ module Fluent::Plugin
         end
 
         def handle_notify
-          return if limit_bytes_per_second_reached?
-          return if group_watcher&.limit_lines_reached?(@path)
+          if limit_bytes_per_second_reached? || group_watcher&.limit_lines_reached?(@path)
+            @metrics.throttled.inc
+            return
 
           with_io do |io|
             begin
@@ -1220,6 +1223,7 @@ module Fluent::Plugin
 
                     if group_watcher_limit || limit_bytes_per_second_reached? || should_shutdown_now?
                       # Just get out from tailing loop.
+                      @metrics.throttled.inc
                       read_more = false
                       break
                     end
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -1224,7 +1224,7 @@ module Fluent::Plugin
 
                     if group_watcher_limit || limit_bytes_per_second_reached? || should_shutdown_now?
                       # Just get out from tailing loop.
-                      @metrics.throttled.inc
+                      @metrics.throttled.inc if group_watcher_limit || limit_bytes_per_second_reached?
                       read_more = false
                       break
                     end
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -1878,13 +1878,6 @@ class TailInputTest < Test::Unit::TestCase
     plugin.instance_eval do
       @pf = Fluent::Plugin::TailInput::PositionFile.load(sio, EX_FOLLOW_INODES, {}, logger: $log)
       @loop = Coolio::Loop.new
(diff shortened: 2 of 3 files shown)


Athishpranav2003:
@ashie Please take a look at this. I am not sure which sort of metric is important here. Once this is merged and released i will add the changes to prometheus plugin too, I will create the PR for it too in some time

Athishpranav2003:
@ashie ignore the fails alone
there is a small mistake and i cant run the checks locally because its too heavy
i will push the change that fixes that now

Athishpranav2003:
@ashie can you rerun the checks?

ashie:
Here is a reply for your comment you seem deleted already:

>  i cant run the checks locally because its too heavy

Yes, fluentd test takes too long time to run whole tests.
e.g.) It takes 1035 seconds in https://github.com/fluent/fluentd/actions/runs/9925766505/job/27418533059

So even we fluentd maintainers also don't often run all tests locally.
Instead, run only partial tests related with the change.

e.g.) Run only in_tail's multiline related tests, and show test names currently running (`-v` option).
```console
$ bundle exec rake test TEST=test/plugin/test_in_tail.rb TESTOPTS="-v --testcase=\"TailInputTest::multiline\""
```

You can see the explanation of test options by the following command:
```console
$ bundle exe rake test TESTOPTS="--help"
    -r, --runner=RUNNER              Use the given RUNNER.
                                     (c[onsole], e[macs], x[ml])
        --collector=COLLECTOR        Use the given COLLECTOR.
                                     (de[scendant], di[r], l[oad], o[bject]_space)
    -n, --name=NAME                  Runs tests matching NAME.
                                     Use '/PATTERN/' for NAME to use regular expression.
                                     Regular expression accepts options.
                                     Example: '/taRget/i' matches 'target' and 'TARGET'
        --ignore-name=NAME           Ignores tests matching NAME.
                                     Use '/PATTERN/' for NAME to use regular expression.
                                     Regular expression accepts options.
                                     Example: '/taRget/i' matches 'target' and 'TARGET'
    -t, --testcase=TESTCASE          Runs tests in TestCases matching TESTCASE.
                                     Use '/PATTERN/' for TESTCASE to use regular expression.
                                     Regular expression accepts options.
                                     Example: '/taRget/i' matches 'target' and 'TARGET'
        --ignore-testcase=TESTCASE   Ignores tests in TestCases matching TESTCASE.
                                     Use '/PATTERN/' for TESTCASE to use regular expression.
                                     Regular expression accepts options.
                                     Example: '/taRget/i' matches 'target' and 'TARGET'
        --location=LOCATION          Runs tests that defined in LOCATION.
                                     LOCATION is one of PATH:LINE, PATH or LINE.
        --attribute=EXPRESSION       Runs tests that matches EXPRESSION.
                                     EXPRESSION is evaluated as Ruby's expression.
                                     Test attribute name can be used with no receiver in EXPRESSION.
                                     EXPRESSION examples:
                                       !slow
                                       tag == 'important' and !slow
        --[no-]priority-mode         Runs some tests based on their priority.
        --default-priority=PRIORITY  Uses PRIORITY as default priority
                                     (h[igh], i[mportant], l[ow], m[ust], ne[ver], no[rmal])
    -I, --load-path=DIR[:DIR...]     Appends directory list to $LOAD_PATH.
        --color-scheme=SCHEME        Use SCHEME as color scheme.
                                     (d[efault])
        --config=FILE                Use YAML format FILE content as configuration file.
        --order=ORDER                Run tests in a test case in ORDER order.
                                     (a[lphabetic], d[efined], r[andom])
        --max-diff-target-string-size=SIZE
                                     Shows diff if both expected result string size and actual result string size are less than or equal SIZE in bytes.
                                     (1000)
        --[no-]stop-on-failure       Stops immediately on the first non success test
                                     (false)
        --[no-]debug-on-failure      Run debugger if available on failure
                                     (false)
        --[no-]gc-stress             Enable GC.stress only while each test is running
                                     (false)
    -v, --verbose=[LEVEL]            Set the output level (default is normal).
                                     (i[mportant-only], n[ormal], p[rogress], s[ilent], v[erbose])
        --[no-]use-color=[auto]      Uses color output
                                     (default is auto)
        --progress-row-max=MAX       Uses MAX as max terminal width for progress mark
                                     (default is auto)
        --progress-style=STYLE       Uses STYLE as progress style
                                     (f[ault-only], i[nplace], m[ark]
        --no-show-detail-immediately Shows not passed test details immediately.
                                     (default is yes)
        --[no-]reverse-output        Shows fault details in reverse.
                                     (default is yes for tty output, no otherwise)
        --output-file-descriptor=FD  Outputs to file descriptor FD
        --                           Stop processing options so that the
                                     remaining options will be passed to the
                                     test.
    -h, --help                       Display this help.

Deprecated options:
        --console                    Console runner (use --runner).
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -1220,6 +1224,7 @@ def handle_notify
 
                     if group_watcher_limit || limit_bytes_per_second_reached? || should_shutdown_now?
                       # Just get out from tailing loop.
+                      @metrics.throttled.inc
```
ashie, line 1227: `should_shutdown_now?` isn't related with throttling so it should be treated separately.

ashie (changes requested):
Thanks for your work.
It looks good.
I commented on one thing that needs to be fixed.

Athishpranav2003:
@ashie addressed the comments

test/plugin/test_in_tail.rb
```diff
@@ -1878,13 +1878,6 @@ def test_z_refresh_watchers
     plugin.instance_eval do
       @pf = Fluent::Plugin::TailInput::PositionFile.load(sio, EX_FOLLOW_INODES, {}, logger: $log)
       @loop = Coolio::Loop.new
-      opened_file_metrics = Fluent::Plugin::LocalMetrics.new
-      opened_file_metrics.configure(config_element('metrics', '', {}))
-      closed_file_metrics = Fluent::Plugin::LocalMetrics.new
-      closed_file_metrics.configure(config_element('metrics', '', {}))
-      rotated_file_metrics = Fluent::Plugin::LocalMetrics.new
-      rotated_file_metrics.configure(config_element('metrics', '', {}))
-      @metrics = Fluent::Plugin::TailInput::MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics)
```
ashie, line 1887: Since this pull request adds `throttling_metrics`, it should be added here too.
On the other hand, I've noticed this initialization isn't needed originally since `create_driver` runs `configure` and it initializes `@metrics`. I overlooked it in reviewing #3504.

test/plugin/test_in_tail.rb
```diff
@@ -2245,15 +2238,6 @@ def test_should_close_watcher_after_rotate_wait
...
       d = create_driver(config, false)
-      d.instance.instance_eval do
-        opened_file_metrics = Fluent::Plugin::LocalMetrics.new
-        opened_file_metrics.configure(config_element('metrics', '', {}))
-        closed_file_metrics = Fluent::Plugin::LocalMetrics.new
-        closed_file_metrics.configure(config_element('metrics', '', {}))
-        rotated_file_metrics = Fluent::Plugin::LocalMetrics.new
-        rotated_file_metrics.configure(config_element('metrics', '', {}))
-        @metrics = Fluent::Plugin::TailInput::MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics)
-      end
```
ashie, line 2256: Ditto.

ashie (approved):
LGTM

Although It would be better to add tests for it, we don't have tests for metrics originally. So I don't always require adding them in this pull request. Let's tackle on it in other pull request(s).

I'll wait merging this until tomorrow to see if other maintainers have comments.

ashie:
> I'll wait merging this until tomorrow to see if other maintainers have comments.

(And waiting CI)

test/plugin/test_in_tail.rb
Athishpranav2003: I see
I didnt notice this when i was navigating through code

cosmo0920 (approved):
Looks attractive. Thanks!

ashie:
Thanks!

PR #4580, opened 2024-08-06
Config File Syntax: Extend Embedded Ruby Code support for Hashes and Arrays

### LINKED ISSUE #4385: Enable string interpolation for hash-type parameters

The [FAQ](https://docs.fluentd.org/quickstart/faq#how-can-i-use-environment-variables-to-configure-parameters-dynamically) mentions that you can use string interpolation to configure parameters using environment variables, e.g. `some_field "#{ENV['FOO_HOME']}"`, but this doesn't seem to apply to keys or values in `hash`-type parameters.

It would be nice if string values (or keys for that matter) in `hash` parameters could also get interpolated in the way `string` parameters already do. For example let's say I want to use an HTTP output with an authorization header that includes a token passed in through the environment:

```
<match>
  @type http
  headers {"Authorization": "Bearer #{ENV['SOME_ENV_VAR']}"}
</match>
```

For my particular example, a first-class feature to support bearer tokens in the HTTP output plugin (#3587) would likely solve my immediate problem, but also implementing more general-purpose string interpolation for hash parameters would remove the need for updates to that output plugin

### Additional context

_No response_

Athishpranav2003:
Fixes #4385 

Support [Embedded Ruby Code](https://docs.fluentd.org/configuration/config-file#embedded-ruby-code) for Hashes and Arrays.

```
key ["foo","#{1 + 1}"] => key ["foo","2"]
key {"foo":"#{1 + 1}"} => key {"foo":"2"}
```

This is not backward compatible.
We can disable this feature by surrounding the entire value with single quotes.

```
key `["foo","#{1 + 1}"]` => key ["foo","#{1 + 1}"]
key `{"foo":"#{1 + 1}"}` => key {"foo":"#{1 + 1}"}
```

**Note: this feature is for JSON literals**

This feature is for literals that using `[]` or `{}`.

* We need to sort out the `literal` and `value` of the Fluentd Config Syntax.
  * Fluentd first interprets `literal` by parsing the config file.
  * `literal` is String or JSON.
  * Then, Fluentd interprets `value` by parsing the `literal`.
  * `value` has [various types](https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values).
* **This feature has nothing to do with parsing the `value`.**

For example, we can specify Array/Hash `value` by using **String** `literal`
We don't necessarily need to use **JSON** `literal` to specify Array/Hash `value`.
The following formats works from previous versions, and there is no specification change at all.

```
key foo,bar
key foo:bar
key "foo,#{1 + 1}"
key "foo:#{1 + 1}"
```

**Note: support only String of JSON**

For example, this does not support Number of JSON.

```
key ["foo",2] => key ["foo",2]
key ["foo",#{1 + 1}] => ERROR
```

It is because JSON `literal` supports multiple lines and comments.
`#` not in String is considered the start of a comment line.

```
key ["foo", # comment
"bar", # comment
"boo" # comment
]
```

**Docs Changes**:
TODO

**Release Note**: 
* Config File Syntax: Extend Embedded Ruby Code support for Hashes and Arrays
  * Example: `key {"foo":"#{1 + 1}"} => key {"foo":"2"}`
  * Please note that this is not backward compatible, although we assume that this will never affect to actual existing configs.
  * In case the behavior changes unintentionally, you can disable this feature by surrounding the entire value with single quotes.
    * `key '{"foo":"#{1 + 1}"}' => key {"foo":"#{1 + 1}"}`

### DIFF
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -254,7 +254,7 @@ EOM
               buffer << line_buffer + "\n"
               line_buffer = ""
             else
-              if @ss.exist?(/\{[^}]+\}/)
+              if @ss.exist?(/^\{[^}]+\}/)
                 # if it's interpolated string
                 skip(/\{/)
                 line_buffer << eval_embedded_code(scan_embedded_code)
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -155,11 +155,11 @@ module Fluent
       end
 
       def scan_embedded_code
-        src = '"#{'+@ss.rest+"\n=begin\n=end\n}"
+        src = '#{'+@ss.rest+"\n=begin\n=end\n}"
 
         seek = -1
         while (seek = src.index('}', seek + 1))
-          unless Ripper.sexp(src[0..seek] + '"').nil? # eager parsing until valid expression
+          unless Ripper.sexp(src[0..seek]).nil? # eager parsing until valid expression
             break
           end
         end
@@ -168,7 +168,7 @@ module Fluent
           raise Fluent::ConfigParseError, @ss.rest
         end
 
-        code = src[3, seek-3]
+        code = src[2, seek-2]
 
         if @ss.rest.length < code.length
           @ss.pos += @ss.rest.length
@@ -255,7 +255,12 @@ EOM
               line_buffer = ""
             else
               # '#' is a char in json string
-              line_buffer << char
+              if skip(/\{/)
+                line_buffer << eval_embedded_code(scan_embedded_code)
+                skip(/\}/)
+              else
+                line_buffer << char
+              end
             end
 
             next # This char '#' MUST NOT terminate json object.
@@ -280,6 +285,7 @@ EOM
         end
 
         JSON.dump(result)
+
       end
     end
   end
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -176,7 +176,7 @@ module Fluent
         end
 
         @ss.pos += code.length
-
+        p code
         '"#{' + code + '}"'
       end
 
@@ -254,11 +254,13 @@ EOM
               buffer << line_buffer + "\n"
               line_buffer = ""
             else
-              # '#' is a char in json string
-              if skip(/\{/)
+              if @ss.exist?(/\{[^}]+\}/)
+                # if it's interpolated string
+                skip(/\{/)
                 line_buffer << eval_embedded_code(scan_embedded_code)
                 skip(/\}/)
               else
+                # '#' is a char in json string
                 line_buffer << char
               end
             end
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -155,11 +155,11 @@ module Fluent
       end
 
       def scan_embedded_code
-        src = '#{'+@ss.rest+"\n=begin\n=end\n}"
+        src = '"#{'+@ss.rest+"\n=begin\n=end\n}"
 
         seek = -1
         while (seek = src.index('}', seek + 1))
-          unless Ripper.sexp(src[0..seek]).nil? # eager parsing until valid expression
+          unless Ripper.sexp(src[0..seek] + '"').nil? # eager parsing until valid expression
             break
           end
         end
@@ -168,7 +168,7 @@ module Fluent
           raise Fluent::ConfigParseError, @ss.rest
         end
 
-        code = src[2, seek-2]
+        code = src[3, seek-3]
 
         if @ss.rest.length < code.length
           @ss.pos += @ss.rest.length
@@ -176,6 +176,7 @@ module Fluent
         end
 
         @ss.pos += code.length
+
         '"#{' + code + '}"'
       end
 
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -176,7 +176,6 @@ module Fluent
         end
 
         @ss.pos += code.length
-        p code
         '"#{' + code + '}"'
       end
 
--- a/lib/fluent/config/literal_parser.rb
+++ b/lib/fluent/config/literal_parser.rb
@@ -285,7 +285,6 @@ EOM
         end
 
         JSON.dump(result)
-
       end
     end
   end
--- a/test/config/test_literal_parser.rb
+++ b/test/config/test_literal_parser.rb
@@ -259,6 +259,17 @@ module Fluent::Config
       test('[ "a" , "b" ]') { assert_text_parsed_as_json(["a","b"], '[ "a" , "b" ]') }
       test("[\n\"a\"\n,\n\"b\"\n]") { assert_text_parsed_as_json(["a","b"], "[\n\"a\"\n,\n\"b\"\n]") }
       test('["ab","cd"]') { assert_text_parsed_as_json(["ab","cd"], '["ab","cd"]') }
+      test('["a","#{v1}"') { assert_text_parsed_as_json(["a","#{v1}"], '["a","#{v1}"]') }
+      test('["a","#{v1}","#{v2}"]') { assert_text_parsed_as_json(["a","#{v1}","#{v2}"], '["a","#{v1}","#{v2}"]') }
+      test('["a","#{v1} #{v2}"]') { assert_text_parsed_as_json(["a","#{v1} #{v2}"], '["a","#{v1} #{v2}"]') }
+      test('["a","#{hostname}"]') { assert_text_parsed_as_json(["a","#{Socket.gethostname}"], '["a","#{hostname}"]') }
+      test('["a","foo#{worker_id}"]') {
+        ENV.delete('SERVERENGINE_WORKER_ID')
+        assert_text_parsed_as('["a","foo"]', '["a","foo#{worker_id}"]')
+        ENV['SERVERENGINE_WORKER_ID'] = '1'
+        assert_text_parsed_as('["a","foo1"]', '["a","foo#{worker_id}"]')
+        ENV.delete('SERVERENGINE_WORKER_ID')
+      }
       json_array_with_js_comment = <<EOA
 [
  "a", // this is a
@@ -296,6 +307,18 @@ EOA
       test('{"a":"b","c":"d"}') { assert_text_parsed_as_json({"a"=>"b","c"=>"d"}, '{"a":"b","c":"d"}') }
       test('{ "a" : "b" , "c" : "d" }') { assert_text_parsed_as_json({"a"=>"b","c"=>"d"}, '{ "a" : "b" , "c" : "d" }') }
       test('{\n\"a\"\n:\n\"b\"\n,\n\"c\"\n:\n\"d\"\n}') { assert_text_parsed_as_json({"a"=>"b","c"=>"d"}, "{\n\"a\"\n:\n\"b\"\n,\n\"c\"\n:\n\"d\"\n}") }
+      test('{"a":"b","c":"#{v1}"}') { assert_text_parsed_as_json({"a"=>"b","c"=>"#{v1}"}, '{"a":"b","c":"#{v1}"}') }
+      test('{"a":"b","#{v1}":"d"}') { assert_text_parsed_as_json({"a"=>"b","#{v1}"=>"d"}, '{"a":"b","#{v1}":"d"}') }
+      test('{"a":"#{v1}","c":"#{v2}"}') { assert_text_parsed_as_json({"a"=>"#{v1}","c"=>"#{v2}"}, '{"a":"#{v1}","c":"#{v2}"}') }
+      test('{"a":"b","c":"d #{v1} #{v2}"}') { assert_text_parsed_as_json({"a"=>"b","c"=>"d #{v1} #{v2}"}, '{"a":"b","c":"d #{v1} #{v2}"}') }
+      test('{"a":"#{hostname}"}') { assert_text_parsed_as_json({"a"=>"#{Socket.gethostname}"}, '{"a":"#{hostname}"}') }
+      test('{"a":"foo#{worker_id}"}') {
+        ENV.delete('SERVERENGINE_WORKER_ID')
+        assert_text_parsed_as('{"a":"foo"}', '{"a":"foo#{worker_id}"}')
+        ENV['SERVERENGINE_WORKER_ID'] = '1'
+        assert_text_parsed_as('{"a":"foo1"}', '{"a":"foo#{worker_id}"}')
+        ENV.delete('SERVERENGINE_WORKER_ID')
+      }
       json_hash_with_comment = <<EOH
 {
  "a": 1, # this is a
--- a/test/config/test_literal_parser.rb
+++ b/test/config/test_literal_parser.rb
@@ -319,6 +319,9 @@ EOA
         assert_text_parsed_as('{"a":"foo1"}', '{"a":"foo#{worker_id}"}')
         ENV.delete('SERVERENGINE_WORKER_ID')
       }
+      test('no quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'test'}, '{"a":"b","c":"#{v1}"}') }
+      test('single quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'#{v1}'}, '\'{"a":"b","c":"#{v1}"}\'') }
+      test('double quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'test'}, '"{\"a\":\"b\",\"c\":\"#{v1}\"}"') }
       json_hash_with_comment = <<EOH
 {
  "a": 1, # this is a


Athishpranav2003:
@ashie please check this whenever you are free

daipom (commented):
Thanks for this PR!

As I commented on the original issue, we need to consider the specifications in `LiteralParser` carefully.
I'd like to think about it some more, but at the moment it doesn't seem like a good design decision to give `types.rb` the same functionality as `LiteralParser`.

I would like to hear from other maintainers.

https://github.com/fluent/fluentd/issues/4385#issuecomment-2270749268

> We need to consider this as an issue of [Embedded Ruby Code](https://docs.fluentd.org/configuration/config-file#embedded-ruby-code). Using environment variables is just one way of it.
> 
> `Embedded Ruby Code` feature is implemented in `LiteralParser`:
> 
> https://github.com/fluent/fluentd/blob/e763c0761c44d9734b6aa374371387a2e8406522/lib/fluent/config/literal_parser.rb#L183-L199
> 
> https://github.com/fluent/fluentd/blob/e763c0761c44d9734b6aa374371387a2e8406522/lib/fluent/config/literal_parser.rb#L108-L110
> 
> Since it is for `double_quoted_string`, we need to use double quotes for the entire value to use this feature.
> 
> It may be possible to address this as a `types` problem, as in #4580, but the specifications in `LiteralParser` need to be carefully considered.

ashie:
> but at the moment it doesn't seem like a good design decision to give `types.rb` the same functionality as `LiteralParser`.

I don't yet see the detail but I feel same smell.

Athishpranav2003:
@daipom do you suggest that we should handle this scenario in `literal parser.rb`?

daipom:
> @daipom do you suggest that we should handle this scenario in `literal parser.rb`?

Yes. We should start by considering whether there are points in `LiteralParser` that can be improved.

Athishpranav2003:
Oh ok
let me look at it today and check if i can move this processing to it

daipom:
@Athishpranav2003 Thanks so much!

Athishpranav2003:
@daipom not sure of this latest changes i have pushed.
2 UTs are failing for the latest code changes which i have made sure i have reused the code in clean form.
Please suggest on what to do for this

Athishpranav2003:
The 2 UTs for which my code is failing are
```ruby
  test('"#{"}"}"') { assert_text_parsed_as("}", '"#{"}"}"') }
  test('"#{#}"') { assert_parse_error('"#{#}"') }  # error in embedded ruby code
```
Seems like these corner cases where written specific to the previous Embedded Ruby code parsing (which i felt needs slight revamp)

Athishpranav2003:
@daipom ignore the prev comments
I have corrected it and pushed
Also tested locally in my side. Please check it and let me know if there is any issue

Athishpranav2003:
@daipom not sure why 2 workflows alone failed, could you check that part alone?

daipom:
> @daipom not sure why 2 workflows alone failed, could you check that part alone?

No problem! They have nothing to do with this PR!
Sorry, some tests are unstable and sometimes fail.

ashie:
Could you add unit test for this?

Athishpranav2003:
@ashie added UTs for the same

daipom:
@Athishpranav2003 
> @daipom ignore the prev comments I have corrected it and pushed Also tested locally in my side. Please check it and let me know if there is any issue

Thanks!
Could you please clarify the specification change?
The focus is on what to do with the specifications.

The current specification is:

* String
  * non-quote: Simple one-line string without spaces.
  * single-quote: Allow spaces or multiple lines.
  * double-quote: Allow escaped characters or embedded codes.
* JSON
  * The specification seems unclear to me, but looks like it is based on `JSON.parse()` and does not support escaped characters or embedded codes.

As far as seeing changes in the code, embedded codes will always be supported, right?
If so, I have 2 concerns:

* It will be impossible to write `#{...}` as a raw string in JSON.
* Not backward compatible.

We need to consider whether the new specifications are acceptable while taking into account these concerns.

Athishpranav2003:
Strings stay the same - only double quoted strings support string interpolation

For Hashes and Arrays - We do string interpolation for anything that has `#{...}` in it. What i can do is maybe restrict this to double quoted strings inside hashes but not sure if its needed. Please let me know on this alone

Tho we have one benefit that the docs present for now we didn't restrict the interpolation to strings so it wont go against the specifications from the past

daipom:
> What i can do is maybe restrict this to double quoted strings inside hashes but not sure if its needed. Please let me know on this alone

This is the difficult part.
JSON uses double quotes for strings.
So, we can't switch the enabled and disabled of embedded codes in with the type of quates...

daipom:
> Strings stay the same - only double quoted strings support string interpolation
> 
> For Hashes and Arrays - We do string interpolation for anything that has `#{...}` in it.

OK. I think this direction is possible for us!
There are concerns(https://github.com/fluent/fluentd/pull/4580#issuecomment-2272700592):
 
>     * It will be impossible to write `#{...}` as a raw string in JSON.
>     * Not backward compatible.

but I think these points are less important for this feature.
To write `#{...}` as is, we can use single-quote to the entire string, such as `param '{"k","#{foo}"}'`.
So, there is a workaround.

We need to hear from other maintainers on this point.

Athishpranav2003:
Fine
We can wait for other's to put their opinions

daipom:
The focus is on what extent we care about the possibility that existing users are using settings like:

```xml
<match test>
  @type http
  endpoint foo
  headers {"foo": "foo#{bar"}
</match>
```

When `#{` exists, this change can break the existing setting.

lib/fluent/config/literal_parser.rb
```diff
@@ -254,8 +254,15 @@ def scan_json(is_array)
               buffer << line_buffer + "\n"
               line_buffer = ""
             else
-              # '#' is a char in json string
-              line_buffer << char
+              if @ss.exist?(/\{[^}]+\}/)
```
Athishpranav2003, line 257: @daipom this check can be changed to capture the case you mentioned. So here we can make sure that `}` is not being matched with the json's ending `}`. In that way we avoid capturing the strings that dont have closing braces for interpolation

ashie:
> > ```
> > * It will be impossible to write `#{...}` as a raw string in JSON.
> > * Not backward compatible.
> > ```
> 
> but I think these points are less important for this feature. To write `#{...}` as is, we can use single-quote to the entire string, such as `param '{"k","#{foo}"}'`. So, there is a workaround.

It seems that such workaround isn't available for Array & Hash types.
Following change breaks the test, since it's tried to parse as String type which isn't matched with declaration in `config_param`.

```diff
diff --git a/test/config/test_config_parser.rb b/test/config/test_config_parser.rb
index fe29028e..b0c18fc9 100644
--- a/test/config/test_config_parser.rb
+++ b/test/config/test_config_parser.rb
@@ -483,8 +483,8 @@ module Fluent::Config
           param_size 4k
           param_bool true
           param_time 10m
-          param_hash { "key1": "value1", "key2": 2 }
-          param_array ["value1", "value2", 100]
+          param_hash '{ "key1": "value1", "key2": 2 }'
+          param_array '["value1", "value2", 100]'
           param_regexp /pattern/
         ])
         target = AllTypes.new.configure(conf)
```

ashie:
> I'm not sure but it seems that such workaround isn't available for Array & Hash types. For example following change breaks the test, since it's tried to parse as String type which isn't matched with declaration in `config_param`.

Sorry, please ignore my comment.
Even it's parsed as String, it will be parsed again as Array or Hash at config/types.rb.

ashie:
> We need to hear from other maintainers on this point.

In general, we intend to keep compatibility so that users can continue to use same major version without changing settings. When we have to change behaviour in any reason,  it would be better to show announcements or warnings before making the change, and wait a few minor versions to make the change in actual.

On the other hand, using `#{...}` in fluentd's setting is expected to evaluate Ruby code in most cases and I guess using such syntax in other purpose seems very rare. If we investigate major plugins and it seems unlikely that such syntax will be used, we can made this change in next minor version (without adding any warning) I guess.

ashie:
> > I'm not sure but it seems that such workaround isn't available for Array & Hash types. For example following change breaks the test, since it's tried to parse as String type which isn't matched with declaration in `config_param`.
> 
> Sorry, please ignore my comment. Even it's parsed as String, it will be parsed again as Array or Hash at config/types.rb.

It would be better adding a test for this.

ashie:
> On the other hand, using `#{...}` in fluentd's setting is expected to evaluate Ruby code in most cases and I guess using such syntax in other purpose seems very rare. If we investigate major plugins and it seems unlikely that such syntax will be used, we can made this change in next minor version (without adding any warning) I guess.

BTW my impression is very positive for including this change in the next minor release (with addressing some corner cases as @daipom mentioned).
It seems it's much natural than before for users.

Athishpranav2003:
> > > I'm not sure but it seems that such workaround isn't available for Array & Hash types. For example following change breaks the test, since it's tried to parse as String type which isn't matched with declaration in `config_param`.
> > 
> > 
> > Sorry, please ignore my comment. Even it's parsed as String, it will be parsed again as Array or Hash at config/types.rb.
> 
> It would be better adding a test for this.

A test to check whether its parsed as array/hash or string?

ashie:
> A test to check whether its parsed as array/hash or string?

A test that checks whether a single quoted string is correctly parsed as array/hash when the declared type is array/hash.

Athishpranav2003:
@ashie i guess current code renders both " and ' quoted hashes the same :( . This is because we dont handle the case seperately since its optional for config to enclose hash/array in quotes. Not sure which is correct way to solve this

lib/fluent/config/literal_parser.rb
daipom: Hmm, we would be happy to reduce the risk simply, but isn't it difficult to narrow down the exact scope of the search for the end brace?
(Don't we need to analyze JSON structure?)

daipom:
> @ashie i guess current code renders both " and ' quoted hashes the same :( . This is because we dont handle the case seperately since its optional for config to enclose hash/array in quotes. Not sure which is correct way to solve this

When we use `"` or `'` to an entire literal, that literal will be handled as String even if it is actually JSON(Array).
LiteralParser does not necessarily parse JSON(Array) as JSON because it will be parsed as JSON later when the type is applied.

https://github.com/fluent/fluentd/blob/5618dd0e86bf0bc47a6bed2da50e37943ac2c8e9/lib/fluent/config/types.rb#L203-L207

lib/fluent/config/literal_parser.rb
Athishpranav2003: I guess only nearest thing we can do here is make sure we dont capture end braces with the interpolation braces.
TBH this problem boils down to proper parsing problem like how compilers do if we are thinking in that rigorous perspective

Athishpranav2003:
https://github.com/fluent/fluentd/blob/master/lib/fluent/config/literal_parser.rb#L67-L78
@daipom 
So here we are giving higher precedence to Arrays and JSON than strings. So even if we specify the quotes its gonna get analysed as JSON/Arrays. 
After the analysis it `JSON.dumps` the array which once converts it back to string which is then again handled in `config/types.rb`

daipom:
> https://github.com/fluent/fluentd/blob/master/lib/fluent/config/literal_parser.rb#L67-L78 @daipom So here we are giving higher precedence to Arrays and JSON than strings. So even if we specify the quotes its gonna get analysed as JSON/Arrays. After the analysis it `JSON.dumps` the array which once converts it back to string which is then again handled in `config/types.rb`

@Athishpranav2003 
If the first character of the literal is `[` or `{`, the literal will be parsed by `scan_json()`.
On the other hand, if the first character is not `[` or `{`, the literal will be parsed by `scan_string()`.

https://github.com/fluent/fluentd/blob/5618dd0e86bf0bc47a6bed2da50e37943ac2c8e9/lib/fluent/config/literal_parser.rb#L70-L76

So, if we use quotes to the entire JSON literal, such as `'{"foo":"bar"}'` , it will be parsed by `scan_string()`.
So, it will be a workaround to write `#{...}` as is.

Athishpranav2003:
But isn't the code using skip to identify the starting { and [ ?
Since it's using the skip the " at the start will anyways match?

ashie:
@Athishpranav2003 Please try this:
```diff
diff --git a/test/config/test_literal_parser.rb b/test/config/test_literal_parser.rb
index c5b9b188..275e468e 100644
--- a/test/config/test_literal_parser.rb
+++ b/test/config/test_literal_parser.rb
@@ -319,6 +319,9 @@ EOA
         assert_text_parsed_as('{"a":"foo1"}', '{"a":"foo#{worker_id}"}')
         ENV.delete('SERVERENGINE_WORKER_ID')
       }
+      test('no quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'test'}, '{"a":"b","c":"#{v1}"}') }
+      test('single quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'#{v1}'}, '\'{"a":"b","c":"#{v1}"}\'') }
+      test('double quote') { assert_text_parsed_as_json({'a'=>'b','c'=>'test'}, '"{\"a\":\"b\",\"c\":\"#{v1}\"}"') }
       json_hash_with_comment = <<EOH
 {
  "a": 1, # this is a
```

```console
$ bundle exec rake test TEST=test/config/test_literal_parser.rb
```

All tests will be passed.
It means

* test single quote: Embedded ruby code isn't evaludated, the string `#{v1}` is output as is.
* test double quote: Embedded ruby code is evaludated, the string `#{v1}` is expanded as `test`.

daipom:
> But isn't the code using skip to identify the starting { and [ ? 
> Since it's using the skip the " at the start will anyways match?

I'm talking about this behavior, is this as expected?

```rb
$ irb -r strscan
irb(main):001> ss = StringScanner.new('{"foo":"bar"}')
=> #<StringScanner 0/13 @ "{\"foo...">
irb(main):002> ss.rest
=> "{\"foo\":\"bar\"}"
irb(main):003> ss.skip(/\{/)
=> 1
irb(main):004> ss = StringScanner.new('\'{"foo":"bar"}\'')
=> #<StringScanner 0/15 @ "'{\"fo...">
irb(main):005> ss.rest
=> "'{\"foo\":\"bar\"}'"
irb(main):006> ss.skip(/\{/)
=> nil
irb(main):007> ss.skip(/'/)
=> 1
```

Athishpranav2003:
@ashie i tried your tests
Its passing
I guess this seems fine to me now

Athishpranav2003:
> > But isn't the code using skip to identify the starting { and [ ?
> > Since it's using the skip the " at the start will anyways match?
> 
> I'm talking about this behavior, is this as expected?
> 
> ```ruby
> $ irb -r strscan
> irb(main):001> ss = StringScanner.new('{"foo":"bar"}')
> => #<StringScanner 0/13 @ "{\"foo...">
> irb(main):002> ss.rest
> => "{\"foo\":\"bar\"}"
> irb(main):003> ss.skip(/\{/)
> => 1
> irb(main):004> ss = StringScanner.new('\'{"foo":"bar"}\'')
> => #<StringScanner 0/15 @ "'{\"fo...">
> irb(main):005> ss.rest
> => "'{\"foo\":\"bar\"}'"
> irb(main):006> ss.skip(/\{/)
> => nil
> irb(main):007> ss.skip(/\'/)
> => 1
> ```

Not sure
I can check this but the problem seems to be solved now tho :0

daipom:
Thanks! I'm checking the acutual behavior.

daipom:
# Array behavior

Applied to only explicit arrays, except for single quoted arrays.

Config:

```xml
<source>
  @type sample
  tag test1
  sample {"k1":"v1","k2":"v2","k3":"v3"}
</source>

<source>
  @type sample
  tag test2
  sample {"k1":"v1","k2":"v2","k3":"v3"}
</source>

<source>
  @type sample
  tag test3
  sample {"k1":"v1","k2":"v2","k3":"v3"}
</source>

# Not applied to shorthand array
<filter test1>
  @type record_transformer
  remove_keys k1, "#{ENV['FOO']}"
</filter>

# Applied to array
<filter test2>
  @type record_transformer
  remove_keys ["k1", "#{ENV['FOO']}"]
</filter>

# Not applied to single quoted array
<filter test3>
  @type record_transformer
  remove_keys '["k1", "#{ENV[\'FOO\']}"]'
</filter>

<match test*>
  @type stdout
</match>
```

Parsed result:

```xml
<ROOT>
  <source>
    @type sample
    tag "test1"
    sample {"k1":"v1","k2":"v2","k3":"v3"}
  </source>
  <source>
    @type sample
    tag "test2"
    sample {"k1":"v1","k2":"v2","k3":"v3"}
  </source>
  <source>
    @type sample
    tag "test3"
    sample {"k1":"v1","k2":"v2","k3":"v3"}
  </source>
  <filter test1>
    @type record_transformer
    remove_keys k1, "#{ENV['FOO']}"
  </filter>
  <filter test2>
    @type record_transformer
    remove_keys ["k1","k2"]
  </filter>
  <filter test3>
    @type record_transformer
    remove_keys ["k1", "#{ENV['FOO']}"]
  </filter>
  <match test*>
    @type stdout
  </match>
</ROOT>
```

Output:

```
... test1: {"k2":"v2","k3":"v3"}
... test2: {"k3":"v3"}
... test3: {"k2":"v2","k3":"v3"}
```

daipom:
# Hash behavior

Applied to only explicit hashes, except for single quoted hashes.

It should be noted that it can also be applied to non-string values of JSON.
It may feel strange because non-String values are not surrounded by double quotes.
The same thing is true for Array.
However, this behavior looks acceptable to me because non-string values of JSON would rarely be used.

Config:

```xml
# Not Applied to shorthand hash
<match test>
  @type http
  endpoint foo
  headers key:#{ENV['FOO']}
</match>

# Applied to String value of hash
<match test>
  @type http
  endpoint foo
  headers {"key":"#{ENV['FOO']}"}
</match>

# Applied to non-String value of hash
<match test>
  @type http
  endpoint foo
  headers {"key":#{ENV['FOO']}}
</match>

# Not Applied to single quoted hash
<match test>
  @type http
  endpoint foo
  headers '{"key":"#{ENV[\'FOO\']}"}'
</match>
```

Parsed result (`FOO=0`):

```xml
<ROOT>
  <match test>
    @type http
    endpoint "foo"
    headers key:#{ENV['FOO']}
  </match>
  <match test>
    @type http
    endpoint "foo"
    headers {"key":"0"}
  </match>
  <match test>
    @type http
    endpoint "foo"
    headers {"key":0}
  </match>
  <match test>
    @type http
    endpoint "foo"
    headers {"key":"#{ENV['FOO']}"}
  </match>
</ROOT>
```

Note

* `in_sample` has the original logic to parse the `sample` param, so it is unsuitable for testing this behavior.
* `worker_id` is not set when checking config on launching, so it can cause a parse error if used for a non-string value of JSON.
  * We will need some fixes in the future for this problem.

```xml
<match test>
  @type http
  endpoint foo
  headers {"key":#{worker_id}}
</match>
```

```
.../home/daipom/work/fluentd/fluentd/lib/fluent/config/basic_parser.rb:92:in `parse_error!': got incomplete JSON hash configuration at 4.conf line 7,1 (Fluent::ConfigParseError)
```

daipom:
https://github.com/fluent/fluentd/pull/4580#issuecomment-2277207381
> It should be noted that it can also be applied to non-string values of JSON.
> It may feel strange because non-string values are not surrounded by double quotes.
> The same thing is true for Array.
> However, this behavior looks acceptable to me because non-string values of JSON would rarely be used.

Considering this point, it is inaccurate to describe this PR's specification as `string interpolation fix`.

https://github.com/fluent/fluentd/pull/4580#issuecomment-2272820976
> For Hashes and Arrays - We do string interpolation for anything that has #{...} in it.

We should describe this specification as 

* Support [Embedded Ruby Code](https://docs.fluentd.org/configuration/config-file#embedded-ruby-code) for Hashes and Arrays.
  * Now, we can evaluate the Ruby code with `#{}` in `"` quoted string or in explicit Hash/Array.
  * This can be disabled by surrounding the entire value with single quotes.

Athishpranav2003:
@daipom i will change the subject of PR as you mentioned. 
For the worker_id should we also add that check or should we give that responsibility to users?

daipom:
> i will change the subject of PR as you mentioned. 

Thanks!!

> For the worker_id should we also add that check or should we give that responsibility to users?

I think this PR does not need to care about the `worker_id` problem!
It does not affect the existing users and it should be a rare case.

daipom:
@Athishpranav2003 Thanks! This looks good to me!
Let me discuss the compatibility concerns a bit more with the other maintainers.

Athishpranav2003:
> @Athishpranav2003 Thanks! This looks good to me! Let me discuss the compatibility concerns a bit more with the other maintainers.

Peace :). 
Let me know if anything needs to be changed

lib/fluent/config/literal_parser.rb
ashie, line 257:
```suggestion
              if @ss.exist?(/^\{[^}]+\}/)
```
Need to  match `{` just after `#`.

ashie (changes requested):

Athishpranav2003:
@ashie i have added the changes you suggested.

daipom (approved):
LGTM. Thanks!

ashie (approved):
LGTM

Waiting to start merge window for v1.18.0.

daipom:
I have updated [the description](https://github.com/fluent/fluentd/pull/4580#issue-2450026010).

daipom:
TODO document.

cosmo0920:
Yay, it's a really good enhancement! :+1 :+1:

PR #4581, opened 2024-08-06
github: set job permissions

kenhys:

Fixes N/A

It is related token permission topic
https://scorecard.dev/viewer/?uri=github.com/fluent/fluentd

By default, it is enough to give read permission.
For stale actions, it require manipulate issues,
so write permission is given explicitly.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/.github/workflows/stale-actions.yml
+++ b/.github/workflows/stale-actions.yml
@@ -10,6 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       issues: write
+      pull-requests: write
     steps:
     - uses: actions/stale@v8
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/stale-actions.yml
```diff
@@ -3,9 +3,13 @@ on:
   schedule:
   - cron: "00 10 * * *"
 
+permissions: read-all
+
 jobs:
   stale:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
ashie, line 12:
```suggestion
      issues: write
      pull-requests: write
```

Isn't it needed also `pull-requests`?

kenhys: Thanks, I've overlooked it.

ashie (approved):

PR #4583, opened 2024-08-07
CI: Set continue-on-error false for Ruby Head

ashie:
N/A

When a test is failed, the job should be also treated as failed.
But the current setting ignore errors.

Please see https://github.com/fluent/fluent-plugin-sql/actions/workflows/linux-ruby-head.yml
(Some tests are always failed with recent Ruby Head but jobs are all green.)

**Docs Changes**:
N/A

**Release Note**: 
N/A


daipom (approved):
LGTM.

PR #4586, opened 2024-08-14
in_sample: duplicate record to support destructive change

### LINKED ISSUE #4575: remove_key_name_field true when key_name is message causes ArgumentError error="message does not exist"
### Describe the bug

When trying to parse a json log in message to a new field, if you try to remove the message field it works the first time, then fails every time after it runs

### To Reproduce

copy the config file below and restart td-agent

### Expected behavior

no error. message field gone and data field only existing every time sample runs
2024-07-31 16:12:44.030559702 -0400 log: {"data":{"field":"value"}}

### Your Environment

```markdown
- Fluentd version: 1.17.0 (I attempted to update this using td-agent-gem)
- Package version: # td-agent --version
td-agent 4.4.1 fluentd 1.15.2 (c32842297ed2c306f1b841a8f6e55bdd0f1cb27f)
- Operating system: Ubuntu
- Kernel version: Linux 5.15.0-1065-aws #71~20.04.1-Ubuntu
```

### Your Configuration

```apache
<source>
    @type sample
    tag log
    sample {"message": "{\"field\":\"value\"}"}
</source>
<filter log>
    @type parser
    key_name message
    reserve_data true
    remove_key_name_field true
    hash_value_field data
    <parse>
        @type json
    </parse>
</filter>
<match log>
  @type stdout
</match>
```

### Your Error Log

```shell
2024-07-31 16:12:41 -0400 [info]: init supervisor logger path="/var/log/td-agent/td-agent.log" rotate_age=nil rotate_size=nil
2024-07-31 16:12:41 -0400 [info]: parsing config file is succeeded path="/etc/td-agent/td-agent.conf"
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-beats' version '1.1.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-calyptia-monitoring' version '0.1.3'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-elasticsearch' version '5.2.3'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-explode_filter' version '0.2.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-filter_empty_keys' version '0.0.4'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-flowcounter-simple' version '0.1.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-grok-parser' version '2.6.2'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-kafka' version '0.18.1'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-kinesis-aggregation' version '0.4.1'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-kv-parser' version '1.0.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-metrics-cmetrics' version '0.1.2'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-mutate_filter' version '1.0.7'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-opensearch' version '1.0.7'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-prometheus' version '2.0.3'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-prometheus_pushgateway' version '0.1.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-record-modifier' version '2.2.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-record-modifier' version '2.1.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-rename-key' version '0.4.1'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-rewrite-tag-filter' version '2.4.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-s3' version '1.7.1'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-sd-dns' version '0.1.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-statsd-output' version '1.4.2'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-systemd' version '1.0.5'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-td' version '1.2.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-utmpx' version '0.5.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluent-plugin-webhdfs' version '1.5.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluentd' version '1.17.0'
2024-07-31 16:12:42 -0400 [info]: gem 'fluentd' version '1.15.2'
2024-07-31 16:12:42 -0400 [info]: gem 'fluentd' version '0.14.25'
2024-07-31 16:12:42 -0400 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2024-07-31 16:12:42 -0400 [warn]: both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2024-07-31 16:12:42 -0400 [info]: using configuration file: <ROOT>
  <source>
    @type sample
    tag "log"
    sample {"message":"{\"field\":\"value\"}"}
  </source>
  <filter log>
    @type parser
    key_name "message"
    reserve_data true
    remove_key_name_field true
    hash_value_field "data"
    <parse>
      @type "json"
    </parse>
  </filter>
  <match log>
    @type stdout
  </match>
</ROOT>
2024-07-31 16:12:42 -0400 [info]: starting fluentd-1.17.0 pid=1879243 ruby="2.7.6"
2024-07-31 16:12:42 -0400 [info]: spawn command to main:  cmdline=["/opt/td-agent/bin/ruby", "-Eascii-8bit:ascii-8bit", "/opt/td-agent/bin/fluentd", "--log", "/var/log/td-agent/td-agent.log", "--daemon", "/var/run/td-agent/td-agent.pid", "--under-supervisor"]
2024-07-31 16:12:42 -0400 [info]: #0 init worker0 logger path="/var/log/td-agent/td-agent.log" rotate_age=nil rotate_size=nil
2024-07-31 16:12:42 -0400 [info]: adding filter pattern="log" type="parser"
2024-07-31 16:12:42 -0400 [info]: adding match pattern="log" type="stdout"
2024-07-31 16:12:43 -0400 [info]: adding source type="sample"
2024-07-31 16:12:43 -0400 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2024-07-31 16:12:43 -0400 [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
2024-07-31 16:12:43 -0400 [info]: #0 starting fluentd worker pid=1879252 ppid=1879249 worker=0
2024-07-31 16:12:43 -0400 [info]: #0 fluentd worker is now running worker=0
2024-07-31 16:12:44.030559702 -0400 log: {"data":{"field":"value"}}
2024-07-31 16:12:45 -0400 [warn]: #0 dump an error event: error_class=ArgumentError error="message does not exist" location=nil tag="log" time=2024-07-31 16:12:45.033706463 -0400 record={}
2024-07-31 16:12:45.033706463 -0400 log: {"data":{}}
2024-07-31 16:12:46 -0400 [warn]: #0 dump an error event: error_class=ArgumentError error="message does not exist" location=nil tag="log" time=2024-07-31 16:12:46.035347126 -0400 record={}
2024-07-31 16:12:46.035347126 -0400 log: {"data":{}}
2024-07-31 16:12:47 -0400 [warn]: #0 dump an error event: error_class=ArgumentError error="message does not exist" location=nil tag="log" time=2024-07-31 16:12:47.037108884 -0400 record={}
2024-07-31 16:12:47.037108884 -0400 log: {"data":{}}
2024-07-31 16:12:48 -0400 [warn]: #0 dump an error event: error_class=ArgumentError error="message does not exist" location=nil tag="log" time=2024-07-31 16:12:48.041690410 -0400 record={}
2024-07-31 16:12:48.041690410 -0400 log: {"data":{}}
```

### Additional context

I didn't see anywhere in the fluentd docs where message has to exist. The other ticket I found relating to this error looked specific to an elasticsearch output plugin

kenhys:

Fixes #4575

When in_sample plugin is used with filter parser which uses remove_key_name_field, it raises the following error repeatedly.

```
  #0 dump an error event: error_class=ArgumentError error="message does not exist"
```
This kind of error occurs when key_name and remove_key_name_field removes key from record with destructive change in filter parser affects generated sample data.

**Docs Changes**:

TODO: Add `reuse_record` parameter.

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/in_sample.rb
+++ b/lib/fluent/plugin/in_sample.rb
@@ -39,6 +39,8 @@ module Fluent::Plugin
     config_param :auto_increment_key, :string, default: nil
     desc "The boolean to suspend-and-resume incremental value after restart"
     config_param :suspend, :bool, default: false,deprecated: 'This parameters is ignored'
+    desc "Reuse the sample data to reduce the load when sending large amounts of data. You can enable it if filter does not do destructive change"
+    config_param :reuse_record, :bool, default: false
     desc "The sample data to be generated. An array of JSON hashes or a single JSON hash."
     config_param :sample, alias: :dummy, default: [{"message" => "sample"}] do |val|
       begin
@@ -117,15 +119,19 @@ module Fluent::Plugin
       end
     end
 
-    def generate
-      d = @sample[@sample_index]
-      unless d
-        @sample_index = 0
-        d = @sample[@sample_index]
-      end
+    def next_sample
+      d = @reuse_record ? @sample[@sample_index] : @sample[@sample_index].dup
       @sample_index += 1
+      return d if d
+
+      @sample_index = 0
+      next_sample
+    end
+
+    def generate
+      d = next_sample
       if @auto_increment_key
-        d = d.dup
+        d = d.dup if @reuse_record
         d[@auto_increment_key] = @storage.update(:auto_increment_value){|v| v + 1 }
       end
       d


kenhys:
checking whether it does not break existing test cases with CI.

daipom (changes requested):
Thanks for this fix!
Shouldn't we remove this `dup`?

https://github.com/fluent/fluentd/blob/c9dcf25b9215e32084a5df53cb0888d31416fc50/lib/fluent/plugin/in_sample.rb#L127-L130

daipom (approved):
I have tested the impact on performance.
There seems to be a noticeable load on the CPU and RES for large data flows.

It seems acceptable to me because it is hard to imagine that the performance of `in_sample` would be a problem.

This may affect users using `in_sample` to test performance.
So, it would be enough to report at the release that the `in_sample` has become a bit overloaded when flowing large data for testing.

I would like to hear one more person's opinion just to be sure.

---

# the impact on performance

```xml
<source>
  @type sample
  tag test
  size xxx
  rate xxx
</source>
```

## size: 10000, rate: 100

Current

```console
$ top -o RES -p xxx
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1610784 daipom    20   0  273060  61036  11152 S  41.6   0.2   0:10.14 ruby
```

This PR

```console
$ top -o RES -p xxx
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1610963 daipom    20   0  274264  69680  11260 S  57.0   0.2   0:09.73 ruby
```

## size: 100000, rate: 100

Current

```console
$ top -o RES -p xxx
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1611452 daipom    20   0  289000  80492  11072 S 101.0   0.2   0:16.52 ruby
```

This PR

```console
$ top -o RES -p xxx
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1611243 daipom    20   0  304720 148012  11152 S 100.0   0.5   0:11.36 ruby
```

kenhys:
Thanks for checking in point of performance.
It is expected that there is performance regression in some extent.

If user uses in_sample for their own plugin benchmark, it affects because baseline was changed.
But I guess that it is very limited use case.
(It is enough to mention in changelog explicitly IMHO)

daipom:
I'll wait a bit to see if anyone else has an opinion.

kenhys:
Added renew_record parameter like this:

```
<source>
    @type sample
    tag test
    size 100000
    rate 100
    #renew_record false
    renew_record true
</source>
<match test>
  @type null
</match>
```

size: 10000

    current master
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     201025 kenhys    20   0  340940  54800  13648 S  24.6   0.1   0:04.62 ruby

    renew_record: false
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     202342 kenhys    20   0  340948  54732  13576 S  25.7   0.1   0:06.83 ruby

    renew_record: true
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     203402 kenhys    20   0  343288  63556  13500 S  30.7   0.1   0:06.86 ruby

size: 100000

    current master
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     204681 kenhys    20   0  364324  80880  13448 S 100.0   0.1   0:14.46 ruby

    renew_record: false
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     205560 kenhys    20   0  364332  80868  13448 S  99.7   0.1   0:31.34 ruby

    renew_record: true
     PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
     207133 kenhys    20   0  365804 113320  13576 S  99.7   0.2 0:13.87 ruby

With added conditional switch, there is a slightly performance regression with previous behavior.
but it is better to give user a choice which behavior is suitable.

kenhys:
Changed to more conservative approach.

What do you think? @daipom

lib/fluent/plugin/in_sample.rb
```diff
@@ -39,6 +39,8 @@ class SampleInput < Input
     config_param :auto_increment_key, :string, default: nil
     desc "The boolean to suspend-and-resume incremental value after restart"
     config_param :suspend, :bool, default: false,deprecated: 'This parameters is ignored'
+    desc "It force to re-generate a new record by default. You can disable it if filter does not do destructive change"
+    config_param :renew_record, :bool, default: true
```
daipom, line 43: If possible, I'd like to invert the meaning and set the default value to `false`.
It is because Fluentd's `bool` value can be omitted and true/false can be represented just by the presence or absence of a config name.

lib/fluent/plugin/in_sample.rb
```diff
@@ -118,14 +120,22 @@ def emit(num)
     end
 
     def generate
-      d = @sample[@sample_index]
+      if @renew_record
+        d = @sample[@sample_index].dup
+      else
+        d = @sample[@sample_index]
+      end
```
daipom, line 127: The code will be simpler if this can be extracted into another method, such as `sample(index)`.

daipom (commented):
Thanks!

lib/fluent/plugin/in_sample.rb
daipom: `disable_copy`, for example?

kenhys: Thanks, fixed.

kenhys: Oops, I've overlooked your comment.

lib/fluent/plugin/in_sample.rb
```diff
@@ -39,6 +39,8 @@ class SampleInput < Input
     config_param :auto_increment_key, :string, default: nil
     desc "The boolean to suspend-and-resume incremental value after restart"
     config_param :suspend, :bool, default: false,deprecated: 'This parameters is ignored'
+    desc "The boolean to control how to generate a record. You can enable it if filter does not do destructive change"
```
daipom, line 42: It would be easier to understand if this explained what this option does and its purpose.

```suggestion
    desc "Reuse the sample data to reduce the load when sending large amounts of data. You can enable it if filter does not do destructive change"
```

lib/fluent/plugin/in_sample.rb
```diff
@@ -117,15 +119,23 @@ def emit(num)
       end
     end
 
+    def sample
+      if @reuse_record
+        @sample[@sample_index]
+      else
+        @sample[@sample_index].dup
+      end
+    end
```
daipom, line 128: It would be better if the responsibilities of methods were more divided.

In the current design, the `sample` method depends on `@sample_index`, but `@sample_index` is handled by the `genetate` method.
This appears to have unnecessarily increased the mutual dependence of these two methods.

There would be two possible ways to improve this.

* Way 1:  This method handles the index as well.

```rb
    def next_sample
      d = @reuse_record ? @sample[@sample_index] : @sample[@sample_index].dup
      @sample_index += 1
      return d if d

      @sample_index = 0
      next_sample
    end
```

* Way 2: This method does not depend on `@sample_index`.

```rb
    def sample(index)
      if @reuse_record
        @sample[index]
      else
        @sample[index].dup
      end
    end
```

daipom (commented):
Thanks! It's very clear with `reuse_record`!

Sorry for the minor points..., but the following are points of concern.

lib/fluent/plugin/in_sample.rb
```diff
@@ -117,15 +119,23 @@ def emit(num)
...
+      d = sample
       unless d
         @sample_index = 0
-        d = @sample[@sample_index]
+        d =  sample
       end
       @sample_index += 1
       if @auto_increment_key
-        d = d.dup
+        d = d.dup unless @reuse_record
```
daipom, line 138: Should it be inverted?

```suggestion
        d = d.dup if @reuse_record
```

kenhys: It should be, thanks.

daipom (approved):
LGTM. Thanks!

PR #4589, opened 2024-08-15
Add logger gem as dependency for Ruby 3.5

Watson1978:
Fixes #

logger gem was marked as bundled gem for Ruby 3.5. Therefore, it will not work on Ruby 3.5 unless it is added as a dependency.
Ref. https://github.com/ruby/ruby/commit/d7e558e3c48c213d0e8bedca4fb547db55613f7c

(Currently, this patch will suppress `warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.` message.)

Similar with GH-4411

**Docs Changes**:

**Release Note**:


daipom (approved):
LGTM. Thanks!!

Issue #4585, opened 2024-08-09
`warning: literal string will be frozen in the future` since Ruby 3.4.0


Watson1978:
### Describe the bug

`warning: literal string will be frozen in the future` since Ruby 3.4.0

According to https://bugs.ruby-lang.org/issues/20205,
String literals may be frozen by default in the future (maybe Ruby 4.0+).

A transition period will be provided for this change,
the warning is only displayed when Ruby is started with the -w option enabled at present.

In fact, if we run the test in Ruby HEAD, we will encounter many warnings.
https://github.com/fluent/fluentd/actions/runs/10277576606

While the performance gains are attractive,
it will affect all Ruby code used by fluentd if string literals were frozen by default:

- Fluentd
- Dependent gem libraries
- Fluentd plugins

Then, I investigate logs in https://github.com/fluent/fluentd/actions/runs/10277576606,
seems that the only Fluentd main codes encounter the warnings when running tests at present.

```
/home/runner/work/fluentd/fluentd/lib/fluent/command/ctl.rb
/home/runner/work/fluentd/fluentd/lib/fluent/command/plugin_config_formatter.rb
/home/runner/work/fluentd/fluentd/lib/fluent/compat/record_filter_mixin.rb
/home/runner/work/fluentd/fluentd/lib/fluent/config/element.rb
/home/runner/work/fluentd/fluentd/lib/fluent/config/literal_parser.rb
/home/runner/work/fluentd/fluentd/lib/fluent/config/types.rb
/home/runner/work/fluentd/fluentd/lib/fluent/log.rb
/home/runner/work/fluentd/fluentd/lib/fluent/match.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/buffer/file_chunk.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/buffer/memory_chunk.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/compressable.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/formatter_csv.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/formatter_ltsv.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/formatter_out_file.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/in_http.rb
/home/runner/work/fluentd/fluentd/lib/fluent/plugin/in_monitor_agent.rb
[...]

PR #4599, opened 2024-08-16
Add `frozen_string_literal: true` in order to suppress 'warning: literal string will be frozen in the future'

Watson1978:
Fixes #

To supress `warning: literal string will be frozen in the future`,
we will require to fix with following proposals.

1. Add `# frozen_string_literal: true` in the beginning of the file
2. Add `# frozen_string_literal: false` in the beginning of the file
3. Add `RUBYOPT="-disable=frozen_string_literal"` as launch option

These suggestions may be combined.

This PR adds `# frozen_string_literal: true` to all required files.
Seems CI will be passed,
however, Fluentd might raise `FrozenError` exception by unexpected frozen string
if it has not sufficient test suites...

Related to GH-4585

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/command/ctl.rb
+++ b/lib/fluent/command/ctl.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #
 # Fluentd
 #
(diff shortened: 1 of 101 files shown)


ashie:
Thanks for clarifying the issue!
Once Ruby's direction is clearer, we'll determine on our direction.

ashie:
ref:
https://byroot.github.io/ruby/performance/2025/10/28/string-literals.html
https://techracho.bpsinc.jp/hachi8833/2025_11_21/154691

PR #4512, opened 2024-05-29
Backport(v1.16): out_file: add warn message for symlink_path setting (#4502)

daipom:
Backported from 74b2e3d7f86f656d65951c4ea095d68c20b0b858. (#4502)

Issue #4269, opened 2023-08-16
-p overwrites the default plugin path


Bonko:
### Describe the bug

When updating from `fluent/fluentd-kubernetes-daemonset:v1.14.6-debian-logzio-amd64-1.0` to `fluent/fluentd-kubernetes-daemonset:v1.16.2-debian-logzio-amd64-1.0` our setup stopped working.

During investigation it turned out that plugins in the default plugin directory (`/etc/fluent/plugin/`) are not processed anymore when using the `-p` parameter.

### To Reproduce

Add custom pluins into `/etc/fluent/plugin/` directory and ensure that they are used by the config:
```
root@27cf7d7bb6eb:/home/fluent# ls -1  /etc/fluent/plugin/
filter_empty_field.rb
filter_field_size.rb
filter_field_type.rb
filter_message.rb
```

When starting fluentd with the default options from the [Dockerfile command](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.16/debian-logzio/entrypoint.sh) (`fluentd -c /test/fluentd.conf -p /fluentd/plugins/  --gemfile /fluentd/Gemfile`) I am getting this error:
```
2023-08-16 10:20:01 +0000 [error]: config error file="/test/fluentd.conf" error_class=Fluent::NotFoundPluginError error="Unknown filter plugin 'message'. Run 'gem search -rd fluent-plugin' to find plugins"
```
I can get rid of this error by either not specifying `-p` at all:
```
fluentd -c /test/fluentd.conf  --gemfile /fluentd/Gemfile
```
or by adding the default plugin explicitly:
```
fluentd -c /test/fluentd.conf -p /fluentd/plugins/ -p /etc/fluent/plugin  --gemfile /fluentd/Gemfile
```

### Expected behavior

The `-p` parameter should add plugin directories as documented without unsetting the default directory.

### Your Environment

```markdown
- Fluentd version: `1.16.2`
- Operating system: `Debian 11`
- Kernel version: `5.15.49-linuxkit-pr`
```

### Your Configuration

```apache
<label @FLUENT_LOG>
  <match fluent.**>
    @type null
  </match>
</label>

<source>
  @type tail
  @id in_tail_container_logs
  path /var/log/containers/*.log
  tag "kubernetes.*"
  pos_file /var/log/containers/fluentd-containers.log.pos
  read_from_head true
  <parse>
    @type "json"
[...]

PR #4600, opened 2024-08-16
v1.16.6

kenhys:

Fixes #

Bump version to v1.16.6.

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,17 @@
 
 ### Bug Fix
 
-* Fix LoadError with console gem v1.25
-  https://github.com/fluent/fluentd/pull/4510
 * out_file: Add warn message for symlink_path setting
   https://github.com/fluent/fluentd/pull/4512
 * parser_json: Fix wrong LoadError warning
   https://github.com/fluent/fluentd/pull/4592
 * `fluentd` command: Fix `--plugin` (`-p`) option not to overwrite default value
   https://github.com/fluent/fluentd/pull/4605
+
+### Misc
+
+* Keep console gem v1.23 to avoid LoadError
+  https://github.com/fluent/fluentd/pull/4510
 
 ## Release v1.16.5 - 2024/03/27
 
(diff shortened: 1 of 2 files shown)


daipom:
I will work on #4269. If it can be fixed quickly, I'd like to include it.

* #4269

daipom:
> I will work on #4269. If it can be fixed quickly, I'd like to include it.
> 
>     * [-p overwrites the default plugin path #4269](https://github.com/fluent/fluentd/issues/4269)

Done,

daipom:
Rebased to the latest v1.16.
I will add a changelog of #4269.

daipom:
I updated some points.
@kenhys Could you please check?

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # v1.16
 
+## Release v1.16.6 - 2024/08/16
+
+### Bug Fix
+
+* YAML config syntax: Fix issue where `$log_level` element was not supported correctly
+  https://github.com/fluent/fluentd/pull/4486
+* out_file: Add warn message for symlink_path setting
+  https://github.com/fluent/fluentd/pull/4512
```
kenhys, line 10: Should it be Enhancement or Misc category?

daipom: Oh, I see.
Indeed it can be.
It is hard to decide.
Given that it is a Stable branch, I hesitate to add a `Enhancement` element easily, but that might be my overthinking.

#4512 is a measure to deal with an undesirable behavior.
From that perspective, it would be more like `Bug Fix` than `Enhancement`.

However, #4512 did not fix the behavior itself.
From that perspective, it can be `Misc` or `Enhancement`.

What do you think?

kenhys: In this context, Misc may be better.

daipom: OK! Let's put it in `Misc`!

kenhys: In v1.17.1 side, fixed by https://github.com/fluent/fluentd/pull/4607.

daipom: Thanks! I have fixed this side.

daipom (approved):
LGTM. Thanks!

PR #4604, opened 2024-08-19
system_config: add path attribute in system/log section

### LINKED ISSUE #4418: Set log_path in system config

I'm setting up AWX to install FluentD on VMs (not in Containers). Ansible's "ansible.builtin.service" doesn't have command line arguments like "-o" because it's using systemctl. So the logs can only be defaulted to /var/log.

I'd simply like "log_path" to be an attribute I can set in the <system> section of FluentD's config file. It's an attribute that can be set with command line arguments, but not in the config file.

I could modify the service file to include the "-o" command line argument, or have Ansible modify the code directly like in the system_config.rb file and so on so that attribute becomes available. In either case though it would just be a hacky temporary solution.

### Additional context

My /var/log directory is mounted to a low capacity drive because we expected to be able to store all fluent related data in our /opt directory. For now I can set the log rotation very tightly to keep it from filling up, but I think the log path being configurable in the global settings file is a deserving feature that I'm surprised doesn't exist already.

Watson1978:
Fixes #4418

This patch will add `path` attribute in sytem/log section due to configure log file path  by System Configuration.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -518,7 +518,6 @@ module Fluent
       @inline_config = opt[:inline_config]
       @use_v1_config = opt[:use_v1_config]
       @conf_encoding = opt[:conf_encoding]
-      @log_path = opt[:log_path]
       @show_plugin_config = opt[:show_plugin_config]
       @libs = opt[:libs]
       @plugin_dirs = opt[:plugin_dirs]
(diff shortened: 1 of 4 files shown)


ashie (approved):
LGTM

daipom (approved):
LGTM. Thanks!

daipom:
@Watson1978 If you have time, it would be great if you could update the document.

PR #4619, opened 2024-08-28
http_server: Ready to support Async 2.0 gem

Watson1978:
Fixes #

http_server plugin helper doesn't work with Async 2.x gem due to using obsolete usage.
This PR updates it to follow current documented way:
https://github.com/socketry/async-http/blob/0a65acd7cf7486e1877f0da86580e1692cd8207b/readme.md#usage
It's applicable both Async 2.x & Async 1.x.
~~But this PR still stay on Async 1.x because io-event gem which is required by Async 2.x can't build on Windows.~~

- TODO
  - [x] Fix Build error with Ruby 3.1 on Windows
    - https://github.com/socketry/io-event/issues/111
  - [x] Solve https://github.com/socketry/async-http/issues/176 
  - [ ] ~~Replace `cool.io` gem to something~~
    - cool.io introduces IO::Buffer which conflicts with Ruby core classes. It affects the behavior of the Async HTTP server.
    - Ref. https://github.com/socketry/async-http/issues/175
  - [x] https://github.com/socketry/cool.io/pull/82

**Docs Changes**:

**Release Note**:

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -63,6 +63,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency("test-unit", ["~> 3.3"])
   gem.add_development_dependency("test-unit-rr", ["~> 1.0"])
   gem.add_development_dependency("oj", [">= 2.14", "< 4"])
+  gem.add_development_dependency("async", "~> 1.23")
   gem.add_development_dependency("async-http", ">= 0.50.0")
   gem.add_development_dependency("aws-sigv4", ["~> 1.8"])
   gem.add_development_dependency("aws-sdk-core", ["~> 3.191"])
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -63,7 +63,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency("test-unit", ["~> 3.3"])
   gem.add_development_dependency("test-unit-rr", ["~> 1.0"])
   gem.add_development_dependency("oj", [">= 2.14", "< 4"])
-  gem.add_development_dependency("async", "~> 1.23")
   gem.add_development_dependency("async-http", ">= 0.50.0")
   gem.add_development_dependency("aws-sigv4", ["~> 1.8"])
   gem.add_development_dependency("aws-sdk-core", ["~> 3.191"])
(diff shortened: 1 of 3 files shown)


ioquatix:
How are you using `cool.io`? Can you replace it with Async?

ioquatix:
By the way, I'm sorry that this is complicated, I did not intend for this to become a big problem. I'm happy to advise or help out.

ashie:
> How are you using `cool.io`? Can you replace it with Async?

Thanks for your comment!
We have an idea to migrate to Async for quite some time:
https://github.com/fluent/fluentd/wiki/Roadmap-and-Development-tasks#v1
but we haven't started migrating it yet because Fluentd depends on cool.io deeply, it's our fundamental event loop mechanism.

We may need to start considering it now.

ioquatix:
Is it exposed to user code or is it all internal implementation?

ashie:
> Is it exposed to user code or is it all internal implementation?

Partially wrapped by our API but Coolio's classes are exposed to plugins.
Probably many built-in & third-party plugins use them.
e.g.)
https://github.com/fluent/fluentd/blob/7f13c7a6833f46523bf20356844147021f027b1c/lib/fluent/plugin/in_tail.rb#L821-L849

ioquatix:
I see, it's a tricky change then.

One option would be to change `cool.io` to rename `IO::Buffer` to it's own namespace, e.g. `cool.io::IOBuffer` (which would make more sense anyway).

Watson1978:
Rebased on master in order to run CI with `cool.io` gem v1.9.0
Related to https://github.com/socketry/cool.io/pull/82#issuecomment-2388000360

ioquatix:
Ah, we also need to fix `io-event` compiling on Windows, I'll sort it out.

Watson1978:
> Ah, we also need to fix `io-event` compiling on Windows, I'll sort it out.

@ioquatix If you have a time, I want you will take an action for https://github.com/socketry/io-event/pull/112 🙏🏻

ashie (approved):
I've tried it with in_monitor_agent, it works fine for me.
Thanks a lot!

ioquatix:
Awesome!

PR #4620, opened 2024-08-29
filter_parser: Transition from single record to stream

### LINKED ISSUE #4100: JSON parser plugin can sometimes output records as strings rather than `Hash`

Athishpranav2003:
Fixes #4100

Transition `filter_parser` from using `filter_with_time` to `filter_stream` logic
**Docs Changes**:
N/A
**Release Note**: 
N/A

### DIFF
--- a/lib/fluent/plugin/filter_parser.rb
+++ b/lib/fluent/plugin/filter_parser.rb
@@ -54,29 +54,32 @@ module Fluent::Plugin
       @parser = parser_create
     end
 
-    FAILED_RESULT = [nil, nil].freeze # reduce allocation cost
     REPLACE_CHAR = '?'.freeze
 
-    def filter_with_time(tag, time, record)
-      raw_value = @accessor.call(record)
-      if raw_value.nil?
-        if @emit_invalid_record_to_error
-          router.emit_error_event(tag, time, record, ArgumentError.new("#{@key_name} does not exist"))
-        end
-        if @reserve_data
-          return time, handle_parsed(tag, record, time, {})
-        else
-          return FAILED_RESULT
+    def filter_stream(tag, es)
+      new_es = Fluent::MultiEventStream.new
+      es.each do |time, record|
+        begin
+          raw_value = @accessor.call(record)
+          if raw_value.nil?
+            new_es.add(time, handle_parsed(tag, record, time, {})) if @reserve_data
+            raise ArgumentError, "#{@key_name} does not exist"
+          else
+            filter_one_record(tag, time, record, raw_value) do |result_time, result_record|
+              new_es.add(result_time, result_record)
+            end
+          end
+        rescue => e
+          router.emit_error_event(tag, time, record, e) if @emit_invalid_record_to_error
         end
       end
-      begin
-        # Note: https://github.com/fluent/fluentd/issues/4100
-        # If the parser returns multiple records from one raw_value,
-        # this returns only the first one record.
-        # This should be fixed in the future version.
-        result_time = nil
-        result_record = nil
+      new_es
+    end
 
+    private
+
+    def filter_one_record(tag, time, record, raw_value)
+      begin
         @parser.parse(raw_value) do |t, values|
           if values
             t = if @reserve_time
@@ -85,38 +88,17 @@ module Fluent::Plugin
                   t.nil? ? time : t
                 end
             @accessor.delete(record) if @remove_key_name_field
-            r = handle_parsed(tag, record, t, values)
-
-            if result_record.nil?
-              result_time = t
-              result_record = r
-            else
-              if @emit_invalid_record_to_error
-                router.emit_error_event(tag, t, r, Fluent::Plugin::Parser::ParserError.new(
-                  "Could not emit the event. The parser returned multiple results, but currently filter_parser plugin only returns the first parsed result. Raw data: '#{raw_value}'"
-                ))
-              end
-            end
           else
-            if @emit_invalid_record_to_error
-              router.emit_error_event(tag, time, record, Fluent::Plugin::Parser::ParserError.new("pattern not matched with data '#{raw_value}'"))
-            end
-
+            router.emit_error_event(tag, time, record, Fluent::Plugin::Parser::ParserError.new("pattern not matched with data '#{raw_value}'")) if @emit_invalid_record_to_error
             next unless @reserve_data
-            next unless result_record.nil?
-
-            result_time = time
-            result_record = handle_parsed(tag, record, time, {})
+            t = time
+            values = {}
           end
+          yield(t, handle_parsed(tag, record, t, values))
         end
 
-        return result_time, result_record
       rescue Fluent::Plugin::Parser::ParserError => e
-        if @emit_invalid_record_to_error
-          raise e
-        else
-          return FAILED_RESULT
-        end
+        raise e
       rescue ArgumentError => e
         raise unless @replace_invalid_sequence
         raise unless e.message.index("invalid byte sequence in") == 0
@@ -124,15 +106,9 @@ module Fluent::Plugin
         raw_value = raw_value.scrub(REPLACE_CHAR)
         retry
       rescue => e
-        if @emit_invalid_record_to_error
-          raise Fluent::Plugin::Parser::ParserError, "parse failed #{e.message}"
-        else
-          return FAILED_RESULT
-        end
+        raise Fluent::Plugin::Parser::ParserError, "parse failed #{e.message}"
       end
     end
-
-    private
 
     def handle_parsed(tag, record, t, values)
       if values && @inject_key_prefix
--- a/test/plugin/test_filter_parser.rb
+++ b/test/plugin/test_filter_parser.rb
@@ -206,6 +206,23 @@ class ParserFilterTest < Test::Unit::TestCase
 
   end
 
+  def test_filter_with_multiple_records
+    d1 = create_driver(%[
+      key_name data
+      <parse>
+        @type json
+      </parse>
+    ])
+    time = Fluent::EventTime.from_time(@default_time)
+    d1.run(default_tag: @tag) do
+      d1.feed(time, {'data' => '[{"xxx_1":"first","yyy":"second"}, {"xxx_2":"first", "yyy_2":"second"}]'})
+    end
+    filtered = d1.filtered
+    assert_equal 2, filtered.length
+    assert_equal ({"xxx_1"=>"first", "yyy"=>"second"}), filtered[0][1]
+    assert_equal ({"xxx_2"=>"first", "yyy_2"=>"second"}), filtered[1][1]
+  end
+
   data(:keep_key_name => false,
        :remove_key_name => true)
   def test_filter_with_reserved_data(remove_key_name)


Athishpranav2003:
@daipom i have raised this as a draft. This needs other changes like changing the tests around this, exception handling and etc. 

For now lets have discussion on the base logic and in meantime i will add them once we fix it

daipom:
Thanks! I will see this this weekend!

Athishpranav2003:
@daipom thank you. The tests needs to be revamped for filter parser since the tests involve few instance variables. Do check it in freetime and let me know

Athishpranav2003:
@daipom any opinion on this logic? Should we go ahead with this?

daipom:
@Athishpranav2003 Sorry for my late response.
I got COVID-19 last weekend and only got a little work done this week :cry: 
I will see this now, please wait a few more days :cry:

Athishpranav2003:
Take care @daipom 

Cool, will work on it once overall design is finalised

lib/fluent/plugin/filter_parser.rb
```diff
@@ -57,78 +57,74 @@ def configure(conf)
...
+                  result_time = t
+                  result_record = r
                 else
-                  t.nil? ? time : t
+                  if @emit_invalid_record_to_error
+                    router.emit_error_event(tag, t, r, Fluent::Plugin::Parser::ParserError.new(
+                      "Could not emit the event. The parser returned multiple results, but currently filter_parser plugin only returns the first parsed result. Raw data: '#{raw_value}'"
+                    ))
+                  end
                 end
```
daipom, line 97: We don't need this logic anymore!

daipom (commented):
@Athishpranav2003 Sorry for my late response.
Basically, it looks good to me! Thanks for this fix!

We would like to address the tests while considering the following points.

* The behavior of `@emit_invalid_record_to_error`.
  * The reason why the old code raises errors would be that the following base class logic handled the errors.
    * https://github.com/fluent/fluentd/blob/919e91b6a9251ac227ac62a778f58cae25bdb2ab/lib/fluent/plugin/filter.rb#L86-L91
  * In the new code, we would need to handle it on the plugin side.
* Code readability.
  * It appears that adding some inner methods would improve code readability.
  * For example ...
    ```rb
    def filter_stream(tag, es)
      new_es = Fluent::MultiEventStream.new
      es.each do |time, record|
        begin
          filter_one_record(tag, time, record) do |filtered_time, filtered_record|
            new_es.add(filtered_time, filtered_record)
          end
        rescue => e
          router.emit_error_event(tag, time, record, e)
        end
      end
      new_es
    end

    def filter_one_record(tag, time, record)
      ...
      yield result_time, result_record
      ...
    end
    ```

daipom:
> It appears that adding some inner methods would improve code readability.
> For example ... 

I guess we need to think a little better about handling error events...

Athishpranav2003:
@daipom yah
I wanted to make sure we are in same page. I will refactor the code and change the error part after that

daipom:
> @daipom yah I wanted to make sure we are in same page. I will refactor the code and change the error part after that

Thanks! This would be the right direction!

Athishpranav2003:
@daipom i have done the changes. 
I will go ahead and few UTs alone for this case meanwhile

Athishpranav2003:
@daipom you can check this now

Athishpranav2003:
@daipom I guess the CI test fail is not related to this change

daipom:
Thanks!! I will see this!

> I guess the CI test fail is not related to this change

Yes, They are not related.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -85,55 +91,33 @@ def filter_with_time(tag, time, record)
...
       rescue Fluent::Plugin::Parser::ParserError => e
         if @emit_invalid_record_to_error
-          raise e
-        else
-          return FAILED_RESULT
+          router.emit_error_event(tag, time, record, e)
         end
       rescue ArgumentError => e
-        raise unless @replace_invalid_sequence
-        raise unless e.message.index("invalid byte sequence in") == 0
```
daipom, line 122:
```rb
raise unless e.message.index("invalid byte sequence in") == 0
```

There appears to be no corresponding process for this.

lib/fluent/plugin/filter_parser.rb
```diff
@@ -85,55 +91,33 @@ def filter_with_time(tag, time, record)
...
-
-        raw_value = raw_value.scrub(REPLACE_CHAR)
-        retry
-      rescue => e
         if @emit_invalid_record_to_error
-          raise Fluent::Plugin::Parser::ParserError, "parse failed #{e.message}"
```
daipom, line 128: There appears to be no corresponding process for this.

test/plugin/test_filter_parser.rb
```diff
@@ -211,6 +211,24 @@ def test_filter
 
   data(:keep_key_name => false,
        :remove_key_name => true)
```
daipom, line 213: This `data` should be before `test_filter_with_reserved_data()`.

daipom (changes requested):
Great! Thanks!
Looks good to me for normal cases.

I'm concerned about some abnormal cases.
The behavior of some of `rescue` cases appears to have changed.
I would like to have it checked.

I would be very happy if you could add some automated tests for the `rescue` cases if possible, as there seems to be a lack of those tests.

Athishpranav2003:
@daipom 
https://github.com/fluent/fluentd/blob/51b860b1be2eb59076d706fda12d55657a614c8f/test/plugin/test_filter_parser.rb#L659-L666
could you please point which emit_event is this call part of

Athishpranav2003:
After doing some verification 
https://github.com/fluent/fluentd/blob/51b860b1be2eb59076d706fda12d55657a614c8f/lib/fluent/plugin/filter.rb#L82-L104

The emit error event is in line 90/99. This implies that earlier we raise the exception in filter one record and capture in the filter stream. Now since we have implemented stream directly so the raise not needed
```
Failure: test_call_emit_error_event_when_parser_error(ParserFilterTest):
  in mock 'flexmock(Fluent::Test::Driver::TestEventRouter)': no matching handler found for emit_error_event("test", 1272942121, {"data"=>"{\"time\":[], \"f1\":\"v1\"}"}, #<Fluent::Plugin::Parser::ParserError: value must be a string or a number: [](Array)>)
  Defined expectations:
    should_receive(:emit_error_event).with("String", Integer, Hash, Fluent::Plugin::Parser::ParserError).once.
  <false> is not true.
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/test_unit_integration.rb:57:in `make_assertion'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/test_unit_integration.rb:65:in `check'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/core_class_methods.rb:104:in `check'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/expectation_director.rb:41:in `call'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/core.rb:150:in `block in method_missing'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/core.rb:289:in `flexmock_wrap'
/home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/core.rb:146:in `method_missing'
(eval at /home/aggressive_racer1/projects/fluentd/vendor/bundle/ruby/3.3.0/gems/flexmock-2.3.8/lib/flexmock/partial_mock.rb:429):3:in `emit_error_event'
/home/aggressive_racer1/projects/fluentd/lib/fluent/plugin/filter.rb:90:in `rescue in block in filter_stream'
/home/aggressive_racer1/projects/fluentd/lib/fluent/plugin/filter.rb:86:in `block in filter_stream'
/home/aggressive_racer1/projects/fluentd/lib/fluent/event.rb:110:in `each'
/home/aggressive_racer1/projects/fluentd/lib/fluent/plugin/filter.rb:85:in `filter_stream'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/filter.rb:49:in `block (2 levels) in instance_hook_after_started'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/event_feeder.rb:43:in `feed_to_plugin'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/event_feeder.rb:87:in `feed'
/home/aggressive_racer1/projects/fluentd/test/plugin/test_filter_parser.rb:664:in `block in test_call_emit_error_event_when_parser_error'
     661:     flexmock(d.instance.router).should_receive(:emit_error_event).
     662:       with("String", Integer, Hash, ParserError).once
     663:     d.run do
  => 664:       d.feed(@tag, Fluent::EventTime.now.to_i, {'data' => '{"time":[], "f1":"v1"}'})
     665:     end
     666:   end
     667: 
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/base.rb:204:in `block in run_actual'
/usr/share/ruby/timeout.rb:186:in `block in timeout'
/usr/share/ruby/timeout.rb:41:in `handle_timeout'
/usr/share/ruby/timeout.rb:195:in `timeout'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/base.rb:203:in `run_actual'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/base.rb:96:in `run'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/base_owner.rb:130:in `run'
/home/aggressive_racer1/projects/fluentd/lib/fluent/test/driver/event_feeder.rb:39:in `run'
/home/aggressive_racer1/projects/fluentd/test/plugin/test_filter_parser.rb:663:in `test_call_emit_error_event_when_parser_error'
```
I used the following jugad way to find the function call

daipom:
> Now since we have implemented stream directly so the raise not needed

Yes, we don't need `raise` in the new implementation.
However, `router.emit_error_event` should be called as before.

daipom:
> https://github.com/fluent/fluentd/blob/51b860b1be2eb59076d706fda12d55657a614c8f/test/plugin/test_filter_parser.rb#L659-L666
> 
> could you please point which emit_event is this call part of

In the current implementation:

https://github.com/fluent/fluentd/blob/78a7972bfe6b421f08472701f04f00515ed24bee/lib/fluent/plugin/filter_parser.rb#L114-L119

In the new implementation:

https://github.com/fluent/fluentd/blob/9ec5a95c964cb2eb6340b7538e25c8547d4bbaec/lib/fluent/plugin/filter_parser.rb#L105-L108

This change is OK because `emit_error_event` is called similarly after all.

Athishpranav2003:
@daipom 
i am not sure what is this issue
```
Failure: test_filter_key_not_exist(ParserFilterTest):
  in mock 'flexmock(Fluent::Test::Driver::TestEventRouter)': no matching handler found for emit_error_event("test", 1272942121, {"foo"=>"bar"}, #<ArgumentError: data does not exist>)
  Defined expectations:
    should_receive(:emit_error_event).with(String, Integer, Hash, #<ArgumentError: data does not exist>).once.
  <false> is not true.
```

daipom:
> @daipom i am not sure what is this issue
> 
> ```
> Failure: test_filter_key_not_exist(ParserFilterTest):
>   in mock 'flexmock(Fluent::Test::Driver::TestEventRouter)': no matching handler found for emit_error_event("test", 1272942121, {"foo"=>"bar"}, #<ArgumentError: data does not exist>)
>   Defined expectations:
>     should_receive(:emit_error_event).with(String, Integer, Hash, #<ArgumentError: data does not exist>).once.
>   <false> is not true.
> ```

I see. This test code specifies the content of expected arguments too strictly.
It appears that this requires all values of the instance match, but the stacktrace data can be different.
We should soften the conditions.

```diff
diff --git a/test/plugin/test_filter_parser.rb b/test/plugin/test_filter_parser.rb
index ec36c836..4201cec8 100644
--- a/test/plugin/test_filter_parser.rb
+++ b/test/plugin/test_filter_parser.rb
@@ -650,7 +650,7 @@ class ParserFilterTest < Test::Unit::TestCase
   def test_filter_key_not_exist
     d = create_driver(CONFIG_NOT_IGNORE)
     flexmock(d.instance.router).should_receive(:emit_error_event).
-      with(String, Integer, Hash, ArgumentError.new("data does not exist")).once
+      with(String, Integer, Hash, ArgumentError).once
     assert_nothing_raised {
       d.run do
         d.feed(@tag, Fluent::EventTime.now.to_i, {'foo' => 'bar'})
```

lib/fluent/plugin/filter_parser.rb
```diff
@@ -57,26 +57,30 @@ def configure(conf)
     FAILED_RESULT = [nil, nil].freeze # reduce allocation cost
```
daipom, line 57:
```suggestion
```

lib/fluent/plugin/filter_parser.rb
```diff
@@ -57,26 +57,30 @@ def configure(conf)
...
-        else
-          return FAILED_RESULT
+    def filter_stream(tag, es)
+      new_es = Fluent::MultiEventStream.new
+      es.each do |time, record|
+        begin
+          raw_value = @accessor.call(record)
+          if raw_value.nil?
+            new_es.add(time, handle_parsed(tag, record, time, {})) if @reserve_data
+            router.emit_error_event(tag,time,record,ArgumentError.new("#{@key_name} does not exist")) if @emit_invalid_record_to_error
```
daipom, line 67:
```suggestion
            raise ArgumentError, "#{@key_name} does not exist"
```

lib/fluent/plugin/filter_parser.rb
```diff
@@ -57,26 +57,30 @@ def configure(conf)
...
+          else
+            filter_one_record(tag, time, record, raw_value) do |result_time, result_record|
+              new_es.add(result_time, result_record) if result_record
```
daipom, line 70:
```suggestion
              new_es.add(result_time, result_record)
```

lib/fluent/plugin/filter_parser.rb
```diff
@@ -57,26 +57,30 @@ def configure(conf)
...
+            end
+          end
+        rescue => e
+          router.emit_error_event(tag, time, record, e)
```
daipom, line 74:
```suggestion
          router.emit_error_event(tag, time, record, e) if @emit_invalid_record_to_error
```

lib/fluent/plugin/filter_parser.rb
```diff
@@ -85,32 +89,15 @@ def filter_with_time(tag, time, record)
...
+            t = time
+            values = {}
           end
+          yield(t, handle_parsed(tag, record, t, values))
         end
 
-        return result_time, result_record
       rescue Fluent::Plugin::Parser::ParserError => e
         if @emit_invalid_record_to_error
           raise e
```
daipom, line 103: Looks to me this would be enough.
(We wouldn't need `FAILED_RESULT`.)

```ruby
      rescue Fluent::Plugin::Parser::ParserError
        raise
      rescue ArgumentError => e
        raise unless @replace_invalid_sequence
        raise unless e.message.index("invalid byte sequence in") == 0

        raw_value = raw_value.scrub(REPLACE_CHAR)
        retry
      rescue => e
        raise Fluent::Plugin::Parser::ParserError, "parse failed #{e.message}"
      end
```

test/plugin/test_filter_parser.rb
daipom, line 1: We need to soften the condition for the above code change.

```diff
diff --git a/test/plugin/test_filter_parser.rb b/test/plugin/test_filter_parser.rb
index ec36c836..4201cec8 100644
--- a/test/plugin/test_filter_parser.rb
+++ b/test/plugin/test_filter_parser.rb
@@ -650,7 +650,7 @@ class ParserFilterTest < Test::Unit::TestCase
   def test_filter_key_not_exist
     d = create_driver(CONFIG_NOT_IGNORE)
     flexmock(d.instance.router).should_receive(:emit_error_event).
-      with(String, Integer, Hash, ArgumentError.new("data does not exist")).once
+      with(String, Integer, Hash, ArgumentError).once
     assert_nothing_raised {
       d.run do
         d.feed(@tag, Fluent::EventTime.now.to_i, {'foo' => 'bar'})
```

daipom (changes requested):
Thanks!
I think it could be simpler, so I would suggest the following code change.

lib/fluent/plugin/filter_parser.rb
Athishpranav2003: This is failing UTs @daipom, I have added the output in the prev comment. Can you check

lib/fluent/plugin/filter_parser.rb
Athishpranav2003: Actually @daipom . The previous behaviour was we raise exception if emit error to event. Here also I followed the same. So your change will not have impact and changes diff structure. Isnit fine if it stays same

test/plugin/test_filter_parser.rb
Athishpranav2003: Oh I see
Got the problem
Now I can raise the exception itself

lib/fluent/plugin/filter_parser.rb
Athishpranav2003: Are yo suggesting to raise exception but only emit in rescue?
Fine will shift to that convention

Athishpranav2003:
Got it @daipom 
Will make the changes in sometime

daipom:
Thanks!

daipom (commented):
Thanks!! The code change looks good to me!
I'd like to confirm the behavior a little more.
We would need more tests of anomaly scenarios.
I will add some tests later.

daipom:
I made another PR to add some test cases for abnormal cases.

* #4638

Let's rebase after it is merged.

I confirmed the test results on my local for this PR.
Basically, there is no problem, but a few tests fail.
It is because this PR removes the following `rescue`, and it changes the error message a little bit.

https://github.com/fluent/fluentd/blob/51b860b1be2eb59076d706fda12d55657a614c8f/lib/fluent/plugin/filter_parser.rb#L114-L119

With this, I believe we should stop removing this `rescue`, even if this is redundant.
It is preferable to keep the error message as unchanged as possible because it could cause a disadvantage to existing users.

The result of #4638 tests applied to this PR:

```
Failure: test_parser_error[invalid format with default](ParserFilterTest::abnormal cases)
/home/daipom/work/fluentd/fluentd/test/plugin/test_filter_parser.rb:546:in `run_and_assert'
/home/daipom/work/fluentd/fluentd/test/plugin/test_filter_parser.rb:691:in `test_parser_error'
     688:         </parse>
     689:       EOC
     690: 
  => 691:       run_and_assert(driver, **data.except(:additional_config))
     692:     end
     693: 
     694:     data(
<[[],
 [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
 [[Fluent::Plugin::Parser::ParserError,
   "value must be a string or a number: [](Array)"]]]> expected but was
<[[],
 [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
 [[Fluent::Plugin::Parser::ParserError,
   "parse failed value must be a string or a number: [](Array)"]]]>

diff:
  [[],
   [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
   [[Fluent::Plugin::Parser::ParserError,
?    "parse failed value must be a string or a number: [](Array)"]]]
======================================================================================================
F
======================================================================================================
Failure: test_parser_error[mixed valid and invalid with default](ParserFilterTest::abnormal cases)
/home/daipom/work/fluentd/fluentd/test/plugin/test_filter_parser.rb:546:in `run_and_assert'
/home/daipom/work/fluentd/fluentd/test/plugin/test_filter_parser.rb:691:in `test_parser_error'
     688:         </parse>
     689:       EOC
     690: 
  => 691:       run_and_assert(driver, **data.except(:additional_config))
     692:     end
     693: 
     694:     data(
<[[{"f1"=>"v1"}],
 [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
 [[Fluent::Plugin::Parser::ParserError,
   "value must be a string or a number: [](Array)"]]]> expected but was
<[[{"f1"=>"v1"}],
 [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
 [[Fluent::Plugin::Parser::ParserError,
   "parse failed value must be a string or a number: [](Array)"]]]>

diff:
  [[{"f1"=>"v1"}],
   [{"data"=>"{\"time\":[], \"f1\":\"v1\"}"}],
   [[Fluent::Plugin::Parser::ParserError,
?    "parse failed value must be a string or a number: [](Array)"]]]
```

Athishpranav2003:
@daipom thanks for this help. I had difficulty in figuring out the edgecases. Thanks for doing it along with this to make it move ahead

Athishpranav2003:
@daipom i have reverted the change with redundant rescue. Can you point out the change in the error messages if its still present in this pr?

daipom (dismissed):
Thanks! I have confirmed this change passes tests on #4638.
I believe we are now ready to merge! (after #4638)
Thanks so much for this fix!

daipom:
@ashie @kenhys 
Which version should we release this on?

This solves the limitation of #4474, which is released on v1.17.0.
It's ambiguous whether it's a bug fix or an enhancement.

I think it is possible to release this on v1.17.
On the other hand, since this is a large fix for `filter_parser`, 1.18 may be safer, given the risk of regression.

kenhys:
> On the other hand, since this is a large fix for filter_parser, 1.18 may be safer, given the risk of regression.

I'll second for v1.18.

daipom:
> Thanks!
> I have confirmed this change passes tests on #4638.
> I believe we are now ready to merge! (after #4638) Thanks so much for this fix!

Now #4638 is merged.
So, I rebased this.
After checking CI, I merge this.
Thanks!

kenhys (approved):
LGTM.

daipom (approved):
LGTM.

Issue #3549, opened 2021-11-02
gracefulReloads are leaking memory again


alex-vmw:
### Describe the bug

Sometime ago I have filed https://github.com/fluent/fluentd/issues/3342, which was determined to be an issue with the fluentd-systemd plugin. However, we are now running with the fixed version of the fluentd-systemd plugin, but seeing a similar issue where gracefulReloads are leaking memory.

Issued 10 gracefulReloads in ~30 minutes.
```
root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 22:53:32 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 22:56:13 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 22:59:22 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:03:04 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:07:53 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:11:11 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:14:58 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:18:06 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:21:59 UTC 2021
{"ok":true}root [ / ]# date; curl http://127.0.0.1:24444/api/config.gracefulReload
Tue Nov  2 23:25:11 UTC 2021
{"ok":true}root [ / ]#
```

Observed that fluentd container memory went from 153MB to 283MB (85% increase) after 10 gracefulReloads.
![image](https://user-images.githubusercontent.com/36181540/139965700-30c7f1d2-90fb-4ed5-aae5-b4d7e5d8bf3a.png)

### To Reproduce

Issue multiple gracefulReloads and observe fluentd memory utilization.

### Expected behavior

[...]

Issue #3469, opened 2021-07-26
The reload configuration with the worker tag failed


xidiandb:
### Describe the bug

I used a configuration with the worker tag, which runs on startup but not on reload.

### To Reproduce

  Use my configuration to start and reload

### Expected behavior

fluent/log.rb:371:error: Failed to reload config file: specified worker_id<0> collisions is detected on <worker> directive. Available worker id(s): []

### Your Environment

```markdown
- Fluentd version: 1.11.4
- TD Agent version:
- Operating system: ubuntu 18.04
- Kernel version:5.4.61-050461-generic
```

### Your Configuration

    <system>
      workers 1
      rpc_endpoint "#{ENV['POD_IP']}:24444"
    </system>
    <label @FLUENT_LOG>
    <match fluent.*>
      @type null
    </match>
    </label>   
    <worker 0>
    <source>
      @type sample
      sample {"hello borg ooo":"world"}
      rate 1
      tag sample.ni.hao
    </source>
    </worker>
    <worker 0>
    <match sample.*.*>
      @type stdout
    </match>
    </worker>

### Your Error Log

```shell
fluent/log.rb:371:error: Failed to reload config file: specified worker_id<0> collisions is detected on <worker> directive. Available worker id(s): []
```

kenhys:
Use the following, do not define multiple `<worker 0>`.

```
<system>
  workers 1
  rpc_endpoint "#{ENV['POD_IP']}:24444"
</system>
<label @FLUENT_LOG>
  <match fluent.*>
    @type null
  </match>
</label>   
<worker 0>
  <source>
    @type sample
    sample {"hello borg ooo":"world"}
    rate 1
    tag sample.ni.hao
  </source>
  <match sample.*.*>
    @type stdout
  </match>
</worker>
```

xidiandb:
@kenhys  But the first time it started, there was no problem, only on reload ，Moreover, my configuration is very complex, which is divided into multiple files, and some configurations cannot be written to a worker tag. I would like to know if this boot is ok and if it is a bug when it reloads

xidiandb:
I want to add configuration dynamically by adding files, each with worker tags. Is that not supportive? But why is there no problem at startup, only on reload

kenhys:
Hmm, I've overlooked it.

kenhys:
https://github.com/fluent/fluentd/blob/master/lib/fluent/supervisor.rb#L290-L303
[...]

Issue #2259, opened 2019-01-09
graceful shutdown looks like lose the received events


HatsuneMiku3939:
- fluentd or td-agent version.
fluentd 1.3.3

- Environment information
linux
ruby 2.5.1

- configuration
```
  <source>
    @type forward
  </source>
  <match test>
    @type file
    path "/tmp/fluent-shutdown-test/output/${tag}"
    append true
    <buffer tag>
      @type "memory"
      path /tmp/fluent-shutdown-test/output/${tag}
    </buffer>
  </match>
```

* test code - https://github.com/HatsuneMiku3939/fluent-shutdown-test/blob/master/test.rb
  * https://travis-ci.org/HatsuneMiku3939/fluent-shutdown-test/builds/477185717

graceful shutdown immediately after post a large number of events looks like lose received events.

- case 1
post 10 events -> graceful shutdown immediately -> file output contains 10 events

- case 2
post 100000 events -> graceful shutdown immediately -> file output doesn't contains 100000 events

- desired behavior of case 2
post 100000 events -> graceful shutdown immediately -> file output contains 100000 events

PR #4624, opened 2024-08-30
SIGUSR2:  zero downtime restart

### LINKED ISSUE #4622: Update/Reload without downtime

Updating Fluentd or reloading a config causes downtime.
Plugins that receive data as a server, such as `in_udp`, `in_tcp`, and `in_syslog`, cannot receive data during this time.
This means that the data sent by a client is lost during this time unless the client has a re-sending feature.
This makes updating Fluentd or reloading a config difficult in some cases.

Add a new feature: Update/Reload without downtime.

For example, implement a mechanism similar to nginx's feature for upgrading on the fly.

* http://nginx.org/en/docs/control.html#upgrade

The main problem is that Fluentd can't run in parallel with the same config.
(It causes some conflicts, such as buffer files)

Because of this problem, it is very difficult to support all plugins.
However, it is possible to support only plugins that can run in parallel.

Based on the above, the following mechanism would be a good way to achieve this.

1. The current supervisor receives a signal.
1. The current supervisor sends signals to its workers, and the workers stop all plugins that cannot run in parallel.
1. The current supervisor starts a new supervisor.
   * => Old processes and new processes run in parallel.
1. After the new supervisor and its workers start to work, the current supervisor and its workers stop.

More specifically, it would be better to run only limited Input plugins in parallel, such as `in_tcp`, `in_udp`, and `in_syslog`.
Stop all plugins except those Input plugins, and prepare a dedicated file buffer for Output.
After the new workers start, they load the file buffer and route those events to the `@ROOT` label.

None.

### Additional context

I have already started to create a PoC.

* https://github.com/fluent/fluentd/pull/4624

daipom:
* Fixes #4622

This replaces the current `SIGUSR2` (#2716) with the new feature.
(Not supported on Windows).

* Restart the new process with zero downtime

The primary motivation is to enable the update of Fluentd without data loss of plugins such as `in_udp`.
(Please See #4622).

Specification:

* 2 ways to trigger this feature (non-Windows):
  * [Signal](https://docs.fluentd.org/deployment/signals): `SIGUSR2` to the supervisor.
    * Sending `SIGUSR2` to the workers triggers [the traditional GracefulReload](https://github.com/fluent/fluentd/pull/2716).
      * (Leave the traditional way, just in case)
  * [RPC](https://docs.fluentd.org/deployment/rpc): `/api/processes.zeroDowntimeRestart`
    * Leave `/api/config.gracefulReload` for the traditional feature.
* This starts the new supervisor and workers with zero downtime for some plugins.
  * Input plugins with `zero_downtime_restart` supported work in parallel. 
    * Supported input plugins:
      * `in_tcp`
      * `in_udp`
      * `in_syslog`
  * The old processes stop after 10s.
* The new supervisor works in `source-only` mode (#4661) until the old processes stop.
  * After the old processes stop, the data handled by the new processes are loaded and processed.
  * If need, you can configure `source_only_buffer` (see #4661).
* Windows: Not affected at all. Remains [the traditional GracefulReload](https://github.com/fluent/fluentd/pull/2716).

# Mechanism

1. The supervisor receives SIGUSR2.
2. Spawn a new supervisor.
3. Take over shared sockets.
4. Launch new workers, and stop old processes in parallel.
   * Launch new workers with source-only mode
     * Limit to zero_downtime_restart_ready? input plugin
   * Send SIGTERM to the old supervisor after 10s delay from 3.
5. The old supervisor stops and sends SIGWINCH to the new one.
6. The new workers run fully.

375586975-1ec2b716-03c3-4db6-90c0-f9e7e8a14a17

Needs following:

* #4661
* https://github.com/treasure-data/serverengine/pull/150

Conditions under which `zero_downtime_restart_ready?` can be enabled:

* Must be able to work in parallel with another Fluentd instance.
* Notes:
  * The sockets provided by [server helper](https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-server) are shared with the new Fluentd instance.
  * Input plugins managing a position such as `in_tail` should not enable its `zero_downtime_restart_ready?`.
    * Such input plugins do not cause data loss on restart, so there is no need to enable this in the first place.
  * `in_http` and `in_forward` could also be supported. Not supporting them this time is simply a matter of time to consider.

# The appropriateness of replacing the traditional SIGUSR2

There are the following reasons:

* The traditional SIGUSR2 feature has some limitations and issues.
  * Limitations:
    > it has 2 limitations.
    > 1. A change to system_config is ignored because it needs to restart(kill/spawn) process.
    > 1. All plugins must not use class variable when restarting.
  * Issues:
    * #2259
    * #3469
    * #3549
* This new feature allows restarts without downtime and such limitations.
  * Although supported plugins are limited, that is not a problem for many plugins.
    (The problem is with server-based input plugins where the stop results in data loss).
* This new feature has a big advantage that it can also be used to update Fluentd.
  * In the future, fluent-package will use this feature to allow update with zero downtime by default.
    * https://github.com/fluent/fluent-package-builder/pull/713
* If needed, we can still use the traditional feature by RPC or directly sending `SIGUSR2` to the workers.

# ETC

**Docs Changes**:
TODO

**Release Note**: 
* Add zero-downtime-restart feature for non-Windows (`USR2` signal and `/api/processes.zeroDowntimeRestart` RPC API)

**TODO**:

- [x] Some implementation TODO referred in code comment.
- [x] Tests
- [ ] Document

### DIFF
--- a/lib/fluent/plugin/in_syslog.rb
+++ b/lib/fluent/plugin/in_syslog.rb
@@ -156,6 +156,10 @@ module Fluent::Plugin
       true
     end
 
+    def zero_downtime_restart_ready?
+      true
+    end
+
     def start
       super
 
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -48,7 +48,35 @@ module Fluent
   class RootAgent < Agent
     ERROR_LABEL = "@ERROR".freeze # @ERROR is built-in error label
 
-    def initialize(log:, system_config: SystemConfig.new)
+    class SourceOnlyMode
+      DISABLED = 0
+      NORMAL = 1
+      ONLY_ZERO_DOWNTIME_RESTART_READY = 2
+
+      def initialize(with_source_only, start_in_parallel)
+        if start_in_parallel
+          @mode = ONLY_ZERO_DOWNTIME_RESTART_READY
+        elsif with_source_only
+          @mode = NORMAL
+        else
+          @mode = DISABLED
+        end
+      end
+
+      def enabled?
+        @mode != DISABLED
+      end
+
+      def only_zero_downtime_restart_ready?
+        @mode == ONLY_ZERO_DOWNTIME_RESTART_READY
+      end
+
+      def disable!
+        @mode = DISABLED
+      end
+    end
+
+    def initialize(log:, system_config: SystemConfig.new, start_in_parallel: false)
       super(log: log)
 
       @labels = {}
@@ -56,7 +84,7 @@ module Fluent
       @suppress_emit_error_log_interval = 0
       @next_emit_error_log_time = nil
       @without_source = system_config.without_source || false
-      @with_source_only = system_config.with_source_only || false
+      @source_only_mode = SourceOnlyMode.new(system_config.with_source_only, start_in_parallel)
       @source_only_buffer_agent = nil
       @enable_input_metrics = system_config.enable_input_metrics || false
 
@@ -67,7 +95,7 @@ module Fluent
     attr_reader :labels
 
     def source_only_router
-      raise "[BUG] 'RootAgent#source_only_router' should not be called when 'with_source_only' is false" unless @with_source_only
+      raise "[BUG] 'RootAgent#source_only_router' should not be called when 'with_source_only' is false" unless @source_only_mode.enabled?
       @source_only_buffer_agent.event_router
     end
 
@@ -154,7 +182,7 @@ module Fluent
 
       super
 
-      setup_source_only_buffer_agent if @with_source_only
+      setup_source_only_buffer_agent if @source_only_mode.enabled?
 
       # initialize <source> elements
       if @without_source
@@ -187,9 +215,12 @@ module Fluent
     end
 
     def lifecycle(desc: false, kind_callback: nil, kind_or_agent_list: nil)
+      only_zero_downtime_restart_ready = false
+
       unless kind_or_agent_list
-        if @with_source_only
+        if @source_only_mode.enabled?
           kind_or_agent_list = [:input, @source_only_buffer_agent]
+          only_zero_downtime_restart_ready = @source_only_mode.only_zero_downtime_restart_ready?
         elsif @source_only_buffer_agent
           # source_only_buffer_agent can re-reroute events, so the priority is equal to output_with_router.
           kind_or_agent_list = [:input, :output_with_router, @source_only_buffer_agent, @labels.values, :filter, :output].flatten
@@ -214,6 +245,9 @@ module Fluent
                  end
           display_kind = (kind == :output_with_router ? :output : kind)
           list.each do |instance|
+            if only_zero_downtime_restart_ready
+              next unless instance.respond_to?(:zero_downtime_restart_ready?) and instance.zero_downtime_restart_ready?
+            end
             yield instance, display_kind
           end
         end
@@ -257,7 +291,7 @@ module Fluent
     end
 
     def cancel_source_only!
-      unless @with_source_only
+      unless @source_only_mode.enabled?
         log.info "do nothing for canceling with-source-only because the current mode is not with-source-only."
         return
       end
@@ -285,7 +319,7 @@ module Fluent
       setup_source_only_buffer_agent(flush: true)
       start(kind_or_agent_list: [@source_only_buffer_agent])
 
-      @with_source_only = false
+      @source_only_mode.disable!
     end
 
     def shutdown(kind_or_agent_list: nil)
@@ -378,7 +412,7 @@ module Fluent
       # See also 'fluentd/plugin/input.rb'
       input.context_router = @event_router
       input.configure(conf)
-      input.event_emitter_apply_source_only if @with_source_only
+      input.event_emitter_apply_source_only if @source_only_mode.enabled?
       if @enable_input_metrics
         @event_router.add_metric_callbacks(input.plugin_id, Proc.new {|es| input.metric_callback(es) })
       end
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -43,11 +43,16 @@ module Fluent
       @rpc_endpoint = nil
       @rpc_server = nil
       @counter = nil
+      @socket_manager_server = nil
+      @starting_new_supervisor_with_zero_downtime = false
+      @new_supervisor_pid = nil
+      start_in_parallel = ENV.key?("FLUENT_RUNNING_IN_PARALLEL_WITH_OLD")
+      @zero_downtime_restart_mutex = Mutex.new
 
       @fluentd_lock_dir = Dir.mktmpdir("fluentd-lock-")
       ENV['FLUENTD_LOCK_DIR'] = @fluentd_lock_dir
 
-      if config[:rpc_endpoint]
+      if config[:rpc_endpoint] and not start_in_parallel
         @rpc_endpoint = config[:rpc_endpoint]
         @enable_get_dump = config[:enable_get_dump]
         run_rpc_server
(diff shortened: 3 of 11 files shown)


daipom:
The basic implementation is done.

lib/fluent/plugin/in_syslog.rb
```diff
@@ -156,6 +156,10 @@ def multi_workers_ready?
       true
     end
 
+    def restart_without_downtime_ready?
```
Watson1978, line 159:
```suggestion
    def zero_downtime_restart_ready?
```

daipom: Thanks! It would be more concise.

lib/fluent/root_agent.rb
```diff
@@ -47,24 +48,57 @@ module Fluent
...
+        if start_in_parallel
+          @mode = RESTART_WITHOUT_DOWNTIME_READY_ONLY
+        elsif with_source_only
+          @mode = NORMAL
+        else
+          @mode = DISABELD
+        end
+      end
+
+      def source_only?
```
Watson1978, line 66: I prefer `enabled?` or `available?` as this method name

daipom: Thanks! It would be more concise.

daipom: Fixed.

lib/fluent/plugin/in_syslog.rb
daipom: Fixed.

lib/fluent/supervisor.rb
```diff
@@ -59,24 +64,37 @@ def before_run
...
 
     def after_run
       stop_windows_event_thread if Fluent.windows?
       stop_rpc_server if @rpc_endpoint
       stop_counter_server if @counter
       cleanup_lock_dir
-      Fluent::Supervisor.cleanup_resources
+      Fluent::Supervisor.cleanup_socketmanager_path unless @starting_new_supervisor_with_zero_downtime
+
+      notify_new_supervisor_that_old_one_has_stopped if @starting_new_supervisor_with_zero_downtime
```
Watson1978, line 97:
```suggestion
      if @starting_new_supervisor_with_zero_downtime
        notify_new_supervisor_that_old_one_has_stopped
      else
        Fluent::Supervisor.cleanup_socketmanager_path
      end
```

lib/fluent/supervisor.rb
```diff
@@ -109,6 +127,11 @@ def run_rpc_server
         end
         nil
       }
+      @rpc_server.mount_proc('/api/processes.zeroDowntimeRestart') { |req, res|
+        $log.debug "fluentd RPC got /api/processes.zeroDowntimeRestart request"
+        Process.kill :USR2, Process.pid
+        nil
+      } unless Fluent.windows?
```
Watson1978, line 134:
```suggestion
      unless Fluent.windows?
        @rpc_server.mount_proc('/api/processes.zeroDowntimeRestart') { |req, res|
          $log.debug "fluentd RPC got /api/processes.zeroDowntimeRestart request"
          Process.kill :USR2, Process.pid
          nil
        }
      end
```

daipom: Thanks!

`/api/config.getDump` has the same format, so I fixed this together.
https://github.com/fluent/fluentd/compare/c714d9ca10542cccafc01b010a784c897f9c4954..a6a14de66001e39415e4537dc0b85702b5076988

lib/fluent/root_agent.rb
```diff
@@ -48,15 +48,43 @@ module Fluent
   class RootAgent < Agent
     ERROR_LABEL = "@ERROR".freeze # @ERROR is built-in error label
 
-    def initialize(log:, system_config: SystemConfig.new)
+    class SourceOnlyMode
+      DISABELD = 0
```
kenhys, line 52:
```suggestion
      DISABLED = 0
```

lib/fluent/root_agent.rb
```diff
@@ -48,15 +48,43 @@ module Fluent
...
+      NORMAL = 1
+      ONLY_ZERO_DOWNTIME_RESTART_READY = 2
+
+      def initialize(with_source_only, start_in_parallel)
+        if start_in_parallel
+          @mode = ONLY_ZERO_DOWNTIME_RESTART_READY
+        elsif with_source_only
+          @mode = NORMAL
+        else
+          @mode = DISABELD
```
kenhys, line 62:
```suggestion
          @mode = DISABLED
```

lib/fluent/root_agent.rb
```diff
@@ -48,15 +48,43 @@ module Fluent
...
+        end
+      end
+
+      def enabled?
+        @mode != DISABELD
```
kenhys, line 67:
```suggestion
        @mode != DISABLED
```

lib/fluent/root_agent.rb
```diff
@@ -48,15 +48,43 @@ module Fluent
...
+      end
+
+      def only_zero_downtime_restart_ready?
+        @mode == ONLY_ZERO_DOWNTIME_RESTART_READY
+      end
+
+      def disable!
+        @mode = DISABELD
```
kenhys, line 75:
```suggestion
        @mode = DISABLED
```

lib/fluent/supervisor.rb
Watson1978:
```ruby
    def after_run
      stop_windows_event_thread if Fluent.windows?
      stop_rpc_server if @rpc_endpoint
      stop_counter_server if @counter
      cleanup_lock_dir

      Fluent::Supervisor.cleanup_socketmanager_path unless @starting_new_supervisor_with_zero_downtime
    ensure
      notify_new_supervisor_that_old_one_has_stopped if @starting_new_supervisor_with_zero_downtime
    end
```

I think it make more readable if `notify_new_supervisor_that_old_one_has_stopped` will moved into `ensure`.

daipom: Thanks! That's right!
I will fix it.

daipom: Fixed: https://github.com/fluent/fluentd/compare/1b52de8ad8a564e359dc0d0714b2b8be4430e50f..d7e68dba4442e0a5804e4ca8345186bb8442458c
Thanks!

Watson1978 (approved):

daipom:
Thanks for your review!

kenhys (changes requested):

kenhys:
during zeroDowntimeRetart,  other HTTP endpoints result in non-guarded state. it it intentional?

lib/fluent/root_agent.rb
daipom: Thanks!
FIxed: https://github.com/fluent/fluentd/compare/d7e68dba4442e0a5804e4ca8345186bb8442458c..e52d3bd5d0dd66eea569c8251f3a1dd6cc5a76ed

lib/fluent/root_agent.rb
daipom: Thanks!
FIxed: https://github.com/fluent/fluentd/compare/d7e68dba4442e0a5804e4ca8345186bb8442458c..e52d3bd5d0dd66eea569c8251f3a1dd6cc5a76ed

lib/fluent/root_agent.rb
daipom: Thanks!
FIxed: https://github.com/fluent/fluentd/compare/d7e68dba4442e0a5804e4ca8345186bb8442458c..e52d3bd5d0dd66eea569c8251f3a1dd6cc5a76ed

lib/fluent/root_agent.rb
daipom: Thanks!
FIxed: https://github.com/fluent/fluentd/compare/d7e68dba4442e0a5804e4ca8345186bb8442458c..e52d3bd5d0dd66eea569c8251f3a1dd6cc5a76ed

daipom:
> during zeroDowntimeRetart, other HTTP endpoints result in non-guarded state. it it intentional?

Yes.
The old Fluentd should continue to work as is until it receives SIGTERM at `4.`.
(Even if the new Fluentd does not work as expected).

The new Fluentd RPC starts at `5.`, so there is no conflict.

If the old Fluentd receives `/api/processes.killWorkers`, it causes just a quick transition to `5.`.

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #4638, opened 2024-09-16
ci: filter_parser: reorganize tests for abnormal cases

daipom:
None.

Adds and reorganizes tests of `filter_parser` for abnormal cases.

The following PR will significantly change the implementation of `filter_parser`.

* #4620

However, currently, there aren't enough tests for abnormal cases.
This condition makes it difficult to safely make such a large change.
This PR solves the problem.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.
(Or not needed because this changes only the test codes.)

### DIFF
--- a/test/plugin/test_filter_parser.rb
+++ b/test/plugin/test_filter_parser.rb
@@ -538,38 +538,6 @@ class ParserFilterTest < Test::Unit::TestCase
     assert_equal t, filtered[2][0]
   end
 
-  CONFIG_NOT_IGNORE = %[
-    key_name data
-    hash_value_field parsed
-    <parse>
-      @type json
-    </parse>
-  ]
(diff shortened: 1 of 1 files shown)


daipom:
Ruby 2.7 tests fail because Ruby 2.7 does not support `Hash#except`.
I'm fixing it.

daipom:
@kenhys Could you please review this?

kenhys (changes requested):
please split into meaningful commits making it easy to review.

daipom:
Rebased to the latest master.
I'll split this fix into several commits.

kenhys (approved):
LGTM.

daipom:
Thanks!

PR #4641, opened 2024-09-20
Use each instead of each_with_index

kenhys:

Fixes #

It is trivial, but apparently index is not used at all in this method.

It does not change internal logic at all.

**Docs Changes**:

N/A

**Release Note**: 

N/A


kenhys:
This code is not changed since 2014...

daipom (approved):
LGTM. Thanks!
Waiting CI.

PR #4649, opened 2024-09-27
plugin_helper/server: Add receive_buffer_size parameter in transport section

Watson1978:
Fixes #

This patch will allow to adjust receive buffer size of UDP socket like in_udp plugin.
Related to https://github.com/fluent/fluentd/pull/1789

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/518

**Release Note**: 
The same as the title


daipom:
Thanks! Is this only for UDP? Is there any reason?

Watson1978:
Make sense.
I just referred to the in_udp plugin...
I might need to add receive_buffer_size parameter in in_tcp plugin too.

daipom:
I see!
Should this be `server` helper's option so that all plugins using the helper can use this option?

`SO_LINGER` Example:

* https://github.com/fluent/fluentd/pull/3644

PR #4661, opened 2024-10-08
Add with-source-only feature

daipom:
None.
Related to

* #4622

This feature may be useful on its own, but mainly, this feature will help #4622 feature.

Add `--with-source-only` option to `fluentd` command.
(Ref: Similar to existing [--without-source](https://docs.fluentd.org/deployment/command-line-option) option)
(We can use system-config option `with_source_only` as well)

It launches Fluentd with Input plugins only.
Here is the specification.

* Those Input plugins emits the events to SourceOnlyBufferAgent.
* The events is kept in the buf_file during the source-only-mode.
* Send SIGWINCH to the supervisor to cancels the mode.
* After canceled, the new agent starts to load the buffer.
* Not supported on Windows. (It would be possible, but I want to limit the cost of implementation and testing for now)

Screenshot from 2024-11-07 16-19-16

**Docs Changes**:
TODO

**Release Note**: 
* Add with-source-only feature
  * `fluentd` command: add `--with-source-only` option
  * system config: add `with_source_only` option
* embedded plugin: add `out_buffer` plugin, which can be used for buffering and relabeling events.

# new options for system config

We can use `with_source_only` system config option instead of `--with-source-only` command line option.

We can customize some buffer configs by setting `source_only_buffer` section.
If there is no need to customize, we don't need to set these configs.

* `with_source_only`: default `nil`
* `source_only_buffer` section
  * `flush_thread_count`: default `1`
  * `overflow_action`: default `drop_oldest_chunk`
  * `path`: default `nil` (use unique path for the instance) 
  * `flush_interval`: default `nil` (use the default of `BufferOutput` plugin: `10`)
  * `chunk_limit_size`: default `nil` (use the default of `BufferOutput` plugin: `256 * 1024 * 1024`)
  * `total_limit_size`: default `nil` (use the default of `BufferOutput` plugin: `64 * 1024 * 1024 * 1024`)
  * `compress`: default `nil` (use the default of `BufferOutput` plugin: `text`)

example:

```xml
<system>
  <source_only_buffer>
    chunk_limit_size 100m
    compress gzip
    flush_thread_count 4
  </source_only_buffer>
</system>
```

# new plugin: out_buffer
This feature internally uses the new plugin `out_buffer`.
This plugin can be used for buffering and relabeling events.

Example: in_udp -> out_buffer -> out_stdout

```xml
<source>
  @type udp
  tag foo.udp
  @label @buffer
  <parse>
    @type none
  </parse>
</source>

<match foo.**>
  @type stdout
</match>

<label @buffer>
  <match **>
    @type buffer
    @label @ROOT
    <buffer>
      path /path/to/buffer
    </buffer>
  </match>
</label>
```

# Limitaion

The following warning logs occur when the agent starts to load the buffer after sending SIGWINCH.

```
2024-11-07 18:45:07 +0900 [warn]: #0 restoring buffer file: path = /tmp/fluent/source-only-buffer/...
```

This is because the setting of `flush_at_shutdown` is changed from `false` to `true` internally.

# Behavior example

Config

```xml
<source>
  @type sample
  tag test
  auto_increment_key number
</source>

<match test.**>
  @type stdout
</match>
```

Behavior

```console
$ bundle exec fluentd -c test.conf --with-source-only
...
{datetime} [info]: starting fluentd-1.17.1 pid=195838 ruby="3.2.2"
{datetime} [info]: spawn command to main:  cmdline=[...]
{datetime} [info]: #0 init worker0 logger path=nil rotate_age=nil rotate_size=nil
{datetime} [info]: adding match pattern="test.**" type="stdout"
{datetime} [info]: adding match pattern="**" type="buffer"
{datetime} [info]: #0 with-source-only: the emitted data will be stored in the buffer files under
/tmp/fluent/source-only-buffer/470d7c43-f10c-4f7f-9d00-75731d4fdc3b. You can send SIGWINCH to
the supervisor process to cancel with-source-only mode and process data.
{datetime} [info]: adding source type="sample"
{datetime} [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
{datetime} [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
{datetime} [info]: #0 starting fluentd worker pid=195858 ppid=195838 worker=0
{datetime} [info]: #0 fluentd worker is now running worker=0

### (Wait a moment and send SIGWINCH to the supervisor (kill -WINCH 195838)) ###

{datetime} [info]: #0 try to cancel with-source-only mode
{datetime} [info]: #0 cancel with-source-only mode and start the other plugins
{datetime} test: {"message":"sample","number":11}
{datetime} test: {"message":"sample","number":12}
{datetime} test: {"message":"sample","number":13}
{datetime} test: {"message":"sample","number":14}
{datetime} test: {"message":"sample","number":15}
{datetime} [info]: #0 shutting down output plugin type=:buffer plugin_id="object:d48"
{datetime} [info]: #0 starts the loading agent for with-source-only
{datetime} [info]: adding match pattern="**" type="buffer"
{datetime} [warn]: #0 restoring buffer file: path =
/tmp/fluent/source-only-buffer/470d7c43-f10c-4f7f-9d00-75731d4fdc3b/buffer.b626515249c8f8e1b5b487087c033fad5.log
{datetime} test: {"message":"sample","number":16}
{datetime} test: {"message":"sample","number":0} # <--- The data while source-only is loaded here!!
{datetime} test: {"message":"sample","number":1}
{datetime} test: {"message":"sample","number":2}
{datetime} test: {"message":"sample","number":3}
{datetime} test: {"message":"sample","number":4}
{datetime} test: {"message":"sample","number":5}
{datetime} test: {"message":"sample","number":6}
{datetime} test: {"message":"sample","number":7}
{datetime} test: {"message":"sample","number":8}
{datetime} test: {"message":"sample","number":9}
{datetime} test: {"message":"sample","number":10}
{datetime} test: {"message":"sample","number":17}
{datetime} test: {"message":"sample","number":18}
{datetime} test: {"message":"sample","number":19}

### Ctrl + C ###

{datetime} [info]: Received graceful stop
{datetime} test: {"message":"sample","number":20}
{datetime} [info]: #0 fluentd worker is now stopping worker=0
{datetime} [info]: #0 shutting down fluentd worker worker=0
{datetime} [info]: #0 shutting down input plugin type=:sample plugin_id="object:d70"
{datetime} [info]: #0 shutting down output plugin type=:buffer plugin_id="object:e10"
{datetime} [info]: #0 shutting down output plugin type=:stdout plugin_id="object:d20"
{datetime} [info]: Worker 0 finished with status 0
```

Behavior (Stop Fluentd without sending SIGWINCH)

```console
$ bundle exec fluentd -c test.conf --with-source-only
...
{datetime} [info]: starting fluentd-1.17.1 pid=196617 ruby="3.2.2"
{datetime} [info]: spawn command to main:  cmdline=[...]
{datetime} [info]: #0 init worker0 logger path=nil rotate_age=nil rotate_size=nil
{datetime} [info]: adding match pattern="test.**" type="stdout"
{datetime} [info]: adding match pattern="**" type="buffer"
{datetime} [info]: #0 with-source-only: the emitted data will be stored in the buffer files under
/tmp/fluent/source-only-buffer/bbd9006d-bc41-418b-b346-f80888641dda. You can send SIGWINCH to
the supervisor process to cancel with-source-only mode and process data.
{datetime} [info]: adding source type="sample"
{datetime} [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
{datetime} [warn]: #0 both of Plugin @id and path for <storage> are not specified. Using on-memory store.
{datetime} [info]: #0 starting fluentd worker pid=196638 ppid=196617 worker=0
{datetime} [info]: #0 fluentd worker is now running worker=0

### Ctrl + C ###

{datetime} [info]: Received graceful stop
{datetime} [info]: #0 fluentd worker is now stopping worker=0
{datetime} [info]: #0 shutting down fluentd worker worker=0
{datetime} [info]: #0 shutting down input plugin type=:sample plugin_id="object:d70"
{datetime} [info]: #0 shutting down output plugin type=:buffer plugin_id="object:d48"
{datetime} [warn]: #0 some buffer files remain in /tmp/fluent/source-only-buffer/bbd9006d-bc41-418b-b346-f80888641dda.
Please consider recovering or saving the buffer files in the directory. To recover them, you can set
the buffer path manually to system config and retry, i.e., restart Fluentd with with-source-only mode
and send SIGWINCH again. Config Example:
<system>
  <source_only_buffer>
    path /tmp/fluent/source-only-buffer/bbd9006d-bc41-418b-b346-f80888641dda
  </source_only_buffer>
</system>

{datetime} [info]: Worker 0 finished with status 0
```

# TODO

- [x] Consider race condition
- [x] Fix and add tests
- [x] Consider how to stop the loading agent
  - Not supported because it is not mandatory.
- [x] Consider configurablity of SourceOnlyBufferAgent.
- [x] Consider Windows support
  - Don't support Windows for now. (It would be possible, but I want to limit the cost of implementation and testing for now)

### DIFF
--- a/lib/fluent/engine.rb
+++ b/lib/fluent/engine.rb
@@ -43,6 +43,8 @@ module Fluent
       @system_config = SystemConfig.new
 
       @supervisor_mode = false
+
+      @root_agent_mutex = Mutex.new
     end
 
     MAINLOOP_SLEEP_INTERVAL = 0.3
@@ -133,7 +135,15 @@ module Fluent
     end
 
     def flush!
-      @root_agent.flush!
+      @root_agent_mutex.synchronize do
+        @root_agent.flush!
+      end
+    end
+
+    def cancel_source_only!
+      @root_agent_mutex.synchronize do
+        @root_agent.cancel_source_only!
+      end
     end
 
     def now
@@ -144,7 +154,9 @@ module Fluent
     def run
       begin
         $log.info "starting fluentd worker", pid: Process.pid, ppid: Process.ppid, worker: worker_id
-        start
+        @root_agent_mutex.synchronize do
+          start
+        end
 
         @fluent_log_event_router.start
 
@@ -158,47 +170,51 @@ module Fluent
         raise
       end
 
-      stop_phase(@root_agent)
+      @root_agent_mutex.synchronize do
+        stop_phase(@root_agent)
+      end
     end
 
     # @param conf [Fluent::Config]
     # @param supervisor [Bool]
     # @reutrn nil
     def reload_config(conf, supervisor: false)
-      # configure first to reduce down time while restarting
-      new_agent = RootAgent.new(log: log, system_config: @system_config)
-      ret = Fluent::StaticConfigAnalysis.call(conf, workers: system_config.workers)
+      @root_agent_mutex.synchronize do
+        # configure first to reduce down time while restarting
+        new_agent = RootAgent.new(log: log, system_config: @system_config)
+        ret = Fluent::StaticConfigAnalysis.call(conf, workers: system_config.workers)
 
-      ret.all_plugins.each do |plugin|
-        if plugin.respond_to?(:reloadable_plugin?) && !plugin.reloadable_plugin?
-          raise Fluent::ConfigError, "Unreloadable plugin plugin: #{Fluent::Plugin.lookup_type_from_class(plugin.class)}, plugin_id: #{plugin.plugin_id}, class_name: #{plugin.class})"
+        ret.all_plugins.each do |plugin|
+          if plugin.respond_to?(:reloadable_plugin?) && !plugin.reloadable_plugin?
+            raise Fluent::ConfigError, "Unreloadable plugin plugin: #{Fluent::Plugin.lookup_type_from_class(plugin.class)}, plugin_id: #{plugin.plugin_id}, class_name: #{plugin.class})"
+          end
         end
+
+        # Assign @root_agent to new root_agent
+        # for https://github.com/fluent/fluentd/blob/fcef949ce40472547fde295ddd2cfe297e1eddd6/lib/fluent/plugin_helper/event_emitter.rb#L50
+        old_agent, @root_agent = @root_agent, new_agent
+        begin
+          @root_agent.configure(conf)
+        rescue
+          @root_agent = old_agent
+          raise
+        end
+
+        unless @suppress_config_dump
+          $log.info :supervisor, "using configuration file: #{conf.to_s.rstrip}"
+        end
+
+        # supervisor doesn't handle actual data. so the following code is unnecessary.
+        if supervisor
+          old_agent.shutdown      # to close thread created in #configure
+          return
+        end
+
+        stop_phase(old_agent)
+
+        $log.info 'restart fluentd worker', worker: worker_id
+        start_phase(new_agent)
       end
-
-      # Assign @root_agent to new root_agent
-      # for https://github.com/fluent/fluentd/blob/fcef949ce40472547fde295ddd2cfe297e1eddd6/lib/fluent/plugin_helper/event_emitter.rb#L50
-      old_agent, @root_agent = @root_agent, new_agent
-      begin
-        @root_agent.configure(conf)
-      rescue
-        @root_agent = old_agent
-        raise
-      end
-
-      unless @suppress_config_dump
-        $log.info :supervisor, "using configuration file: #{conf.to_s.rstrip}"
-      end
-
-      # supervisor doesn't handle actual data. so the following code is unnecessary.
-      if supervisor
-        old_agent.shutdown      # to close thread created in #configure
-        return
-      end
-
-      stop_phase(old_agent)
-
-      $log.info 'restart fluentd worker', worker: worker_id
-      start_phase(new_agent)
     end
 
     def stop
--- a/lib/fluent/env.rb
+++ b/lib/fluent/env.rb
@@ -14,6 +14,8 @@
 #    limitations under the License.
 #
 
+require 'securerandom'
+
 require 'serverengine/utils'
 require 'fluent/oj_options'
 
@@ -25,6 +27,7 @@ module Fluent
   DEFAULT_OJ_OPTIONS = Fluent::OjOptions.load_env
   DEFAULT_DIR_PERMISSION = 0755
   DEFAULT_FILE_PERMISSION = 0644
+  INSTANCE_ID = ENV['FLUENT_INSTANCE_ID'] || SecureRandom.uuid
 
   def self.windows?
     ServerEngine.windows?
--- a/lib/fluent/plugin/out_buffer.rb
+++ b/lib/fluent/plugin/out_buffer.rb
@@ -1,0 +1,40 @@
+#
+# Fluentd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'fluent/plugin/output'
+
+module Fluent::Plugin
+  class BufferOutput < Output
+    Fluent::Plugin.register_output("buffer", self)
+    helpers :event_emitter
+
+    config_section :buffer do
+      config_set_default :@type, "file"
+      config_set_default :chunk_keys, ["tag"]
+      config_set_default :flush_mode, :interval
+      config_set_default :flush_interval, 10
+    end
+
+    def multi_workers_ready?
+      true
+    end
+
+    def write(chunk)
+      return if chunk.empty?
+      router.emit_stream(chunk.metadata.tag, Fluent::MessagePackEventStream.new(chunk.read))
+    end
+  end
+end
--- a/lib/fluent/plugin_helper/event_emitter.rb
+++ b/lib/fluent/plugin_helper/event_emitter.rb
@@ -26,6 +26,9 @@ module Fluent
 
       def router
         @_event_emitter_used_actually = true
+
+        return Engine.root_agent.source_only_router if @_event_emitter_force_source_only_router
+
         if @_event_emitter_lazy_init
           @router = @primary_instance.router
         end
@@ -46,6 +49,14 @@ module Fluent
 
       def event_emitter_used_actually?
         @_event_emitter_used_actually
+      end
+
+      def event_emitter_apply_source_only
+        @_event_emitter_force_source_only_router = true
+      end
+
+      def event_emitter_cancel_source_only
+        @_event_emitter_force_source_only_router = false
       end
 
       def event_emitter_router(label_name)
@@ -72,6 +83,7 @@ module Fluent
         super
         @_event_emitter_used_actually = false
         @_event_emitter_lazy_init = false
+        @_event_emitter_force_source_only_router = false
         @router = nil
       end
 
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -22,6 +22,7 @@ require 'fluent/label'
 require 'fluent/plugin'
 require 'fluent/system_config'
 require 'fluent/time'
+require 'fluent/source_only_buffer_agent'
 
 module Fluent
   #
@@ -54,16 +55,21 @@ module Fluent
       @inputs = []
       @suppress_emit_error_log_interval = 0
       @next_emit_error_log_time = nil
-      @without_source = false
-      @enable_input_metrics = false
+      @without_source = system_config.without_source || false
+      @with_source_only = system_config.with_source_only || false
+      @source_only_buffer_agent = nil
+      @enable_input_metrics = system_config.enable_input_metrics || false
 
       suppress_interval(system_config.emit_error_log_interval) unless system_config.emit_error_log_interval.nil?
-      @without_source = system_config.without_source unless system_config.without_source.nil?
-      @enable_input_metrics = !!system_config.enable_input_metrics
     end
 
     attr_reader :inputs
     attr_reader :labels
+
+    def source_only_router
+      raise "[BUG] 'RootAgent#source_only_router' should not be called when 'with_source_only' is false" unless @with_source_only
+      @source_only_buffer_agent.event_router
+    end
 
     def configure(conf)
       used_worker_ids = []
@@ -148,6 +154,8 @@ module Fluent
 
       super
 
+      setup_source_only_buffer_agent if @with_source_only
+
       # initialize <source> elements
       if @without_source
         log.info :worker0, "'--without-source' is applied. Ignore <source> sections"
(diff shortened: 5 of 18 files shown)


lib/fluent/env.rb
```diff
@@ -25,6 +27,7 @@ module Fluent
   DEFAULT_OJ_OPTIONS = Fluent::OjOptions.load_env
   DEFAULT_DIR_PERMISSION = 0755
   DEFAULT_FILE_PERMISSION = 0644
+  COMMON_UUID = ENV['FLUENT_COMMON_UUID'] || SecureRandom.uuid
```
Watson1978, line 30: I don't know what it's for. 
I think `COMMON_UUID` should be more specific.

lib/fluent/engine.rb
```diff
@@ -133,7 +135,15 @@ def emit_stream(tag, es)
...
-      @root_agent.flush!
+      @root_agent_mutex.synchronize do
+        @root_agent.flush!
+      end
+    end
+
+    def cancel_source_only!
+      @root_agent_mutex.synchronize do
+        @root_agent.cancel_source_only!
+      end
```
Watson1978, line 146: @root_agent is replaced in the following places. I think that part would also require a mutex.

```ruby
      # Assign @root_agent to new root_agent
      # for https://github.com/fluent/fluentd/blob/fcef949ce40472547fde295ddd2cfe297e1eddd6/lib/fluent/plugin_helper/event_emitter.rb#L50
      old_agent, @root_agent = @root_agent, new_agent
      begin
        @root_agent.configure(conf)
      rescue
        @root_agent = old_agent
        raise
      end
```

lib/fluent/source_only_buffer_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+        ])
+      )
+
+      @base_buffer_dir = buffer_config['path']
+      # It can be "#{@base_buffer_dir}/worker#{fluentd_worker_id}/" when using multiple workers
+      @actual_buffer_dir = File.dirname(outputs[0].buffer.path)
+
+      log.info "with-source-only: the emitted data will be stored in the buffer files under" +
+               " #{@base_buffer_dir}. You can send SIGWINCH to the supervisor process to cancel" +
+               " with-source-only mode and process data." unless flush
```
Watson1978, line 55: I think it should use postfix `unless` only when the code is only one line.

lib/fluent/env.rb
daipom: This is a unique ID for that Fluentd instance (the supervisor and workers).

What names can we think of?

Watson1978: At least, I think  FLUENT_COMMON_UUID is not appropriate...

lib/fluent/source_only_buffer_agent.rb
daipom: Indeed! I will fix this!

lib/fluent/root_agent.rb
```diff
@@ -231,13 +264,45 @@ def flush!
...
+      lifecycle_control_list[:input].each(&:event_emitter_cancel_source_only)
+
+      # Want to make sure that the source_only_router finishes all process before
+      # shutting down the agent.
+      # Strictly speaking, it would be necessary to have exclusive lock between
+      # EventRouter and the shutting down process of this agent.
+      # However, adding lock to EventRouter would worsen its performance, and
+      # the entire shutting down process does not care about it either.
+      # So, just sleep a few seconds here, though it may not be strictly safe.
+      sleep 5
```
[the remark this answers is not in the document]
daipom: As commented, we cannot guarantee safety, no matter how large this value is.

[The entire shutting down process](https://github.com/fluent/fluentd/blob/0e96d5485912a099c0e445b62bbab9963c579601/lib/fluent/root_agent.rb#L298-L371) does not care about this (EventRouter must finish all process before filter/output plugins shutdowns.)
So, it would not actually cause problems, but I put `sleep` for a few seconds just to be safe.

lib/fluent/env.rb
daipom: Hmm, wouldn't it be natural to have a common ID as an instance?

Currently, this is only used for the `BUFFER_DIR_NAME` of `SourceOnlyBufferAgent`.

https://github.com/fluent/fluentd/blob/0e96d5485912a099c0e445b62bbab9963c579601/lib/fluent/source_only_buffer_agent.rb#L21-L24

However, I feel there would be no need to limit its use.

If we limit this ID to this buffer name, it would be better not to define it in `env.rb`.
I think it would be also somewhat inappropriate to put such a limited thing into the entire environment variables.

Then, there would be no good way to share the same temporal directory among workers.
It would be because workers did not have a common unique ID between them.

I don't know if it will actually be used for other purposes in the future, but it wouldn't be so bad to have a unique ID for the instance as a whole.

daipom: @kenhys gave us advice.
`FLUENT_INSTANCE_ID` and `Fluent::INSTANCE_ID` would be better.

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: > If it doesn't cause problems, isn't it unnecessary?

I'm not sure.
I am only assuming that, based on existing implementations, such problems will basically not occur.
I expect some `sleep` would reduce the possibility to almost zero.

> I don't want to put in a magic number that I'll be stuck dealing with when I read it later.

I understand that. So, I wrote the comments there.
Is it insufficient? 

Since we cannot deny the possibility that this `sleep` may contribute to the stability of the operation, it is hard to say whether it is really better to remove it.

I understand that an implementation that does not require comments is better, but should we prioritize it over one that may contribute to stability of the operation?

Maybe we should verify this effect.

lib/fluent/source_only_buffer_agent.rb
```diff
@@ -0,0 +1,89 @@
...
+                  " Config Example:\n#{config_example_to_recover(@base_buffer_dir)}"
+        return
+      end
+
+      begin
+        FileUtils.remove_dir(@base_buffer_dir)
+      rescue Errno::ENOENT
+        # This worker doesn't need to do anything. Another worker may remove the dir first.
+      rescue => e
+        log.warn "failed to remove the buffer dirctory: #{@base_buffer_dir}", error: e
```
Watson1978, line 73:
```suggestion
        log.warn "failed to remove the buffer directory: #{@base_buffer_dir}", error: e
```

lib/fluent/plugin_helper/event_emitter.rb
```diff
@@ -48,6 +51,14 @@ def event_emitter_used_actually?
         @_event_emitter_used_actually
       end
 
+      def event_emitter_set_source_only
```
Watson1978, line 54: I don't know an antonym for `cancel`,
but ... I don't know if `set` is appropriate....

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: Hmm, as commented in the codes,  it would be tough to solve the root cause.

> Strictly speaking, it would be necessary to have exclusive lock between
> EventRouter and the shutting down process of this agent.
> However, adding lock to EventRouter would worsen its performance

lib/fluent/plugin_helper/event_emitter.rb
Watson1978: I think `event_emitter_apply_source_only` is slightly better...

daipom: Indeed! I will fix this. Thanks!

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: > it would be necessary to have exclusive lock between EventRouter and the shutting down process.

I'm checking if this is true.

lib/fluent/root_agent.rb
```diff
@@ -54,17 +55,30 @@ def initialize(log:, system_config: SystemConfig.new)
...
-      @without_source = false
-      @enable_input_metrics = false
+      @without_source = system_config.without_source || false
+      @source_only_buffer_agent = nil
+      @enable_input_metrics = system_config.enable_input_metrics || false
+
+      @with_source_only = false
+      if system_config.with_source_only
+        if Fluent.windows?
+          log.warn "with-source-only is disabled because it is not supported on Windows"
```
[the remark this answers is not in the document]
daipom: Thanks!
It may be better to raise an Exception here.

If you are talking about not running the `config_param` in SystemConfig on Windows, that control would not be necessary. 

Because...

* About command line options, Fluentd doesn't start with nonexistent options.
* On the other hand, about configs, Fluentd ignores nonexistent settings.

So, there is no actual example of such a thing, at least not in this repository.

Example:

https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/plugin/out_forward.rb#L213-L220

https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/plugin/out_file.rb#L171-L174

So, we should consider whether to disable it or stop starting Fluentd when it is set on Windows.

lib/fluent/engine.rb
daipom: Thanks! That's right.
Also, it might be necessary to consider separating mutex from the new agent and the old agent.
I will consider it.

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: I have checked this.
The conclusion is that ...

* We should `sleep` for 1 sec, at least.
* We should overwrite `handle_emits_error` of `SourceOnlyBufferAgent`.

Here are the details.

https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/event_router.rb#L114-L118

I can reproduce it by adding `sleep` to `EventRouter#emit_stream()`.

```diff
    def emit_stream(tag, es)
+     sleep 10
      match(tag).emit_events(tag, es)
```

Sending SIGWINCH will terminate the buffer.
Then, some crucial variables become `nil`, and it causes `nil` error.

https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/plugin/buffer.rb#L285-L292
https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/plugin/buffer.rb#L328-L330
https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/plugin/buffer.rb#L294-L296

The problem is that we can not confirm the error in the current implementation.
We should overwrite `handle_emits_error` of `SourceOnlyBufferAgent`.

https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/event_router.rb#L114-L123
https://github.com/fluent/fluentd/blob/eebc7e1da6e0916be2ccdfab79fe0cffef296d21/lib/fluent/agent.rb#L162-L166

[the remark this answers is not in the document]
daipom: > We should sleep for 1 sec, at least.

`1s` would be enough because the router has only one output plugin when using source-only.
So, `EventRouter#emit_stream()` would immediately pass the data to the output plugin.

In the worst case, i.e., when the buffer terminates before the output plugin finishes `execute_chunking`, we lose the data.
Since we can expect almost no chance of that occurring, it would be better to make that a limitation.

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: Fixed to raise ConfigError on Windows.

lib/fluent/env.rb
daipom: Fixed.

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: Fixed.

lib/fluent/plugin_helper/event_emitter.rb
daipom: Fixed.

lib/fluent/source_only_buffer_agent.rb
daipom: Fixed.

lib/fluent/source_only_buffer_agent.rb
daipom: Fixed.

lib/fluent/engine.rb
daipom: Fixed.

lib/fluent/root_agent.rb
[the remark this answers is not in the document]
daipom: > In the worst case, i.e., when the buffer terminates before the output plugin finishes execute_chunking, we lose the data.
Since we can expect almost no chance of that occurring, it would be better to make that a limitation.

If this occurs, the following warning log occurs.

```
SourceOnlyBufferAgent: emit transaction failed: error_class=NoMethodError error="undefined method `get' for nil:NilClass" location=...
```

daipom:
I have added tests.

When making tests, I noticed `cancel_source_only!` didn't start plugins correctly, so I fixed it (production code diff).
(It was because the `@with_source_only` flag became `false` in the last of `cancel_source_only!`).

https://github.com/fluent/fluentd/pull/4661/files#diff-55c812cbe4793ae507d193d9dd642ce06c21d8e7bb41cd385740667e4f9e5033R266-R267

lib/fluent/plugin/out_buffer.rb
```diff
@@ -0,0 +1,40 @@
...
+module Fluent::Plugin
+  class BufferOutput < Output
+    Fluent::Plugin.register_output("buffer", self)
+    helpers :event_emitter
+
+    config_section :buffer do
+      config_set_default :@type, "file"
+      config_set_default :chunk_keys, ["tag"]
+      config_set_default :flush_mode, :interval
+      config_set_default :flush_interval, 10
```
kenhys, line 28: Is it appropriate default value?

kenhys (approved):
LGTM.

lib/fluent/plugin/out_buffer.rb
daipom: > Is it appropriate default value?

It can be flushed immediately if we only think about things after cancelling source-only.
However, if we think about things during source-only, we should put some amount of data together into one chunk.
So, for example, we should not use `flush_mode immediate` here.

daipom:
Thanks for your review!
@Watson1978 @kenhys

PR #4689, opened 2024-10-29
ci: add cancel-in-progress setting

Watson1978:
Fixes #

Often I forget to Signed-off and push new commit again. I would like the github action related to the PR to be automatically cancelled at that time. I want the test to run on the newly pushed commit immediately.

Ref. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs

**Docs Changes**:

**Release Note**:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,10 @@ on:
       - '*.md'
       - 'lib/fluent/version.rb'
 
+concurrency:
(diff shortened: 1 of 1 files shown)


daipom (approved):
Thanks!
It would reduce unnecessary test run.

kenhys:
Not backported yet.

PR #4712, opened 2024-11-20
Change "adding ..." log level to debug

Alainx277:

It reduces the amount of info logs on startup when using many filters or workers. In normal usage these logs do not provide any valuable information, so we only print them when debugging.

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/agent.rb
+++ b/lib/fluent/agent.rb
@@ -125,7 +125,7 @@ module Fluent
 
     def add_match(type, pattern, conf)
       log_type = conf.for_this_worker? ? :default : :worker0
-      log.info log_type, "adding match#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
+      log.debug log_type, "adding match#{@context.nil? ? '' : " in #{@context}"}", pattern: pattern, type: type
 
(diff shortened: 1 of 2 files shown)


daipom:
Thanks for this PR!
I see.

Leaving aside the need to address CI, this change looks a bit controversial to me.
It is because these logs only occur when starting and don't repeat themselves.

It would be nice to know the list of plugins actually running, in case of anomalies.
However, we can confirm it by looking at the configuration.

If anyone else has an opinion, please let me know.

By the way, there was some talk that when not setting `@id`, it might be a good idea to log the automatically determined plugin-id here.

daipom:
@Alainx277 
Thanks for this PR, but I'm considering not merging this.

The motivation for this change is the following:

> In normal usage these logs do not provide any valuable information, so we only print them when debugging.

However, it is not correct.
These logs tell us the list of plugins actually running.
That's useful information to some extent in case of anomalies.
(From the config, we can't tell if the config loaded correctly and the plugin actually loaded.)

As an alternative, it might be good to be able to turn off these logs by adding an option to the `system_config`.
Please discuss such points with us in another issue or PR.

Thanks again!

PR #4715, opened 2024-11-26
system_config: add source_only_buffer to SYSTEM_CONFIG_PARAMETERS

daipom:
* Additional fix for #4661

Need this to the section handled in `SystemConfig#dup`.
This should be added to SYSTEM_CONFIG_PARAMETERS, though this is not affected to the actual behavior,

Note: I don't understand why `log` section is not added. Maybe it should be added as well.
(It should be done by another PR).

**Docs Changes**:
Not needed

**Release Note**: 
Not needed


kenhys (approved):
LGTM.

daipom:
Thanks!

PR #4748, opened 2024-12-26
ci: exclude test case until Ruby 3.4 is installable

kenhys:

Fixes #

Ruby 3.4 binary on Windows is not ready yet.
Exclude it from matrix until it is available.

**Docs Changes**:

N/A

**Release Note**: 

N/A


kenhys:
or keep it as is?

daipom (approved):
Thanks!

PR #4750, opened 2024-12-27
ci: use latest rubygems again for Ruby 3.1 on windows

kenhys:

Fixes #

It seems that it fixes failure with Ruby 3.1 on windows.

```
   undefined method `ignored?' for #<Gem::Specification:0x000001dccaf43168
```

**Docs Changes**:

N/A

**Release Note**: 

N/A


daipom (approved):
Thanks!
It appears that Ruby 3.1 on Windows has the same issue.

PR #4759, opened 2025-01-06
Use JSON instead of Yajl to improve performance of store/load files

Watson1978:
Fixes #

When I read 10 GB file using in_tail plugin, it calls Yajl methods many times when it stores to buffer file.

Recently, Ruby's JSON has huge improvement and it is much faster.
Ref. https://byroot.github.io/ruby/json/2024/12/15/optimizing-ruby-json-part-1.html

* Before
  * It spent 88.533409329 sec to handle 10 GB file
* After
  * It spent 72.324643557 sec to handle 10 GB file

* config
```
<source>
  @type tail
  path /home/watson/prj/sandbox/fluentd/log/access*.log
  pos_file /home/watson/prj/sandbox/fluentd/log/access.log.pos
  tag log
  read_from_head true
  follow_inodes true
  # rotate_wait 0
  refresh_interval 5
  # open_on_every_update true
  <parse>
    @type none
  </parse>
</source>

<match **>
  @type file
  path /home/watson/prj/sandbox/fluentd/log/log
</match>
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/formatter_out_file.rb
+++ b/lib/fluent/plugin/formatter_out_file.rb
@@ -16,7 +16,7 @@
 
 require 'fluent/plugin/formatter'
 require 'fluent/time'
-require 'yajl'
+require 'json'
 
 module Fluent
   module Plugin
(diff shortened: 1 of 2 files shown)


daipom (approved):
LGTM! Thanks!

kenhys:
NOTE: will not be backported to v1.16

PR #4763, opened 2025-01-09
in_tail: reduce the lifespan of the buffer area

Watson1978:
Fixes #

When log files are rotated at high speed, the lifespan of `IOHandler` objects may become extended.
Along with this, the lifespan of `@iobuf` also increases.

Since `@iobuf` uses 8 KB of memory, if many such objects exist simultaneously, memory usage might significantly increase.

**Docs Changes**:

**Release Note**:


daipom (approved):
LGTM. Thanks!

PR #4764, opened 2025-01-09
output: use Hash.new {|hash, key| ... } for default value

Watson1978:
Fixes #

This will remove evaluating `meta_and_data[meta] ||= []` in the iterator every time to improves performance slightly.

The results of actually reading a 10 GB file are as follows.

* Before: 67.50262675 sec
* After: 63.862033033 sec

Here is micro benchmark code.
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
end

Benchmark.ips do |x|
  h1 = {}
  h2 = Hash.new { |h, k| h[k] = [] }
  key = :foo

  x.report("1") {
    h1[key] ||= []
    h1[key][0] = 1
  }

  x.report("2") {
    h2[key][0] = 1
  }

  x.compare!
end
```

```
                   1     1.776M i/100ms
                   2     2.239M i/100ms
                   1     18.426M (± 1.5%) i/s   (54.27 ns/i) -     92.366M in   5.013927s
                   2     23.545M (± 0.8%) i/s   (42.47 ns/i) -    118.653M in   5.039795s

Comparison:
                   2: 23544712.1 i/s
                   1: 18426072.0 i/s - 1.28x  slower
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -1036,17 +1036,17 @@ module Fluent
       #   iteration of event stream, and it should be done just once even if total event stream size
       #   is bigger than chunk_limit_size because of performance.
       def handle_stream_with_custom_format(tag, es, enqueue: false)
-        meta_and_data = {}
+        meta_and_data = Hash.new { |h, k| h[k] = [] }
         records = 0
         es.each(unpacker: Fluent::MessagePackFactory.thread_local_msgpack_unpacker) do |time, record|
           meta = metadata(tag, time, record)
-          meta_and_data[meta] ||= []
           res = format(tag, time, record)
           if res
             meta_and_data[meta] << res
             records += 1
           end
         end
+        meta_and_data.default_proc = nil
         write_guard do
           @buffer.write(meta_and_data, enqueue: enqueue)
         end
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -1036,17 +1036,17 @@ def generate_format_proc
...
         es.each(unpacker: Fluent::MessagePackFactory.thread_local_msgpack_unpacker) do |time, record|
           meta = metadata(tag, time, record)
-          meta_and_data[meta] ||= []
           res = format(tag, time, record)
           if res
             meta_and_data[meta] << res
             records += 1
           end
         end
+        meta_and_data.default_proc = nil
```
daipom, line 1049: > meta_and_data.default_proc = nil

Do we need this?
Looking at the following, it seems like it would work without this.

https://github.com/fluent/fluentd/blob/17d9322fadf7fad28cfa44118c7831c7a00d96ea/lib/fluent/plugin/buffer.rb#L343-L345

https://github.com/fluent/fluentd/blob/17d9322fadf7fad28cfa44118c7831c7a00d96ea/lib/fluent/plugin/buffer.rb#L653-L654

daipom (commented):
Thanks! Remarkable improvement!
I commented on one point.

lib/fluent/plugin/output.rb
Watson1978: As long as there are no incorrect references on the plug-in side, there will be no problems.
If a plugin makes an incorrect reference, it will return nil and the plugin will be aware of the problem.

To avoid worrying about the implementation of plug-ins, it pass objects that behave the same as before.

daipom: I see.
I understood it as reverting to the default behavior so as not to have an unnecessary impact on subsequent processing.

Note: Since this hash is only written to a buffer here, we would not need to worry about processing by unspecified plugins. However, we revert to the default behavior, just in case.

daipom (approved):
LGTM! Thanks!

PR #4769, opened 2025-01-14
Use `#match?` method instead of `#match`

Watson1978:
Fixes #

#match method returns MatchData, it has some cost to create the MatchData object.
If it does not use MatchData object, it is better to use #match? method instead. 

* verify
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'benchmark-memory'
end

Benchmark.ips do |x|
  pattern = /\{.*,.*\}/
  path = 'path/to/file'

  x.report("match") {
    pattern.match(path)
  }

  x.report("match?") {
    pattern.match?(path)
  }

  x.compare!
end
```

```
               match     1.082M i/100ms
              match?     1.683M i/100ms
               match     11.116M (± 0.8%) i/s   (89.96 ns/i) -     56.266M in   5.062181s
              match?     16.779M (± 0.5%) i/s   (59.60 ns/i) -     84.142M in   5.014786s

Comparison:
              match?: 16779085.3 i/s
               match: 11115820.6 i/s - 1.51x  slower
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -147,7 +147,7 @@ module Fluent::Plugin
         raise Fluent::ConfigError, "cannot use glob_policy as always with the default path_delimitor: `,\""
       end
 
-      if @glob_policy == :extended && /\{.*,.*\}/.match(@path) && extended_glob_pattern(@path)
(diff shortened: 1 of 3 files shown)


daipom (approved):
LGTM. Thanks!!

PR #4745, opened 2024-12-26
Bump required ruby version

kenhys:
**Which issue(s) this PR fixes**: 

Fixes #

**What this PR does / why we need it**: 

Raise baseline version of Ruby to 3.2.
fluent-package v5 (LTS) adopts ruby 3.2, so it is reasonable to set minimum version to Ruby 3.2. 

NOTE: these kind of changes might be better to apply in v1.19.0

**Docs Changes**:

https://docs.fluentd.org/quickstart/faq

https://github.com/fluent/fluentd-docs-gitbook/pull/524

**Release Note**: 

N/A

PR #4770, opened 2025-01-15
out_file: remove workaround for Ruby 2.7 and 3.0 on macOS

Watson1978:
Fixes #

Next fluentd version will drop support of Ruby 3.1 or below.
Ref. https://github.com/fluent/fluentd/pull/4745

So, we can remove the workaround for Ruby 2.7 and 3.0 on macOS that added at https://github.com/fluent/fluentd/pull/3579

**Docs Changes**:

**Release Note**:


daipom (approved):
LGTM. Thanks!

daipom:
Note: original issue:
* #3569

PR #4771, opened 2025-01-15
server: remove codes for async v1.x

Watson1978:
Fixes #

Async v2.0 just supports Ruby 3.1 or above.
When we had been supporting Ruby 2.7 and 3.0, we need async v1.x and v2.x both.
Related to https://github.com/fluent/fluentd/pull/4619

Next fluentd version will drop support of Ruby 3.1 or below.
So, now we can remove the codes for async v1.x
Ref. https://github.com/fluent/fluentd/pull/4745

**Docs Changes**:

**Release Note**:

### DIFF
--- a/fluentd.gemspec
+++ b/fluentd.gemspec
@@ -65,7 +65,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency("test-unit", ["~> 3.3"])
   gem.add_development_dependency("test-unit-rr", ["~> 1.0"])
   gem.add_development_dependency("oj", [">= 2.14", "< 4"])
-  gem.add_development_dependency("async-http", ">= 0.50.0")
+  gem.add_development_dependency("async-http", "~> 0.86")
   gem.add_development_dependency("aws-sigv4", ["~> 1.8"])
   gem.add_development_dependency("aws-sdk-core", ["~> 3.191"])
   gem.add_development_dependency("rexml", ["~> 3.2"])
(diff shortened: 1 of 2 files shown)


fluentd.gemspec
```diff
@@ -65,6 +65,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency("test-unit", ["~> 3.3"])
   gem.add_development_dependency("test-unit-rr", ["~> 1.0"])
   gem.add_development_dependency("oj", [">= 2.14", "< 4"])
+  gem.add_development_dependency("async", ">= 2.0")
```
daipom, line 68: Does this need to be `development_dependency`?

Watson1978: If not specified, async v1.x would also be available. However, this does not work in practice, as the corresponding code is erased.

Or, it increases the version of async-http. 
It would  increase the required version of async.
https://github.com/fluent/fluentd/blob/c2b09d4396c0f065f60099d3e84ca52ec5faf4f8/fluentd.gemspec#L68

https://rubygems.org/gems/async-http/versions/0.50.0

Watson1978: I adjust the version of `async-http` instead.

daipom: Thanks!
Note:

* `async` is not mandatory for Fluentd. If there is no `async`, `webrick` is used for `HttpServer` plugin helper.
  * This should be fixed in another PR in the future....
* Fluentd uses `async` and `async-http`, and `async-http` depends on `async`. So it is enough to manage `async-http` version.

PR #4781, opened 2025-01-27
github: Add link to feedback use-case/testimonial

kenhys:
**Which issue(s) this PR fixes**: 
Fixes #

**What this PR does / why we need it**: 

Add contact information to collect Fluentd's use-case

It help to collect Fluentd use-case widely.

**Docs Changes**:

N/A

**Release Note**: 

N/A

PR #4787, opened 2025-01-28
github: remove needless label

kenhys:

Follow-up #4781 

It is harmful to add extra label.
Item should be placed under contact_links.

See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

**Docs Changes**:

N/A

**Release Note**: 

N/A


daipom (approved):
Oh, I see! Thanks!

Issue #4674, opened 2024-10-18
CI: Fails with "Errno::EACCES: Permission denied - bind(2)" on Windows


Watson1978:
### Describe the bug

Sometimes, CI fails with "Errno::EACCES: Permission denied - bind(2)" on Windows platform, like

```
 1) Error: test: skip_invalid_event[tcp](ForwardInputTest::packed forward): Errno::EACCES: Permission denied - bind(2) for 127.0.0.1:53297
C:/hostedtoolcache/windows/Ruby/3.1.6/x64/lib/ruby/3.1.0/socket.rb:179:in `bind'
C:/hostedtoolcache/windows/Ruby/3.1.6/x64/lib/ruby/3.1.0/socket.rb:179:in `bind'
D:/a/fluentd/fluentd/lib/fluent/plugin_helper/server.rb:390:in `server_create_udp_socket'
D:/a/fluentd/fluentd/lib/fluent/plugin_helper/server.rb:174:in `server_create'
D:/a/fluentd/fluentd/lib/fluent/plugin/in_forward.rb:183:in `start'
D:/a/fluentd/fluentd/lib/fluent/test/driver/base.rb:120:in `instance_start'
D:/a/fluentd/fluentd/lib/fluent/test/driver/base.rb:78:in `run'
D:/a/fluentd/fluentd/lib/fluent/test/driver/base_owner.rb:130:in `run'
D:/a/fluentd/fluentd/test/plugin/test_in_forward.rb:491:in `block (2 levels) in <class:ForwardInputTest>'



28) Error: test: Create new connection per send_data(ForwardOutputTest): Errno::EACCES: Permission denied - bind(2) for 127.0.0.1:53356
C:/hostedtoolcache/windows/Ruby/3.1.6/x64/lib/ruby/3.1.0/socket.rb:179:in `bind'
C:/hostedtoolcache/windows/Ruby/3.1.6/x64/lib/ruby/3.1.0/socket.rb:179:in `bind'
D:/a/fluentd/fluentd/lib/fluent/plugin_helper/server.rb:390:in `server_create_udp_socket'
D:/a/fluentd/fluentd/lib/fluent/plugin_helper/server.rb:174:in `server_create'
D:/a/fluentd/fluentd/lib/fluent/plugin/in_forward.rb:183:in `start'
D:/a/fluentd/fluentd/lib/fluent/test/driver/base.rb:120:in `instance_start'
[...]

PR #4793, opened 2025-01-29
test_cat: use proper protocol in unused_port method (#4686)

kenhys:

Backport #4686

Fixes an issue where an inappropriate protocol was specified in the unused_port method.
See https://github.com/fluent/fluentd/issues/4674#issuecomment-2440522606

**Docs Changes**:

**Release Note**:


daipom (approved):
LGTM.

PR #4813, opened 2025-01-31
parser_json: use JSON as fallback parser instead of Yajl for performance

Watson1978:
Fixes #

Recently, Ruby's json has incredible performance improvements.
It might be faster than oj gem.
So, I think json is a suitable as fallback.

This is similar with https://github.com/fluent/fluentd/pull/4759

Here is easily benchmark. (I used same log file in #4759)

* Before
  * It spent 90.50963662 sec to handle 10 GB file 
* After
  * It spent 74.624230077 sec to handle 10 GB file

* config
```
<source>
  @type tail
  tag log
  read_from_head true
  <parse>
    @type json
  </parse>
</source>

<match **>
  @type file
  path "#{File.expand_path '~/tmp/fluentd/log'}"
</match>
```

FYI)
* https://byroot.github.io/ruby/json/2024/12/15/optimizing-ruby-json-part-1.html
* https://byroot.github.io/ruby/json/2024/12/18/optimizing-ruby-json-part-2.html
* https://byroot.github.io/ruby/json/2024/12/27/optimizing-ruby-json-part-3.html
* https://byroot.github.io/ruby/json/2024/12/29/optimizing-ruby-json-part-4.html
* https://byroot.github.io/ruby/json/2025/01/04/optimizing-ruby-json-part-5.html
* https://byroot.github.io/ruby/json/2025/01/12/optimizing-ruby-json-part-6.html
* https://byroot.github.io/ruby/json/2025/01/14/optimizing-ruby-json-part-7.html

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/560

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -52,8 +52,8 @@ module Fluent
         when :oj
           return [Oj.method(:load), Oj::ParseError] if Fluent::OjOptions.available?
 
-          log&.info "Oj is not installed, and failing back to Yajl for json parser"
-          configure_json_parser(:yajl)
+          log&.info "Oj is not installed, and failing back to JSON for json parser"
+          configure_json_parser(:json)
         when :json then [JSON.method(:load), JSON::ParserError]
         when :yajl then [Yajl.method(:load), Yajl::ParseError]
         else
--- a/test/plugin/test_parser_json.rb
+++ b/test/plugin/test_parser_json.rb
@@ -28,12 +28,12 @@ class JsonParserTest < ::Test::Unit::TestCase
 
       result = @parser.instance.configure_json_parser(:oj)
 
-      assert_equal [Yajl.method(:load), Yajl::ParseError], result
+      assert_equal [JSON.method(:load), JSON::ParserError], result
       logs = @parser.logs.collect do |log|
         log.gsub(/\A\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2} [-+]\d{4} /, "")
       end
       assert_equal(
-        ["[info]: Oj is not installed, and failing back to Yajl for json parser\n"],
+        ["[info]: Oj is not installed, and failing back to JSON for json parser\n"],
         logs
       )
     end


kenhys:
we are happy if we could observe micro benchmark result.

Watson1978:
## Micro benchmark
### benchmark code
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'oj'
  gem 'json'
  gem 'yajl-ruby', require: 'yajl'
end

Benchmark.ips do |x|
  json = '{"a":"Alpha","b":true,"c":12345,"d":[true,[false,[-123456789,null],3.9676,["Something else.",false],null]],"e":{"zero":null,"one":1,"two":2,"three":[3],"four":[0,1,2,3,4]},"f":null,"h":{"a":{"b":{"c":{"d":{"e":{"f":{"g":null}}}}}}},"i":[[[[[[[null]]]]]]]}'

  x.report('Oj.load') { Oj.load(json) }
  x.report('JSON.load') { JSON.load(json) }
  x.report('Yajl.load') { Yajl.load(json) }

  x.compare!
end
```

### result
```
             Oj.load    31.383k i/100ms
           JSON.load    35.513k i/100ms
           Yajl.load    13.839k i/100ms
             Oj.load    319.187k (± 1.5%) i/s    (3.13 μs/i) -      1.601M in   5.015522s
           JSON.load    364.950k (± 0.5%) i/s    (2.74 μs/i) -      1.847M in   5.060214s
           Yajl.load    137.015k (± 2.5%) i/s    (7.30 μs/i) -    691.950k in   5.053679s

Comparison:
           JSON.load:   364950.5 i/s
             Oj.load:   319187.1 i/s - 1.14x  slower
           Yajl.load:   137015.2 i/s - 2.66x  slower
```

kenhys:
By the way, which json gem version did you tested with?
Do we need to use specific version of json gem?

Watson1978:
> By the way, which json gem version did you tested with? Do we need to use specific version of json gem?

I tested with json v2.9.1 (latest version), it is default version in Ruby 3.4.1
https://github.com/ruby/ruby/blob/master/doc/NEWS/NEWS-3.4.0.md

If we will use Ruby 3.4 or later in the next LTS, hmmm, I think we don't need to specify the version...

test/plugin/test_parser_json.rb
```diff
@@ -28,12 +28,12 @@ def test_fall_back_oj_to_yajl_if_oj_not_available
 
       result = @parser.instance.configure_json_parser(:oj)
 
-      assert_equal [Yajl.method(:load), Yajl::ParseError], result
+      assert_equal [JSON.method(:load), JSON::ParserError], result
```
byroot, line 31: You probably want `JSON.method(:parse)`: https://bugs.ruby-lang.org/issues/19528

It's also very marginally faster.

Watson1978: Thanks.

kenhys:
NOTE: will not be backported to v1.16

lib/fluent/plugin/parser_json.rb
```diff
@@ -52,9 +52,9 @@ def configure_json_parser(name)
         when :oj
           return [Oj.method(:load), Oj::ParseError] if Fluent::OjOptions.available?
 
-          log&.info "Oj is not installed, and failing back to Yajl for json parser"
-          configure_json_parser(:yajl)
-        when :json then [JSON.method(:load), JSON::ParserError]
+          log&.info "Oj is not installed, and failing back to JSON for json parser"
+          configure_json_parser(:json)
+        when :json then [JSON.method(:parse), JSON::ParserError]
```
daipom, line 57: Thanks for this enhancement!

> when :json then [JSON.method(:parse), JSON::ParserError]

This change means that `JSON#parse` is faster than `JSON#load`, right?
If so, could you please separate the PR?

test/plugin/test_parser_json.rb
daipom: Thanks byroot!
Let's fix this in another PR.

Watson1978: OK, I will separate the PR

daipom (approved):
LGTM. Thanks!!
Waiting CI.

Watson1978:
I created separated PR in https://github.com/fluent/fluentd/pull/4817

orgads:
Following this, can yajl-ruby be a development_dependency, so that it is not installed by default?

PR #4817, opened 2025-02-03
parser_json: use JSON.parse instead of .load

Watson1978:
Fixes #

Ref. https://github.com/fluent/fluentd/pull/4813#discussion_r1937321034
Fix wrong usage of `JSON.load` method.
We should use `JSON.parse` instead.

JSON.load performs unnecessary deserialisation.

```
irb(main):001> JSON.load('{ "json_class": "String", "raw": [72, 101, 108, 108, 111] }')
=> "Hello"
irb(main):002> JSON.parse('{ "json_class": "String", "raw": [72, 101, 108, 108, 111] }')
=> {"json_class" => "String", "raw" => [72, 101, 108, 108, 111]}
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/parser_json.rb
+++ b/lib/fluent/plugin/parser_json.rb
@@ -54,7 +54,7 @@ module Fluent
 
           log&.info "Oj is not installed, and failing back to JSON for json parser"
           configure_json_parser(:json)
-        when :json then [JSON.method(:load), JSON::ParserError]
+        when :json then [JSON.method(:parse), JSON::ParserError]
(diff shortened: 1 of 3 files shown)


daipom (approved):
LGTM. Thanks!

daipom:
I think backporting to LTS is unnecessary because no specific issues have been reported with this.

I cannot assume how much this could change Fluentd's behavior.
So, I'm a little concerned about backporting this to LTS.
I think we should not backport this unless we have a strong reason.

If anyone has an opinion that we should backport, please comment.

Issue #4822, opened 2025-02-05
Set permissions and ownership of parent directories of ".pos" files


lifecrisis:
### Is your feature request related to a problem? Please describe.

Greetings:

I am investigating how directories are created when using the `tail` input plugin. It would be super helpful if I could get some guidance on how this works since I'm having trouble getting clarity from the docs.

Relevant documentation can be found [here](https://docs.fluentd.org/input/tail).

The documentation gives an example at the top, which I will modify and reproduce here to describe my problem...

```
<source>
  @type tail
  path /var/log/httpd-access.log

  # NOTE: Here I have inserted the additional /httpd/ directory.
  pos_file /var/log/td-agent/httpd/httpd-access.log.pos

  tag apache.access
  <parse>
    @type apache2
  </parse>
</source>
```

My problem is as follows: When `td-agent` starts up, it will create the `/var/log/td-agent/httpd/` directory as needed to hold the `.pos` file under it. I find that I am unable to use the configuration files to specify the ownership and mode for this new directory. I have verified that the `td-agent` user on my machine can set the ownership and mode of this directory, but I can't do it automatically in the config files.

For example, when the dirctory holding the `.pos` file is created, the mode is `00755` when I would like it to be `00770`. I would also like it to have a specific user and group ownership.

I am not sure if this would be a new feature, or if I'm just missing how to make this small change from reading the existing docs. If I am missing something, I apologize!

Many thanks in advance, please let me know if more information is needed and I will try to provide it.

Best,
lifecrisis

### Describe the solution you'd like

I would like to be able to set custom ownership and permissions on directories created to hold `.pos` files.

### Describe alternatives you've considered

I tried setting the permissions external to the `td-agent` process, but this is cumbersome with many servers. It is best to specify the permissions in the config files so that I can rely on `td-agent` to set itself up properly. :)

kenhys:
https://docs.fluentd.org/deployment/command-line-option
Fluentd's --umask command line option might help what you want.

egemenkus:
> https://docs.fluentd.org/deployment/command-line-option Fluentd's --umask command line option might help what you want.
> 
> https://docs.fluentd.org/deployment/system-config#dir_permission https://docs.fluentd.org/deployment/system-config#file_permission

I recently submitted a PR to add explicit permission control for pos_file directories in Fluentd. However, since there is already a pending umask PR, I’m wondering if this feature is still necessary.

[...]

Issue #4810, opened 2025-01-30
File permissions can't have full accesss


The-LittleTeapot:
### Describe the bug

Unable to set 

<system>
  file_permission 777
</system>

Files are created as 755 due to umask.

### To Reproduce

add to config:
<system>
  file_permission 777
</system>

### Expected behavior

files should be created with full permissions

### Your Environment

```markdown
- Fluentd version: v1.16.6-debian-1.0
- Operating system:GNU/Linux
- Kernel version:3.10.0-1160.76.1.el7.x86_64
```

### Your Configuration

```apache
<system>
  file_permission 777
</system>

<source>
  @type forward
  port 24224
  bind 0.0.0.0
</source>

<match **>
  @type copy
  
  <store>
  @type file
    path /app/log/webui-testing/webui-debug-log3
    append true
    format out_file
    <buffer>
      path /app/buffer/webui/webui-debug-buffer
      flush_interval 5s
      flush_mode interval
      chunk_limit_size 512m
      total_limit_size 1g
      overflow_action block
      @type file
    </buffer>
  </store>
</match>
```

### Your Error Log

```shell
NO error message,

Files are created with r-x permissions (755) due to the default umask.
```

### Additional context

If changing the permissions is not possible how can we change the default umask?

daipom:
Fluentd has `--umask` command line option.

* https://docs.fluentd.org/deployment/command-line-option

However, we cannot set this in `system_config`.
We should add this setting to `system_config`.

daipom:
For docker images, we can change this `ENTRYPOINT`.

https://github.com/fluent/fluentd-docker-image/blob/89d5c78641633b5ea8acff6958f484ef75c45160/v1.18/debian/Dockerfile#L64

Example:

```
ENTRYPOINT ["tini",  "--", "/bin/entrypoint.sh", "--umask", "0000"]
```

daipom:
Sorry. There are some limitations for `--umask`.
It would not work for Fluentd Docker images.

* Fluetnd v1.14.6 ~ v1.18.0: `--umask` works only when using `--no-supervisor`.
  * This will be fixed in #4836.
* However, even if that PR is merged, it would not work for Fluentd Docker images.
  * In [ServerEngine](https://github.com/treasure-data/serverengine), it appears that the umask value is applied only when using `daemonize`.
  * So, it would work for fluent-package, but not for Docker images.
 
I wonder why ServerEngine has such specifications.
[...]

PR #4829, opened 2025-02-10
Add umask option to Fluentd system configuration & Fix Tests

### LINKED ISSUE #4816: System config: add umask option

Example:

* #4810

### Additional context

_No response_

egemenkus:
**Which issue(s) this PR fixes:**
Fixes #4816

**What this PR does / why we need it:**
This PR introduces an umask option in Fluentd system configuration.
Previously, users had to use the `--umask` command-line argument, which was inconvenient for service deployments and containerized environments.
With this PR, users can now define umask directly in `fluent.conf`, making it easier to manage in service configurations.

Additionally, this PR fixes previous test issues and reopens the PR after correcting an earlier mistake. The previous PR contained incorrect assertions in the test cases, which have now been corrected to ensure that:

The configured umask value is correctly applied.
New files are created during testing, and their permissions are checked to verify that the umask is applied properly.
Invalid umask values raise a `Fluent::ConfigError` as expected.

Based on my tests, I believe this implementation will work as expected in the actual program. However, if any issues are found during the review process, I am more than happy to make the necessary adjustments.

**Docs Changes:**
No documentation changes required.

**Release Note:**
Added `umask` option to Fluentd system configuration.
Fixed test cases related to `umask` handling and file permissions.
Implemented additional tests to verify `umask` application by creating new files and checking permissions.

### DIFF
--- a/fluent.conf
+++ b/fluent.conf
@@ -1,6 +1,11 @@
 # In v1 configuration, type and id are @ prefix parameters.
 # @type and @id are recommended. type and id are still available for backward compatibility
 
+## System-wide configurations
+<system>
+  umask 0027
+</system>
+
 ## built-in TCP input
 ## $ echo <json> | fluent-cat <tag>
 <source>
@@ -12,7 +17,6 @@
 #<source>
 #  @type unix
 #</source>
-
 # HTTP input
 # http://localhost:8888/<tag>?json=<json>
 <source>
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -180,6 +180,33 @@ module Fluent
       label_configs.each { |name, e| @labels[name].configure(e) }
       setup_error_label(error_label_config) if error_label_config
 
+      # Get umask value from system config
+      system_conf = conf.elements.find { |e| e.name == "system" }
+      if system_conf
+        if system_conf.has_key?("umask")
+          umask_str = system_conf.to_h["umask"]
+          begin
+            umask_value = Integer(umask_str, 8)  # ArgumentError for invalid octal
+            if umask_value > 0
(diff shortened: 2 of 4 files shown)


daipom:
@egemenkus Thanks for this improvement!
I will review this.

daipom (changes requested):
@egemenkus Sorry for my late response.
In the first place, I found a bug in `--umask` options.

* #4836

It appears that we need to fix the bug first.
As commented next, even if #4836 is merged, the umask will not be applied in the normal launch.

* https://github.com/fluent/fluentd/issues/4810#issuecomment-2674030900

It will be applied when using `--daemon` option.
It is the specification of [ServerEngine](https://github.com/treasure-data/serverengine).

Based on the above, could you please fix this PR entirely as follows?

* Add `umask` option to SystemConfig.
  * https://github.com/fluent/fluentd/blob/master/lib/fluent/system_config.rb
* Assuming [#4836](https://github.com/fluent/fluentd/pull/4836), replace the passing value with the value of SystemConfig's option.
  * It would be better to start the fix after [#4836](https://github.com/fluent/fluentd/pull/4836) is merged.
  * SystemConfig fetches command line options as follows.
    * https://github.com/fluent/fluentd/blob/644c921078827328affe9143301a834e8b3e7dcc/lib/fluent/supervisor.rb#L1168-L1184

As with #4836, it is sufficient for now to pass the configuration value to ServerEngine.

egemenkus:
Alright, thank you. I will wait and try to contribute again after reviewing it in more detail.

daipom:
Thanks!

daipom:
@egemenkus I commented here. Please check.

* https://github.com/fluent/fluentd/issues/4822#issuecomment-2693107945

PR #4835, opened 2025-02-18
formatter_json: use JSON as fallback parser instead of Yajl for performance

Watson1978:
Fixes #

Recently, Ruby's json has incredible performance improvements.
It might be faster than oj gem.
So, I think json is a suitable as fallback.

This is similar with https://github.com/fluent/fluentd/pull/4813

* Before
  * It spent 70.708872503 sec to handle 10 GB file 
* After
  * It spent 54.428017716 sec to handle 10 GB file

* config
```
<source>
  @type tail
  path "#{File.expand_path('~/tmp/fluentd/access*.log')}"
  pos_file "#{File.expand_path('~/tmp/fluentd/access.log.pos')}"
  tag log
  read_from_head true
  <parse>
    @type none
  </parse>
</source>

<match **>
  @type file
  path "#{File.expand_path('~/tmp/fluentd/output/log')}"

  # use formatter_json
  format json
</match>
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/formatter_json.rb
+++ b/lib/fluent/plugin/formatter_json.rb
@@ -34,11 +34,11 @@ module Fluent
           if Fluent::OjOptions.available?
             @dump_proc = Oj.method(:dump)
           else
-            log.info "Oj isn't installed, fallback to Yajl as json parser"
-            @dump_proc = Yajl.method(:dump)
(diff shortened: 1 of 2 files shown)


daipom (approved):
Thanks!

PR #3671, opened 2022-03-13
Added --umask command line parameter

syedriko:
Signed-off-by: Sergey Yedrikov <syedriko@redhat.com>

**Which issue(s) this PR fixes**: 
None

**What this PR does / why we need it**: 
I needed to control permissions on files and directories fluentd creates on disk, to make them accessible only to the user running fluentd:
```
sh-4.4# find /var/lib/fluentd -ls
100978405      0 drwxr-xr-x   5  root     root           53 Feb  4 04:52 /var/lib/fluentd
157295736      0 drwx------   2  root     root           59 Feb  4 05:01 /var/lib/fluentd/pos
157295739     12 -rw-------   1  root     root        11117 Feb  4 05:02 /var/lib/fluentd/pos/es-containers.log.pos
157295855      4 -rw-------   1  root     root          137 Feb  4 05:01 /var/lib/fluentd/pos/journal_pos.json
 48242762      0 drwxr-xr-x   2  root     root          115 Feb  4 05:02 /var/lib/fluentd/default
 48257477      8 -rw-------   1  root     root         5156 Feb  4 05:02 /var/lib/fluentd/default/buffer.b5d72a27e600fe3fec22c832f88f14303.log
 48257478      4 -rw-------   1  root     root          203 Feb  4 05:02 /var/lib/fluentd/default/buffer.b5d72a27e600fe3fec22c832f88f14303.log.meta
 50664000      0 drwxr-xr-x   2  root     root            6 Feb  4 04:52 /var/lib/fluentd/retry_default
sh-4.4# 
```
Setting the appropriate umask looked like an elegant solution. However, I found it impossible to control the umask of fluentd when it runs with --no-supervisor, as fluentd overrides the umask set in the parent shell.

**Docs Changes**:
--umask UMASK           change umask
needs to be added to https://docs.fluentd.org/deployment/command-line-option

**Release Note**:

PR #4836, opened 2025-02-21
chumask: fix issue where it only works with standalone worker

daipom:
* Fixes #3671

Fix #3671 (88a72608f36e8dd7287b44f1908f74fe19e5034a)

After #3671, the value of `chumask` passed to ServerEngine was always `nil`.
The `chumask` value was being used only when standalone. It appears to be unintentional.

This fixes it.

* By default, `0` as `chumask` is passed to ServerEngine.
* If specifying `--umask` option, that value is passed.
* This `chumask` value is applied when using `daemonize`.
  * When not using `daemonize`, the value is not used.
  * This is the specification of ServerEngine.

Specification change:
(Just undo the specification that #3671 unintentionally changed.)

* This changes the default `umask` when using `daemonize`.
  * fluent-package (RPM/DEB), for example.
  * Before: system default (`nil` for ServerEngine)
  * After: `0`

**Docs Changes**:
Not needed.
(There is not enough information on https://docs.fluentd.org/deployment/command-line-option. It would be good to add more information.)

**Release Note**: 
* Fixed a bug where the default `umask` was not set to `0` when using `--daemon` (td-agent, fluent-package) since v1.14.6.
* `--umask` command line option: Fixed so that it is applied when Fluentd runs with `--daemon` (fluent-package) as well as when Fluentd runs with `--no-supervisor`.


daipom:
It would be necessary for us to solve this issue before making the following enhancement:

* #4816

Watson1978 (approved):
👍🏻

ashie (approved):

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

kenhys:
MEMO: will not backport to v1.16 because behavior is changed.

PR #4845, opened 2025-03-04
memory_chunk: clear internal memory use in buffer string

### LINKED ISSUE #1657: Fluentd retains excessive amounts of memory after handling traffic peaks
When setting up a simple, 2-part Fluentd (TCP -> forwarder) -> (forwardee -> disk configuration) and giving it 5 million JSON objects to process all at once, resident-set memory consumption jumps from an inital 30MB to between 200-450MB, and does not come back down after computation is complete.  This is observed using version 2.3.5-1.el7 of the TD Agent RPM package running on CentOS 7.  (The version of Fluentd in that package is 0.12.36.)

Steps to reproduce:
```sh
# define and start a simple TCP forwarder
$ cat > test-in.conf <<'EOF'
<source>
  @type  tcp
  tag    testing
  format json
  port   10130
</source>

<match **>
  @type forward
  require_ack_response true
  flush_interval 10s

  <server>
    host 127.0.0.1
    port 10131
  </server>
</match>
EOF
$ td-agent --no-supervisor -c test-in.conf &

# define and start a simple forwadee that logs to disk
$ cat > test-out.conf <<'EOF'
<source>
   @type forward
   port  10131
</source>

<match **>
  @type              file
  format             json
  path               "/tmp/fluent-test"
  flush_interval     10s
  buffer_chunk_limit 16m
  compress           gzip
</match>
EOF
$ td-agent --no-supervisor -c test-out.conf &

# observe initial memory consumption
$ ps -o pid,vsz,rss,cmd | grep '[t]d-'
 4254 431724 30544 /opt/td-agent/embedded/bin/ruby /sbin/td-agent --no-supervisor -c test-out.conf
 4259 498816 30052 /opt/td-agent/embedded/bin/ruby /sbin/td-agent --no-supervisor -c test-in.conf

# pump a bunch of data through it
$ seq 1 5000000 | sed 's/.*/{"num": &, "filler": "this is filler text to make the event larger"}/' > /dev/tcp/localhost/10130 &

# wait for all data to flush to disk and CPU to return to idle
$ watch -n1 'ls -lt /tmp/fluent-test.* | head -5'

# observe final memory consumption
$ ps -o pid,rss,cmd | grep '[t]d-'
 4338 628248 288972 /opt/td-agent/embedded/bin/ruby /sbin/td-agent --no-supervisor -c test-out.conf
 4343 1023868 461616 /opt/td-agent/embedded/bin/ruby /sbin/td-agent --no-supervisor -c test-in.conf
```

As you can see from the RSS numbers, each td-agent process started out around 30MB and they ended at ~290MB and ~460MB, respectively.  Neither process will release that memory if you wait a while.  (In the real-world staging system we initially discovered this on, memory consumption of the `test-out.conf`-equivalent configuration reached over 3GB, and the `test-in.conf`-equivalent was a Fluent Bit instance exhibiting a recently-fixed [duplication bug](https://github.com/fluent/fluent-bit/issues/342).)

Reviewing a [Fluentd-related kubernetes issue](https://github.com/kubernetes/kubernetes/issues/32762) during our own diagnostics, we noticed that the behavior we observed seemed similar to the Fluentd behavior described there when built without jemalloc.  This led us to check if the td-agent binary we were using was in fact linked with jemalloc.  According to the [FAQ](https://www.fluentd.org/faqs), jemalloc is used when building the Treasure Data RPMs, and though we found jemalloc libraries installed on the system, we couldn't find any existence of jemalloc in the process running in memory.  Specifically, we tried the following things:

```sh
# given a td-agent process with PID 4343...

# no jemalloc shared library mentioned in the memory mapping
$ pmap 4343 | grep jemalloc
# ... so it doesn't look like it's dynamically linked

# grab the entire memory space and search it for references to jemalloc
$ gcore 4343
$ strings core.4343 | fgrep jemalloc

# ... but if it were statically linked, you'd expect to find some of these strings
$ strings /opt/td-agent/embedded/lib/libjemalloc.a | fgrep jemalloc
```

In short, this leads us to wonder... are the binaries invoked by `td-agent` actually linked with jemalloc?  If they are not, is the old memory fragmentation problem that jemalloc solved what we are observing here?  (And if they aren't, am I raising this issue in the wrong place, and if so where should I raise it?)

Watson1978:
Fixes #1657

Seems that a full GC have to run in order to collect the wasted String object in `@chunk`.
However, full GC is not performed frequently.

This patch purges internal memory in wasted String object immediately.

Here is memory usage retrieved by `ObjectSpace.memsize_of_all`.

> [!NOTE]
> Ruby has various optimization, such as reusing freed heap areas by pooling them.
> For this reason, it may not be reflected in the process's memory usage immediately even if it clear the object.

## Reproduce
* Configuration
```
<source>
  @type tcp
  tag testing
  format json
  port 10130
</source>
<match **>
  @type forward
  require_ack_response true
  <server>
    host 127.0.0.1
    port 10131
  </server>

  ## Tune the buffer parameters
  <buffer tag>
    flush_mode interval
    flush_interval 1s
  </buffer>
</match>
```

* test data
```ruby
require "socket"
require "json"

i = 0

Thread.new do
  s = TCPSocket.open("localhost", 10130)
  data = {"message": "a" * 512 * 1024}

  loop do
    puts "[#{i}]"
    data["number"] = i
    s.puts data.to_json
    i += 1
    sleep 0.01
  end

  s.close
end

sleep 20
```

**Docs Changes**:
Not needed.

**Release Note**: 
buf_memory: improve memory performance

### DIFF
--- a/lib/fluent/plugin/buffer/memory_chunk.rb
+++ b/lib/fluent/plugin/buffer/memory_chunk.rb
@@ -68,7 +68,7 @@ module Fluent
 
         def purge
           super
-          @chunk = ''.force_encoding("ASCII-8BIT")
+          @chunk.clear
           @chunk_bytes = @size = @adding_bytes = @adding_size = 0
           true
         end
--- a/lib/fluent/plugin/buffer/memory_chunk.rb
+++ b/lib/fluent/plugin/buffer/memory_chunk.rb
@@ -74,7 +74,7 @@ module Fluent
         end
 
         def read(**kwargs)
-          @chunk
+          @chunk.dup
         end
 
         def open(**kwargs, &block)
--- a/test/plugin/test_output_as_buffered_compress.rb
+++ b/test/plugin/test_output_as_buffered_compress.rb
@@ -122,12 +122,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
     assert_equal :gzip, @i.buffer.compress
 
     @i.register(:write) do |c|
-      compressed_data = c.instance_variable_get(:@chunk)
-      if compressed_data.is_a?(File)
-        compressed_data.seek(0, IO::SEEK_SET)
-        compressed_data = compressed_data.read
-      end
-      compressed_data = compressed_data.dup
+      compressed_data = c.read(compressed: :gzip)
       c.write_to(io)
     end
 
@@ -161,12 +156,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
 
     @i.register(:format) { |tag, time, record| "#{record}\n" }
     @i.register(:write) { |c|
-      compressed_data = c.instance_variable_get(:@chunk)
-      if compressed_data.is_a?(File)
-        compressed_data.seek(0, IO::SEEK_SET)
-        compressed_data = compressed_data.read
-      end
-      compressed_data = compressed_data.dup
+      compressed_data = c.read(compressed: :gzip)
       c.write_to(io)
     }
 
--- a/test/plugin/test_output_as_buffered_compress.rb
+++ b/test/plugin/test_output_as_buffered_compress.rb
@@ -127,6 +127,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
         compressed_data.seek(0, IO::SEEK_SET)
         compressed_data = compressed_data.read
       end
+      compressed_data = compressed_data.dup
       c.write_to(io)
     end
 
@@ -165,6 +166,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
         compressed_data.seek(0, IO::SEEK_SET)
         compressed_data = compressed_data.read
       end
+      compressed_data = compressed_data.dup
       c.write_to(io)
     }
 


lib/fluent/plugin/buffer/memory_chunk.rb
```diff
@@ -68,13 +68,13 @@ def empty?
...
           super
-          @chunk = ''.force_encoding("ASCII-8BIT")
+          @chunk.clear
           @chunk_bytes = @size = @adding_bytes = @adding_size = 0
           true
         end
 
         def read(**kwargs)
-          @chunk
+          @chunk.dup
```
Watson1978, line 77: Even if it duplicates the string, it just increase a few memory usage.
`String#dup` will share the internal heap area for new object.

```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-memory'
end

Benchmark.memory do |x|
  x.report('single large string') {
    str1 = 'a' * 1_000
  }

  x.report('duplicate large string') {
    str1 = 'a' * 1_000
    str1.dup
  }

  x.report('duble large string') {
    str1 = 'a' * 1_000
    str2 = 'b' * 1_000
  }
end
```

```
 single large string     1.081k memsize (     0.000  retained)
                         2.000  objects (     0.000  retained)
                         2.000  strings (     0.000  retained)
duplicate large string
                         1.161k memsize (     1.041k retained)
                         4.000  objects (     1.000  retained)
                         2.000  strings (     1.000  retained)
  duble large string     2.162k memsize (     0.000  retained)
                         4.000  objects (     0.000  retained)
                         4.000  strings (     0.000  retained)
```

test/plugin/test_output_as_buffered_compress.rb
```diff
@@ -127,6 +127,7 @@ def waiting(seconds)
         compressed_data.seek(0, IO::SEEK_SET)
         compressed_data = compressed_data.read
       end
+      compressed_data = compressed_data.dup
```
daipom, line 130:
```diff
--- a/test/plugin/test_output_as_buffered_compress.rb
+++ b/test/plugin/test_output_as_buffered_compress.rb
@@ -122,12 +122,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
     assert_equal :gzip, @i.buffer.compress
 
     @i.register(:write) do |c|
-      compressed_data = c.instance_variable_get(:@chunk)
-      if compressed_data.is_a?(File)
-        compressed_data.seek(0, IO::SEEK_SET)
-        compressed_data = compressed_data.read
-      end
-      compressed_data = compressed_data.dup
+      compressed_data = c.read(compressed: :gzip)
       c.write_to(io)
     end
```

test/plugin/test_output_as_buffered_compress.rb
```diff
@@ -165,6 +166,7 @@ def waiting(seconds)
...
```
daipom, line 169:
```diff
@@ -161,12 +156,7 @@ class BufferedOutputCompressTest < Test::Unit::TestCase
 
     @i.register(:format) { |tag, time, record| "#{record}\n" }
     @i.register(:write) { |c|
-      compressed_data = c.instance_variable_get(:@chunk)
-      if compressed_data.is_a?(File)
-        compressed_data.seek(0, IO::SEEK_SET)
-        compressed_data = compressed_data.read
-      end
-      compressed_data = compressed_data.dup
+      compressed_data = c.read(compressed: :gzip)
       c.write_to(io)
     }
```

daipom (commented):
Thanks for this improvement!
Basically looks good to me!

I have commented on some points for the test code.
It could be simpler.
Please check them.

test/plugin/test_output_as_buffered_compress.rb
Watson1978: Nice. Thanks!

daipom:
Thanks! Waiting CI...

daipom (approved):

PR #4849, opened 2025-03-05
test: fix unstable buffer test

daipom:
None.

Fix occasional test failure: https://github.com/fluent/fluentd/actions/runs/13650069134/job/38156420830?pr=4846#step:6:4985

The max size of the buffer queue is `1` by default.
So, `try_flush` will be called with `flush_thread_interval`, not `flush_thread_burst_interval`.

The previous test code assumes that the interval is 0.1s, but it was actually 1s.
This would be the cause of the occasional failure.

    test: #write is called per value combination of variables, per flush_interval & chunk sizes, and buffer chunk is purged(BufferedOutputTest::buffered output feature with variables):
            assert{ @i.buffer.stage.size == 0 }
                    |  |      |     |    |
                    |  |      |     |    false
                    |  |      |     1
                    |  |      {...}
                    |  #<...>
                    #<...>
    /Users/runner/work/fluentd/fluentd/test/plugin/test_output_as_buffered.rb:1459:in `block (2 levels) in <class:BufferedOutputTest>'
         1456:       @i.enqueue_thread_wait
         1457:       @i.flush_thread_wakeup
         1458:
      => 1459:       assert{ @i.buffer.stage.size == 0 }
         1460:
         1461:       waiting(4) do
         1462:         Thread.pass until @i.write_count > 1

**Docs Changes**:
Not needed.

**Release Note**: 
Not needed.


Watson1978 (approved):
👍🏻

PR #4854, opened 2025-03-06
v1_parser: use URI#open instead of URI.open

Watson1978:
Fixes #

By [CodeQL documentation](https://codeql.github.com/codeql-query-help/ruby/rb-non-constant-kernel-open/), it is safer to avoid using `URI.open`.

This is similar with https://github.com/fluent/fluentd/pull/4848

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -172,7 +172,7 @@ module Fluent
           require 'open-uri'
           basepath = '/'
           fname = path
-          data = URI.open(uri) { |f| f.read }
+          data = u.open { |f| f.read }
           data.force_encoding('UTF-8')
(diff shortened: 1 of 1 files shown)


daipom (commented):
Thanks.

It appears to me that v1parser can recognize Windows absolute paths (`C:/foo`).
Can't we just use `u.open`?

Watson1978:
Thanks.

```
        if u.scheme == 'file' || (!u.scheme.nil? && u.scheme.length == 1) || u.path == uri.tr(' ', '+') # file path
          # When the Windows absolute path then u.scheme.length == 1
```

Indeed, seems it recognize Windows path.
I will fix the patch

daipom (approved):
Thanks!

PR #4855, opened 2025-03-06
parser: fix file path into URI#open for Windows

Watson1978:
Fixes #

This patch will fix the problem in https://github.com/fluent/fluentd/pull/4848
`u.path` has the file path without a drive letter in Windows environment.

If Fluentd is worked on `c` drive and config file in `d` drive, it can't parse config file, I think.

**Docs Changes**:

**Release Note**:


daipom (approved):
Thanks!

PR #4858, opened 2025-03-07
Add logging for errors about loading dependencies on startup

daipom:
None.

Adds error handling similar to worker process.
https://github.com/fluent/fluentd/blob/8df10863769da32459aba2a1408134d0b8e7d9f4/lib/fluent/supervisor.rb#L1149-L1164

`Fluent::Engine.run_configure` calls `Fluent::Registry#search`.
It can run `Kernel.require`.
It can raise `LoadError`, `ConflictError` and others.

Without this error handling, Fluentd cannot log those errors.
This makes a situation where Fluentd fails to start, but there is no error log in the log file.
This solves that problem.

This appears to be an oversight of the following fix:

* https://github.com/fluent/fluentd/pull/2651

**Docs Changes**:
Not needed.

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -721,6 +721,18 @@ module Fluent
         $log.error 'config error', file: @config_path, error: e
         $log.debug_backtrace
         exit!(1)
+      rescue ScriptError => e # LoadError, NotImplementedError, SyntaxError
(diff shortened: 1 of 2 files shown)


Watson1978 (approved):
👍🏻

Issue #4063, opened 2023-02-18
CI: Some tests about signal handlers sometimes fail


daipom:
### Describe the bug

`test_cont_in_main_process_signal_handlers` and `test_cont_in_supervisor_signal_handler` of `SupervisorTest` sometimes fail.

```
===============================================================================
Failure: test_cont_in_main_process_signal_handlers(SupervisorTest):
  class()
  Called 0 times.
  Expected 1 times.
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:74:in `block in dump_object_count'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:73:in `each_object'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:73:in `dump_object_count'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:18:in `block in dump'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:136:in `open'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:136:in `_open_dump_path'
/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/gems/3.0.0/gems/sigdump-0.2.4/lib/sigdump.rb:14:in `dump'
/home/runner/work/fluentd/fluentd/lib/fluent/supervisor.rb:1027:in `dump_non_windows'
/home/runner/work/fluentd/fluentd/lib/fluent/supervisor.rb:943:in `block in install_main_process_signal_handlers'
/home/runner/work/fluentd/fluentd/test/test_supervisor.rb:221:in `kill'
/home/runner/work/fluentd/fluentd/test/test_supervisor.rb:221:in `test_cont_in_main_process_signal_handlers'
     218:     sv = Fluent::Supervisor.new(opts)
     219:     sv.send(:install_main_process_signal_handlers)
     220: 
[...]

PR #4862, opened 2025-03-11
Backport(v1.16) test: improve Sigdump test for Windows (#4847)

daipom:
* Backport #4847

Sigdump.dump iterates ObjectSpace.
If there are DoubleDefinition objects of RR, it can set some expectations of RR unexpectedly.
It can cause test failure.

By doing GC, we can remove DoubleDefinition objects made in other tests. This allows us to simplify this test.

See https://github.com/fluent/fluentd/issues/4063

**Docs Changes**:
Not needed.

**Release Note**:
Not needed.


Watson1978 (approved):
👍🏻

PR #4864, opened 2025-03-14
formatter_csv: fix memory leak

### LINKED ISSUE #3627: in_exec with out_exec in one config create file descriptors leak

Watson1978:
Fixes #3627

Seems `in_exec` plugin creats Thread object each interval and  it invoke formatter_csv on the thread.

If it use a Thread object as hash key when caching with formatter_csv, that thread will be permanently referenceable and will not be collected by the GC.

Here is memory usage of Ruby's process.
chart

### Reproduce
See #3627
```
<system>
  process_name fluentd_monitor
</system>

<source>
  @type exec
  tag hardware.memory
  run_interval 0.1
  command /home/watson/prj/sandbox/fluentd/get_memory_metrics.sh
  <parse>
    @type json
    types mem_available:float,swap_total:float,swap_free:float
  </parse>
</source>

<match **>
  @type exec
  command /home/watson/prj/sandbox/fluentd/psql_memory.sh
  <format>
    @type csv
    force_quotes false
    delimiter |
    fields time_local,mem_available,swap_total,swap_free
  </format>
  <buffer time>
    @type file
    path /home/watson/prj/sandbox/fluentd/log
    timekey 50s
    timekey_wait 0s
    flush_mode lazy
    total_limit_size 200MB
    retry_type periodic  # exponential backoff is broken https://github.com/fluent/fluentd/issues/3609
    retry_max_times 0  # need to go to secondary as soon as possible https://github.com/fluent/fluentd/issues/3610
  </buffer>
</match>
```

**Docs Changes**:

**Release Note**: 
formatter_csv: fix memory leak

### DIFF
--- a/lib/fluent/compat/formatter.rb
+++ b/lib/fluent/compat/formatter.rb
@@ -104,15 +104,6 @@ module Fluent
 
         # Do not cache because it is hard to consider the thread key correctly.
         # (We can try, but it would be low priority.)
-        def csv_for_thread
-          CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts)
-        end
-
-        # It is hard to consider the thread key correctly for the compat layer.
-        def csv_thread_key
-          raise NotImplementedError, "Compat CsvFormatter does not support CSV cache. Do not use this method."
-        end
-
         def csv_cacheable?
           false
         end
--- a/lib/fluent/compat/formatter.rb
+++ b/lib/fluent/compat/formatter.rb
@@ -112,6 +112,10 @@ module Fluent
         def csv_thread_key
           raise NotImplementedError, "Compat CsvFormatter does not support CSV cache. Do not use this method."
         end
+
+        def csv_cacheable?
+          false
+        end
       end
 
       class SingleValueFormatter < Fluent::Plugin::SingleValueFormatter
--- a/lib/fluent/compat/formatter.rb
+++ b/lib/fluent/compat/formatter.rb
@@ -101,6 +101,17 @@ module Fluent
 
       class CsvFormatter < Fluent::Plugin::CsvFormatter
         # TODO: warn when deprecated
+
+        # Do not cache because it is hard to consider the thread key correctly.
+        # (We can try, but it would be low priority.)
+        def csv_for_thread
+          CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts)
+        end
+
+        # It is hard to consider the thread key correctly for the compat layer.
+        def csv_thread_key
+          raise NotImplementedError, "Compat CsvFormatter does not support CSV cache. Do not use this method."
+        end
       end
 
       class SingleValueFormatter < Fluent::Plugin::SingleValueFormatter
--- a/lib/fluent/plugin/formatter_csv.rb
+++ b/lib/fluent/plugin/formatter_csv.rb
@@ -52,11 +52,11 @@ module Fluent
         @generate_opts = {col_sep: @delimiter, force_quotes: @force_quotes, headers: @fields,
                           row_sep: @add_newline ? :auto : "".force_encoding(Encoding::ASCII_8BIT)}
         # Cache CSV object per thread to avoid internal state sharing
-        @cache = {}
+        Thread.current[:csv_formatter] = nil
       end
 
       def format(tag, time, record)
-        csv = (@cache[Thread.current] ||= CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts))
+        csv = (Thread.current[:csv_formatter] ||= CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts))
         line = (csv << record).string.dup
         # Need manual cleanup because CSV writer doesn't provide such method.
         csv.rewind
@@ -65,7 +65,7 @@ module Fluent
       end
 
       def format_with_nested_fields(tag, time, record)
-        csv = (@cache[Thread.current] ||= CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts))
+        csv = (Thread.current[:csv_formatter] ||= CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts))
         values = @accessors.map { |a| a.call(record) }
         line = (csv << values).string.dup
         # Need manual cleanup because CSV writer doesn't provide such method.
--- a/lib/fluent/plugin/formatter_csv.rb
+++ b/lib/fluent/plugin/formatter_csv.rb
@@ -36,7 +36,7 @@ module Fluent
       config_param :add_newline, :bool, default: true
 
       def csv_cacheable?
-        owner ? true : false
+        !!owner
       end
 
       def csv_thread_key
--- a/lib/fluent/plugin/formatter_csv.rb
+++ b/lib/fluent/plugin/formatter_csv.rb
@@ -36,7 +36,12 @@ module Fluent
       config_param :add_newline, :bool, default: true
 
       def thread_key
-        "#{owner.plugin_id}_csv_formatter_#{@usage}_csv"
+        if owner
+          "#{owner.plugin_id}_csv_formatter_#{@usage}_csv"
+        else
+          # For Fluent::Compat::TextFormatter
+          "csv_formatter_#{@usage}_csv"
+        end
       end
 
       def configure(conf)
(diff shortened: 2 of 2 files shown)


daipom (commented):
Thanks! Good catch!

It seems that in #2529, input plugins were assumed to spawn only a fixed number of threads for emitting.
However, `in_exec` for example, makes threads for emitting without limit.
The cache would then become infinitely large.
We need to fix this cache.

Watson1978:
@daipom Thanks. I fixed the key name at last commit.

lib/fluent/plugin/formatter_csv.rb
```diff
@@ -52,11 +57,11 @@ def configure(conf)
         @generate_opts = {col_sep: @delimiter, force_quotes: @force_quotes, headers: @fields,
                           row_sep: @add_newline ? :auto : "".force_encoding(Encoding::ASCII_8BIT)}
         # Cache CSV object per thread to avoid internal state sharing
-        @cache = {}
+        Thread.current[thread_key] = nil
```
daipom, line 60: Just a note.

>         # Cache CSV object per thread to avoid internal state sharing
>        Thread.current[thread_key] = nil

Now, this code would be unnecessary since the key becomes unique for the instance.
However, it may be the correct specification to clear the cache when `configure` is performed.
(I don't know if there is a case for performing `configure` multiple times.)

lib/fluent/plugin/formatter_csv.rb
```diff
@@ -35,6 +36,10 @@ class CsvFormatter < Formatter
       config_param :fields, :array, value_type: :string
       config_param :add_newline, :bool, default: true
 
+      def thread_key
+        "#{plugin_id}_csv"
```
daipom, line 40:
```suggestion
        "#{owner.plugin_id}_csv_formatter_#{usage}_csv"
```

Formatter plugins (owned plugins) are unique to `owner.plugin_id` and `usage`.

daipom (changes requested):
Sorry, I forgot that this is a `owned` plugin.
It would not be expected to have an owned plugin with `PluginId`.

So, could you please fix the following points?

(Owned plugins are unique to the owner's id and `usage`.)

Watson1978:
When it invoke `CsvFormatter` via `Fluent::Compat::TextFormatter`, looks like it does not set `owner`...

daipom:
> When it invoke `CsvFormatter` via `Fluent::Compat::TextFormatter`, looks like it does not set `owner`...

Sorry, I missed this point.
I have pushed another idea to fix this issue.

* 80c02fbb385e51afe925b5083cba92fc8c27ab55

@Watson1978 Could you please review this commit?
If there seem to be some problems, we can revert it.

The aim of this commit is to ensure that the compat layer does not affect this cache feature.
(Because I'm not sure how to make the key unique on the compat layer context.)

lib/fluent/compat/formatter.rb
```diff
@@ -101,6 +101,21 @@ class LabeledTSVFormatter < Fluent::Plugin::LabeledTSVFormatter
...
+        # (We can try, but it would be low priority.)
+        def csv_for_thread
+          CSV.new("".force_encoding(Encoding::ASCII_8BIT), **@generate_opts)
+        end
+
+        # It is hard to consider the thread key correctly for the compat layer.
+        def csv_thread_key
+          raise NotImplementedError, "Compat CsvFormatter does not support CSV cache. Do not use this method."
+        end
+
```
daipom, line 115:
```suggestion
```

Looks like overriding `csv_cacheable?` is enough.

lib/fluent/plugin/formatter_csv.rb
```diff
@@ -35,6 +35,22 @@ class CsvFormatter < Formatter
       config_param :fields, :array, value_type: :string
       config_param :add_newline, :bool, default: true
 
+      def csv_cacheable?
+        owner ? true : false
```
daipom, line 39:
```suggestion
        not owner.nil?
```

Simpler?

daipom (commented):
Thanks for fixing the CI failures my commit caused!

It seems possible that there is no owner. (FluentBinlogReader)
It would be a good solution to disable the cache feature in such a case.

I commented on some minor points.
Please check them.

lib/fluent/compat/formatter.rb
Watson1978: Indeed. Thanks.

lib/fluent/plugin/formatter_csv.rb
Watson1978: Hmm, I think using ternary operator is readable for me...

kenhys: one more suggestion: `!!owner`

Watson1978: Thanks. Replaced with `!!owner`
https://github.com/fluent/fluentd/pull/4864/commits/eae43cae2f08bb94d4601599129ee416bb7038e3

daipom (approved):
LGTM! Thanks for your contribution!

PR #4865, opened 2025-03-14
Add warning for directory permission to in_tail.rb

moatom:
~~Fixes #~~

This PR provides a warning for typical configuration mistakes that prevent log transfer from starting.

When using the tail type as a source, it is common for directories with inappropriate partitions to be mistakenly included in the source path.

In my opinion, it would be preferable to issue a warning in such cases to help users avoid these mistakes.

Steps to Reproduce:

```shell
$ sudo mkdir /var/log/noaccess
$ sudo mkdir /var/log/noaccess/noaccess2
$ sudo touch /var/log/noaccess/noaccess2/fluentd.log
$ sudo chmod 0000 /var/log/noaccess
$ sed -Ei 's|^([[:space:]]*)path .*|\1path /var/log/noaccess/noaccess2/fluentd.log|' example/in_tail.conf
$ bundle exec fluentd -c example/in_tail.conf
{no warning...}
```

**Docs Changes**:
No documentation changes required.

**Release Note**: 
in_tail: add warning for directory permission.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -165,6 +165,7 @@ module Fluent::Plugin
         @path_formatters = @paths.map{|path| [path, Fluent::Timezone.formatter(@path_timezone, path)]}.to_h
         @exclude_path_formatters = @exclude_path.map{|path| [path, Fluent::Timezone.formatter(@path_timezone, path)]}.to_h
       end
+      check_dir_permission unless Fluent.windows?
 
       # TODO: Use plugin_root_dir and storage plugin to store positions if available
       if @pos_file
@@ -210,6 +211,20 @@ module Fluent::Plugin
       rotated_file_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_rotated_total", help_text: "Total number of rotated files")
       throttling_metrics = metrics_create(namespace: "fluentd", subsystem: "input", name: "files_throttled_total", help_text: "Total number of times throttling occurs per file when throttling enabled")
       @metrics = MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics, throttling_metrics)
+    end
+
+    def check_dir_permission
+      expand_paths_raw.select { |path|
+        not File.exist?(path)
+      }.each { |path|
+        inaccessible_dir = Pathname.new(File.expand_path(path))
+          .ascend
+          .reverse_each
+          .find { |p| p.directory? && !p.executable? }
+        if inaccessible_dir
+          log.warn "Skip #{path} because '#{inaccessible_dir}' lacks execute permission."
+        end
+      }
     end
 
     def configure_tag
@@ -321,7 +336,7 @@ module Fluent::Plugin
       end
     end
 
-    def expand_paths
+    def expand_paths_raw
       date = Fluent::EventTime.now
       paths = []
       @paths.each { |path|
@@ -367,10 +382,14 @@ module Fluent::Plugin
                end
         use_glob?(path) ? Dir.glob(path) : path
       }.flatten.uniq
+      paths - excluded
+    end
+
+    def expand_paths
       # filter out non existing files, so in case pattern is without '*' we don't do unnecessary work
       hash = {}
-      (paths - excluded).select { |path|
-        FileTest.exist?(path)
+      expand_paths_raw.select { |path|
+        File.exist?(path)
       }.each { |path|
         # Even we just checked for existence, there is a race condition here as
         # of which stat() might fail with ENOENT. See #3224.
--- a/test/plugin/test_in_tail.rb
+++ b/test/plugin/test_in_tail.rb
@@ -2591,6 +2591,46 @@ class TailInputTest < Test::Unit::TestCase
     d.instance_shutdown if d && d.instance
   end
 
+  def test_warn_without_directory_permission
+    omit "Cannot test with root user" if Process::UID.eid == 0
+    omit "NTFS doesn't support UNIX like permissions" if Fluent.windows?
+    path = "#{@tmp_dir}/noaccess/tail.txt"
+    begin
+      FileUtils.mkdir_p("#{@tmp_dir}/noaccess")
+      FileUtils.touch(path)
+      FileUtils.chmod(0400, path)
+      FileUtils.chmod(0600, "#{@tmp_dir}/noaccess")
+      config = config_element('', '', {
+                                'tag' => "tail",
+                                'path' => path,
+                                'format' => 'none',
+                                "pos_file" => "#{@tmp_dir}/tail.pos",
+                              })
+      d = create_driver(config, false)
+      fname = File.expand_path("#{@tmp_dir}/noaccess")
+      assert(d.logs.any?{|log| log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
+    end
+  end
+
+  def test_no_warn_with_directory_permission
+    omit "Cannot test with root user" if Process::UID.eid == 0
+    omit "NTFS doesn't support UNIX like permissions" if Fluent.windows?
+    path = "#{@tmp_dir}/noaccess/tail.txt"
+    begin
+      FileUtils.mkdir_p("#{@tmp_dir}/noaccess")
+      FileUtils.chmod(0100, "#{@tmp_dir}/noaccess")
+      config = config_element('', '', {
+                                'tag' => "tail",
+                                'path' => path,
+                                'format' => 'none',
+                                "pos_file" => "#{@tmp_dir}/tail.pos",
+                              })
+      d = create_driver(config, false)
+      fname = File.expand_path("#{@tmp_dir}/noaccess")
+      assert(d.logs.all?{|log| !log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
+    end
+  end
+
   def test_shutdown_timeout
     Fluent::FileWrapper.open("#{@tmp_dir}/tail.txt", "wb") do |f|
       # Should be large enough to take too long time to consume


lib/fluent/plugin/in_tail.rb
```diff
@@ -385,6 +385,13 @@ def expand_paths
           $log.warn "expand_paths: stat() for #{path} failed with #{e.class.name}. Skip file."
         end
       }
+      if check_permission
+        (paths - excluded).map { |path| File.dirname(path) }.uniq.each { |dir|
```
moatom, line 389: It might be better to add `select { |path| !FileTest.exist?(path) }` before map.

moatom:
> [Test / Ruby 3.4 on macos-latest (pull_request)](https://github.com/fluent/fluentd/actions/runs/13865592763/job/38858881719?pr=4865)Failing after 23m

https://github.com/fluent/fluentd/actions/runs/13865592763/job/38858881719?pr=4865#step:6:13416
```
4) Failure: test_unwatched_files_should_be_removed(TailInputTest::path)
/Users/runner/work/fluentd/fluentd/test/plugin/test_in_tail.rb:1757:in 'test_unwatched_files_should_be_removed'
     1754:       waiting(20) { sleep 0.1 until Dir.glob("#{@tmp_dir}/*.txt").size == 0 } # Ensure file is deleted on Windows
     1755:       waiting(5) { sleep 0.1 until d.instance.instance_variable_get(:@tails).keys.size <= 0 }
     1756: 
  => 1757:       assert_equal(
     1758:         {
     1759:           files: [],
     1760:           tails: []
<{:files=>[], :tails=>[]}> expected but was
<{:files=>[],
 :tails=>
  ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/6085abb32771a48075e1/tail.txt"]}>

diff:
? {:files=>[], :tails=>[]}
+  :tails=>
+   ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/6085abb32771a48075e1/tail.txt"]}
Error: <{:files=>[], :tails=>[]}> expected but was
<{:files=>[],
 :tails=>
  ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/6085abb32771a48075e1/tail.txt"]}>.

diff:
- {:files=>[], :tails=>[]}
+ {:files=>[],
+  :tails=>
+   ["/Users/runner/work/fluentd/fluentd/test/plugin/../tmp/tail/6085abb32771a48075e1/tail.txt"]}
```

👀

daipom:
> This PR provides a warning for typical configuration mistakes that prevent log transfer from starting.
When using the tail type as a source, it is common for directories with inappropriate partitions to be mistakenly included in the source path.

I see.

In some cases, `Permission denied` warning logs appear.
In other cases, they do not.

If the directory does not have the `x` permission, then warning logs do not appear.
This PR will solve that problem.

daipom:
@moatom 
Thanks for this PR!
The direction of checking the readability of the parent directory seems to have some effect.
However, I have some concerns.

* A: It does not seem to be able to support the case of using glob (such as `*`).
* B: We do not want to warn for paths that do not yet exist.

About A, it would be possible that this feature supports only static paths that do not use wildcards, etc.

About B, I feel the essential difficulty of this issue.
There could be cases where the mid-level parent directory, rather than the direct parent directory, does not have `x` permissions.
As long as there is no `x` permission, it would be difficult to check whether the path does not exist yet or whether the permission is inappropriate.
If we do this strictly, it will be necessary to find the directories whose existence can be verified and check their permissions.

I don't think it necessarily needs to be strictly confirmed.
However, it would be better to take these considerations into account and be clear about which cases are to be supported.

For example, I want to avoid false positives, such as generating warning logs for paths that just do not yet exist.

daipom:
> However, it would be better to take these considerations into account and be clear about which cases are to be supported.
> 
> For example, I want to avoid false positives, such as generating warning logs for paths that just do not yet exist.

To avoid false positives, we can consider the following direction for example.

* Instead of checking directory permissions, generate a warning log in the next process when the path does not exist.

https://github.com/fluent/fluentd/blob/e3dcebca2d3ec6b23affd30b743d42850b737ab9/lib/fluent/plugin/in_tail.rb#L372-L374

Log message:

```
... [warn] ... Unable to confirm the existence of #{path}.
You can ignore this warning log if the file is created later.
If the file already exists, please check the configuration and directory permissions.
```

(This supports only static paths that do not use wildcards, etc.)

moatom:
@daipom I'm sorry for my late response🙇

> If the directory does not have the `x` permission, then warning logs do not appear.

Oops. I thought the required permission was `r`. `FileTest.readable?(dir)` in the commit might not be appropriate...



> (This supports only static paths that do not use wildcards, etc.)

The following part seems to expand the glob, so I believe this issue shouldn't occur (if my understanding is correct):

https://github.com/fluent/fluentd/pull/4865/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802L333-R346



> I don't think it necessarily needs to be strictly confirmed.
>  However, it would be better to take these considerations into account and be clear about which cases are to be supported.
>
> For example, I want to avoid false positives, such as generating warning logs for paths that just do not yet exist.
>
> ...
>
> To avoid false positives, we can consider the following direction for example.
>
> Log message:
>
> ```
> ... [warn] ... Unable to confirm the existence of #{path}.
> You can ignore this warning log if the file is created later.
> If the file already exists, please check the configuration and directory permissions.
> ```

I completely agree with this direction! I refined this idea as follows:

---

Experienced users are the ones who primarily use complex paths, and they are likely to notice issues on their own. Therefore, in this PR I would like to focus on adding a warning for the following simple case: 

​	`/p_1/p_2/.../p_t/.../p_n` 

where 

- `p_i` is a directory for `0<=i<=n-1`
  - `p_0 := /` (the root directory)
-  `p_n` is the log file
- **`p_t` is the shallowest directory (or file, if `t = n`) that lacks the x permission in the path.**

For this specific case, the conditions for generating warnings can be summarized in the following table:

| ↓ All Permissions in the Path \ → File (`p_n`) Actually Exist | Exist                   | Not Exist               |
| ------------------------------------------------- | ----------------------- | ----------------------- |
| Granted (`p_t` doesn't exist)                     | ○ (No Warning)          | ○ (No Warning)          |
| Not Granted (`p_t` exists)                        | ☓ (Warning about `p_t`) | ☓ (Warning about `p_t`) |

As seen in the table, there is no concern of false positives because `Not Granted`×`Not Exist` seems to be also warned.

Based on this, how about modifying the warning message to explicitly mention `p_t`? (Though I’m not sure about the effort required to detect `p_t`...)

```
... [warn] ... Unable to confirm the existence of #{path} because some required permissions are missing from #{`/p1/p2/.../p_t`} in the path.
```

moatom:
>>    (This supports only static paths that do not use wildcards, etc.)
>
>The following part seems to expand the glob, so I believe this issue shouldn't occur (if my understanding is correct):
>
>https://github.com/fluent/fluentd/pull/4865/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802L333-R346

I might have missed cases where logs are emitted only at the debug level:
https://github.com/fluent/fluentd/pull/4865/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802R352-R355

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,6 +370,19 @@ def expand_paths
...
+              break true unless FileTest.executable?(current_path)
+              break false if current_path == "/"
+            end
+          end
+          if is_bad_permission
+            $log.warn "Skip #{path} because a directory in the path lacks execute permission."
+          end
+        end
         FileTest.exist?(path)
       }.each { |path|
```
moatom, line 387: This should work fine unless I've overlooked something.

daipom:
@moatom 
Thanks for the update! I will see it soon.

> > (This supports only static paths that do not use wildcards, etc.)
> 
> The following part seems to expand the glob, so I believe this issue shouldn't occur (if my understanding is correct):
> 
> https://github.com/fluent/fluentd/pull/4865/files#diff-456fdeb51bc472beb48891caac0d063e0073655dba7ac2b72e6fdc67dc6ac802L333-R346

If there are not enough permissions for the directory, `Dir.glob()` does not return the files under that directory.
So, it would be very difficult to tell if the files do not exist or if there are not enough permissions.

daipom:
> Experienced users are the ones who primarily use complex paths, and they are likely to manage to notice issues on their own. Therefore, in this PR, I would like to focus on adding a warning for the following simple case (if glob expansion does not occur, I want to ignore such cases):

This would be no problem!

> We are concerned about false positives for `Granted`×`Not Exist` (We need this test case).
> Can we avoid it by not emitting warnings when we cannot detect `p_t` from `p_0 .. p_{n-1}`?

Yes. False positives are my concern.
`break true unless FileTest.executable?(current_path)` breaks when the `current_path` does not exist.
So, it causes false positives.

> > To avoid false positives, we can consider the following direction for example.
> >
> > Log message:
> >
> > ```
> > ... [warn] ... Unable to confirm the existence of #{path}.
> > You can ignore this warning log if the file is created later.
> > If the file already exists, please check the configuration and directory permissions.
> > ```
> I agree with this direction if the above approach is not effective!

If it is difficult to resolve this false positive, let's go in this direction.

I explain specific examples of the false positive below. 

# Case

For example, consider the following case.

```console
$ sudo ls -lR /tmp/2
/tmp/2:
合計 12
drwxrwx--- 2 root root 4096  3月 28 13:47 d---
drwxrwx--x 2 root root 4096  3月 28 13:47 d--x
drwxrwxr-x 2 root root 4096  3月 28 13:46 dr-x

/tmp/2/d---:
合計 8
-rw-r----- 1 root root 6  3月 28 13:47 file_---.log
-rw-r--r-- 1 root root 5  3月 28 13:47 file_r--.log

/tmp/2/d--x:
合計 8
-rw-r----- 1 root root 6  3月 28 13:47 file_---.log
-rw-r--r-- 1 root root 5  3月 28 13:47 file_r--.log

/tmp/2/dr-x:
合計 8
-rw-rw---- 1 root root 6  3月 28 13:46 file_---.log
-rw-rw-r-- 1 root root 5  3月 28 13:37 file_r--.log
```

Let in_tail collect these 6 patterns and non-existent path (totally 7 paths).

```xml
<source>
  @type tail
  tag test
  path /tmp/2/dr-x/file_r--.log, /tmp/2/dr-x/file_---.log, /tmp/2/d--x/file_r--.log, /tmp/2/d--x/file_---.log, /tmp/2/d---/file_r--.log, /tmp/2/d---/file_---.log, /tmp/2/not_exist/test.log
  read_from_head true
  refresh_interval 5s
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

# Result: master

|         path         | read | warning |
|----------------------|------|---------|
| `dr-x/file_r--.log`  | y    | -       |
| `dr-x/file_---.log`  | n    | y       |
| `d--x/file_r--.log`  | y    | -       |
| `d--x/file_---.log`  | n    | y       |
| `d---/file_r--.log`  | n    | n       |
| `d---/file_---.log`  | n    | n       |
| `not_exist/test.log` | -    | n       |

```
[info]: #0 following tail of /tmp/2/dr-x/file_r--.log
{datetime} test: {"message":"test"}
[info]: #0 following tail of /tmp/2/dr-x/file_---.log
[warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
[info]: #0 following tail of /tmp/2/d--x/file_r--.log
{datetime} test: {"message":"test"}
[info]: #0 following tail of /tmp/2/d--x/file_---.log
[warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
```

# Result: PR

|         path         | read |  warning   |
|----------------------|------|------------|
| `dr-x/file_r--.log`  | y    | -          |
| `dr-x/file_---.log`  | n    | y          |
| `d--x/file_r--.log`  | y    | -          |
| `d--x/file_---.log`  | n    | y          |
| `d---/file_r--.log`  | n    | **y**      |
| `d---/file_---.log`  | n    | **y**      |
| `not_exist/test.log` | -    | **y** (FP) |

```
[warn]: #0 Skip /tmp/2/d---/file_r--.log because a directory in the path lacks execute permission.
[warn]: #0 Skip /tmp/2/d---/file_---.log because a directory in the path lacks execute permission.
[warn]: #0 Skip /tmp/2/not_exist/test.log because a directory in the path lacks execute permission.
[info]: #0 following tail of /tmp/2/dr-x/file_r--.log
{datetime} test: {"message":"test"}
[info]: #0 following tail of /tmp/2/dr-x/file_---.log
[warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
[info]: #0 following tail of /tmp/2/d--x/file_r--.log
{datetime} test: {"message":"test"}
[info]: #0 following tail of /tmp/2/d--x/file_---.log
[warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
```

moatom:
> If it is difficult to resolve this false positive, let's go in this direction.

Thanks for explaining the case!
I'll try to investigate this when I have time.

I'm thinking it might work if I switch the traversal direction in the code above to top-down instead.
https://github.com/fluent/fluentd/pull/4865#discussion_r2008038867

moatom:
@daipom 
I have fixed this PR to work for all cases!
If it looks good, should I squash the commits and add all 7 test cases to `test/plugin/test_in_tail.rb`?

<h1 dir="auto">Result: Latest PR</h1>


```xml
<source>
  @type tail
  tag test
  path /tmp/2/dr-x/file_r--.log, /tmp/2/dr-x/file_---.log, /tmp/2/d--x/file_r--.log, /tmp/2/d--x/file_---.log, /tmp/2/d---/file_r--.log, /tmp/2/d---/file_---.log, /tmp/2/not_exist/test.log
  pos_file /tmp/fluentd-test.pos
  read_from_head true
  refresh_interval 5s
  <parse>
    @type none
  </parse>
</source>

<match test.**>
  @type stdout
</match>
```

path | read | warning
-- | -- | --
`dr-x/file_r--.log` | y | -
`dr-x/file_---.log` | n | y
`d--x/file_r--.log` | y | -
`d--x/file_---.log` | n | y
`d---/file_r--.log` | n | **y**
`d---/file_---.log` | n | **y**
`not_exist/test.log` | - | **n**

```shell
$ sudo ls -lR /tmp/2
/tmp/2:
total 12
drwxrwx--- 2 root root 4096 Apr 11 14:09 d---
drwxrwx--x 2 root root 4096 Apr 11 14:09 d--x
drwxrwxr-x 2 root root 4096 Apr 11 14:09 dr-x

/tmp/2/d---:
total 8
-rw-r----- 1 root root 5 Apr 11 14:11 file_---.log
-rw-r--r-- 1 root root 5 Apr 11 14:11 file_r--.log

/tmp/2/d--x:
total 8
-rw-r----- 1 root root 5 Apr 11 14:11 file_---.log
-rw-r--r-- 1 root root 5 Apr 11 14:11 file_r--.log

/tmp/2/dr-x:
total 8
-rw-rw---- 1 root root 5 Apr 11 14:11 file_---.log
-rw-rw-r-- 1 root root 5 Apr 11 14:11 file_r--.log

$ bundle exec fluentd -c in_tail.conf
025-04-11 15:44:22 +0000 [warn]: #0 Skip /tmp/2/d---/file_r--.log because '/tmp/2/d---' lacks execute permission.
2025-04-11 15:44:22 +0000 [warn]: #0 Skip /tmp/2/d---/file_---.log because '/tmp/2/d---' lacks execute permission.
2025-04-11 15:44:22 +0000 [info]: #0 following tail of /tmp/2/dr-x/file_r--.log
2025-04-11 15:44:22.441945749 +0000 test: {"message":"test"}
2025-04-11 15:44:22 +0000 [info]: #0 following tail of /tmp/2/dr-x/file_---.log
2025-04-11 15:44:22 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-11 15:44:22 +0000 [info]: #0 following tail of /tmp/2/d--x/file_r--.log
2025-04-11 15:44:22.442540302 +0000 test: {"message":"test"}
2025-04-11 15:44:22 +0000 [info]: #0 following tail of /tmp/2/d--x/file_---.log
2025-04-11 15:44:22 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
```

daipom:
@moatom Thanks! Excellent! This direction looks good to me!
I will comment on some minor code improvements later!
We may need to consider how it works on Windows.

> If it looks good, should I squash the commits and add all 7 test cases to test/plugin/test_in_tail.rb

Sqush would be good!
For test cases, the current two would be sufficient.

> P.S. I'm not sure, but it might be a false negative when even / lacks the required permissions in this approach. (Should I mention this in a comment?)

Do you mean the case where we just specify `/` as `path`?

moatom:
> Do you mean the case where we just specify / as path?

I was referring to the case where the path is `/not_exist_file.log`, and the root directory `/` lacks the permissions.

In my PR, the logic relies on finding the first directory in the path hierarchy for which `!p.executable?` holds. However, in the situation where `/` itself lacks the permissions, it is uncertain whether `p.directory?` can reliably determine if `/` is a directory in the first place.

This scenario is rare and it’s unclear whether such a case should be officially supported.

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,7 +370,17 @@ def expand_paths
       # filter out non existing files, so in case pattern is without '*' we don't do unnecessary work
       hash = {}
       (paths - excluded).select { |path|
-        FileTest.exist?(path)
+        if check_permission && !File.readable?(path)
```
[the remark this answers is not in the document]
moatom: `File.readable?` => `File.exist?` holds.
`File.exist?` => `File.readable?` does not hold because the log file in the the path might lack the r permission.

Therefore, `File.exist?` appears likely more appropriate!

lib/fluent/plugin/in_tail.rb
```diff
@@ -370,7 +370,17 @@ def expand_paths
...
+          inaccessible_dir = Pathname.new(File.expand_path(path))
```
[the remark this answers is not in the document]
moatom: > Do we need File.expand_path?

I believe so.

We want `Pathname.new(path).ascend.reverse_each.find` to start ascending from the root directory `/`.
If `path` is a relative path, `ascend` begins from the current working directory `.`, not from `/`.

lib/fluent/plugin/in_tail.rb
```diff
@@ -253,7 +253,7 @@ def start
         FileUtils.mkdir_p(pos_file_dir, mode: @dir_perm) unless Dir.exist?(pos_file_dir)
         @pf_file = File.open(@pos_file, File::RDWR|File::CREAT|File::BINARY, @file_perm)
         @pf_file.sync = true
-        @pf = PositionFile.load(@pf_file, @follow_inodes, expand_paths, logger: log)
+        @pf = PositionFile.load(@pf_file, @follow_inodes, expand_paths(true), logger: log)
```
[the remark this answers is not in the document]
moatom: It appears likely that configure (and thus the warnings) is called twice in this implementation.
What do you think about this?


```bash
$ bundle exec fluentd -c example/in_tail.conf
2025-04-18 17:06:33 +0000 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2025-04-18 17:06:33 +0000 [info]: parsing config file is succeeded path="example/in_tail.conf"
2025-04-18 17:06:33 +0000 [info]: gem 'fluentd' version '1.18.0'
2025-04-18 17:06:33 +0000 [warn]: Skip /tmp/2/d---/file_r--.log because '/tmp/2/d---' lacks execute permission.
2025-04-18 17:06:33 +0000 [warn]: Skip /tmp/2/d---/file_---.log because '/tmp/2/d---' lacks execute permission.
2025-04-18 17:06:33 +0000 [info]: using configuration file: <ROOT>
  <source>
    @type tail
    tag "test"
    path "/tmp/2/dr-x/file_r--.log, /tmp/2/dr-x/file_---.log, /tmp/2/d--x/file_r--.log, /tmp/2/d--x/file_---.log, /tmp/2/d---/file_r--.log, /tmp/2/d---/file_---.log, /tmp/2/not_exist/test.log"
    read_from_head true
    refresh_interval 5s
    pos_file "/tmp/fluentd-test-5.pos"
    <parse>
      @type "none"
      unmatched_lines 
    </parse>
  </source>
  <match test.**>
    @type stdout
  </match>
</ROOT>
2025-04-18 17:06:33 +0000 [info]: starting fluentd-1.18.0 pid=49840 ruby="3.2.8"
2025-04-18 17:06:33 +0000 [info]: spawn command to main:  cmdline=["/usr/local/bin/ruby", "-r/usr/local/rvm/gems/default/gems/bundler-2.6.5/lib/bundler/setup", "-Eascii-8bit:ascii-8bit", "/workspaces/fluentd/vendor/bundle/ruby/3.2.0/bin/fluentd", "-c", "example/in_tail.conf", "--under-supervisor"]
2025-04-18 17:06:34 +0000 [info]: #0 init worker0 logger path=nil rotate_age=nil rotate_size=nil
2025-04-18 17:06:34 +0000 [info]: adding match pattern="test.**" type="stdout"
2025-04-18 17:06:34 +0000 [info]: adding source type="tail"
2025-04-18 17:06:34 +0000 [warn]: #0 Skip /tmp/2/d---/file_r--.log because '/tmp/2/d---' lacks execute permission.
2025-04-18 17:06:34 +0000 [warn]: #0 Skip /tmp/2/d---/file_---.log because '/tmp/2/d---' lacks execute permission.
2025-04-18 17:06:34 +0000 [info]: #0 starting fluentd worker pid=49850 ppid=49840 worker=0
2025-04-18 17:06:34 +0000 [info]: #0 following tail of /tmp/2/dr-x/file_r--.log
2025-04-18 17:06:34.458192953 +0000 test: {"message":"test"}
2025-04-18 17:06:34 +0000 [info]: #0 following tail of /tmp/2/dr-x/file_---.log
2025-04-18 17:06:34 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-18 17:06:34 +0000 [info]: #0 following tail of /tmp/2/d--x/file_r--.log
2025-04-18 17:06:34.458481605 +0000 test: {"message":"test"}
2025-04-18 17:06:34 +0000 [info]: #0 following tail of /tmp/2/d--x/file_---.log
2025-04-18 17:06:34 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
2025-04-18 17:06:34 +0000 [info]: #0 fluentd worker is now running worker=0
2025-04-18 17:06:35 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-18 17:06:35 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
2025-04-18 17:06:36 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-18 17:06:36 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
2025-04-18 17:06:37 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-18 17:06:37 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
2025-04-18 17:06:38 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/dr-x/file_---.log
2025-04-18 17:06:38 +0000 [warn]: #0 Permission denied @ rb_sysopen - /tmp/2/d--x/file_---.log
^C2025-04-18 17:06:38 +0000 [info]: Received graceful stop
```

lib/fluent/plugin/in_tail.rb
```diff
@@ -165,6 +165,7 @@ def configure(conf)
         @path_formatters = @paths.map{|path| [path, Fluent::Timezone.formatter(@path_timezone, path)]}.to_h
         @exclude_path_formatters = @exclude_path.map{|path| [path, Fluent::Timezone.formatter(@path_timezone, path)]}.to_h
       end
+      check_dir_permission
```
moatom, line 168: >  We may need to consider how it works on Windows.

I am not very familiar with Windows permissions.
Is `check_dir_permission unless Fluent.windows?` good enough for now?

moatom:
@daipom 
Hi! I've fixed all the points.
If everything looks good to you, I'll squash the commits.

daipom:
Thanks! Sorry for the delay, I am seeing this. Please wait some more.

lib/fluent/plugin/in_tail.rb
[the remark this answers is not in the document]
daipom: I see!
It's hard to imagine a case where we specify a relative path, but maybe we should do this just in case.

lib/fluent/plugin/in_tail.rb
daipom: Yes!
Let's skip this check on Windows.

I did a quick check, and it looks like the `executable?` confirmation doesn't make sense on Windows.

moatom: @daipom
Fixed this and squashed the commits!

lib/fluent/plugin/in_tail.rb
```diff
@@ -212,6 +213,20 @@ def configure(conf)
       @metrics = MetricsInfo.new(opened_file_metrics, closed_file_metrics, rotated_file_metrics, throttling_metrics)
     end
 
+    def check_dir_permission
+      expand_paths_raw.select { |path|
+        if !File.exist?(path)
```
daipom, line 218:
```suggestion
      expand_paths_raw.select { |path|
        not File.exist?(path)
      }.each { |path|
```

Sorry for my late response!

It would be more natural to separate `select` like this, or use `each` from the beginning.

moatom: @daipom Fixed!

lib/fluent/plugin/in_tail.rb
```diff
@@ -212,6 +213,20 @@ def configure(conf)
...
+    def check_dir_permission
+      expand_paths_raw.select { |path|
+        not File.exist?(path)
+      }.each { |path|
+        inaccessible_dir = Pathname.new(File.expand_path(path))
+          .ascend
+          .reverse_each
+          .find { |p| p.directory? && !p.executable? }
+        if inaccessible_dir
+          $log.warn "Skip #{path} because '#{inaccessible_dir}' lacks execute permission."
```
daipom, line 225:
```suggestion
          log.warn "Skip #{path} because '#{inaccessible_dir}' lacks execute permission."
```

test/plugin/test_in_tail.rb
```diff
@@ -2591,6 +2591,50 @@ def test_EACCES
...
+      FileUtils.chmod(0600, "#{@tmp_dir}/noaccess")
+      config = config_element('', '', {
+                                'tag' => "tail",
+                                'path' => path,
+                                'format' => 'none',
+                                "pos_file" => "#{@tmp_dir}/tail.pos",
+                              })
+      d = create_driver(config, false)
+      fname = File.expand_path("#{@tmp_dir}/noaccess")
+      assert($log.out.logs.any?{|log| log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
```
daipom, line 2611:
```suggestion
      assert(d.logs.any?{|log| log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
```

test/plugin/test_in_tail.rb
```diff
@@ -2591,6 +2591,50 @@ def test_EACCES
...
+    end
+  ensure
+    d.instance_shutdown if d && d.instance
```
daipom, line 2614:
```suggestion
```

test/plugin/test_in_tail.rb
```diff
@@ -2591,6 +2591,50 @@ def test_EACCES
...
+      FileUtils.chmod(0100, "#{@tmp_dir}/noaccess")
+      config = config_element('', '', {
+                                'tag' => "tail",
+                                'path' => path,
+                                'format' => 'none',
+                                "pos_file" => "#{@tmp_dir}/tail.pos",
+                              })
+      d = create_driver(config, false)
+      fname = File.expand_path("#{@tmp_dir}/noaccess")
+      assert($log.out.logs.all?{|log| !log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
```
daipom, line 2632:
```suggestion
      assert(d.logs.all?{|log| !log.include?("Skip #{path} because '#{fname}' lacks execute permission.\n") })
```

test/plugin/test_in_tail.rb
```diff
@@ -2591,6 +2591,50 @@ def test_EACCES
...
+    end
+  ensure
+    d.instance_shutdown if d && d.instance
```
daipom, line 2635:
```suggestion
```

daipom (commented):
@moatom 
Sorry for the repeated reviews.
After we fix this, I think we can merge this!

Review points:

* The `log` accessor is provided for plugin class, so we should use it in plugin implementations.
  * We should not use the global logger in plugins. Some existing codes are wrong... :cry: 
  * Sorry I didn't notice this before.
* In tests, we can use `Driver#logs` to get the logs.
  * If the plugin uses the global logger, this cannot be used.
  * We want to avoid using the global logger for tests as much as possible.
* We don't need to call `Driver#instance_shutdown` manually unless we call  `Driver#run` with `shutdown: false`.

moatom:
@daipom 
> Sorry for the repeated reviews.

No problem at all. I really appreciate your prompt responses!

> After we fix this, I think we can merge this!

I've fixed the points. Could you confirm at your convenience?

daipom (approved):
LGTM! Thanks so much for this improvement!

PR #4866, opened 2025-03-16
in_http: allow empty Origin header requests to pass CORS checks

dlackty:

Some requests, such as those made by apps, certain automated scripts, or older browsers, may not include an Origin header. Previously, such requests were blocked by the CORS check, even though they may not necessarily be cross-origin.

For CORS, the server is responsible for reporting the allowed origins. The web browser is responsible for enforcing that requests are only sent from allowed domains. So this change updates the CORS handling logic to allow requests with an empty Origin header to pass, ensuring compatibility with legitimate non-browser clients while maintaining security.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/574

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -504,8 +504,9 @@ module Fluent::Plugin
         # ==========
         # For every incoming request, we check if we have some CORS
         # restrictions and allow listed origins through @cors_allow_origins.
+        # If origin is empty, it's likely a server-to-server request and considered safe.
         unless @cors_allow_origins.nil?
-          unless @cors_allow_origins.include?('*') || include_cors_allow_origin
+          unless @cors_allow_origins.include?('*') || include_cors_allow_origin || @origin.nil?
             send_response_and_close(RES_403_STATUS, {'Connection' => 'close'}, "")
             return
           end
--- a/test/plugin/test_in_http.rb
+++ b/test/plugin/test_in_http.rb
@@ -940,6 +940,26 @@ class HttpInputTest < Test::Unit::TestCase
     end
   end
 
+  def test_cors_with_nil_origin
+    d = create_driver(config + %[
+      cors_allow_origins ["http://foo.com"]
+    ])
+    assert_equal ["http://foo.com"], d.instance.cors_allow_origins
+
+    time = event_time("2011-01-02 13:14:15 UTC")
+    event = ["tag1", time, {"a"=>1}]
+    res_code = nil
+
+    d.run do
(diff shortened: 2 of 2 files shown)


daipom:
@dlackty Thanks for this PR!
I will review this soon.

daipom:
> For CORS, the server is responsible for reporting the allowed origins. The web browser is responsible for enforcing that requests are only sent from allowed domains.

I see.
I am not familiar with CORS, but it certainly seems to be true.
CORS does not protect the server.
The browser makes the decision on whether to allow the communication.

Given this CORS specification, this `in_http`'s specification, which returns 403 depending on `cors_allow_origins`, would not necessarily be required.
So, it would be possible to relax the condition as this PR.

On the other hand, it would also be possible that `in_http` intentionally checks the condition strictly, although it is not a CORS specification.
If there is a possibility that some users are intentionally using the current strict feature, it may be better not to change the default behavior in order to keep compatibility.

I would like to hear opinions on this point.

Note: 

* https://docs.fluentd.org/input/http#cors_allow_origins
  * > If you set ["domain1", "domain2"] to cors_allow_origins, in_http returns 403 to access from other domains.
  * Looks like `in_http` intentionally checks the condition strictly, although it is not a CORS specification.
* #473
  * Looks like it was initially implemented not as a CORS, but simply as a allow list. It doesn't look like this was intentional.
* #882
  * Looks like the proper CORS behavior was added later.

daipom (approved):
At least, there would be no problem to relax the condition like this PR.
(In the first place, I don't think it is necessary to return 403.)

I will wait a while, and if there are no objections, I will merge this.
Thanks!

dlackty:
Suggestion: To avoid breaking compatibility while still allowing requests with an empty Origin header, we could consider allowing `cors_allow_origins` to include `nil` as a valid value.

For example:

```ruby
cors_allow_origins [nil, "example.com", "another-domain.com"]
```

This way:
	•	Existing strict behavior remains unchanged by default.
	•	Users who explicitly want to allow empty Origin headers can opt in by adding nil.
	•	This keeps the logic flexible without affecting current users who rely on the strict enforcement.

Would love to hear your thoughts on this approach! 🚀

daipom:
I see! Thanks!
So, we have a workaround. We can use this workaround for the current version.

Even if we have a workaround, I see no problem with this change.

If anyone has a different opinion, please let me know.
I will wait a while, and if there are no objections, I will merge this for v1.19.0.

daipom:
Thanks @dlackty !

PR #4870, opened 2025-03-18
chunk: rename open method name to open_io

Watson1978:
Fixes #

If we have used `super` in `open` method, CodeQL scan detects that it might call `Kernel.#open` and it is judged as danger.

https://github.com/fluent/fluentd/blob/7dbf9b36c392c9e4ad769b7b462c90ee4f5a36d3/lib/fluent/plugin/buffer/chunk.rb#L269-L279

But `Kernel.#open` is never called in this super, so, I think CodeQL judgement is a false positive...

However, this judgement may be misunderstood for many people, so we will avoid this judgement by changing the method name.

**Docs Changes**:

**Release Note**:


Watson1978:
I closed the alert with dismiss alert

PR #4872, opened 2025-03-19
Draft: chunk: fix vulnerability warning for kernel open

daipom:
* Another idea for #4870

The same as #4870

**Docs Changes**:
Not needed.

**Release Note**: 
TODO


daipom:
It appears that CodeQL does not recognize this guard code.

Issue #4808, opened 2025-01-29
Memory spike while using suppress and throttling plugin


akshaysharama:
### Describe the bug

Using these plugins to discard the duplicate logs from the fluentd.

Trying to remove duplicate logs from the k8s cluster microservices, reading logs from the /var/log/containers and removing duplicate log to improve readbility.

### To Reproduce

<filter kubernetes.**>
          @type suppress
          log_suppress_interval 10      # Suppress duplicates for 10 seconds
          num 1                          # Suppress after 2 identical logs
          max_slot_num 100000           # Max unique logs to track
          attr_keys short_message         # Attributes to check for duplicates
          add_tag_prefix sp.            # Prefix for suppressed logs
</filter>
<filter sp.*>
          @type throttle
          threshold 1                   # Limit to 5 logs per interval
          interval 300                   # Time interval in seconds
          add_tag_prefix th.            # Prefix for throttled logs
          enable_throttle true
          discard_throttled true
</filter>

### Expected behavior

Memory shoot-up should not happen

### Your Environment

```markdown
- Fluentd version:5.9.11 (Chart: https://charts.bitnami.com/bitnami)
- Package version:
- Operating system:Ubuntu 22.04 LTS
- Kernel version::5.15.0-124-generic
```

### Your Configuration

```apache
<filter kubernetes.**>
          @type suppress
          log_suppress_interval 10      # Suppress duplicates for 10 seconds
          num 1                          # Suppress after 2 identical logs
          max_slot_num 100000           # Max unique logs to track
          attr_keys short_message         # Attributes to check for duplicates
          add_tag_prefix sp.            # Prefix for suppressed logs
</filter>
<filter sp.*>
          @type throttle
          threshold 1                   # Limit to 5 logs per interval
          interval 300                   # Time interval in seconds
          add_tag_prefix th.            # Prefix for throttled logs
          enable_throttle true
          discard_throttled true
</filter>.
```

### Your Error Log

```shell
No error logs only seeing memory spike
```

akshaysharama:
@daipom can you please here? sorry to tag you directly, I have gone through issues raised and observed you are actively looks on plugin issues.  sorry once again, but memory spike is messing our test bed so we need some help here.

Watson1978:
I found fluent-plugin-suppress increase memory usage when many long log messages are duplicated

akshaysharama:
@Watson1978 , will be getting a fix soon or how will proceed currently?

[...]

PR #4881, opened 2025-03-25
in_tail: reduce memory usage slightly

Watson1978:
Fixes #

Previous, I changed lifespan of `iobuf` at https://github.com/fluent/fluentd/pull/4763 to work GC collector well.

In addition, it seems that clearing `iobuf` manually when it is no longer needed, it might improve memory efficiency.

memory_usage

* config
```
<source>
  @type tail
  tag log
  refresh_interval 5s
  <parse>
    @type none
  </parse>
</source>

<match **>
  @type file
</match>
```

* script to generate log data
```
require "json"

path = File.expand_path("~/tmp/access.log")

File.open(path, "w") do |f|
  loop do
    log = { time: Time.now, message: "a" * 10_000 }.to_json
    f.puts log
    sleep 0.05
  end
end
```

Related to https://github.com/fluent/fluentd/issues/4808

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -1238,6 +1238,8 @@ module Fluent::Plugin
                   end
                 rescue EOFError
                   @eof = true
+                ensure
+                  iobuf.clear
                 end
               end
 


daipom (approved):
Thanks!

PR #4884, opened 2025-03-26
formatter_out_file: reduce memory usage

Watson1978:
Fixes #

When dealing with long logs at fast, it might improve memory efficient to clear the original strings used in string interpolation.

chart

* config
```
<source>
  @type tail
  tag log
  refresh_interval 5s
  <parse>
    @type none
  </parse>
</source>

<match **>
  @type file
</match>
```

* script to generate log data
```ruby
require "json"

path = File.expand_path("~/tmp/access.log")

File.open(path, "w") do |f|
  loop do
    log = { time: Time.now, message: "a" * 50 * 1024 }.to_json
    f.puts log
    sleep 0.0001
  end
end
```

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/formatter_out_file.rb
+++ b/lib/fluent/plugin/formatter_out_file.rb
@@ -43,10 +43,13 @@ module Fluent
       end
 
       def format(tag, time, record)
+        json_str = JSON.generate(record)
         header = ''
         header << "#{@timef.format(time)}#{@delimiter}" if @output_time
         header << "#{tag}#{@delimiter}" if @output_tag
-        "#{header}#{JSON.generate(record)}#{@newline}"
+        "#{header}#{json_str}#{@newline}"
+      ensure
+        json_str&.clear
       end
     end
   end


Watson1978:
Here is a simple memory benchmark.
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-memory'
end

Benchmark.memory do |x|
  x.report('interpolation') {
    begin
      str1 = 'a' * 1_000
      "#{str1}\n"
    end
  }

  x.report('interpolation + clear') {
    begin
      str1 = 'a' * 1_000
      "#{str1}\n"
    ensure
      str1.clear
    end
  }
end
```

```
$ ruby -v bench.rb
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
       interpolation     2.124k memsize (     0.000  retained)
                         3.000  objects (     0.000  retained)
                         3.000  strings (     0.000  retained)
interpolation + clear
                         1.123k memsize (     0.000  retained)
                         3.000  objects (     0.000  retained)
                         3.000  strings (     0.000  retained)
```

daipom:
I talked to @Watson1978.

> Hmm, `String#clear` frees the allocated memory. So, it may have a certain effect.
> 
> However, it appears to me that normally this free should be done by minor GC.

The aim of this PR is to improve memory consumption when large records are input in a short period of time.

> It could be not effective in cases there is memory pressure, but rather may be overloaded with record-by-record freeing.

Since they are released by GC anyway, wouldn't the load be the same if they were released for each record?
(We don't know the details. Might need to check throughput if possible.)

Watson1978:
With 100 KB message.

```ruby
require "json"

path = File.expand_path("~/tmp/access.log")

File.open(path, "w") do |f|
  loop do
    log = { time: Time.now, message: "a" * 100 * 1024 }.to_json
    f.puts log
    sleep 0.0001
  end
end
```
chart

daipom:
Worker process RSS

Purple: master
Green: PR

Screenshot from 2025-03-27 10-59-38

```xml
<source>
  @type sample
  tag test
  sample {"message":"#{"a"*100_000}"}
  size 100
</source>

<match test.**>
  @type file
  path /dev/null
  append
  add_path_suffix false
  <buffer []>
    @type memory
    flush_mode interval
    flush_interval 3s
    chunk_limit_size 1g
    chunk_limit_size 3g
  </buffer>
</match>
```

Watson1978:
I tried with your config.

size 100 | size 500
--|--
 | 

- size 100
```
<source>
  @type sample
  tag test
  sample {"message":"#{"a"*100_000}"}
  size 100
</source>
```

- size 500
```
<source>
  @type sample
  tag test
  sample {"message":"#{"a"*100_000}"}
  size 500
</source>
```

daipom:
As a check of processing speed, I checked the metrics of `emit_records` of the `out_file`.
(Emit so many records that emit is not completed in 1 second)

Against my concerns, I confirm that this PR increases the speed.

Screenshot from 2025-03-27 12-30-39

```xml
<source>
  @type sample
  tag test
  sample {"message":"#{"a"*1000}"}
  size 1_000_000
</source>

<source>
  @type monitor_agent
</source>

<match test.**>
  @type file
  @id output
  path /dev/null
  append
  add_path_suffix false
  <buffer []>
    @type memory
    flush_mode interval
    flush_interval 1s
    chunk_limit_size 5g
    chunk_limit_size 10g
  </buffer>
</match>
```

lib/fluent/plugin/formatter_out_file.rb
```diff
@@ -43,10 +43,13 @@ def configure(conf)
...
 
       def format(tag, time, record)
+        json_str = JSON.generate(record)
         header = ''
         header << "#{@timef.format(time)}#{@delimiter}" if @output_time
         header << "#{tag}#{@delimiter}" if @output_tag
-        "#{header}#{JSON.generate(record)}#{@newline}"
+        "#{header}#{json_str}#{@newline}"
+      ensure
+        json_str.clear
```
daipom, line 52:
```suggestion
        result = "#{header}#{json_str}#{@newline}"
        json_str.clear
        result
```

I'm concerned that `json_str` can be `nil`.
There would be 2 cases:

* `JSON.generate` raises an exception.
* The process is killed during `JSON.generate`.
  * I encountered this case during testing performance.
    * ``[warn]: #0 emit transaction failed: error_class=NoMethodError error="undefined method `clear' for nil:NilClass"``

It would not be a problem if `String#clear` cannot be performed.
So, can we remove `ensure`?

Watson1978: Indeed.

I will fix that using `&.`.

daipom (approved):
Thanks!

PR #4886, opened 2025-04-01
formatter_json: reduce memory usage

Watson1978:
Fixes #

When dealing with long logs at fast, it might improve memory efficient to clear the original strings used in string interpolation.

This is similar with #4884 .

chart

* config
```
<source>
  @type sample
  tag test
  sample {"message": "#{'a' * 100 * 1024}"}
  size 100
</source>

<match **>
  @type file
  format json
</match>
```

**Docs Changes**:

**Release Note**:


daipom (approved):
Thanks!

PR #4893, opened 2025-04-04
v1parser: support built-in config files

kenhys:

This is just a PoC.

Currently `@include` directive supports to reuse configuration files, but no way to
apply specific configuration files by default (without specifying to).

For example, if this feature was merged, a specific configuration could be applied in any case for fluent-package.
By putting some conf files under /etc/fluent/conf.d/*.conf. 

**Docs Changes**:

T.B.D.

**Release Note**:

### DIFF
--- a/lib/fluent/config/v1_parser.rb
+++ b/lib/fluent/config/v1_parser.rb
@@ -16,6 +16,7 @@
 
 require 'strscan'
 require 'uri'
+require 'find'
 
 require 'fluent/config/error'
 require 'fluent/config/basic_parser'
@@ -38,6 +39,7 @@ module Fluent
         @include_basepath = include_basepath
         @fname = fname
         @logger = defined?($log) ? $log : nil
+        @included_conf_files = []
       end
 
       def parse!
@@ -57,6 +59,19 @@ module Fluent
       RESERVED_PARAMS = %W(@type @id @label @log_level)
 
       def parse_element(root_element, elem_name, attrs = {}, elems = [])
+        begin
+          Find.find("#{@include_basepath}/conf.d/") do |path|
+            next if File.directory?(path)
+            unless @included_conf_files.include?(File.expand_path(path))
+              config_full_path = File.expand_path(path)
+              @included_conf_files << config_full_path
+              @logger.info "include built-in #{config_full_path}" if @logger
+              eval_include(attrs, elems, path)
+            end
+          end
+        rescue Errno::ENOENT
+          @logger.info "no built-in config files under #{@include_basepath}/conf.d/*" if @logger
+        end
         while true
           spacing
           if eof?
@@ -143,6 +158,9 @@ module Fluent
 
       def parse_include(attrs, elems)
         uri = scan_string(LINE_END)
+        if @included_conf_files.include?(File.expand_path(File.join(@include_basepath, uri)))
+          return
+        end
         eval_include(attrs, elems, uri)
         line_end
       end


kenhys:
This is just a PoC.

lib/fluent/system_config.rb
```diff
@@ -58,6 +59,7 @@ class SystemConfig
     config_param :dir_permission, default: nil do |v|
       v.to_i(8)
     end
+    config_param :config_include_dir, default: ENV["FLUENT_CONF_DIR"] || '/etc/fluent/conf.d'
```
Watson1978, line 62: It might be better to declare this as a constant in the https://github.com/fluent/fluentd/blob/510b891de1fb8a81065c8016bc5ac4f0eece4d63/lib/fluent/env.rb#L23-L26

lib/fluent/supervisor.rb
```diff
@@ -806,6 +807,53 @@ def configure(supervisor: false)
 
       $log.info :supervisor, 'parsing config file is succeeded', path: @config_path
 
+      # merge additional configuration
+      config_type = if File.extname(@config_path) == '.yaml' || File.extname(@config_path) == '.yml'
+                      :yaml
+                    else
+                      nil
+                    end
```
Watson1978, line 815: As the file extensions are checked some times, I think it is betther to create a method, like:

```ruby
def config_type(config_path)
  if File.extname(config_path) == '.yaml' || File.extname(config_path) == '.yml'
    :yaml
  else
    nil
  end
end


  # merge additional configuration
  config_type = config_type(@config_path)

  ...

    if config_type == :yaml
      if config_type(path) != :yaml
        $log.warn :supervisor, 'skip loading configuration file', path: path
        next
      end
...
```

daipom (commented):
It would be necessary to add more tests.
At least, we want to confirm that some additional configs are merged correctly.

lib/fluent/system_config.rb
kenhys: Thanks, it is reasonable.

lib/fluent/supervisor.rb
kenhys: Thanks, I've took a bit differenct approach.

lib/fluent/supervisor.rb
```diff
@@ -806,6 +811,53 @@ def configure(supervisor: false)
...
+              elements = Fluent::Config.build(config_path: path,
+                                              encoding: @conf_encoding,
+                                              use_v1_config: @use_v1_config,
+                                              type: config_type).elements
+              @conf.merge_elements(elements)
+            end
+          end
+        rescue Errno::ENOENT
+          $log.warn :supervisor, 'no default configuration include directory was disabled', path: @system_config.config_include_dir
+        end
```
daipom, line 858: > no default configuration include directory was disabled

What kind of case do you assume?

kenhys: It was broken and inappropriate. Fixed it. thanks!

kenhys: I've intended the following case:

```
3.2.4 :003 > Find.find("/404") do |path| end
/home/kenhys/.rvm/rubies/ruby-3.2.4/lib/ruby/3.2.0/find.rb:43:in `block in find': No such file or directory - /404 (Errno::ENOENT)
        from /home/kenhys/.rvm/rubies/ruby-3.2.4/lib/ruby/3.2.0/find.rb:43:in `collect!'
        from /home/kenhys/.rvm/rubies/ruby-3.2.4/lib/ruby/3.2.0/find.rb:43:in `find'
        from (irb):3:in `<main>'                                                    
        from /home/kenhys/.rvm/rubies/ruby-3.2.4/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /home/kenhys/.rvm/rubies/ruby-3.2.4/bin/irb:25:in `load'               
        from /home/kenhys/.rvm/rubies/ruby-3.2.4/bin/irb:25:in `<main>'             
        from /home/kenhys/.rvm/gems/ruby-3.2.4@fluentd/bin/ruby_executable_hooks:22:in `eval'
        from /home/kenhys/.rvm/gems/ruby-3.2.4@fluentd/bin/ruby_executable_hooks:22:in `<main>'
```

daipom: I see! Thanks!

kenhys:
> At least, we want to confirm that some additional configs are merged correctly.

It should be, thanks! I'll try.

lib/fluent/config/element.rb
```diff
@@ -71,6 +71,10 @@ def add_element(name, arg = '')
         e
       end
 
+      def merge_elements(elements)
+        @elements.concat(elements)
+      end
+
```
daipom, line 77: Maybe we don't need this new method.

We can merge `Fluent::Config::Element` by

```rb
          @conf += Fluent::Config.build(config_path: path,
                                        encoding: @conf_encoding,
                                        use_v1_config: @use_v1_config,
                                        type: @config_file_type)
```

lib/fluent/supervisor.rb
```diff
@@ -806,6 +811,53 @@ def configure(supervisor: false)
...
 
+      # merge additional configuration
+      config_type = if File.extname(@config_path) == '.yaml' || File.extname(@config_path) == '.yml'
+                      :yaml
+                    else
+                      nil
+                    end
+      if @system_config.config_include_dir&.empty?
+        $log.warn :supervisor, 'default configuration include directory was disabled'
+      else
```
daipom, line 822: It looks to me that we don't need to warn a user, since the user has explicitly disabled this feature.

lib/fluent/supervisor.rb
```diff
@@ -806,6 +811,53 @@ def configure(supervisor: false)
...
+        begin
+          Find.find(@system_config.config_include_dir) do |path|
+            next if path == @system_config.config_include_dir
```
daipom, line 825:
```suggestion
            next unless File.file?(path)
```

`Find.find` search entries recursively including sub-directories, right?
If so, we need to skip all directories.

lib/fluent/supervisor.rb
daipom, line 1: We need to consider `reload`.

https://github.com/fluent/fluentd/blob/510b891de1fb8a81065c8016bc5ac4f0eece4d63/lib/fluent/supervisor.rb#L1080-L1105

lib/fluent/supervisor.rb
```diff
@@ -806,6 +811,53 @@ def configure(supervisor: false)
...
+              $log.info :supervisor, 'loading additional configuration file', path: path
+              # wrap additional YAML configuration to make it parsed correctly
+              buffer = "config:\n"
+              buffer << File.readlines(path).collect do |line|
+                "  " + line.chomp
+              end.join("\n")
+              Tempfile.open do |file|
+                file.puts(buffer)
+                file.flush
+                elements = Fluent::Config::YamlParser.parse(file.path).elements
```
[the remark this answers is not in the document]
daipom: Note: We talked about the fact that this feature is not compatible with the existing `include` feature, so its priority is low.
(The existing `include` feature can share parameters, but this feature does not support it.)

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1208,30 @@ def build_system_config(conf)
...
+                                        encoding: @conf_encoding,
+                                        use_v1_config: @use_v1_config,
+                                        type: "guess")
+          end
+        rescue Errno::ENOENT
+          $log.warn :supervisor, 'inaccessible include directory was specified', path: @system_config.config_include_dir
+        end
+      end
+      conf
+    end
```
daipom, line 1233: Don't we need to assume multiple files?

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1208,30 @@ def build_system_config(conf)
       system_config
     end
 
+    def build_additional_configuration
+      conf = nil
+      if @system_config.config_include_dir&.empty?
+        $log.info :supervisor, 'default configuration include directory was disabled'
```
daipom, line 1214: How about using early `return`?

```rb
if @system_config.config_include_dir&.empty?
  $log.info :supervisor, 'default configuration include directory was disabled'
  return nil
end
```

test/test_supervisor.rb
```diff
@@ -1054,6 +1054,47 @@ def test_stop_parallel_old_supervisor_after_delay
     end
   end
 
+  sub_test_case "include additional configuration" do
+    setup do
+      omit "Not supported on Windows" if Fluent.windows?
```
daipom, line 1059: > omit "Not supported on Windows" if Fluent.windows?

Is this feature not supported on Windows?

lib/fluent/supervisor.rb
kenhys: Thanks, it should be.

lib/fluent/supervisor.rb
kenhys: I've overlooked it, thanks.

test/test_supervisor.rb
kenhys: Not limited.
Initially skipped.

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1208,35 @@ def build_system_config(conf)
...
+          $log.info :supervisor, 'loading additional configuration file', path: path
+          element = Fluent::Config.build(config_path: path,
+                                         encoding: @conf_encoding,
+                                         use_v1_config: @use_v1_config,
+                                         type: :guess)
+          if conf
+            conf += element
+          else
+            conf = element
+          end
```
daipom, line 1232: How about using `yield` to avoid managing `conf`?

```rb
yield Fluent::Config.build(config_path: path,
                           encoding: @conf_encoding,
                           use_v1_config: @use_v1_config,
                           type: :guess) 
```

```rb
build_additional_configurations do |additional_conf|
  @conf += additional_conf
end
```

* Codes become simpler.
* if `return` without `yield`, then the block is not called, so we can omit `nil` check too.
* build_additional_configuration**s** would be better?

kenhys: Thanks, good idea.

lib/fluent/supervisor.rb
```diff
@@ -1090,6 +1095,10 @@ def reload_config
             type: @config_file_type,
           )
 
+          build_additional_configuration do |additional_conf|
+            @conf += additional_conf if additional_conf
```
daipom, line 1099:
```suggestion
            conf += additional_conf if additional_conf
```

This variable should be local here.

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1205,28 @@ def build_system_config(conf)
       system_config
     end
 
+    def build_additional_configuration
```
daipom, line 1208:
```suggestion
    def build_additional_configurations
```

How about a name that shows there can be more than one?

lib/fluent/env.rb
```diff
@@ -21,6 +21,7 @@
 
 module Fluent
   DEFAULT_CONFIG_PATH = ENV['FLUENT_CONF'] || '/etc/fluent/fluent.conf'
+  DEFAULT_CONFIG_INCLUDE_DIR = ENV["FLUENT_CONF_INCLUDE_DIR"] || '/etc/fluent/conf.d'
```
daipom, line 24: I'm sorry, maybe I'm not understanding something, but could you tell me why this needs to be an environment variable?

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1205,28 @@ def build_system_config(conf)
...
+          # NOTE: both types of normal config (.conf) and YAML will be loaded.
+          # Thus, it does not care whether @config_path is .conf or .yml.
+          $log.info :supervisor, 'loading additional configuration file', path: path
+          yield Fluent::Config.build(config_path: path,
+                                         encoding: @conf_encoding,
+                                         use_v1_config: @use_v1_config,
+                                         type: :guess)
+        end
+      rescue Errno::ENOENT
+        $log.warn :supervisor, 'inaccessible include directory was specified', path: @system_config.config_include_dir
```
daipom, line 1226:
```suggestion
        $log.info :supervisor, '"config_include_dir" does not exist. Does not load additional configuration files.', path: @system_config.config_include_dir
```

There are 2 points.

* How about changing the log level to `info`.
* It would be better to make the message specific so that the user knows what the warning is about.

In the current implementation, the following warning occurs when I run raw Fluentd by default.

```
2025-04-17 18:26:42 +0900 [warn]: inaccessible include directory was specified
 path="/etc/fluent/conf.d"
```

This would be confusing because a user specifies nothing, but it warns about a path the user may not know.

If we use `warn` level here, it would be better to limit it to cases where the user has explicitly configured it.
Mayby we don't have to go that far because it can be assumed that there are few cases where a user actually configures this path explicitly, other than to disable this feature in the package.

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1205,28 @@ def build_system_config(conf)
...
+        $log.info :supervisor, 'default configuration include directory was disabled'
+        return
+      end
+      begin
+        Find.find(@system_config.config_include_dir) do |path|
+          next if File.directory?(path)
+          next unless [".conf", ".yaml", ".yml"].include?(File.extname(path))
+          # NOTE: both types of normal config (.conf) and YAML will be loaded.
+          # Thus, it does not care whether @config_path is .conf or .yml.
+          $log.info :supervisor, 'loading additional configuration file', path: path
```
daipom, line 1219: > $log.info :supervisor, 'loading additional configuration file', path: path

I have noticed that these logs are not logged in the case of `reload_config`.

It is because `:supervisor` limits the message to be logged only when the process is a supervisor.

Since `reload_config` is not recommended, this may be acceptable as a specification.
I shared it anyway.

lib/fluent/supervisor.rb
```diff
@@ -806,6 +807,10 @@ def configure(supervisor: false)
 
       $log.info :supervisor, 'parsing config file is succeeded', path: @config_path
 
+      build_additional_configuration do |additional_conf|
+        @conf += additional_conf if additional_conf
```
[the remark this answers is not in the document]
kenhys: Thanks.

lib/fluent/supervisor.rb
kenhys: Oh, good catch.

lib/fluent/supervisor.rb
kenhys: Fixed, thanks.

lib/fluent/env.rb
kenhys: Mainly introduced customizing for fluent-package windows edition.
On windows /etc/fluent/conf.d is not appropriate.

daipom: I see! Thanks!

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1205,28 @@ def build_system_config(conf)
       system_config
     end
 
+    def build_additional_configurations
+      if @system_config.config_include_dir&.empty?
+        $log.info :supervisor, 'default configuration include directory was disabled'
```
daipom, line 1210:
```suggestion
        $log.info :supervisor, 'configuration include directory is disabled'
```

* It would be clearer if we omitted the word `default`.
  * In this case, the feature itself is disabled.
* Using the present tense would be clearer.

lib/fluent/supervisor.rb
```diff
@@ -1196,6 +1205,28 @@ def build_system_config(conf)
...
+        Find.find(@system_config.config_include_dir) do |path|
+          next if File.directory?(path)
+          next unless [".conf", ".yaml", ".yml"].include?(File.extname(path))
+          # NOTE: both types of normal config (.conf) and YAML will be loaded.
+          # Thus, it does not care whether @config_path is .conf or .yml.
+          $log.info :supervisor, 'loading additional configuration file', path: path
+          yield Fluent::Config.build(config_path: path,
+                                         encoding: @conf_encoding,
+                                         use_v1_config: @use_v1_config,
+                                         type: :guess)
```
daipom, line 1223:
```suggestion
          yield Fluent::Config.build(config_path: path,
                                     encoding: @conf_encoding,
                                     use_v1_config: @use_v1_config,
                                     type: :guess)
```

test/command/test_fluentd.rb
```diff
@@ -1536,4 +1536,76 @@ def send_end(port)
...
+
+    test "inaccessible include directory error" do
+      conf_path = create_config_include_dir_configuration("inaccessible_include.conf", "/nonexistent")
+      assert_log_matches(create_cmdline(conf_path),
+                         "[info]: inaccessible include directory was specified")
+    end
+
+    data("include additional configuration with relative conf.d" => [{"relative_path" => true}],
+         "include additional configuration with full-path conf.d" => [{"relative_path" => false}])
+    test "additional configuration file (conf.d/child.conf) was loaded" do |(option)|
```
daipom, line 1575:
```suggestion
    data("include additional configuration with relative conf.d" => {"relative_path" => true},
         "include additional configuration with full-path conf.d" => {"relative_path" => false})
    test "additional configuration file (conf.d/child.conf) was loaded" do |option|
```

test/command/test_fluentd.rb
```diff
@@ -1536,4 +1536,76 @@ def send_end(port)
...
+
+    test "inaccessible include directory error" do
+      conf_path = create_config_include_dir_configuration("inaccessible_include.yml", "/nonexistent", true)
+      assert_log_matches(create_cmdline(conf_path),
+                         "[info]: inaccessible include directory was specified")
+    end
+
+    data("include additional YAML configuration with relative conf.d" => [{"relative_path" => true}],
+         "include additional YAML configuration with full path conf.d" => [{"relative_path" => false}])
+    test "additional relative configuration file (conf.d/child.yml) was loaded" do |(option)|
```
daipom, line 1603:
```suggestion
    data("include additional YAML configuration with relative conf.d" => {"relative_path" => true},
         "include additional YAML configuration with full path conf.d" => {"relative_path" => false})
    test "additional relative configuration file (conf.d/child.yml) was loaded" do |option|
```

kenhys:
Rebased with master and squash feedback commits.

test/test_supervisor.rb
```diff
@@ -1054,6 +1054,129 @@ def test_stop_parallel_old_supervisor_after_delay
...
+    setup do
+      @config_include_dir = File.join(@tmp_dir, "conf.d")
+      FileUtils.mkdir_p(@config_include_dir)
+    end
+
+    test "no additional configuration" do
+      c = Fluent::Config::Element.new('system', '', { 'config_include_dir' => '' }, [])
+      stub(Fluent::Config).build { config_element('ROOT', '', {}, [c]) }
+      supervisor = Fluent::Supervisor.new({})
+      stub(supervisor).build_spawn_command { "dummy command line" }
```
daipom, line 1067:
```suggestion
```

Since this test does not call `Supervisor#run_supervisor`, we don't need to stub `Supervisor#build_spawn_command`.

kenhys:
Followed feedback (Use stub.proxy)

Watson1978 (approved):
👍🏻

daipom (approved):
LGTM! Thanks!

daipom:
Could you update the description of this PR?

Issue #4758, opened 2025-01-06
Update Forward Protocol Specification for zstd compression support


daipom:
### Is your feature request related to a problem? Please describe.

The following PR supports zstd compression.

* #4657

So, we need to update [Forward Protocol Specification - CompressedPackedForward Mode](https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5#compressedpackedforward-mode).

### Describe the solution you'd like

Update the following description and add `zstd` value to `compressed` option.

* https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5#compressedpackedforward-mode
* https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5#option

Would that be Forward Protocol Specification v1.6?

### Describe alternatives you've considered

Having no idea.

daipom:
I propose changes to the following two sections.

> CompressedPackedForward Mode

https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5#compressedpackedforward-mode

```diff
 ### CompressedPackedForward Mode
 
-It carries a series of events as a msgpack binary, compressed by gzip, on a single request. The supported compression algorithm is only gzip.
+It carries a series of events as a msgpack binary, compressed by gzip or zstd, on a single request.
 
-- `entries` is a gzipped binary chunk of `MessagePackEventStream`, which MAY be a concatenated binary of multiple gzip binary strings.
-- Client MUST send an option with `compressed` key with the value `gzip`.
-- Client MUST send a gzipped chunk as msgpack `bin` format.
+- `entries` is a gzip/zstd binary chunk of `MessagePackEventStream`, which MAY be a concatenated binary of multiple gzip/zstd binary strings.
+- Client MUST send an option with `compressed` key with the value `gzip` or `zstd`.
+- Client MUST send a gzip/zstd chunk as msgpack `bin` format.
 - Server MUST accept `bin` format.
 - Server MAY decompress and decode individual events on demand but MAY NOT do right after request arrival. It means it MAY costs less, compared to `Forward` mode, when decoding is not needed by any plugins.
```

> Option

https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5#option

```diff
 - Server MAY just ignore any options given.
[...]

PR #4657, opened 2024-10-03
compressable:Zstd comp support

Athishpranav2003:
**Which issue(s) this PR fixes**: 
Fixes #4162 

**What this PR does / why we need it**: 
Adds new compression method support to handle messages
**Docs Changes**:
TODO
**Release Note**: 
N/A

PR #4896, opened 2025-04-09
out_forward: add warning for zstd as experimental

daipom:
* Related to #4758

When using `compress zstd` for `out_forward`, output the following warning log.

```
zstd compression feature is an experimental new feature supported since v1.19.0.
Please make sure that the destination server also supports this feature before using it.
in_forward plugin for Fluentd supports it since v1.19.0.
```

In #4758, it was agreed that we would refrain from updating [the forward protocol v1.5](https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1.5) for the time being and instead output a log message indicating that this feature is experimental.

> for now, we will make zstd available as an exprimenatal feature on the Fluentd side, and note in the documentation that it is an experimental feature that cannot be used unless the server supports it.

> Just displaying as a warning log to tell: "This is experimental feature and not standardized yet" or similar message should be enough.

* https://github.com/fluent/fluentd/issues/4758#issuecomment-2608873502
* https://github.com/fluent/fluentd/issues/4758#issuecomment-2608918209
* https://github.com/fluent/fluentd/issues/4758#issuecomment-2609160527

**Docs Changes**:
Not needed. (It should be done as #4657)

**Release Note**: 
Not needed. (A note for #4657 would be enough.)

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -271,9 +271,15 @@ module Fluent::Plugin
       end
 
       raise Fluent::ConfigError, "ack_response_timeout must be a positive integer" if @ack_response_timeout < 1
+
+      if @compress == :zstd
+        log.warn "zstd compression feature is an experimental new feature supported since v1.19.0." +
+                 " Please make sure that the destination server also supports this feature before using it." +
+                 " in_forward plugin for Fluentd supports it since v1.19.0."
+      end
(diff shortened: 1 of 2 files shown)


Watson1978 (approved):

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #4899, opened 2025-04-10
http_server: stop fallback to WEBrick

### LINKED ISSUE #4648: Replace WEBrick because it is no longer recommended for production use
### Describe the bug

Fluentd [depends on `webrick`](https://github.com/fluent/fluentd/blob/7e39526150963beb4bafe1a874a97b536b93095d/fluentd.gemspec#L35) and uses it in [quite a few places](https://github.com/search?q=repo%3Afluent%2Ffluentd%20webrick&type=code) and thus although it's no longer included in ruby, it is required (e.g. [the dockerfile](https://github.com/fluent/fluentd-docker-image/blob/a5280782e6b2682d6e05acd781440d648d90304c/v1.17/alpine/Dockerfile#L14) installs `ruby-webrick`).

There have been a fair few CVEs reported for webrick in recent years, but more worrying is that in response to a recent security vulnerability report, one of the maintainers [said "webrick is not for production"](https://github.com/ruby/webrick/issues/145#issuecomment-2372838285).

If fluentd contines to rely on this, it feels risky (e.g. maybe future reports may not be patched so quickly / at all).

Perhaps it would be wise to migrate each usage away from webrick to a production-suitable replacement?

### To Reproduce

Install fluentd - you will not be able to do this without also installing webrick.

### Expected behavior

We should move away from production use of a package which the maintainers no longer recommend for production use.

### Your Environment

```markdown
- Fluentd version: all recent versions
- Package version: N/A
- Operating system: N/A
- Kernel version: N/A
```

### Your Configuration

```apache
Any configuration.
```

### Your Error Log

```shell
N/A
```

### Additional context

_No response_

daipom:
* Partially Fixes #4648

Make `http_server` plugin helper stop the fallback to WEBrick.

WEBrick is no longer recommended for production use.

We don't need this feature because it was for supporting Ruby < 2.3. (#2447)

**Docs Changes**:
https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-http_server
=> https://github.com/fluent/fluentd-docs-gitbook/pull/575

**Release Note**: 
The same as the title.


kenhys (approved):
LGTM.

Watson1978 (approved):
👍🏻

PR #4900, opened 2025-04-10
in_http: replace WEBrick::HTTPUtils.parse_query with URI

### LINKED ISSUE #4648: Replace WEBrick because it is no longer recommended for production use

daipom:
* Partially Fixes #4648

Replace `WEBrick::HTTPUtils.parse_query` with `URI.decode_www_form` in `in_http`.

WEBrick is no longer recommended for production use.

WEBrick::HTTPUtils is just a util, so it might be harmless.
However, it would be happy to reduce dependence if possible.

There is no impact on the specification.

**Docs Changes**:
Not needed. (There is no impact on the specification.)

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/plugin/in_http.rb
+++ b/lib/fluent/plugin/in_http.rb
@@ -531,12 +531,12 @@ module Fluent::Plugin
         @env['REMOTE_ADDR'] = @remote_addr if @remote_addr
 
         uri = URI.parse(@parser.request_url)
-        params = WEBrick::HTTPUtils.parse_query(uri.query)
+        params = parse_query(uri.query)
 
         if @format_name != 'default'
           params[EVENT_RECORD_PARAMETER] = @body
         elsif /^application\/x-www-form-urlencoded/.match?(@content_type)
-          params.update WEBrick::HTTPUtils.parse_query(@body)
+          params.update parse_query(@body)
         elsif @content_type =~ /^multipart\/form-data; boundary=(.+)/
           boundary = WEBrick::HTTPUtils.dequote($1)
           params.update WEBrick::HTTPUtils.parse_form_data(@body, boundary)
@@ -553,7 +553,7 @@ module Fluent::Plugin
 
         if (@add_query_params)
 
-          query_params = WEBrick::HTTPUtils.parse_query(uri.query)
+          query_params = parse_query(uri.query)
 
           query_params.each_pair {|k,v|
             params["QUERY_#{k.tr('-','_').upcase}"] = v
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_http.rb
```diff
@@ -643,6 +643,10 @@ def include_cors_allow_origin
 
         !r.nil?
       end
+
+      def parse_query(query)
+        query.nil? ? {} : Hash[URI.decode_www_form(query, Encoding::ASCII_8BIT)]
```
kenhys, line 648: ';' separator to be ignored. Is it ok here?
(I'm not sure difference WEBrick::HTTPUtils.parse_query and URI.decode_www_form)

kenhys: ref. https://rubydoc.info/stdlib/webrick/1.8.6/WEBrick%2FHTTPUtils:parse_query

daipom: Oh, I didn't notice that.

If we want to keep that specification, it would be better not to replace it by force.

> `WEBrick::HTTPUtils` is just a util, so it might be harmless.

Watson1978: Seems [puma](https://github.com/puma/puma) that Ruby web server also uses URI.decode_www_form.
https://github.com/puma/puma/blob/97c7d129a940c809fb379b3ecf314d39a18a332b/lib/puma/util.rb#L29

daipom: Thanks! I see!

Watson1978 (approved):
👍🏻

daipom:
Thanks!

PR #4909, opened 2025-04-21
winsvc: Stop the service when the supervisor is dead

ashie:
None.

Even if the supervisor process is dead due to config error or etc, the service process will continue running unexpectedly because no one monitor it. This commit fix this issue.

**Docs Changes**:
Not needed.

**Release Note**: 
Same with the title.

### DIFF
--- a/lib/fluent/winsvc.rb
+++ b/lib/fluent/winsvc.rb
@@ -48,6 +48,8 @@ begin
   end
 
   class FluentdService < Daemon
+    ERROR_WAIT_NO_CHILDREN = 128
+
     @pid = 0
     @service_name = ''
 
@@ -57,18 +59,21 @@ begin
     
     def service_main
       @pid = service_main_start(@service_name)
-      while running?
-        sleep 10
+      begin
+        loop do
+          sleep 5
+          break unless running?
+          raise Errno::ECHILD unless Process.waitpid2(@pid, Process::WNOHANG)
+        end
+      rescue Errno::ECHILD
+        @pid = 0
+        SetEvent(@@hStopEvent)
+        SetTheServiceStatus.call(SERVICE_STOPPED, ERROR_WAIT_NO_CHILDREN, 0, 0)
       end
     end
 
     def service_stop
-      if @pid <= 0
-        set_event(@service_name)
-        return
-      end
-
-      wait_supervisor_finished
+      wait_supervisor_finished if @pid > 0
     end
 
     def service_paramchange


ashie:
Unfortunately it's hard to write tests for windows service.
Instead I tested this code manually and confirmed that it works fine.

Watson1978 (dismissed):
👍🏻

lib/fluent/winsvc.rb
```diff
@@ -57,8 +57,16 @@ def initialize(service_name)
...
-        sleep 10
+      loop do
+        sleep 5
+        break unless running?
+        begin
+          pid, status = Process.waitpid2(@pid, Process::WNOHANG)
+          raise Errno::ECHILD if pid || status
+        rescue Errno::ECHILD
+          exit_code = status ? status.to_i : Errno::ECHILD::Errno
+          SetTheServiceStatus.call(SERVICE_STOPPED, exit_code, 0, 0)
```
daipom, line 68: Thanks for this fix!

> SetTheServiceStatus.call(SERVICE_STOPPED, exit_code, 0, 0)

I'm wondering about the right way to stop the service.
I'm not sure, but it might be preferable to notify the SCM by using `Service.stop`.

https://github.com/chef/win32-service/blob/67a0f7ac46a5e0332d95867f1ee2e7cd733b9117/lib/win32/daemon.rb#L279-L324

```rb
thr = Thread.new do
  begin
    while WaitForSingleObject(@@hStopEvent, 1000) == WAIT_TIMEOUT
      ...
    end

    service_stop if respond_to?("service_stop")
  ensure
    SetEvent(@@hStopCompletedEvent)
  end
end

if respond_to?("service_main")
  service_main(*@@Argv)
end

thr.join
```

https://github.com/chef/win32-service/blob/67a0f7ac46a5e0332d95867f1ee2e7cd733b9117/lib/win32/daemon.rb#L105-L135

https://github.com/chef/win32-service/blob/67a0f7ac46a5e0332d95867f1ee2e7cd733b9117/lib/win32/daemon.rb#L176-L186

daipom: I'm considering this.

```rb
require "win32/service"
```

```rb
rescue Errno::ECHILD
  Service.stop(@service_name)
end
```

daipom: Maybe the current implementation cannot stop the process correctly.

```rb
require "win32/daemon"

class TestService < Win32::Daemon
  def service_main
    count = 0
    loop do
      sleep 5
      break unless running?
      begin
        count += 1
        raise Errno::ECHILD if count >= 3
      rescue Errno::ECHILD
        SetTheServiceStatus.call(SERVICE_STOPPED, Errno::ECHILD::Errno, 0, 0)
      end
    end
  end
end

TestService.new.mainloop
```

```console
$ Get-EventLog System -newest 1 | fl *

EventID            : 7023
MachineName        : LAPTOP-MBJONE5D
Data               : {116, 0, 101, 0...}
Index              : 206242
Category           : (0)
CategoryNumber     : 0
EntryType          : Error
Message            : The testservice service terminated with the following error:
                     %%10
Source             : Service Control Manager
ReplacementStrings : {testservice, %%10}
InstanceId         : 3221232495
TimeGenerated      : 2025/04/22 18:58:54
TimeWritten        : 2025/04/22 18:58:54
UserName           :
Site               :
Container          :
```

```
testservice サービスは、次のエラーで終了しました: 
環境が間違っています。
```

daipom: Looks like this can stop the process correctly.
(Immediate and no error.)

```rb
require "win32/daemon"
require "win32/service"

class TestService < Win32::Daemon
  def service_main
    count = 0
    loop do
      sleep 5
      break unless running?
      begin
        count += 1
        raise Errno::ECHILD if count >= 3
      rescue Errno::ECHILD
        Service.stop("testservice")
      end
    end
  end
end

TestService.new.mainloop
```

ashie: Ah sorry, the current way doesn't seem right as you say.
`SetTheServiceStatus.call` doesn't stop [the command receiver thread](https://github.com/chef/win32-service/blob/67a0f7ac46a5e0332d95867f1ee2e7cd733b9117/lib/win32/daemon.rb#L279-L318).
As a result the ruby process seems still remained.

> Maybe the current implementation cannot stop the process correctly.

> ```
> $ Get-EventLog System -newest 1 | fl *
> ...
> EntryType          : Error
> Message            : The testservice service terminated with the following error:
>                      %%10
> ```
> 
> ```
> testservice サービスは、次のエラーで終了しました: 
> 環境が間違っています。
> ```

The above result is partially intended.
IMHO something an error code should be set because the supervisor is abnormally exited in this case.
(When the supervisor exits normally, the loop will be exited by `break unless running?` before calling `Process.waitpid2`)
`Service.stop` doesn't seem set an error code.
This is the why I use `SetTheServiceStatus.call` here.
But using `Errno::ECHILD` as the error code here is nonsense since it's not compatible with Win32's exit code.

The number of `ECHILD` is `10`:

  https://learn.microsoft.com/ja-jp/cpp/c-runtime-library/errno-constants?view=msvc-170

while the error code `10` is `ERROR_BAD_ENVIRONMENT` (`環境が間違っています`) on Win32:

  https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-

We should use correct Win32 error code here.
In addition, we should stop the command receiver thread correctly as I mentioned above.

daipom: > SetTheServiceStatus.call doesn't stop [the command receiver thread](https://github.com/chef/win32-service/blob/67a0f7ac46a5e0332d95867f1ee2e7cd733b9117/lib/win32/daemon.rb#L279-L318).

Yes! I'm concerned about it!

> The above result is partially intended.

I see!

> We should use correct Win32 error code here.
> In addition, we should stop the command receiver thread correctly as I mentioned above.

I see!

daipom:
Thanks!

I share my check scripts for now.

`/test/run.rb`

```rb
require "win32/daemon"
require "win32/service"

class TestService < Win32::Daemon
  File.open("/test/log/test_service.log", 'a') { |f| f.puts "#{Time.now.inspect}, TestService init" }

  def service_main
    count = 0
    while running?
    File.open("/test/log/test_service.log", 'a') { |f| f.puts "#{Time.now.inspect}, TestService mainloop" }
      sleep 2
      begin
        count += 1
        raise Errno::ECHILD if count >= 2
      rescue Errno::ECHILD
        File.open("/test/log/test_service.log", 'a') { |f| f.puts "#{Time.now.inspect}, TestService trigger stop" }
        SetTheServiceStatus.call(SERVICE_STOPPED, Errno::ECHILD::Errno, 0, 0)
        # Service.stop("testservice")
        # SetEvent(@@hStopEvent)
        break
      end
    end
  end

  def service_stop
    File.open("/test/log/command_receiver_thread.log", 'a') { |f| f.puts "#{Time.now.inspect}, TestService service_stop called" }
  end
end

TestService.new.mainloop
```

* Install debug win32_service
  * https://github.com/fluent-plugins-nursery/win32-service/tree/fix-slow-start-with-ruby3-with-debug-logs
  * `rake install`
* Register service

```rb
require "win32/service"

# Specify ruby bin path
ruby_path = "C:\\Ruby32-x64\\bin\\ruby.exe"

Win32::Service.create(
  service_name: "testservice",
  binary_path_name: "\"#{ruby_path}\" C:\\test\\run.rb",
  display_name: "testservice",
)
```

* Prepare `/test/log/` dir
* Run powershell script

```ps1
echo "Clear logs"
rm /test/log/*

echo "Start the service"
Start-Service testservice

echo "Waiting for a while ..."
sleep 6

echo "Check service status:"
Get-Service testservice

echo "`nCheck Ruby process:"
ps -Name ruby

echo "`nLog:"
cat /test/log/* | sort

echo "`nLatest System EventLogs:"
Get-EventLog System -newest 3 | fl *
```

lib/fluent/winsvc.rb
ashie: > We should use correct Win32 error code here.
> In addition, we should stop the command receiver thread correctly as I mentioned above.

I've fixed these issue.
I confirmed that fluentdwinsvc service is stopped and no ruby process is remained.

daipom: Thanks! I will check it.

lib/fluent/winsvc.rb
```diff
@@ -57,8 +59,16 @@ def initialize(service_name)
...
-        sleep 10
+      begin
+        loop do
+          sleep 5
+          break unless running?
+          pid, status = Process.waitpid2(@pid, Process::WNOHANG)
+          raise Errno::ECHILD if pid || status
+        end
+      rescue Errno::ECHILD
+        set_event(@@hStopEvent)
```
daipom, line 70:
```suggestion
        SetEvent(@@hStopEvent)
```

We should use [SetEvent of Win32 api](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-setevent) here.

`set_event` is a private method of this class to signal to the supervisor.

daipom, line 70:
```suggestion
      rescue Errno::ECHILD
        @pid = 0
        set_event(@@hStopEvent)
```

`service_stop` callback is called after `hStopEvent`.
We should set `@pid` to `0` to skip the callback because the callback will wait for the supervisor process to init, meaninglessly, for a while.

https://github.com/fluent/fluentd/blob/c3ebd6687931298a3d6faf7d0e8383f31d81357c/lib/fluent/winsvc.rb#L65-L72

In addition, it would be better to change this callback as follows.

```rb
 def service_stop 
   return if @pid <= 0 
   wait_supervisor_finished 
 end 
```

lib/fluent/winsvc.rb
```diff
@@ -57,8 +59,16 @@ def initialize(service_name)
...
     def service_main
       @pid = service_main_start(@service_name)
-      while running?
-        sleep 10
+      begin
+        loop do
+          sleep 5
+          break unless running?
+          pid, status = Process.waitpid2(@pid, Process::WNOHANG)
+          raise Errno::ECHILD if pid || status
```
daipom, line 67:
```suggestion
          raise Errno::ECHILD unless Process.waitpid(@pid, Process::WNOHANG)
```

If we don't use `status`, we can simplify the code.

daipom (commented):
@ashie 
Thanks!
This seems to be the right way to stop the service process!

I have commented on some minor points.
Please check them.

lib/fluent/winsvc.rb
ashie: Although I don't think it's supervisor specific, using `set_event` here isn't suitable as you say.
`set_event` expects a event name as the argument but `@@hStopEvent` is an event handle itself.
I'll fix it.

daipom: > Although I don't think it's supervisor specific, 

Oh! Sorry, that's right!

ashie: Done

lib/fluent/winsvc.rb
ashie: Done

lib/fluent/winsvc.rb
ashie: Done. More simplified.

```ruby
def
  wait_supervisor_finished if @pid > 0
end
```

daipom: Thanks! :heart:

daipom:
I'm sorry.
My suggestion was wrong: https://github.com/fluent/fluentd/pull/4909#discussion_r2057663429
The if condition was reversed.
The service accidentally stops after starting, without stopping the supervisor and workers.

I have made a PR to fix this bug and adding tests:

* #4954

PR #4914, opened 2025-04-23
config: change inspect format

kenhys:

Fixes #

In the previous versions, it is hard to identify
the separator of elements.
Added attr: and elements: prefix for properties and mark it <...>.

Before:

```
  name:ROOT, arg:, {}, [name:system, arg:,
  {"config_include_dir"=>"..."}, [], name:source, arg:,
  {"@type"=>"forward"}, [], name:source, arg:, {"@type"=>"tcp"}, []]
```

After:

```
  <name:ROOT, arg:, attrs:{}, elements:[<name:system, arg:,
  attrs:{"config_include_dir"=>"..."}, elements:[]>, <name:source,
  arg:, attrs:{"@type"=>"forward"}, []>, <name:source, arg:,
  atrrs:{"@type"=>"tcp"}, elements:[]>]
```

**Docs Changes**:

N/A

**Release Note**:

### DIFF
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -73,7 +73,7 @@ module Fluent
 
       def inspect
         attrs = super
-        "name:#{@name}, arg:#{@arg}, " + attrs + ", " + @elements.inspect
+        "<name:#{@name}, arg:#{@arg}, attrs:" + attrs + ", elements:" + @elements.inspect + ">"
       end
 
       # Used by PP and Pry
--- a/lib/fluent/config/element.rb
+++ b/lib/fluent/config/element.rb
@@ -73,7 +73,7 @@ module Fluent
 
       def inspect
         attrs = super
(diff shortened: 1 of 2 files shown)


lib/fluent/config/element.rb
```diff
@@ -73,7 +73,7 @@ def add_element(name, arg = '')
 
       def inspect
         attrs = super
-        "name:#{@name}, arg:#{@arg}, " + attrs + ", " + @elements.inspect
+        "<name:#{@name}, arg:#{@arg}, attrs:" + attrs + ", elements:" + @elements.inspect + ">"
```
daipom, line 76:
```suggestion
        "<name:#{@name}, arg:#{@arg}, attrs:#{attrs}, elements:#{@elements.inspect}>"
```

Thanks!
This would be clearer.
I have confirmed this does not change the output format.

kenhys: Thanks!

daipom (approved):
Thanks!

PR #4942, opened 2025-04-30
Backport(v1.16): winsvc: Stop the service when the supervisor is dead (#4909)

kenhys:
Backport #4909
None.

Even if the supervisor process is dead due to config error or etc, the service process will continue running unexpectedly because no one monitor it. This commit fix this issue.

**Docs Changes**:
Not needed.

**Release Note**: 
Same with the title.


daipom (approved):
Thanks!

PR #4954, opened 2025-05-13
winsvc: Fix bug where service accidentally stops after starting

daipom:
This fixes a bug of #4909.

After 53bcd3c83edb2bdc0d98481c86e9955ffdc4f5ef (#4909), the service accidentally stops after starting, without stopping the supervisor and workers.

**Docs Changes**:
Not needed.

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/winsvc.rb
+++ b/lib/fluent/winsvc.rb
@@ -63,7 +63,7 @@ begin
         loop do
           sleep 5
           break unless running?
-          raise Errno::ECHILD unless Process.waitpid2(@pid, Process::WNOHANG)
+          raise Errno::ECHILD if Process.waitpid(@pid, Process::WNOHANG)
         end
       rescue Errno::ECHILD
         @pid = 0
--- a/test/scripts/windows_service_test.ps1
+++ b/test/scripts/windows_service_test.ps1
@@ -1,0 +1,73 @@
+$ErrorActionPreference = "Stop"
+Set-PSDebug -Trace 1
+
+$default_conf_path = (Resolve-Path fluent.conf).Path
+$current_path = (Get-Location).Path
+$log_path = "$current_path/fluentd.log"
+
+ruby bin/fluentd --reg-winsvc i --reg-winsvc-fluentdopt "-c '$default_conf_path' -o '$log_path'"
+
+# Test: must not start automatically
+if ((Get-Service fluentdwinsvc).Status -ne "Stopped") {
+    Write-Error "The service should not start automatically."
+}
+
+Start-Service fluentdwinsvc
+Start-Sleep 30
+
+# Test: the service should be running after started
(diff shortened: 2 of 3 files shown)


lib/fluent/winsvc.rb
```diff
@@ -63,7 +63,7 @@ def service_main
         loop do
           sleep 5
           break unless running?
-          raise Errno::ECHILD unless Process.waitpid2(@pid, Process::WNOHANG)
+          raise Errno::ECHILD if Process.waitpid2(@pid, Process::WNOHANG)
```
ashie, line 66:
```suggestion
          raise Errno::ECHILD if Process.waitpid(@pid, Process::WNOHANG)
```
We don't use the second return value in actual.

daipom: Thanks! Fixed!

daipom:
I do not understand why this trace log is marked as Error.
However, it does not seem to affect the test.
I have confirmed that if there is a matching log, the test enters the `if` statement as expected, and the test fails.

https://github.com/fluent/fluentd/actions/runs/14992579391/job/42119227188?pr=4954#step:5:97
> Error: DEBUG:   26+     if ( >>>> Select-String -Path $_ -Pattern "[warn]", "[error]", "[fatal]" -SimpleMatch -Quiet) {

daipom:
The fails of Windows tests with Ruby 3.2 and 3.3 are nothing to do with this fix.

daipom:
> I do not understand why this trace log is marked as Error. However, it does not seem to affect the test. I have confirmed that if there is a matching log, the test enters the `if` statement as expected, and the test fails.
> 
> https://github.com/fluent/fluentd/actions/runs/14992579391/job/42119227188?pr=4954#step:5:97
> 
> > Error: DEBUG:   26+     if ( >>>> Select-String -Path $_ -Pattern "[warn]", "[error]", "[fatal]" -SimpleMatch -Quiet) {

It appears that GitHub responds to the `[error]` string, and it has nothing to do with the PowerShell process.
We can ignore this.

ashie (dismissed):
LGTM

test/scripts/windows_service_test.ps1
```diff
@@ -0,0 +1,73 @@
...
+    Write-Output $ruby_processes
+    Write-Error "All Ruby processes should stop."
+}
+
+# Test: service should stop when the supervisor fails to launch
+# https://github.com/fluent/fluentd/pull/4909
+$test_setting = @'
+<source>
+  @type sample
+  @id foo
```
kenhys, line 51:
```suggestion
  @id DUPLICATED_ID
```

It might be better to assign more meaningful one instead of `foo`

daipom: Thanks! Fixed!

kenhys (approved):

daipom:
Thanks for your review

PR #4961, opened 2025-05-17
rubocop: use yield without block

kenhys:
Fixes #

This is cosmetic change, it does not change behavior at all. The following rubocop configuration detects it.

```
Performance/RedundantBlockCall:
  Exclude:
    - 'test/**/*'
    #- 'lib/fluent/**/*.rb'
    - 'lib/fluent/plugin_helper/*.rb'
    - 'lib/fluent/plugin/*.rb'
    - 'lib/fluent/compat/*.rb'
    - 'lib/fluent/config/*.rb'
    - 'lib/fluent/*.rb'
  Enabled: true
```

Benchmark result:

```
ruby 3.2.8 (2025-03-26 revision 13f495dc2c) +YJIT [x86_64-linux]
          block call     1.315M i/100ms
   yield with &block     1.721M i/100ms
yield without &block     1.692M i/100ms
          block call     13.572M (± 0.5%) i/s   (73.68 ns/i) -     68.365M in   5.037353s
   yield with &block     17.927M (± 3.8%) i/s   (55.78 ns/i) -     89.490M in   5.000061s
yield without &block     18.191M (± 0.3%) i/s   (54.97 ns/i) -     91.355M in   5.021978s

Comparison:
yield without &block: 18191218.3 i/s
   yield with &block: 17927144.4 i/s - same-ish: difference falls within error
          block call: 13572023.6 i/s - 1.34x  slower
```

Appendix: benchmark script

```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'benchmark-memory'
end

def block_call(s, &block)
  block.call(s) if block_given?
end

def yield_with_block(s, &block)
  yield(s) if block_given?
end

def yield_without_block(s)
  yield(s) if block_given?
end

Benchmark.ips do |x|
  x.report("block call") {
    block_call("dummy") do |message|
      message
    end
  }
  x.report("yield with &block") {
    yield_with_block("dummy") do |message|
      message
    end
  }
  x.report("yield without &block") {
    yield_without_block("dummy") do |message|
      message
    end
  }
  x.compare!
end
```

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/test/driver/base.rb
+++ b/lib/fluent/test/driver/base.rb
@@ -185,7 +185,7 @@ module Fluent
           end
         end
 
-        def run_actual(timeout: DEFAULT_TIMEOUT, &block)
+        def run_actual(timeout: DEFAULT_TIMEOUT)
           if @instance.respond_to?(:_threads)
             sleep 0.1 until @instance._threads.values.all?(&:alive?)
           end
@@ -201,7 +201,7 @@ module Fluent
           return_value = nil
           begin
             Timeout.timeout(timeout * 2) do |sec|
-              return_value = block.call if block_given?
+              return_value = yield if block_given?
             end
           rescue Timeout::Error
             raise TestTimedOut, "Test case timed out with hard limit."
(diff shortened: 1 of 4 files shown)


kenhys:
Not to affect previous behavior, partially fixed Performance/RedundantBlockCall: issue under lib/fluent/test/.

daipom (approved):
Thanks!

PR #4966, opened 2025-05-19
metrics: enable input metrics by default

Watson1978:
Fixes #

I ran fluend with `--enable-input-metrics` option and read 10GB file with the `in_tail` plugin
There was no significant performance degradation.

So, I think it is safe to always enable `--enable-input-metrics`.

### environment
* ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-linux]

### results
* fluentd master branch (efbb51d7)
    * `ruby bin/fluentd -c in_tail.conf`
        * 65.939232375 seconds
    * `ruby bin/fluentd --enable-input-metrics -c in_tail.conf`
        * 67.776236261 seconds
* fluentd 1.16.9 (daccbc6f)
    * `ruby bin/fluentd -c in_tail.conf`
        * 106.312500419 seconds

### config
```
<source>
  @type tail
  tag test
  read_from_head true
  <parse>
    @type json
  </parse>
</source>

<match **>
  @type file
  path "#{File.expand_path '~/tmp/fluentd/log'}"
</match>
```

### script to generate 10GB
```ruby
# frozen_string_literal: true
require "json"
require "fileutils"

def data_generater(str)
  {
    "message": str * 1000,
  }.to_json
end

FILE_MAX_SIZE = 10 * 1024 * 1024 * 1024
FILE_PATH = "/home/watson/tmp/fluentd/access-1.log"

dir = File.dirname(FILE_PATH)
FileUtils.mkdir_p(dir)

File.open(FILE_PATH, "w") do |f|
  data = data_generater('a')
  loop do
    f.puts data
    break if File.size(FILE_PATH) > FILE_MAX_SIZE
  end
end
```

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/578

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/command/fluentd.rb
+++ b/lib/fluent/command/fluentd.rb
@@ -155,7 +155,11 @@ op.on('--strict-config-value', "Parse config values strictly", TrueClass) {|b|
   cmd_opts[:strict_config_value] = b
 }
 
-op.on('--enable-input-metrics', "Enable input plugin metrics on fluentd", TrueClass) {|b|
+op.on('--enable-input-metrics', "[DEPRECATED] Enable input plugin metrics on fluentd", TrueClass) {|b|
+  cmd_opts[:enable_input_metrics] = b
+}
+
+op.on('--disable-input-metrics', "Disable input plugin metrics on fluentd", FalseClass) {|b|
   cmd_opts[:enable_input_metrics] = b
 }
 
--- a/test/command/test_fluentd.rb
+++ b/test/command/test_fluentd.rb
@@ -1608,4 +1608,16 @@ CONF
                          "[info]: loading additional configuration file path=\"#{conf_dir}/child.yml\"")
     end
   end
+
+  test "allow --disable-input-metrics option" do
+    conf = <<CONF
+<system>
+</system>
+CONF
+    conf_path = create_conf_file('input-metrics.conf', conf)
+    assert_log_matches(
+      create_cmdline(conf_path, '--disable-input-metrics'),
+      "#0 fluentd worker is now running worker=0"
+    )
+  end
 end
--- a/test/command/test_fluentd.rb
+++ b/test/command/test_fluentd.rb
@@ -1610,11 +1610,7 @@ CONF
   end
 
   test "allow --disable-input-metrics option" do
-    conf = <<CONF
-<system>
-</system>
-CONF
-    conf_path = create_conf_file('input-metrics.conf', conf)
+    conf_path = create_conf_file('empty.conf', '')
     assert_log_matches(
       create_cmdline(conf_path, '--disable-input-metrics'),
       "#0 fluentd worker is now running worker=0"
--- a/lib/fluent/root_agent.rb
+++ b/lib/fluent/root_agent.rb
@@ -86,7 +86,7 @@ module Fluent
       @without_source = system_config.without_source || false
       @source_only_mode = SourceOnlyMode.new(system_config.with_source_only, start_in_parallel)
       @source_only_buffer_agent = nil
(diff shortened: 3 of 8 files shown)


lib/fluent/command/fluentd.rb
```diff
@@ -155,7 +155,7 @@
   cmd_opts[:strict_config_value] = b
 }
 
-op.on('--enable-input-metrics', "Enable input plugin metrics on fluentd", TrueClass) {|b|
+op.on('--enable-input-metrics', "[DEPRECATED] Enable input plugin metrics on fluentd", TrueClass) {|b|
```
daipom, line 158: Thanks!

It would be possible for us to leave this option as is and only change the default value.
Thus, if this affects performance, a user can disable it.
What do you think about this?

cosmo0920: Ah, that's good idea.
Turn this option by default and leave the choice of disabling this should be better than completely removed it.

Watson1978: I added an option to disable it at startup.

daipom: If possible, I want to use the existing option because it will allow us to avoid making a new deprecated option.

Could we do either of the following?

```rb
--[no-]enable-input-metrics
```
=> a user would be able to disable this by `--no-enable-input-metrics`

or

```rb
--enable-input-metrics [BOOL] ... do |b|
  opt = !!b
end
```
=> a user would be able to disable this by `--enable-input-metrics no`

Watson1978: I'm not sure if either of these are common ways.

cosmo0920 (approved):
I'm pleased this feature to be enabled by default. Thank you for your work! :+1:

lib/fluent/command/fluentd.rb
daipom: I see!

For users, it would be clearer if the feature could be disabled with the option `--disable-input-metrics`.

So, looks good with the current specification. Thanks!

daipom: Oh, sorry, I have one question.
What about systemconfig?

We have `--disable-input-metrics` command line option.
But we still use `enable_input_metrics` in systemconfig?

daipom (commented):
It would be better to add some tests to check if the option for disabling the feature works intendedly.

lib/fluent/command/fluentd.rb
Watson1978: The `--enable-input-metrics` option cannot be set `enable_input_metrics` to false, so I introduced the new `--disable-input-metrics` option.

However, we can set `enable_input_metrics` to false directly with system config.
I think it is sufficient to change the default values about system config.

daipom:
Thanks!
Could you update this test as well? 
https://github.com/fluent/fluentd/blob/a3f4c829b557a8e587571f9ff4996ec002062212/test/test_supervisor.rb#L57

test/command/test_fluentd.rb
```diff
@@ -1608,4 +1608,16 @@ def create_config_include_dir_configuration(config_path, config_dir, yaml_format
                          "[info]: loading additional configuration file path=\"#{conf_dir}/child.yml\"")
     end
   end
+
+  test "allow --disable-input-metrics option" do
+    conf = <<CONF
+<system>
+</system>
+CONF
+    conf_path = create_conf_file('input-metrics.conf', conf)
```
daipom, line 1617:
```suggestion
    conf_path = create_conf_file('empty.conf', '')
```

Do we need this config for this test?

Watson1978: Thanks.

PR #4970, opened 2025-05-20
win32api: remove support for runtime used in old Ruby

Watson1978:
Fixes #

This is just refactoring to clean up the branch for old ruby.

Windows Ruby have started using UCRT since 3.1.
Now, we have been supporting Ruby 3.2+ on master.
https://rubyinstaller.org/2021/12/31/rubyinstaller-3.1.0-1-released.html

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/win32api.rb
+++ b/lib/fluent/win32api.rb
@@ -22,13 +22,7 @@ module Fluent
     require 'fiddle/types'
     extend Fiddle::Importer
 
-    if RUBY_PLATFORM.split('-')[-1] == "ucrt"
-      MSVCRT_DLL = 'ucrtbase.dll'
-    else
-      MSVCRT_DLL = 'msvcrt.dll'
-    end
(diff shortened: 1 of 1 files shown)


ashie:
I'm not sure but CRT isn't always tied with Ruby's version.
Although RubyInstaller uses ucrt, other Ruby build might use msvcrt.

Watson1978:
> other Ruby build might use msvcrt.

I think it probably wouldn't be able to get it unless it build Ruby by yourself on Windows...

ashie:
> I think it probably wouldn't be able to get it unless it build Ruby by yourself on Windows...

`x64-msvcrt-ruby340.dll` is appeared in https://packages.msys2.org/packages/mingw-w64-x86_64-ruby

Watson1978:
OK, I see. Thanks.

PR #4971, opened 2025-05-20
output: add metrics for number of stored chunks in secondary

Watson1978:
Fixes #

This PR will add metrics for number of stored chunks in secondary.
This metrics can be used to determine whether recovery is required from secondary if failure occurs.

**Docs Changes**:

**Release Note**:


kenhys (dismissed):
LGTM.

lib/fluent/plugin/output.rb
```diff
@@ -278,6 +279,7 @@ def configure(conf)
         @emit_records_metrics = metrics_create(namespace: "fluentd", subsystem: "output", name: "emit_records", help_text: "Number of emit records")
         @emit_size_metrics =  metrics_create(namespace: "fluentd", subsystem: "output", name: "emit_size", help_text: "Total size of emit events")
         @write_count_metrics = metrics_create(namespace: "fluentd", subsystem: "output", name: "write_count", help_text: "Number of writing events")
+        @write_secondary_count_metrics = metrics_create(namespace: "fluentd", subsystem: "output", name: "secondary_chunk_count", help_text: "Number of writing events in secondary")
```
daipom, line 282:
```suggestion
        @write_secondary_count_metrics = metrics_create(namespace: "fluentd", subsystem: "output", name: "secondary_write_count", help_text: "Number of writing events in secondary")
```

daipom: or `write_secondary_count`?

lib/fluent/plugin/output.rb
```diff
@@ -1224,6 +1226,7 @@ def try_flush
             dump_chunk_id = dump_unique_id_hex(chunk_id)
             log.trace "adding write count", instance: self.object_id
             @write_count_metrics.inc
+            @write_secondary_count_metrics.inc if using_secondary
```
daipom, line 1229: Don't we also need to count it in the case of `delayed_commit`?

https://github.com/fluent/fluentd/blob/0b72a77d23a667c2c5c648e457e16fc8835c2134/lib/fluent/plugin/output.rb#L1212-L1215

Watson1978: Ah, thanks.

Watson1978: Added https://github.com/fluent/fluentd/pull/4971/commits/70d80b5cc861fffa577a0603205b5015a3e23094

daipom:
Should we add these getters?
I'm concerned that we don't have some already, though...

https://github.com/fluent/fluentd/blob/0b72a77d23a667c2c5c648e457e16fc8835c2134/lib/fluent/plugin/output.rb#L174-L196

daipom:
> Should we add these getters? I'm concerned that we don't have some already, though...
> 
> https://github.com/fluent/fluentd/blob/0b72a77d23a667c2c5c648e457e16fc8835c2134/lib/fluent/plugin/output.rb#L174-L196

It seems that we should add the getter because it is unnatural that `write_second_count` does not have a getter although `write_count` has a getter.

Watson1978:
> Should we add these getters?

I think we can remove these getter methods.
If it defined a getter method, I think we should use appropriately in the following section as well.

https://github.com/fluent/fluentd/blob/d0664384945579fbf23d51bba329440bd42aae37/lib/fluent/plugin/output.rb#L1590-L1601

Look like we have use these getter methods only used in test codes.
Is it really necessary to define it?

test/plugin/test_output_as_buffered_secondary.rb
```diff
@@ -348,6 +349,8 @@ def dummy_event_stream
       assert{ @i.write_count > prev_write_count }
       assert{ @i.num_errors == prev_num_errors }
 
+      assert{ @i.instance_variable_get(:@write_secondary_count_metrics).get > 0 }
```
daipom, line 352:
```suggestion
      assert{ @i.write_secondary_count > 0 }
```

test/plugin/test_output_as_buffered_secondary.rb
```diff
@@ -737,6 +742,7 @@ def dummy_event_stream
 
       assert_equal 0, @i.write_count
       assert_equal 0, @i.num_errors
+      assert_equal 0, @i.instance_variable_get(:@write_secondary_count_metrics).get
```
daipom, line 745:
```suggestion
      assert_equal 0, @i.write_secondary_count
```

test/plugin/test_output_as_buffered_secondary.rb
Watson1978: Thanks!

daipom:
@Watson1978 Could you update https://docs.fluentd.org/input/monitor_agent#output-example?
It looks very old :cry:

Watson1978:
> @Watson1978 Could you update https://docs.fluentd.org/input/monitor_agent#output-example? It looks very old 😢

OK, I see :)

Watson1978:
I created PR to update the doc
https://github.com/fluent/fluentd-docs-gitbook/pull/582

PR #4972, opened 2025-05-20
feat: add umask configuration support

wanglu-cell:
Fixes #

增加 umask 配置支持，使日志文件权限控制更灵活。当用户需要指定非默认文件权限时，可通过配置 umask 参数实现。
**Docs Changes**:
无
**Release Note**: 
新增功能: 支持通过 umask 参数控制输出文件权限


daipom:
Thanks for your contribution!

Could you please clarify what this PR does?

Fluentd already has `umask` command line option.

* https://docs.fluentd.org/deployment/command-line-option

The limitation of this option is:

* There is no system config option yet.
* It works only when using `--daemon` or `--no-supervisor`.
  * For example, it does not work when using Docker.

PR #4979, opened 2025-05-27
multi_output: fix metrics name

Watson1978:
Fixes #

In multi_output, two metrics are created with `emit_records`.
https://github.com/fluent/fluentd/blob/0b72a77d23a667c2c5c648e457e16fc8835c2134/lib/fluent/plugin/multi_output.rb#L94-L95

Since it is not possible to distinguish between them, this PR will fix the name.

**Docs Changes**:
Not needed.

**Release Note**: 
The same as the title.

### DIFF
--- a/lib/fluent/plugin/multi_output.rb
+++ b/lib/fluent/plugin/multi_output.rb
@@ -91,7 +91,7 @@ module Fluent
         super
 
         @num_errors_metrics = metrics_create(namespace: "fluentd", subsystem: "multi_output", name: "num_errors", help_text: "Number of count num errors")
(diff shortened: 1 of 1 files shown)


daipom (approved):
Thanks!

@cosmo0920 
If you know anything, could you please let me know?
Could this name change potentially affect users?
I don't know much about metrics...
As for Fluentd's internal logic, it does not seem to be affected.

cosmo0920:
It is no relevant issues for now. Thanks for the good catch. :+1:

cosmo0920 (approved):

PR #4980, opened 2025-05-28
in_tail: add metrics for file tracked count

Watson1978:
Fixes #

This PR will add a `tracked_file_count` field to the value returned by statistics method.
This will make it easy to see how many log files are being tracked.

**Docs Changes**:

**Release Note**:


daipom (approved):
Thanks! LGTM.

PR #4981, opened 2025-05-28
output: add metrics for dropped oldest chunk count

Watson1978:
Fixes #

When `overflow_action drop_oldest_chunk` is configured, the oldest buffer chunk will be automatically discarded when the total buffer capacity is exhausted.

This PR introduces a new metric, `drop_oldest_chunk_count`, which counts the number of discarded chunks.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/582

**Release Note**: 
Same as the title


daipom (approved):
LGTM. Thanks!

PR #4986, opened 2025-05-30
buffer: add feature to evacuate chunk files when retry limit

daipom:
None.

Add feature to evacuate chunk files when retry limit.
When reached the retry limit, `buf_file` and `buf_file_single` evacuates all the chunk files (and the meta files) in the queue to the following dir before purging.

* `(root_dir)/buffer/(plugin-id)/`

`root_dir` is `system_config.root_dir` if it is configured.
Otherwise, `DEFAULT_BACKUP_DIR` is applied.
(`/tmp/fluent`. We can change this by env var `FLUENT_BACKUP_DIR`)

There is no separate directory for each worker because the IDs of each chunk are entirely unique.
This makes recovery easier.

After the problem with the flush (such as a network issue) is resolved, we can put back the files and restart Fluentd to flush them again.

## Difference from the `backup` feature:

The `backup` feature is for unrecoverable errors, mainly for bad chunks.
On the other hand, this feature is for normal chunks.
The main motivation for this feature is to enable recovery by evacuating buffer files
when the retry limit is reached due to external factors such as network issues.

## Difference from the `secondary` feature:

The `secondary` feature is not suitable for recovery.
It can be difficult to recover files made by `out_secondary_file` because the metadata is lost.
For file buffers, the easiest way for recovery is to evacuate the chunk files as is.
Once the issue is recovered, we can put back the chunk files, and restart Fluentd to load them.
This feature enables it.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/583

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/plugin/buf_file.rb
+++ b/lib/fluent/plugin/buf_file.rb
@@ -229,6 +229,20 @@ module Fluent
         File.unlink(path, path + '.meta') rescue nil
       end
 
+      def evacuate_chunk(chunk)
+        unless chunk.is_a?(Fluent::Plugin::Buffer::FileChunk)
+          raise ArgumentError, "The chunk must be FileChunk, but it was #{chunk.class}."
+        end
+
+        backup_dir = File.join(backup_base_dir, 'buffer', safe_owner_id)
+        FileUtils.mkdir_p(backup_dir, mode: system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION) unless Dir.exist?(backup_dir)
+
+        FileUtils.copy([chunk.path, chunk.meta_path], backup_dir)
+        log.warn "chunk files are evacuated to #{backup_dir}.", chunk_id: dump_unique_id_hex(chunk.unique_id)
+      rescue => e
+        log.error "unexpected error while evacuating chunk files.", error: e
+      end
+
       private
 
       def escaped_patterns(patterns)
--- a/lib/fluent/plugin/buf_file_single.rb
+++ b/lib/fluent/plugin/buf_file_single.rb
@@ -241,6 +241,20 @@ module Fluent
         File.unlink(path) rescue nil
       end
 
+      def evacuate_chunk(chunk)
+        unless chunk.is_a?(Fluent::Plugin::Buffer::FileSingleChunk)
+          raise ArgumentError, "The chunk must be FileSingleChunk, but it was #{chunk.class}."
+        end
+
+        backup_dir = File.join(backup_base_dir, 'buffer', safe_owner_id)
+        FileUtils.mkdir_p(backup_dir, mode: system_config.dir_permission || Fluent::DEFAULT_DIR_PERMISSION) unless Dir.exist?(backup_dir)
+
+        FileUtils.copy(chunk.path, backup_dir)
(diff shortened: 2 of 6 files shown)


lib/fluent/plugin/buf_file.rb
```diff
@@ -229,6 +229,22 @@ def handle_broken_files(path, mode, e)
         File.unlink(path, path + '.meta') rescue nil
       end
 
+      def evacuate_chunk(chunk)
+        unless chunk.is_a?(Fluent::Plugin::Buffer::FileChunk)
+          raise ArgumentError, "The chunk must be FileChunk, but it was #{chunk.class}."
+        end
+        safe_owner_id = owner.plugin_id.gsub(/[ "\/\\:;|*<>?]/, '_')
+        backup_base_dir = system_config.root_dir || DEFAULT_BACKUP_DIR
```
ashie, line 237: It would be better to unify the code that overlaps with buffer.rb.

lib/fluent/plugin/buffer.rb
```diff
@@ -636,6 +637,25 @@ def clear_queue!
...
+        #
+        # Note: Difference from the `secondary` feature.
+        #       The `secondary` feature is not suitable for recovery.
+        #       It can be difficult to recover files made by `out_secondary_file` because the metadata
+        #       is lost.
+        #       For file buffers, the easiest way for recovery is to evacuate the chunk files as is.
+        #       Once the issue is recovered, we can put back the chunk files, and restart Fluentd to
+        #       load them.
+        #       This feature enables it.
+      end
```
Watson1978, line 657: It would be better if this method has guard sentence, it would found the miss easily in future.

Like,
```ruby
  raise NotImplementedError, "BUG: buffer plugins MUST implement this method"
```

lib/fluent/plugin/buf_file_single.rb
```diff
@@ -241,6 +241,20 @@ def handle_broken_files(path, mode, e)
         File.unlink(path) rescue nil
       end
 
+      def evacuate_chunk(chunk)
+        unless chunk.is_a?(Fluent::Plugin::Buffer::FileSingleChunk)
```
Watson1978, line 245: If you need to check strictly, you should use `instance_of?`.

```
irb(main):001> class A; end
=> nil
irb(main):002> class B < A; end
=> nil
irb(main):003> obj = B.new
=> #<B:0x00007ce071d3cc18>
irb(main):004> obj.is_a?(B)
=> true
irb(main):005> obj.is_a?(A)
=> true
irb(main):006> obj.instance_of?(B)
=> true
irb(main):007> obj.instance_of?(A)
=> false
```

`is_a?` method passes the inherited class as well.

lib/fluent/plugin/buffer.rb
daipom: Thanks!
However, it should be optional for buffer plugins to overwrite this method.

The default behavior of this method is to do nothing.
Buffer plugins can overwrite this on demand.

For example, `MemoryBuffer` does not overwrite this method.

If we force plugins to overwrite this method, it destroys the compatibility of existing third-party buffer plugins.

Watson1978 (approved):
👍🏻

daipom:
Thanks for your review!

Issue #4994, opened 2025-06-04
memory leak with `pattern not matched` error


Watson1978:
### Describe the bug

When many logs do not match the pattern, memory usage increases significantly.

### To Reproduce

* conf
```
<source>
  @type tail
  tag test
  path "#{File.expand_path '~/tmp/fluentd/access-*.log'}"
  refresh_interval 5s
  read_from_head true
  <parse>
    @type regexp
    expression /^(?<log>a+)+$/
    timeout 0.1s
  </parse>
</source>

<match **>
  @type stdout
</match>
```

* script to generate log data

```ruby
# frozen_string_literal: true
require "fileutils"

FILE_MAX_SIZE = 50 * 1024 * 1024
FILE_PATH = "/home/watson/tmp/fluentd/access-1.log"

dir = File.dirname(FILE_PATH)
FileUtils.mkdir_p(dir)

File.open(FILE_PATH, "w") do |f|

  data = "#{'a' * 1024}X"
  loop do
    f.puts data
    break if File.size(FILE_PATH) > FILE_MAX_SIZE
  end
end
```

When try to reproduce, fluentd will display the warning log as following and increase memory usage during handling the message.

```
2025-06-04 18:18:37 +0900 [info]: following tail of /home/watson/tmp/fluentd/access-1.log
[...]

PR #4995, opened 2025-06-05
Limit high-memory logging ('pattern not matched' in parsing) to necessary log level

Watson1978:
Fixes #

Currently, the “pattern not matched” warning is always generated regardless of the `log_level` setting. This warning might use a lot of memory when the log is very long.

To reduce memory usage, this PR will generate warnings only when necessary, according to `log_level`.

Related to https://github.com/fluent/fluentd/issues/4994

* config
```
<system>
  log_level error
</system>

<source>
  @type tail
  tag test
  refresh_interval 5s
  read_from_head true
  <parse>
    @type regexp
    expression /^(?<log>a+)+$/
    timeout 0.1s
  </parse>
</source>

<match **>
  @type stdout
</match>
```

Here is memory usage.

**Docs Changes**:
Not needed.

**Release Note**: 
Same as the title.

### DIFF
--- a/lib/fluent/compat/socket_util.rb
+++ b/lib/fluent/compat/socket_util.rb
@@ -148,7 +148,7 @@ module Fluent
         def on_message(msg, addr)
           @parser.parse(msg) { |time, record|
             unless time && record
-              log.warn "pattern not matched: #{msg.inspect}"
+              log.warn { "pattern not matched: #{msg.inspect}" }
               return
             end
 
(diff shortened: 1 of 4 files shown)


daipom:
Thanks!

Note: Long strings are allocated on the heap, so they are likely not released immediately.
It would be better not to generate unnecessary long strings.
(It probably won't affect normal use cases, though)

daipom (approved):

PR #1874, opened 2018-02-28
buf_file: Skip and delete broken file chunks during resume. fix #1760

repeatedly:
On k8s environment, we received the report "file chunks are unexpectedly broken and it causes error loop during resume."
This patch mitigates this problem by ignoring broken chunks in resume.

PR #4998, opened 2025-06-09
file_chunk: add stricter checks for broken meta files

Watson1978:
* Related to #3970

This PR improves meta file corruption checking.

The meta file contains at least the following field values.

https://github.com/fluent/fluentd/blob/fa2eb58922e1c36f83bf1d5243b325a860f72864/lib/fluent/plugin/buffer/file_chunk.rb#L249-L254

This PR reinforces https://github.com/fluent/fluentd/pull/1874.

Without this changes, it might causes following error when launch fluentd every time with broken meta file:

```
2025-06-06 12:11:26 +0900 [error]: unexpected error while checking flushed chunks. ignored. error_class=NoMethodError error="undefined method '<' for nil"
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin/output.rb:1479:in 'block in Fluent::Plugin::Output#enqueue_thread_run'
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin/buffer.rb:548:in 'block in Fluent::Plugin::Buffer#enqueue_all'
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin/buffer.rb:542:in 'Array#each'
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin/buffer.rb:542:in 'Fluent::Plugin::Buffer#enqueue_all'
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin/output.rb:1479:in 'Fluent::Plugin::Output#enqueue_thread_run'
  2025-06-06 12:11:26 +0900 [error]: /Users/watson/src/fluentd/lib/fluent/plugin_helper/thread.rb:78:in 'block in Fluent::PluginHelper::Thread#thread_create'
```

If the timekey value is corrupted, the above error occurs.
Since there is no appropriate way to check timekey directly, check `id`, `c`, and `m` fields instead. This is because when timekey is broken, other fields may also be broken.

It might be possible that the `@size` is 0.
`@unique_id`, `@created_at`, and `@modified_at` are set when FileChunk is initialized, so they definitely have some values.
I think these fields should be written in meta file.

So, this PR adds the `id`, `c`, and `m` fields check.

Previously, it operates using default value if metadata was broken. 
However, it can miss the corruption and result in unexpected errors.
So, this PR enhances the detection of broken metadata files instead of using defalut value.

This change has backward compatible with v0.14 behavior.

**Docs Changes**:
Not necessarily required.

**Release Note**: 
buf_file: reinforce buffer file corruption check

### DIFF
--- a/test/plugin/test_buf_file.rb
+++ b/test/plugin/test_buf_file.rb
@@ -1320,32 +1320,36 @@ class FileBufferTest < Test::Unit::TestCase
       compare_log(@p, 'enqueued meta file is broken')
       assert { not File.exist?(p2) }
       assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(c2id)}.log") }
+    end
 
-      # broken id, c, m fields in meta data
-      c3id, p3 = create_first_chunk('q')
-      metadata = File.read(p3 + '.meta')
-      File.open(p3 + '.meta', 'wb') { |f| f.write(metadata[0..6] + "\0" * (metadata.size - 6)) } # create enqueued broken meta file
-
-      Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
-        @p.start
-      end
-
-      compare_log(@p, 'enqueued meta file is broken')
-      assert { not File.exist?(p3) }
-      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(c3id)}.log") }
-
-      # truncate meta data
-      c4id, p4 = create_first_chunk('q')
-      metadata = File.read(p4 + '.meta')
-      File.open(p4 + '.meta', 'wb') { |f| f.write(metadata[0..-2]) } # create enqueued broken meta file with last byte truncated
+    test '#resume backups enqueued broken metadata which has broken id, c, m fields' do
+      setup_plugins({'path' => @bufpath})
+      cid, path = create_first_chunk('q')
+      metadata = File.read(path + '.meta')
+      File.open(path + '.meta', 'wb') { |f| f.write(metadata[0..6] + "\0" * (metadata.size - 6)) } # create enqueued broken meta file
 
       Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
         @p.start
       end
 
       compare_log(@p, 'enqueued meta file is broken')
-      assert { not File.exist?(p4) }
-      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(c4id)}.log") }
+      assert { not File.exist?(path) }
+      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(cid)}.log") }
+    end
+
+    test '#resume backups enqueued broken metadata by truncated' do
+      setup_plugins({'path' => @bufpath})
+      cid, path = create_first_chunk('q')
+      metadata = File.read(path + '.meta')
+      File.open(path + '.meta', 'wb') { |f| f.write(metadata[0..-2]) } # create enqueued broken meta file with last byte truncated
+
+      Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
+        @p.start
+      end
+
+      compare_log(@p, 'enqueued meta file is broken')
+      assert { not File.exist?(path) }
+      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(cid)}.log") }
     end
 
     test '#resume throws away broken chunk with disable_chunk_backup' do


test/plugin/test_buf_file.rb
```diff
@@ -1180,9 +1180,7 @@ def write_metadata(path, chunk_id, metadata, size, ctime, mtime)
   sub_test_case 'there are existing broken file chunks' do
     setup do
       @id_output = 'backup_test'
-      @bufdir = File.expand_path('../../tmp/broken_buffer_file', __FILE__)
-      FileUtils.rm_rf @bufdir rescue nil
-      FileUtils.mkdir_p @bufdir
+      @bufdir = Dir.mktmpdir
```
Watson1978, line 1183: hmm, `FileUtils.rm_rf` sometimes fails to execute if `@bufdir` already exists on Windows platform.
Therefore `Dir.mktmpdir` is used instead in here.

lib/fluent/plugin/buffer/file_chunk.rb
```diff
@@ -219,13 +219,17 @@ def restore_metadata(bindata)
             # old type of restore
             data = Fluent::MessagePackFactory.msgpack_unpacker(symbolize_keys: true).feed(bindata).read rescue {}
           end
+          raise FileChunkError, "invalid meta data" if data.nil? || !data.is_a?(Hash)
+          raise FileChunkError, "invalid unique_id" unless data[:id]
+          raise FileChunkError, "invalid created_at" unless data[:c].to_i > 0
+          raise FileChunkError, "invalid modified_at" unless data[:m].to_i > 0
```
Watson1978, line 225: `restore_metadata` has been invoked only when `*.meta` file is existed.
This means that `write_metadata` has been invoked before, and the meta file was created.

`write_metadata checks that the file is not corrupted by ensuring that the field to be written has the appropriate value.

In here,  it checks the field to be written by `write_metadata` has the appropriate value.
If the value is not appropriate, it is likely that some kind of error occurred during writing.

test/plugin/test_buf_file.rb
daipom: This is because of the added test code.
The same instance is started multiple times within a single test.
Could you please fix the test `#resume backups enqueued broken metadata`?

daipom: Then, we don't need to change this `setup`.

test/plugin/test_buf_file.rb
```diff
@@ -1320,6 +1319,32 @@ def compare_log(plugin, msg)
...
+      metadata = File.read(p4 + '.meta')
+      File.open(p4 + '.meta', 'wb') { |f| f.write(metadata[0..-2]) } # create enqueued broken meta file with last byte truncated
+
+      Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
+        @p.start
+      end
+
+      compare_log(@p, 'enqueued meta file is broken')
+      assert { not File.exist?(p4) }
+      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(c4id)}.log") }
```
daipom, line 1347: https://github.com/fluent/fluentd/pull/4998#discussion_r2168147899

lib/fluent/plugin/buffer/file_chunk.rb
daipom: Could you please make it clear that this does not cause any compatibility issues with the v0.14 API in https://github.com/fluent/fluentd/pull/4998#issue-3129583759?

test/plugin/test_buf_file.rb
```diff
@@ -1320,6 +1319,32 @@ def compare_log(plugin, msg)
...
+      metadata = File.read(p3 + '.meta')
+      File.open(p3 + '.meta', 'wb') { |f| f.write(metadata[0..6] + "\0" * (metadata.size - 6)) } # create enqueued broken meta file
+
+      Fluent::SystemConfig.overwrite_system_config('root_dir' => @bufdir) do
+        @p.start
+      end
+
+      compare_log(@p, 'enqueued meta file is broken')
+      assert { not File.exist?(p3) }
+      assert { File.exist?("#{@bufdir}/backup/worker0/#{@id_output}/#{@d.dump_unique_id_hex(c3id)}.log") }
```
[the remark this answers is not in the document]
Watson1978: Fixed with https://github.com/fluent/fluentd/pull/4998/commits/b44b0542b95a26393fab0f4199a3f3832e8464f7

test/plugin/test_buf_file.rb
Watson1978: Fixed with https://github.com/fluent/fluentd/commit/b44b0542b95a26393fab0f4199a3f3832e8464f7Fixed with https://github.com/fluent/fluentd/commit/b44b0542b95a26393fab0f4199a3f3832e8464f7

cosmo0920 (commented):
This metadata could be handled as a hash(map) like data structures. 
So, this is just my two cents but it seems that we can add shasum hash for record contents into that metadata?
is it feasible?

lib/fluent/plugin/buffer/file_chunk.rb
daipom: Thanks!

```diff
- However, it causes error with default value.
+ However, it can miss the corruption and result in unexpected errors.
```

I have changed this expression slightly to avoid misunderstandings.

daipom:
> This metadata could be handled as a hash(map) like data structures. So, this is just my two cents but it seems that we can add shasum hash for record contents into that metadata. is it feasible?

Thanks @cosmo0920 !

It would certainly be technically feasible to introduce a checksum for the chunk in the metadata.
However, the implementation cost will likely be significant.
For now, I'm considering this and #5003 as relatively easy improvements in the upcoming v1.19 release.

daipom (approved):
Thanks! LGTM!

PR #5017, opened 2025-07-04
config: add <system> umask setting to fluent.conf

### LINKED ISSUE #4816: System config: add umask option

blazethunderstorm:
Fixes #4816 

This PR adds support for configuring `umask` via the `<system>` directive in `fluent.conf`.  
This allows users to control default file permissions for buffer/output files without needing to set `--umask` on the command line, making container and system deployments cleaner and more flexible.

**Docs Changes**:  
Yes – adds mention of `umask` under `<system>` configuration in the documentation.

**Release Note**:  
Add support for `umask` configuration via the `<system>` directive in `fluent.conf`.

### DIFF
--- a/fluent.conf
+++ b/fluent.conf
@@ -1,6 +1,11 @@
 # In v1 configuration, type and id are @ prefix parameters.
 # @type and @id are recommended. type and id are still available for backward compatibility
 
+<system>
+  umask 0022
+  log_level info
+</system>
+
 ## built-in TCP input
 ## $ echo <json> | fluent-cat <tag>
 <source>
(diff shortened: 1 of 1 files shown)


blazethunderstorm:
@daipom pls review

daipom:
Thanks, but did you check the behavior?
It looks that this just adds the umask setting to the sample config file.
It will not work since Fluentd does not support umask in system config.
What we need to do is to support umask in system config, instead of changing the sample config.

blazethunderstorm:
@daipom alright

PR #5028, opened 2025-07-19
rubocop: use plugin scope logger

kenhys:
Fixes #

* Delay string interpolation for log.debug
* Use plugin scope logger instead of global scope $log

This issue was found by the following rubocop:

```
  Lint/FluentdPluginLogScope:
    AssumeConfigLogLevel: 'info'
    Exclude:
      - 'lib/fluent/*.rb'
      - 'lib/fluent/compat/*.rb'
    Enabled: true
```

**Docs Changes**:

N/A

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/in_monitor_agent.rb
+++ b/lib/fluent/plugin/in_monitor_agent.rb
@@ -204,7 +204,7 @@ module Fluent::Plugin
     def start
       super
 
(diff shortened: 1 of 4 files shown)


Watson1978 (approved):
👍🏻

PR #5036, opened 2025-07-28
out_stdout: support output to $stdout independently of logger

### LINKED ISSUE #5029: Output to Log file changes the destination of stdout output plugin too
### Describe the bug

Requirement:
Fluentd runs as a kubernetes container
 i.) Fluentd's own logs should be redirected to a log file
ii.) Fluentd is expected to read some other data from other plugins like in_http and emit that data onto stdout.

Issue: 
To achieve (i.), -o command-line option is used as suggested at https://docs.fluentd.org/deployment/logging#by-command-line-option-1
```
$ fluentd -o /var/log/fluentd.log
```

To achieve (ii.), a match with [stdout output plugin](https://docs.fluentd.org/output/stdout) is added in the config.
For ex.
```
<source>
  @type http
  port 9880
  bind 0.0.0.0
  tag test
</source>

<match test>
    @type stdout
 </match>
```
If we configure fluentd process to output to a log-file, it also redirects the data (tagged test) intended for stdout, into the  /var/log/fluentd.log file. It does not emit anything on stdou anymore.

Is this expected?

How to avoid it such that the log location and stdout location are separate?

### To Reproduce

1. Configure fluentd config (as shared above)
2. Start fluentd with -o option
 fluentd -o /var/log/fluentd.log
4. Send data on the http input. For ex
`curl -X POST -d 'json={"foo":"bar"}' http://localhost:9880/test`

### Expected behavior

Fluentd own logs must go to logfile.
Records read from http must be present on stdout of fluentd

### Your Environment

```markdown
- Fluentd version: 1.16.9
- Package version:5.0.7
- Operating system: Rocky Linux 8.10
- Kernel version:6.6.87.2-microsoft-standard-WSL2
```

### Your Configuration

```apache
<source>
  @type http
  port 9880
  bind 0.0.0.0
  tag test
</source>

<match test>
    @type stdout
 </match>
```

### Your Error Log

```shell
No error log, there is no data emitted on stdout of fluentd container.
```

### Additional context

_No response_

daipom:
* Fixes #5029

Add `use_logger` option to `out_stdout` plugin.

* Type: `bool`
* Default: `true`

If you set `false` to this option, `out_stdout` outputs events to `$stdout`, instead of `$logger`.
This feature enables events to be output to STDOUT in containerized environments, such as K8s.

This does not change the default behavior.

Note: Although it's preferable for the default value of a `bool` option to be `false` (because it can be used as a flag), I couldn't think of a suitable name for it.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/593

**Release Note**: 
out_stdout: support output to STDOUT independently of Fluentd logger by setting `use_logger` to `false`.

### DIFF
--- a/lib/fluent/plugin/out_stdout.rb
+++ b/lib/fluent/plugin/out_stdout.rb
@@ -25,6 +25,9 @@ module Fluent::Plugin
     DEFAULT_LINE_FORMAT_TYPE = 'stdout'
     DEFAULT_FORMAT_TYPE = 'json'
 
+    desc "If Fluentd logger outputs logs to a file, this plugin outputs events to the file as well."
+    config_param :use_logger, :bool, default: true
+
     config_section :buffer do
       config_set_default :chunk_keys, ['tag']
       config_set_default :flush_at_shutdown, true
@@ -44,6 +47,10 @@ module Fluent::Plugin
       true
     end
 
+    def dest_io
+      @use_logger ? $log : $stdout
+    end
+
     attr_accessor :formatter
 
     def configure(conf)
@@ -57,9 +64,9 @@ module Fluent::Plugin
     def process(tag, es)
       es = inject_values_to_event_stream(tag, es)
       es.each {|time,record|
-        $log.write(format(tag, time, record))
+        dest_io.write(format(tag, time, record))
       }
-      $log.flush
+      dest_io.flush
     end
 
     def format(tag, time, record)
@@ -68,7 +75,7 @@ module Fluent::Plugin
     end
 
     def write(chunk)
-      chunk.write_to($log)
+      chunk.write_to(dest_io)
     end
   end
 end
--- a/lib/fluent/plugin/out_stdout.rb
+++ b/lib/fluent/plugin/out_stdout.rb
@@ -25,7 +25,7 @@ module Fluent::Plugin
     DEFAULT_LINE_FORMAT_TYPE = 'stdout'
     DEFAULT_FORMAT_TYPE = 'json'
 
-    desc "If Fluentd logger outputs logs to a file, this plugin outputs events to the file as well."
+    desc "If Fluentd logger outputs logs to a file (with -o option), this plugin outputs events to the file as well."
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_stdout.rb
```diff
@@ -25,6 +25,9 @@ class StdoutOutput < Output
     DEFAULT_LINE_FORMAT_TYPE = 'stdout'
     DEFAULT_FORMAT_TYPE = 'json'
 
+    desc "If Fluentd logger outputs logs to a file, this plugin outputs events to the file as well."
+    config_param :sync_logger, :bool, default: true
```
Watson1978, line 29: We discussed that it would be better to change this parameter to `use_logger`.

kenhys: Another idea: `follow_logger_output: true`, but `use_logger` might be straightforward enough.

daipom: Thanks!
I have changed the option name to `use_logger`!

Watson1978 (dismissed):
👍🏻

lib/fluent/plugin/out_stdout.rb
```diff
@@ -25,6 +25,9 @@ class StdoutOutput < Output
     DEFAULT_LINE_FORMAT_TYPE = 'stdout'
     DEFAULT_FORMAT_TYPE = 'json'
 
+    desc "If Fluentd logger outputs logs to a file, this plugin outputs events to the file as well."
```
kenhys, line 28:
```suggestion
    desc "If Fluentd logger outputs logs to a file (with -o option), this plugin outputs events to the file as well."
```

It might be redundant, it helps to understand what it means.

kenhys (changes requested):

lib/fluent/plugin/out_stdout.rb
daipom: Thanks! Applied!

kenhys (approved):
LGTM.

PR #5055, opened 2025-08-14
ci: remove reference to mock object after test

Watson1978:
**Which issue(s) this PR fixes**: 
Fixes #5054

**What this PR does / why we need it**: 

**Docs Changes**:

**Release Note**:

Issue #5054, opened 2025-08-12
CI: Failed ObjectSpaceInputTest


Watson1978:
### Describe the bug

Now ObjectSpaceInputTest causes following error.

```
ObjectSpaceInputTest: 
  test_configure:                                                                                                                                               .: (0.001749)
  test_emit:                                                                                                                                                    F
=======================================================================================================================================================================================
Failure: test_emit(ObjectSpaceInputTest):
      assert{ emits.length > 0 }
              |     |      |
              |     |      false
              |     0
              []
/home/watson/src/fluentd/test/plugin/test_in_object_space.rb:58:in `test_emit'
     55:     d.run(expect_emits: 3)
     56: 
     57:     emits = d.events
  => 58:     assert{ emits.length > 0 }
     59: 
     60:     emits.each { |tag, time, record|
     61:       assert_equal d.instance.tag, tag
=======================================================================================================================================================================================
: (300.285955)
```

### To Reproduce

`bundle exec rake`

### Expected behavior

No error

### Your Environment

```markdown
- Fluentd version:
- Package version:
- Operating system:
- Kernel version:
```

### Your Configuration

```apache
none
```

### Your Error Log

```shell
Ref. https://github.com/fluent/fluentd/actions/runs/16896497001
```

Watson1978:
Seems this is related to [flexmock](https://github.com/doudou/flexmock) v2.4.3,
it has a bug.

https://github.com/doudou/flexmock/releases/tag/v2.4.4

Watson1978:
If I downgrade `flexmock` version to `2.4.2`, the tests will be passed.

Watson1978:
Since https://github.com/doudou/flexmock/commit/4c07690005aaa53eb627f37d26ea8f96c85280b0 commit, we can reproduce the problem...

PR #5085, opened 2025-09-10
Backport(v1.16): test: ensure removing old engine on setup (#5057)

daipom:
* Backport #5057

Continued from

* #5054
* #5055 (6cac9f0fc226a6d6046a64a59789a1b7500242da)

Each test should not consider the initialization of `Fluent::Engine`. It should be the responsibility of `Fluent::Test.setup`.

Note: Set `nil` explicitly to ensure that GC can remove the objects, though it is very strange that some objects can still exist after `remove_const` and `GC.start`.

**Docs Changes**:
Not needed.

**Release Note**:
CI improvements.

### DIFF
--- a/lib/fluent/test.rb
+++ b/lib/fluent/test.rb
@@ -40,9 +40,12 @@ module Fluent
 
       $log = dummy_logger
 
-      Fluent.__send__(:remove_const, :Engine)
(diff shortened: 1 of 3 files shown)


daipom:
Need to merge the following PR first.

* #5086

PR #5094, opened 2025-09-11
compressable: Handle zstd frames correctly

cosmo0920:
This change is needed because, without this change, we just received unknown zstd frames like:

forward: len=12984, head=54 d1 00 da cc 74 2e 4e 20 d6 2a ce 01 be f2 c2 | compressed(opt)=2

But, zstd specification always needs to attach the head of magic bytes like:

forward: len=19835, head=28 b5 2f fd a0 c2 14 03 00 a4 d0 00 4a f7 30 39 | compressed(opt)=2

So, we need to attach the head of magic bytes `28 b5 2f fd` in zstd compressed payloads.

This can be dumped with:

```diff
diff --git a/lib/fluent/plugin/out_forward.rb b/lib/fluent/plugin/out_forward.rb
index 4c323bb0..f192d6e3 100644
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -672,6 +672,9 @@ module Fluent::Plugin
         sock.write @sender.forward_header                    # array, size=3
         sock.write tag.to_msgpack                            # 1. tag: String (str)
         chunk.open(compressed: @compress) do |chunk_io|
+          head = chunk_io.read(8) || ''.b
+          @log.info "debug: forward entries head", hex: head.bytes.map { |b| "%02x" % b }.join(' ')
+          chunk_io.rewind
           entries = [0xdb, chunk_io.size].pack('CN')
           sock.write entries.force_encoding(Encoding::UTF_8) # 2. entries: String (str32)
           IO.copy_stream(chunk_io, sock)                     #    writeRawBody(packed_es)
```

None

This could be a known issue after merging https://github.com/fluent/fluentd/pull/4657.
This is because with that PR patch, we wasn't able to decompress zstd compressed insisted payloads **in Fluent Bit side**.
In our side, we need to set up explicit zstd frames with the head of magic bytes: `28 b5 2f fd`.

However, stream writer of zstd-ruby does not wrap up their compressing payloads with that zstd specific payloads.
So, we always experienced this kind of errors by using Fluent Bit's development version of in_forward with zstd compressed insisted payloads.
With gzip compressed payloads, there is no issue but the behavior differences of StreamWrite class between Gzip and Zstd could cause this issue.

The related Fluent Bit's PR is:
https://github.com/fluent/fluent-bit/pull/10710

**Docs Changes**:

**Release Note**:

### DIFF
--- a/lib/fluent/plugin/compressable.rb
+++ b/lib/fluent/plugin/compressable.rb
@@ -26,14 +26,29 @@ module Fluent
         io = output_io || StringIO.new
         if type == :gzip
           writer = Zlib::GzipWriter.new(io)
+          writer.write(data)
+          writer.finish
+          io.string
         elsif type == :zstd
-          writer = Zstd::StreamWriter.new(io)
+          sc = Zstd::StreamingCompress.new(**kwargs)
+          chunk_size = kwargs[:chunk_size] || (256 * 1024)
+
+          if data.is_a?(String)
+            i = 0
+            while i < data.bytesize
+              io << sc.compress(data.byteslice(i, chunk_size))
+              i += chunk_size
+            end
+          else
+            data.each { |d| io << sc.compress(d) }
+          end
+
+          # Need to close frame to prevent unknown frame descriptor errors
+          io << sc.finish
+          io.string
         else
           raise ArgumentError, "Unknown compression type: #{type}"
         end
-        writer.write(data)
-        writer.finish
-        output_io || io.string
       end
 
       # compressed_data is String like `compress(data1) + compress(data2) + ... + compress(dataN)`


cosmo0920:
How's going this PR, mate?
Should we deeply dive into the dependent gem like zstd-ruby?

daipom:
Sorry, I haven’t been able to make time over the past few days.
Thanks for the reproduction steps!
I'll try it and review this change.

daipom:
Sorry for the delay.
It appears this is a bug of `Zstd::StreamWriter`.
I have reported it to the upstream: https://github.com/SpringMT/zstd-ruby/issues/112#issuecomment-3337520133.

cosmo0920:
> ```diff
> ```diff
> 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> It appears we don't need to use the Streaming feature.
> It seems sufficient to just do a single compression and append it to the IO.
> I have confirmed this resolves the error mentioned in #5094 (comment).
> ```

Is this patch works for Enumerable argument of data?

cosmo0920:
Hi,
I also worked and addressed for larger compression cases on zstd-ruby:
https://github.com/SpringMT/zstd-ruby/pull/116

Does this PR resolve this kind of issues?

daipom:
> > It appears we don't need to use the Streaming feature.
> > It seems sufficient to just do a single compression and append it to the IO.
> > I have confirmed this resolves the error mentioned in #5094 (comment).
> 
> Does this patch work for Enumerable type argument of data?

Does `Fluent::Plugin::Compressable#compress()` need to support Enumerable?
(Does the existing code support it? Looks like it does not support it.)

Note: `Fluent::Plugin::Compressable#compress()` is used here (`EventStream#to_msgpack_stream`).

https://github.com/fluent/fluentd/blob/14c0a0bebe06284320d8a1ef096a063a51c65455/lib/fluent/event.rb#L65-L68

Enumerable is used here (`Chunk#append`).
(We need to consider Enumerable here.)

https://github.com/fluent/fluentd/blob/14c0a0bebe06284320d8a1ef096a063a51c65455/lib/fluent/plugin/buffer/chunk.rb#L251-L263

daipom:
> Hi, I also worked and addressed for larger compression cases on zstd-ruby: [SpringMT/zstd-ruby#116](https://github.com/SpringMT/zstd-ruby/pull/116)
> 
> Does this PR resolve this kind of issues without this patch? Because implementing in C would be better to handle zstd compression in large amount of data processing environments. So, I wanted to confirm with Fluentd Core team.

Thanks so much!
Yes! The library fix should solve this issue. I will confirm it.

After this issue itself is resolved, it might still be worth considering the point that we may not need to use the Streaming feature of zstd-ruby.

daipom:
I have confirmed this issue is resolved with zstd-ruby 1.5.7.1!

daipom:
> After this issue itself is resolved, it might still be worth considering the point that we may not need to use the Streaming feature of zstd-ruby.

It may be better to address this point in a separate PR.
Perhaps we can close this PR now.

cosmo0920:
> > After this issue itself is resolved, it might still be worth considering the point that we may not need to use the Streaming feature of zstd-ruby.
> 
> It may be better to address this point in a separate PR. I made the issue for it.
> 
> * [zstd compression: stop using the streaming feature #5111](https://github.com/fluent/fluentd/issues/5111)
> 
> Perhaps we can close this PR now.

Now, we need to take benckmark w/ streaming zstd compression or w/ normal zstd compression.
Do we need to avoid to use this kind of feature?
Even though I fixed the experimental-like feature of streaming compression in zstd-ruby.
So, I mean, your company insists that "we fix OSS bugs in their upstream and we tend to report their bugs with patches as much as possible".
Just avoiding to use the buggy feature seems not to follow your company's credo.

daipom:
> > > After this issue itself is resolved, it might still be worth considering the point that we may not need to use the Streaming feature of zstd-ruby.
> > 
> > 
> > It may be better to address this point in a separate PR. I made the issue for it.
> > 
> > * [zstd compression: stop using the streaming feature #5111](https://github.com/fluent/fluentd/issues/5111)
> > 
> > Perhaps we can close this PR now.
> 
> Now, we need to take benckmark w/ streaming zstd compression or w/ normal zstd compression. Do we need to avoid to use this kind of feature? Even though I fixed the experimental-like feature of streaming compression in zstd-ruby. So, I mean, your company insists that "we fix OSS bugs in their upstream and we tend to report their bugs with patches as much as possible". Just avoiding to use the buggy feature seems not to follow your company's credo.

Ah, there might have been a misunderstanding.

If the logic fundamentally requires the streaming feature, it wouldn’t be desirable to give it up just because the feature is still experimental.
However, my concern is that the logic does not fundamentally need the streaming feature.
This is because the process simply writes the given data and finishes.

cosmo0920:
> > Now, we need to take benckmark w/ streaming zstd compression or w/ normal zstd compression. Do we need to avoid to use this kind of feature? Even though I fixed the experimental-like feature of streaming compression in zstd-ruby. So, I mean, your company insists that "we fix OSS bugs in their upstream and we tend to report their bugs with patches as much as possible". Just avoiding to use the buggy feature seems not to follow your company's credo.
> 
> Ah, there might have been a misunderstanding.
> 
> If the logic fundamentally requires the streaming feature, it wouldn’t be desirable to give it up just because the feature is still experimental. However, my concern is that the logic does not fundamentally need the streaming feature. This is because the process simply writes the given data and finishes.

Got it. The minimum required logic here is just terminating/creating for each zstd frame with the correct data structures and heading magic bytes in zstd with reasonable CPU or memory usages.
So, just to avoid using the streaming feature, it would be encountered for performance issues.
But, in reality, we sometimes don't have enough cycles to fix this kind of feature.
In this case, we could avoid using the streaming feature in zstd compression and ensuring the boundaries of zstd frames. 
Plus, personally, I am concerned about memory consumption when using the streaming feature in zstd compression. This could be a performance trade off which is like processing speed vs memory consumption.

PR #5126, opened 2025-10-20
YAML config: Supports parsing array format

spairy:
None.
  
Supports parsing array of basic data types when analyzing YAML configuration files.

The current behavior of Fluentd:

* Support YAML Array format for `$args` only.
* Other Array options must be specified in a comma-separated format.
  * `retryable_response_codes: 503, 504`
  * If there’s only a single value, it must be specified as String or with the trailing comma.
    * `retryable_response_codes: "503"`
    * `retryable_response_codes: 503,`

This supports YAML Array format for all options.

```
retryable_response_codes:
  - 503
```

```
retryable_response_codes: [503]
```

Note: This PR doesn't address the issue where setting an Int directly to an Array option causes ConfigError: `retryable_response_codes: 503` (The `C` case in https://github.com/fluent/fluentd/pull/5126#issuecomment-3424253314)
It could be a different issue. => #5149

**Docs Changes**: Not Need

**Release Note**: Same as the title.

### DIFF
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -343,6 +343,155 @@ class ConfigTest < Test::Unit::TestCase
       10.times { match_conf['type'] }
       assert_equal before_size, match_conf.unused.size
     end
+
+    data(
+      "One String for $arg" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              buffer:
+                $arg: tag
+                $type: memory
+                flush_mode: immediate
+      CONF
+      "Comma-separated String for $arg" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              buffer:
+                $arg: tag, time
+                $type: memory
+                timekey: 1h
+                flush_mode: immediate
+      CONF
+      "One-liner Array for $arg" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              buffer:
+                $arg: [tag, time]
+                $type: memory
+                timekey: 1h
+                flush_mode: immediate
+      CONF
+      "Multi-liner Array for $arg" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              buffer:
+                $arg:
+                  - tag
+                  - time
+                $type: memory
+                timekey: 1h
+                flush_mode: immediate
+      CONF
+      "One String for normal Array option" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              format:
+                $type: csv
+                fields: message
+      CONF
+      "Comma-separated String for normal Array option" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              inject:
+                time_key: timestamp
+                time_type: string
+              format:
+                $type: csv
+                fields: timestamp, message
+      CONF
+      "One-liner Array for normal Array option" => <<~CONF,
+        config:
+          - source:
+              $type: sample
+              tag: test
+          - match:
+              $type: stdout
+              $tag: test.**
+              inject:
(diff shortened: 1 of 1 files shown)


daipom:
Thanks for this improvement!
Could you add DCO?
Could you explain this feature in more detail?

spairy:
Gett error when config array of basic data types with fluentd.yaml, eg: 
  ```
  retryable_response_codes:
    - 503  
  ```
or
  `  retryable_response_codes: [503]`

 This feature  is to address this issue.

daipom:
I see. There seems to be a bug.
Fluentd should be able to parse YAML Array format: https://docs.fluentd.org/configuration/config-file-yaml#multiline-support-for-quoted-string-array-and-hash-values

I have confirmed the following behaviors.

* A: Can parse Array format for `$tag`.
* B: Cannot parse Array format for normal Array options.
* C: Cannot parse a single Int for normal Array options.
  * It appears that the workaround is to specify a single String. (See A example below)

# A: Can parse Array format for `$tag`

```yaml
config:
  - source:
      $type: http
  - match:
      $type: stdout
      $tag: server.**
      format:
        $type: csv
        fields: message
      buffer:
        $arg:
          - tag
          - time
        $type: memory
        timekey: 1h
        flush_mode: immediate
  - source:
      $type: sample
      tag: client
  - match:
      $type: http
      $tag: client.**
      endpoint: http://localhost:9880/server
      retryable_response_codes: "503"
      buffer:
        flush_mode: immediate
```

# B: Cannot parse Array format for normal Array options.

```
fluentd/lib/fluent/config/yaml_parser/parser.rb:145:in `section_build': undefined method `each' for "message":String (NoMethodError)
```

```yaml
config:
  - source:
      $type: http
  - match:
      $type: stdout
      $tag: server.**
      format:
        $type: csv
        fields:
          - message
      buffer:
        $arg:
          - tag
          - time
        $type: memory
        timekey: 1h
        flush_mode: immediate
  - source:
      $type: sample
      tag: client
  - match:
      $type: http
      $tag: client.**
      endpoint: http://localhost:9880/server
      retryable_response_codes:
        - 503
      buffer:
        flush_mode: immediate
```

# C: Cannot parse a single Int for normal Array options.

```
[error]: config error file="..." error_class=Fluent::ConfigError error="array required but got 503"
```

```yaml
config:
  - source:
      $type: http
  - match:
      $type: stdout
      $tag: server.**
      format:
        $type: csv
        fields: message
      buffer:
        $arg:
          - tag
          - time
        $type: memory
        timekey: 1h
        flush_mode: immediate
  - source:
      $type: sample
      tag: client
  - match:
      $type: http
      $tag: client.**
      endpoint: http://localhost:9880/server
      retryable_response_codes: 503
      buffer:
        flush_mode: immediate
```

daipom:
It’s still unclear whether the current implementation is intentional or an oversight (possibly a bug).
Here’s the current behavior of Fluentd.

* Support YAML Array format for `$args` only.
* Other Array options must be specified in a comma-separated format.
  * `retryable_response_codes: 503, 504`
  * If there’s only a single value, it must be specified as String.
    * `retryable_response_codes: "503"`

daipom:
We supported YAML in the following: 

* https://github.com/fluent/fluentd/pull/3712

There hasn’t been any particular discussion about handling Arrays. There are no test cases for it, but it’s clearly documented.

daipom:
There are two cases for Arrays: one where they should be recognized as sections, and another where they should be treated as single options.
Currently, the internal logic forces them to be recognized as sections.
To support Array options, we need to add cases like in this PR.

lib/fluent/config/yaml_parser/parser.rb
```diff
@@ -164,6 +168,9 @@ def section_build(name, config, indent: 0, arg: nil)
 
           SectionBuilder.new(name, sb, indent, arg)
         end
+        def basic_type?(value)
+          [Integer, Float, String, TrueClass, FalseClass, NilClass].any?{ |type| value.is_a?(type) }
+        end
```
daipom:
```suggestion

        def section?(value)
          value.is_a?(Array) or value.is_a?(Hash)
        end
```

daipom (commented):
Thanks!
The direction of this fix looks good!

Since the current implementation assumes either an Array or a Hash as a section, how about checking it like this?

daipom:
I’d appreciate it if you could add some test cases.
Please let me know if that’s difficult.

https://github.com/fluent/fluentd/blob/dc21d8313852f6bc2dfe1576ca9d97032d4ba3dc/test/test_config.rb#L161

Issue #5141, opened 2025-11-04
CI: Failed with `URI::InvalidComponentError`


Watson1978:
### Describe the bug

Recently, `uri` gem v1.1.0 was released.
https://github.com/ruby/uri/releases/tag/v1.1.0

Then, Fluentd's CI has been failed with `URI::InvalidComponentError` at `test/test_supervisor.rb`.

```
Error: test_rpc_server[ipv6](SupervisorTest): URI::InvalidComponentError: bad component(expected host component): 
/home/watson/.rbenv/versions/3.4.7/lib/ruby/gems/3.4.0/gems/uri-1.1.1/lib/uri/http.rb:73:in 'URI::HTTP#check_host'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/gems/3.4.0/gems/uri-1.1.1/lib/uri/generic.rb:653:in 'URI::Generic#host='
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http/generic_request.rb:230:in 'Net::HTTPGenericRequest#update_uri'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:2475:in 'Net::HTTP#begin_transport'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:2404:in 'Net::HTTP#transport_request'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:2378:in 'Net::HTTP#request'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:2249:in 'Net::HTTP#request_get'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:826:in 'block in Net::HTTP.get_response'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:1626:in 'Net::HTTP#start'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:1064:in 'Net::HTTP.start'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:824:in 'Net::HTTP.get_response'
/home/watson/.rbenv/versions/3.4.7/lib/ruby/3.4.0/net/http.rb:805:in 'Net::HTTP.get'
[...]

PR #5144, opened 2025-11-06
Backport(v1.19): gem: fix uri gem version to keep IPv6 tests (#5142)

Watson1978:
Backport https://github.com/fluent/fluentd/pull/5142

Related to #5141

Since uri 1.1.0, `Net::HTTP` will cause a `URI::InvalidComponentError` if we use IPv6 address.
To avoid the error, this PR will fix uri version to 1.0.x.

**Docs Changes**:

**Release Note**:


daipom (approved):
Thanks!

PR #5231, opened 2026-01-20
plugin_helper/http_server: Ensure request body is closed to prevent socket leaks in POST requests

Watson1978:
Fixes https://github.com/fluent-plugins-nursery/fluent-plugin-opentelemetry/issues/24

When using the `http_server` helper to handle POST requests (e.g., in plugins like `fluent-plugin-opentelemetry`),
a socket leak occurs, eventually leading to `Errno::EMFILE` (Too many open files).

This happens because `Async::HTTP` requires the request body to be fully consumed or explicitly closed to release the resources and allow connection reuse (Keep-Alive). If the body is left unclosed, the socket remains in a CLOSE_WAIT state on the server side indefinitely.

**Docs Changes**:
N/A

**Release Note**: 
* plugin_helper/http_server: Ensure request body is closed to prevent socket leaks in POST requests

### DIFF
--- a/lib/fluent/plugin_helper/http_server/app.rb
+++ b/lib/fluent/plugin_helper/http_server/app.rb
@@ -71,6 +71,8 @@ module Fluent
                 path
               end
             @router.route!(name, canonical_path, req)
+          ensure
+            request.body&.close
           end
         end
       end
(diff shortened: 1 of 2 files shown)


kenhys (approved):
LGTM.

PR #5239, opened 2026-01-27
fix: Support IPv6 bind addresses in http_server helper

I501307:
Fixes # (if there's an issue, or remove this line)

Fixes URI::InvalidURIError when binding to IPv6 addresses. The http_server 
helper was generating invalid URIs like 'http://:::24231' instead of the 
RFC 3986 compliant 'http://[::]:24231'.

**Release Note**: 
http_server helper: Fix IPv6 bind address support

### DIFF
--- a/test/plugin_helper/test_http_server_helper.rb
+++ b/test/plugin_helper/test_http_server_helper.rb
@@ -364,47 +364,22 @@ class HttpHelperTest < Test::Unit::TestCase
       end
     end
     
-    test 'http_server can bind and serve on IPv6 loopback ::1' do
+    data(
+      'IPv6 loopback ::1' => [:ipv6_loopback_test, '::1', '::1'],
+      'IPv6 any address ::' => [:ipv6_any_test, '::', '::1'],
+      'pre-bracketed IPv6 [::1]' => [:ipv6_bracketed_test, '[::1]', '::1']
+    )
+    test 'http_server can bind and serve on IPv6 address' do |data|
       return unless ipv6_enabled?
       
+      test_name, bind_addr, connect_addr = data
+      
       on_driver do |driver|
-        driver.http_server_create_http_server(:ipv6_loopback_test, addr: '::1', port: @port, logger: NULL_LOGGER) do |s|
+        driver.http_server_create_http_server(test_name, addr: bind_addr, port: @port, logger: NULL_LOGGER) do |s|
           s.get('/test') { [200, { 'Content-Type' => 'text/plain' }, 'OK'] }
         end
         
-        # Use Net::HTTP directly with proper IPv6 handling
-        http = Net::HTTP.new('::1', @port)
-        response = http.get('/test')
-        assert_equal('200', response.code)
-        assert_equal('OK', response.body)
-      end
-    end
-
-    test 'http_server can bind on IPv6 any address ::' do
-      return unless ipv6_enabled?
-      
-      on_driver do |driver|
-        driver.http_server_create_http_server(:ipv6_any_test, addr: '::', port: @port, logger: NULL_LOGGER) do |s|
-          s.get('/test') { [200, { 'Content-Type' => 'text/plain' }, 'OK'] }
-        end
-        
-        # Connect via loopback to test the :: bind worked
-        http = Net::HTTP.new('::1', @port)
-        response = http.get('/test')
-        assert_equal('200', response.code)
-        assert_equal('OK', response.body)
-      end
-    end
-
-    test 'http_server handles pre-bracketed IPv6 address [::1]' do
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin_helper/http_server/server.rb
```diff
@@ -37,7 +37,8 @@ def initialize(addr:, port:, logger:, default_app: nil, tls_context: nil)
 
           # TODO: support http2
           scheme = tls_context ? 'https' : 'http'
-          @uri = URI("#{scheme}://#{@addr}:#{@port}").to_s
+          addr_display = @addr.include?(":") ? "[#{@addr}]" : @addr
```
Watson1978: Can you add test for `[::]` in `@addr` ?

I501307: fixed here : 3aea00ea931d2354c235f6c5e5b01355d94b920b

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+## Release v1.19.1 - TBD
+
+### Bug fix
+
+* http_server helper: Fix IPv6 bind address support in URI construction
+  * Fixed `URI::InvalidURIError` when binding to IPv6 addresses (e.g., `::`, `::1`)
+  * IPv6 addresses are now properly bracketed in URIs per RFC 3986 (e.g., `http://[::]:24231`)
+  * Handles pre-bracketed addresses correctly to avoid double-bracketing
+  * Affects all plugins using http_server helper with IPv6 bind addresses
+
```
kenhys: Already v1.19.1 was released, so it should be v1.19.2 or v1.20 series.

I501307: fixed here : https://github.com/fluent/fluentd/pull/5239/changes/77b380aa1a4beb682704d2ebbc067ce8ed9f11a6

I501307: @kenhys I guess we are good to go now 🚀

kenhys (commented):
Thanks @I501307 !!

If you afford to write, it might be safer to write with conditional networking test case additionally.

```
test ... do
  omit ""  unless ipv6_enabled?
  ...
end
```

Watson1978 (dismissed):
Thanks 👍🏻

kenhys (changes requested):
> After: Tests only validate formatting logic → no networking required, works everywhere
Verified locally:

Thanks!

It works great without network, but if plugin helper implementation was changed (in the future), also
need to follow test code too. 
If we forgot to follow (it might be accidentally happen), it becomes meaningless test cases there.
So I have concern about extracting internal logic directly to test code.
What do you think? @Watson1978

I501307:
> > What do you think? @Watson1978
> 
> Sorry, I didn't get to look closely at the test code.
> 
> Indeed.
> 
> It should not only test the logic itself. it should actually run the http_server helper to verify it behaves as expected.

Integration tests are now added https://github.com/fluent/fluentd/pull/5239/changes#diff-0733209104ce038bd67b2898b481ab3b6ff2634063b71948ae8a230bb3d5bcf3R23-R29 🌻

I501307:
> > After: Tests only validate formatting logic → no networking required, works everywhere
> > Verified locally:
> 
> Thanks!
> 
> It works great without network, but if plugin helper implementation was changed (in the future), also need to follow test code too. If we forgot to follow (it might be accidentally happen), it becomes meaningless test cases there. So I have concern about extracting internal logic directly to test code. What do you think? @Watson1978

done here https://github.com/fluent/fluentd/pull/5239/changes#diff-0733209104ce038bd67b2898b481ab3b6ff2634063b71948ae8a230bb3d5bcf3R23-R29 🌻

test/plugin_helper/test_http_server_helper.rb
```diff
@@ -363,6 +363,53 @@ def start_https_request(addr, port, verify: true, cert_path: nil, selfsigned: tr
         end
       end
     end
+    
+    test 'http_server can bind and serve on IPv6 loopback ::1' do
```
kenhys: Thanks.

Could you use `data(...)`  test pattern to reduce redundant test code? (something like)

```
data(
'ipv6 loopback test' => [:ipv6_loopback_test, '::1'],
'ipv6 any test' => [:ipv6_any_test, '::']
...
)
test 'http_server can bind and serve on IPv6 address' do |test_case, addr|
...
end
```

kenhys: If above fix was done, LGTM.

kenhys: Looks good. rebased with HEAD.

kenhys:
@I501307 
Thanks for contribution, greatly appreciated.

PR #5262, opened 2026-02-18
Add send_keepalive_packet option to out_forward plugin

2ZZ:
Fixes #

This PR adds the `send_keepalive_packet` configuration parameter to the `out_forward` plugin to enable TCP keepalive (SO_KEEPALIVE) on forward output connections.

**Background:**
The existing `keepalive` parameter only enables connection pooling/reuse, not TCP-level keepalive probes. This can lead to accumulation of dead connections in CLOSE-WAIT state.

**Changes:**
- Added `send_keepalive_packet` config parameter (default: false)
- Pass the parameter to both `socket_create_tcp` and `socket_create_tls` calls
- Added tests to verify the parameter works correctly

**Related:**
This follows the same pattern as #2352 which added the same feature to `in_forward`.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/608

**Release Note**: 

* out_forward: Add `send_keepalive_packet` option

### DIFF
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -119,6 +119,8 @@ module Fluent::Plugin
     config_param :keepalive, :bool, default: false
     desc "Expired time of keepalive. Default value is nil, which means to keep connection as long as possible"
     config_param :keepalive_timeout, :time, default: nil
+    desc 'Check the remote connection is still available by sending a keepalive packet if this value is true.'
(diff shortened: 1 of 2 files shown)


Watson1978:
@2ZZ Can you add DCO sign in your commit and fix tests ?

Watson1978 (changes requested):
I think send_keepalive_packet should only be set to true when keepalive is set to true.




Related to https://github.com/fluent/fluentd/pull/2352

PR #5272, opened 2026-03-10
in_debug_agent: accept only from local machine by default

kenhys:
Fixes #

Usually, in_debug_agent must be explicitly enabled by users who know what you do.
But, there is an security concern which accepts external access by default even though user must enable it explicitly.

With this commit, change that behavior a bit secure by default.

**Docs Changes**:

Need to update default.

https://docs.fluentd.org/monitoring-fluentd/monitoring-rest-api#datadog-dd-agent-integration

**Release Note**: 

N/A

### DIFF
--- a/lib/fluent/plugin/in_debug_agent.rb
+++ b/lib/fluent/plugin/in_debug_agent.rb
@@ -26,7 +26,7 @@ module Fluent::Plugin
       super
     end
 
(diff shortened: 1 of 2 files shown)


kenhys:
No need to backport? :thinking:

Watson1978 (approved):
👍🏻

PR #5286, opened 2026-03-18
output windows: check shorter service timeout on shutdown

kenhys:
Fixes #

When flush_at_shutdown was enabled, there is a case that buffer flush process takes longer time to be done. In such a case, if service timeout is too short, the fluentd process will be shutting down forcibly.

This commit warn in specific environment.

**Docs Changes**:

N/A

**Release Note**: 

N/A


kenhys:
improved warning message

PR #5307, opened 2026-03-31
maintainers: Update fukuda

daipom:


kenhys:
@daipom Wishing you all the best in your next chapter!

Watson1978:
@daipom Best of luck with your next adventure! 👍🏻

PR #5320, opened 2026-04-10
feat(in_exec): add command_timeout option to limit child process exec…

### LINKED ISSUE #5319: feat(in_exec): add command_timeout option to kill long-running child processes

**Problem**

`in_exec` plugin calls `child_process_execute` without passing `wait_timeout`,
so it defaults to `nil` (infinite wait).

When the external command hangs (e.g. `snmpget` against an unavailable `snmpd`),
the child process stays alive indefinitely. This causes repeated warnings:
[warn] previous child process is still running. skipped.

The warning does not indicate what is actually wrong (the external command hanging),
making troubleshooting difficult.

**Concrete example**

- `run_interval 1s`, command includes `snmpget`
- `snmpd` is down → `snmpget` waits up to 6 seconds (default `-t 1 -r 5`)
- Every 1s interval fires while the previous `snmpget` is still alive
- Fluentd only shows `child process still running`, not the actual cause

**Inconsistency with out_exec**

`out_exec` already exposes `command_timeout` which maps to `wait_timeout` in
`child_process_execute`. `in_exec` lacks this option, creating an inconsistent
interface between the two plugins.

**Proposed fix**

Add `command_timeout` config param to `in_exec`, mirroring `out_exec`:

```ruby
config_param :command_timeout, :time, default: nil

child_process_execute(:exec_input, @command,
  interval: @run_interval,
  wait_timeout: @command_timeout,
  **options,
  &method(:run)
)
```

configuring `run_interval` longer than the command's max duration also
resolves this. However, defensive handling at the plugin level is still valuable.

### Additional context

_No response_

zoklk:
Fixes #5319

Adds `command_timeout` option to `in_exec` plugin, mirroring the existing
`command_timeout` option in `out_exec`.

Currently, `in_exec` calls `child_process_execute` without `wait_timeout`,
defaulting to `nil` (infinite wait). When the external command hangs,
Fluentd repeatedly shows `previous child process is still running. skipped.`
with no indication of the actual cause, making troubleshooting difficult.

`out_exec` already exposes `command_timeout` which maps to `wait_timeout`
in `child_process_execute`. This PR resolves the inconsistency between
the two plugins.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/619

**Release Note**:
Add `command_timeout` option to `in_exec` plugin to kill long-running
child processes and align behavior with `out_exec`.

### DIFF
--- a/test/plugin/test_in_exec.rb
+++ b/test/plugin/test_in_exec.rb
@@ -320,10 +320,11 @@ EOC
         </parse>
       ]
       start_time = Time.now
-      d.run(timeout: 3)
+      d.run(timeout: 5) do
+        sleep 1 # avoid to return test immediately
+      end
       elapsed = Time.now - start_time
-
-      assert elapsed < 4, "command should have been killed by command_timeout"
+      assert (elapsed >= 1.0 and elapsed < 5), "command should have been killed by command_timeout"
     end
 
     test 'command_timeout defaults to nil' do
--- a/test/plugin/test_in_exec.rb
+++ b/test/plugin/test_in_exec.rb
@@ -295,4 +295,47 @@ EOC
       assert_match(/LoadError/, event[2]['message'])
     end
   end
+  sub_test_case 'command_timeout' do
+    test 'configure command_timeout' do
+      d = create_driver %[
+        command ruby -e "sleep 10"
+        tag test
+        run_interval 1s
+        command_timeout 1s
+        <parse>
+          @type none
+        </parse>
+      ]
+      assert_equal 1.0, d.instance.command_timeout
+    end
+
+    test 'command_timeout kills long-running child process' do
+      d = create_driver %[
+        command ruby -e "sleep 10"
+        tag test
+        run_interval 5s
+        command_timeout 1s
+        <parse>
+          @type none
+        </parse>
+      ]
+      start_time = Time.now
+      d.run(timeout: 3)
+      elapsed = Time.now - start_time
+
+      assert elapsed < 4, "command should have been killed by command_timeout"
+    end
+
+    test 'command_timeout defaults to nil' do
+      d = create_driver %[
+        command ruby -e "puts 'hello'"
+        tag test
+        run_interval 1s
+        <parse>
+          @type none
+        </parse>
+      ]
+      assert_nil d.instance.command_timeout
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


Watson1978:
Hi @zoklk, thanks for this great PR!
Could you also send a PR to the [fluentd-docs-gitbook](https://github.com/fluent/fluentd-docs-gitbook) repository to add the documentation for the new `command_timeout` parameter instead of following?

```
**Docs Changes**:
The `command_timeout` option should be added to the in_exec documentation
at https://docs.fluentd.org/input/exec
```

zoklk:
Hi @Watson1978 , I've sent a PR to the fluentd-docs-gitbook repository for the `command_timeout` documentation.

fluent/fluentd-docs-gitbook#619

Watson1978:
@zoklk Thank you so much!

Watson1978 (dismissed):
Looks good to me.

@kenhys Can you also review this PR?

test/plugin/test_in_exec.rb
```diff
@@ -295,4 +295,47 @@ def create_driver(conf)
...
+        command_timeout 1s
+        <parse>
+          @type none
+        </parse>
+      ]
+      start_time = Time.now
+      d.run(timeout: 3)
+      elapsed = Time.now - start_time
+
+      assert elapsed < 4, "command should have been killed by command_timeout"
```
kenhys:
```
      start_time = Time.now
      d.run(timeout: 5) do
        sleep 1 # avoid to return test immediately
      end
      elapsed = Time.now - start_time
      assert (elapsed >= 1.0 and elapsed < 5), "command should have been killed by command_timeout"
```

Without sleep block, test returns immediately (not to wait timeout), so it might be false test case.

kenhys (changes requested):

test/plugin/test_in_exec.rb
kenhys: nitpick: pull request with master branch is not recommended.

zoklk: Hi @kenhys , thank you for the feedback. I have updated the test case by adding a sleep block. I also noted your advice on the branch naming convention for my future contributions.

kenhys:
waiting CI...

kenhys (approved):

kenhys:
@zoklk Thanks!

PR #5357, opened 2026-05-12
output: add a cache for metadata to reduce object allocations

Watson1978:
Fixes #

This PR introduces a 1-item cache to `Output#metadata` to drastically reduce object allocations and improve overall throughput.

In high-volume log streams, logs often share the same `tag` and `timekey` sequentially.
Previously, `Output#metadata` instantiated a new `Fluent::Plugin::Buffer::Metadata` object for every single record, which caused massive memory allocations and garbage collection pressure.

Here is the result when run `rake benchmark:run:in_tail` with 5 GB data.

before:
```
  13.585014   4.170233  17.755247 ( 16.827939)
```

after:
```
  12.648212   4.101712  16.749924 ( 15.615520)
```

**Docs Changes**:
N/A

**Release Note**: 
* output: add a cache for metadata to reduce object allocations

### DIFF
--- a/lib/fluent/plugin/output.rb
+++ b/lib/fluent/plugin/output.rb
@@ -922,29 +922,34 @@ module Fluent
           return Struct.new(:timekey, :tag, :variables).new
         end
 
+        timekey = @chunk_key_time ? calculate_timekey(time) : nil
+        @_metadata_cache ||= MetadataCache.new
+
         # timekey is int from epoch, and `timekey - timekey % 60` is assumed to mach with 0s of each minutes.
         # it's wrong if timezone is configured as one which supports leap second, but it's very rare and
         # we can ignore it (especially in production systems).
         if @chunk_keys.empty?
-          if @chunk_key_time && @chunk_key_tag
-            timekey = calculate_timekey(time)
-            @buffer.metadata(timekey: timekey, tag: tag)
-          elsif @chunk_key_time
-            timekey = calculate_timekey(time)
-            @buffer.metadata(timekey: timekey)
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/output.rb
```diff
@@ -1616,5 +1621,27 @@ def statistics
...
+
+      def initialize
+        @timekey = nil
+        @tag = nil
+        @variables = nil
+        @metadata = nil
+      end
+
+      def cached?(timekey:, tag:, variables: nil)
+        @metadata && @timekey == timekey && @tag == tag && @variables == variables
```
kenhys, line 1636: I'm concerned about the impact on performance when a large number of chunk keys are specified. 🤔 
(usually it might be ignored such a worst case)

Watson1978: I believe the current approach is still a net positive for performance even in such worst-case scenarios

1. I think creating a new `Buffer::Metadata` object and the subsequent GC pressure is generally much more expensive than C-level `Hash#==` evaluation, even for relatively large hashes.
2. As you mentioned, specifying dozens or hundreds of `chunk_keys` is an extreme edge case (and probably an anti-pattern for chunking strategy).

For the vast majority of practical use cases (0 to a few chunk keys), this cache effectively eliminates millions of allocations in the hot path.

kenhys: Maybe I was worrying too much about rare cases.

kenhys (approved):
LGTM.

Watson1978:
Thanks

PR #5368, opened 2026-06-01
docs: Update SECURITY.md to clarify out-of-scope reports

ashie:
N/A

Add a dedicated "Out of Scope / Non-Vulnerabilities" section to reduce noise from automated scanners and invalid bounty reports.

This explicitly clarifies our policy on:
- `in_debug_agent` arbitrary code execution (by design)
- `record_transformer` with `enable_ruby` (by design)
- Legacy ReDoS issues mitigated by Ruby 3.3+ (welcome as standard bugs instead)

**Docs Changes**:

N/A

**Release Note**: 

Same with the title

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -23,3 +23,24 @@ https://github.com/fluent/fluentd/security/advisories
 
 > [!NOTE]
 > If you use a Docker image of Fluentd, please check [Fluentd Docker Image](https://github.com/fluent/fluentd-docker-image/blob/master/SECURITY.md) and report it there.
+
+
+## Out of Scope / Non-Vulnerabilities
+
+Before reporting a vulnerability, please check if it falls under the following expected behaviors, design choices, or configuration issues. Reports regarding these items will be closed immediately as **not a bug** or **out of scope**.
+
+### 1. `in_debug_agent` and arbitrary code execution (`eval`)
+
+* **Description:** The `in_debug_agent` plugin is a developer tool designed specifically for debugging a running Fluentd process via dRuby. It requires explicit activation in the local configuration file and is disabled by default.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -23,3 +23,24 @@ https://github.com/fluent/fluentd/security/advisories
...
+
+### 3. ReDoS (Regular Expression Denial of Service) reproducible only on Ruby 3.2 or older
+
+* **Description:** Potential algorithmic complexity issues in regular expressions that could lead to High CPU usage or Denial of Service on older Ruby runtimes.
+* **Why it is out of scope:** Fluentd officially supports modern Ruby environments. Our primary distribution channels including the official `fluent-package` and official Docker images ship with **Ruby 3.3 or later (e.g., Ruby 3.4)**, which includes built-in, robust mitigation against ReDoS at the language level (such as matching timeouts and engine-level optimizations). We do not accept ReDoS reports that only impact legacy Ruby versions (Ruby 3.2 or older) unless the issue can be demonstrated to cause a practical denial of service on Ruby 3.3+.
```
kenhys, line 45: > and official Docker images ship with **Ruby 3.3 or later (e.g., Ruby 3.4)**,

 **Ruby 3.4 or later** seems correct.

kenhys (dismissed):

Watson1978 (approved):
👍🏻

SECURITY.md
ashie: The current phrasing "Ruby 3.3 or later (e.g., Ruby 3.4)" is actually intentional for two reasons:

* Security Context: The crucial point here is that the language-level ReDoS mitigations were fundamentally introduced in Ruby 3.3. Anyone using Ruby 3.3 or newer is protected, which is the baseline requirement we want to highlight.
* Maintainability: Explicitly writing just "Ruby 3.4" means we would have to update this security policy every time a new major Ruby version is released in the future.

By using "3.3 or later (e.g., 3.4)", we can state the exact protected version while referencing our current release status as an example, without needing to frequently touch this file.

kenhys (approved):

PR #5383, opened 2026-06-09
Replace Yajl to JSON for handling JSON stream

Watson1978:
Fixes #

An experiment to replace Yajl with JSON by leveraging ruby/json#994.

### Behavioral changes
This PR preserves Yajl's behavior with one intentional exception:

* Truncated streams warn instead of raise
  *  When an input stream ends in the middle of a JSON document, Yajl raised a `ParseError`.
  * This PR emits a warning log and discards the incomplete tail instead. All documents completed before the cut are processed as before — only the incomplete trailing portion is affected
  * Rationale: for a log collector, keeping the successfully parsed records flowing while making the data loss visible to operators is preferable to failing the whole stream. The incomplete fragment itself is never logged, as it may contain sensitive record data.

### TODO
- [x] Ensure all CI tests pass
  * Currently monitoring the edge case reported in https://github.com/ruby/json/issues/1008
- [x] Check for memory leaks
  * Ensure that the memory usage of `JSON::ResumableParser` does not grow continuously during long-running stream parsing.
- [x] Test several edge cases for chunk boundaries
  * Chunk split in the middle of a Unicode escape sequence (e.g., `\uXXXX`).
  * Chunk split in the middle of primitive values (number, boolean, null):
    * e.g., `{"key": tr` and `ue}`
    * e.g., `{"key": 12` and `34.56}`
  * Parser recovery
    * Check whether the parser can successfully parse the *next* valid JSON chunk right after a `JSON::ParserError` is raised from a malformed chunk.

**Docs Changes**:

**Release Note**: 
* Replace Yajl to JSON for handling JSON stream


Watson1978:
### Performance Comparison: `yajl-ruby` vs `JSON::ResumableParser`

I conducted a performance test to evaluate the efficiency of `JSON::ResumableParser` compared to the current `yajl-ruby` implementation.

**Test Scenario**

* A client sends a continuous stream of ~1MB JSON messages via TCP at fixed 100ms intervals (approx. 10 MB/s).
* Monitored CPU and Memory (RSS) usage of the receiver during processing.

**Results (Handling ~10 MB/s constant load)**

* `yajl-ruby`
  * CPU Usage: ~70% (Average)
  * Memory (Max RSS): ~126 MB

* `JSON::ResumableParser`
  * CPU Usage: ~5% (Average)
  * Memory (Max RSS): ~127 MB

`JSON::ResumableParser` processes the same data stream vastly more efficiently than `yajl-ruby`. While memory consumption remains nearly identical, the new parser reduces CPU usage from ~70% to just ~5% under the same load. This massive reduction in CPU overhead indicates that `ResumableParser` provides significantly higher potential throughput and overall resource efficiency in stream processing environments.

Code_Generated_Image


### Fluentd conf
```apache
<source>
  @type forward
  bind "0.0.0.0"
  port 24224
</source>

<match **>
  @type null
</match>
```

### Client script
```ruby
require 'socket'
require 'json'
require 'time'
require 'securerandom'

host = '127.0.0.1'
port = 24224

puts "Connecting to #{host}:#{port}..."
sock = TCPSocket.new(host, port)

loop do
  msg = [
    "test.memory",
    Time.now.to_i,
    {
      "id" => SecureRandom.uuid,
      "message" => "This is a memory test for JSON::ResumableParser",
      "dummy_data" => ('a'..'z').to_a.sample * 1024 * 1024
    }
  ]

  json_str = JSON.generate(msg)
  sock.write(json_str)
  sleep 0.1
end

sock.close
```

lib/fluent/compat/exec_util.rb
```diff
@@ -77,10 +76,19 @@ def each_line(line)
...
       class JSONParser < Parser
+        BYTES_TO_READ = 8192
+
         def call(io)
-          y = Yajl::Parser.new
-          y.on_parse_complete = @on_message
-          y.parse(io)
+          parser = JSON::ResumableParser.new({})
+          begin
+            while (chunk = io.readpartial(BYTES_TO_READ))
```
kou: We can reuse chunk buffer:

```suggestion
            chunk = +"".b
            while io.readpartial(BYTES_TO_READ, chunk)
```

Watson1978:
Seems we need to add `allow_duplicate_key: true` for duplicated keys for compatibility.

```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'yajl-ruby', require: 'yajl'
  gem 'json', '>= 2.20'
end

require 'yajl'
require 'json'

chunks = [
  '{"key": "hello", "key": "world"}',
]

puts "** Yajl::Parser **"
y = Yajl::Parser.new
y.on_parse_complete = ->(record){
  p record
}
chunks.each do |chunk|
  y << chunk
end

puts "** JSON::ResumableParser **"
parser = JSON::ResumableParser.new({})
chunks.each do |chunk|
  parser << chunk
  while parser.parse
    p parser.value
  end
end
```

```
$ ruby -w json.rb
** Yajl::Parser **
{"key" => "world"}
** JSON::ResumableParser **
json.rb:27: warning: detected duplicate key "key" in JSON object. This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true` at line 1 column 1
{"key" => "world"}
```

lib/fluent/plugin/parser_json.rb
```diff
@@ -94,11 +91,17 @@ def parser_type
...
 
       def parse_io(io, &block)
-        y = Yajl::Parser.new
-        y.on_parse_complete = ->(record){
-          block.call(parse_time(record), record)
-        }
-        y.parse(io, @stream_buffer_size)
+        parser = JSON::ResumableParser.new({})
+        begin
+          while (chunk = io.readpartial(@stream_buffer_size))
```
kou: We can reuse chunk buffer here too.

Watson1978: Thanks.
I will fix it.

Watson1978: https://github.com/fluent/fluentd/pull/5383/changes/4301e21f09b8c0547c18da4edc150e2f1774203f

lib/fluent/compat/exec_util.rb
```diff
@@ -77,10 +77,29 @@ def each_line(line)
...
+                @on_message.call(parser.value)
+              end
+            end
+          rescue EOFError
+            # `rest` only reflects unconsumed tokenizer bytes, so a cut on a
+            # complete token boundary leaves it empty; `partial_value` covers
+            # those cases. Use `partial_value` only for its nil check: the
+            # partially built object it returns may hold sensitive record data
+            # and must never be logged.
+            if !parser.rest.empty? || !parser.partial_value.nil?
```
Watson1978: We can change this implementation with
```
            if !parser.empty?
```
, after https://github.com/ruby/json/pull/1048 was released

lib/fluent/plugin/parser_json.rb
```diff
@@ -99,11 +96,27 @@ def parser_type
...
+              block.call(parse_time(record), record)
+            end
+          end
+        rescue EOFError
+          # `rest` only reflects unconsumed tokenizer bytes, so a cut on a
+          # complete token boundary leaves it empty; `partial_value` covers
+          # those cases. Use `partial_value` only for its nil check: the
+          # partially built object it returns may hold sensitive record data
+          # and must never be logged.
+          if !parser.rest.empty? || !parser.partial_value.nil?
```
Watson1978: We can change this implementation with
```
            if !parser.empty?
```
, after https://github.com/ruby/json/pull/1048 was released

lib/fluent/compat/exec_util.rb
kenhys: Not describing TODO: in multiple file, how about patch? 

```
class JSON::ResumableParser
  unless method_defined?(:empty?)
    def empty?
      rest.empty? && partial_value.nil?
    end
  end
end
```


Except above parts, LGTM.

kenhys (changes requested):

lib/fluent/compat/exec_util.rb
kou: If we want to add a shim, we should use Refinement not monkey patch to avoid affecting outside of Fluentd.

```ruby
# This is for json < 2.2.1
unless JSON::ResumableParser.method_defined?(:empty?)
  module Fluentd
    module JSONResumableParserEmptyP
      refine JSON::ResumableParser do
        def empty?
          rest.empty? && partial_value.nil?
        end
      end
    end
  end
end
```

```ruby
using FluentD::JSONResumableParerEmptyP # in exec_util.rb and parser_json.rb
```

kenhys (approved):

Watson1978:
Thanks

PR #5388, opened 2026-06-18
engine: remove duplicated word in unreloadable plugin error message

s3onghyun:
Fixes #

The ConfigError raised on reload read "Unreloadable plugin plugin: ..." — removed the duplicated "plugin".

**Docs Changes**:
N/A

**Release Note**: 
* engine: remove duplicated word in unreloadable plugin error message


Watson1978 (dismissed):
Thanks

kenhys (approved):
Good catch!

PR #5390, opened 2026-06-22
system_config: add umask option to <system> directive

### LINKED ISSUE #4816: System config: add umask option

somaz94:
**Which issue(s) this PR fixes**: Fixes #4816

Adds a `umask` option to the `<system>` directive so the process umask can be
set from the config file. Previously this required the `--umask` command line
option, which is inconvenient for services and container images.

It reuses the existing chumask handling: `<system> umask` sets the same value
that `--umask` does, and the command line option takes precedence when both are
given. Both supervised and `--no-supervisor` modes honor it.

    <system>
      umask 0022
    </system>

Tests: added a config round-trip case and a supervisor test asserting the value
reaches ServerEngine, including command-line precedence.

**Docs Changes**:
https://github.com/fluent/fluentd-docs-gitbook/pull/625

**Release Note**:
Add `umask` option to the `<system>` directive.

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -808,6 +808,12 @@ module Fluent
       )
       @system_config = build_system_config(@conf)
 
+      # `<system> umask` sets the process umask, just like the `--umask` command
+      # line option. The command line option takes precedence when both are given.
+      if @system_config.umask && !@cl_opt.key?(:chumask)
+        @chumask = @system_config.umask
+      end
+
       $log.info :supervisor, 'parsing config file is succeeded', path: @config_path
 
       build_additional_configurations(parsed_files) do |additional_conf|
--- a/lib/fluent/system_config.rb
+++ b/lib/fluent/system_config.rb
@@ -30,7 +30,7 @@ module Fluent
       :file_permission, :dir_permission, :counter_server, :counter_client,
       :strict_config_value, :enable_msgpack_time_support, :disable_shared_socket,
       :metrics, :enable_input_metrics, :enable_size_metrics, :enable_jit, :source_only_buffer,
-      :config_include_dir
+      :config_include_dir, :umask
     ]
 
     config_param :workers,   :integer, default: 1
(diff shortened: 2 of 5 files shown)


kenhys:
It is ideal that not only checking serverengine layer, but also actual generated buffer permissions for example. could you afford to add such a test case?

somaz94:
Added an end-to-end test in `test/command/test_fluentd.rb` that boots fluentd with `<system> umask 0077` and a file buffer, then asserts the staged buffer chunk is created as `0600` (0644 & ~umask).

One thing worth noting: the buffer files only pick up the umask in `--no-supervisor` mode, where fluentd calls `File.umask` directly in the worker. In supervised mode the umask reaches only the supervisor process (ServerEngine applies `chumask` there, not the workers' `worker_chumask`), so worker-created buffer files stay `0644` — and this is true for the existing `--umask` option too, not something this PR changes. So the test runs with `--no-supervisor` to deterministically verify the actual file permission. Happy to open a separate issue for the supervised-mode worker umask gap if that's useful.

Watson1978 (approved):
Thanks for seeing this through, and for adding the buffer-permission test in response to the review.

The implementation is nicely minimal — reusing the existing `chumask` path (now that #4836 fixed it reaching ServerEngine) and keeping `umask` as a string so it flows through the same `to_i(8)` handling is the right call. The command-line precedence via `!@cl_opt.key?(:chumask)` is correct too, since `cmd_opts` only gains `:chumask` when `--umask` is actually passed.

For future readers, it's worth being explicit that `<system> umask` takes effect in exactly the same modes as the existing `--umask` option:

- `--no-supervisor` (standalone): applied — Fluentd calls `File.umask` directly in the worker.
- Supervised + `--daemon` (daemonize, e.g. fluent-package): applied — ServerEngine sets the umask before spawning workers, so worker-created files inherit it.
- Supervised without `--daemon` (foreground): not applied to worker-created files — ServerEngine's non-daemonize path never calls `File.umask`, and Fluentd doesn't set `worker_chumask`.

So the common packaged deployment (`--daemon`) does honor it; only the foreground-supervisor case is uncovered. That's existing `--umask` behavior rather than anything this PR changes, so it's not a blocker.

For the foreground-supervisor gap, let's track it in a separate follow-up issue rather than widening this PR — keeping this one scoped to "the config-file equivalent of `--umask`" is cleanest, and matches what you already suggested.

For the docs follow-up, it'd help to spell out which modes the option takes effect in, so users on a foreground supervisor don't expect it to change buffer file permissions.

Thanks.

Watson1978:
@kenhys I'm planning to approve this. The umask applicability matches the existing `--umask` (works with `--no-supervisor` and `--daemon`, not under a foreground supervisor — I verified all three) and I'll track the foreground-supervisor gap in a separate issue. Anything you'd like addressed before I do?

kenhys:
>  I verified all three) and I'll track the foreground-supervisor gap in a separate issue. Anything you'd like addressed before I do?


I'll second that decision.

kenhys (approved):

PR #5413, opened 2026-07-07
Config: accept a bare scalar for an array option in YAML syntax

### LINKED ISSUE #5149: YAML Config File Syntax:  Single int value for an array option causes a config error
### Describe the bug

In YAML syntax, a single int value for an array option causes a config error.

Found in the following fix.

* #5126

> Note: This PR doesn't address the issue where setting an Int directly to an Array option causes ConfigError: `retryable_response_codes: 503` (The `C` case in https://github.com/fluent/fluentd/pull/5126#issuecomment-3424253314)
It could be a different issue.

The current behavior is inconsistent and somewhat unnatural.

Unnatural current behavior:

* An error occurs only when a single int value is specified for an array option.
* It can be avoided by setting it as a String.
* There’s no such issue with options that are initially String arrays.
* There’s no such issue with the normal config syntax.

### To Reproduce

NG(Causes `config error file="..." error_class=Fluent::ConfigError error="array required but got 503"`):

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes: 503
```

OK as YAML syntax:

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes: "503"
```

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes: 503,
```

OK as normal syntax:

```xml
<match test>
  @type http
  endpoint "http://localhost:9880/tag"
  retryable_response_codes 503
</match>
```

OK as YAML syntax since v1.19.1 (after #5126):

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes: [503]
```

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes:
        - 503
```

Please also refer to the `C` case in https://github.com/fluent/fluentd/pull/5126#issuecomment-3424253314.

### Expected behavior

Fluentd can parse a single int value for an array option, such as the following:

```yaml
config:
  - match:
      $type: http
      $tag: test
      endpoint: http://localhost:9880/tag
      retryable_response_codes: 503
```

### Your Environment

```markdown
- Fluentd version: v1.19.1
- Package version: None
- Operating system: Ubuntu 22.04.5 LTS
- Kernel version: 6.8.0-85-generic
```

### Your Configuration

```apache
See `To reproduce`.
```

### Your Error Log

```shell
2025-11-11 14:27:25 +0900 [info]: init supervisor logger path=nil rotate_age=nil rotate_size=nil
2025-11-11 14:27:25 +0900 [info]: parsing config file is succeeded path="/test/fluentd/config/yaml/tmp.yaml"
2025-11-11 14:27:25 +0900 [info]: inaccessible include directory was specified path="/etc/fluent/conf.d"
2025-11-11 14:27:25 +0900 [info]: gem 'fluentd' version '1.19.1'
2025-11-11 14:27:25 +0900 [error]: config error in:
<match test>
  @type http
  endpoint "http://localhost:9880/tag"
  retryable_response_codes 503
</match>

2025-11-11 14:27:25 +0900 [error]: config error file="/test/fluentd/config/yaml/tmp.yaml" error_class=Fluent::ConfigError error="array required but got 503"
```

### Additional context

_No response_

arpitjain099:
Fixes #5149

In YAML config syntax, a single scalar value for an array option is rejected. For example `retryable_response_codes: 503` fails with `array required but got 503`, while the classic syntax `retryable_response_codes 503` works fine. The difference is that the classic path delivers the value as the String `"503"`, which `Config.array_value` already splits into `["503"]`, whereas YAML hands it over as a bare `Integer` that falls straight through to the "array required" check.

This wraps a non-String, non-Array, non-Hash scalar into a one-element array so both config syntaxes behave the same way. The wrapped value then flows through the existing `value_type` coercion, so `retryable_response_codes: 503` becomes `[503]` just like `retryable_response_codes: [503]` does. A Hash still raises as before, so an accidental mapping under an array option is not silently accepted.

I could not run the full suite locally (my Ruby is 2.6 and Fluentd needs 3.2), so I verified the logic by tracing the exact `array_value` path and reproducing it in isolation; CI should confirm. I added test cases in test/config/test_types.rb covering a bare integer with and without `value_type`, and that an existing array is left untouched.

**Docs Changes**:
None.

**Release Note**:
Fix a config error when a single scalar value is given to an array option in YAML config syntax (for example `retryable_response_codes: 503`).

### DIFF
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -229,8 +229,10 @@ module Fluent
 
       param = if val.is_a?(String)
                 val.start_with?('[') ? JSON.parse(val, Fluent::DEFAULT_JSON_PARSE_OPTIONS) : val.strip.split(/\s*,\s*/)
-              else
+              elsif val.is_a?(Array) || val.is_a?(Hash)
                 val
+              else
+                [val]
               end
       if param.class != Array
         raise ConfigError, "array required but got #{val.inspect}"
--- a/lib/fluent/config/types.rb
+++ b/lib/fluent/config/types.rb
@@ -229,10 +229,16 @@ module Fluent
 
       param = if val.is_a?(String)
                 val.start_with?('[') ? JSON.parse(val, Fluent::DEFAULT_JSON_PARSE_OPTIONS) : val.strip.split(/\s*,\s*/)
-              elsif val.is_a?(Array) || val.is_a?(Hash)
+              elsif val.is_a?(Array)
                 val
-              else
+              elsif val.is_a?(Numeric) || val == true || val == false
+                # Wrap only the bare scalars that Psych/YAML actually produces
+                # here (nil is handled by the early return above). Any other
+                # type (Hash, Symbol, Time, arbitrary objects) falls through and
+                # still raises "array required" below.
                 [val]
+              else
+                val
               end
       if param.class != Array
         raise ConfigError, "array required but got #{val.inspect}"
--- a/test/config/test_types.rb
+++ b/test/config/test_types.rb
@@ -378,14 +378,19 @@ class TestConfigTypes < ::Test::Unit::TestCase
          "value_type: :integer w/o quote" => [[1,2,1], '1,2,1', {value_type: :integer}],
          "bare integer (YAML scalar)" => [[503], 503, {}],
          "bare integer w/ value_type" => [[503], 503, {value_type: :integer}],
+         "bare float (YAML scalar)" => [[1.5], 1.5, {}],
+         "bare true (YAML scalar)" => [[true], true, {}],
+         "bare false (YAML scalar)" => [[false], false, {}],
          "already an array is untouched" => [[503], [503], {}])
     test 'array' do |(expected, val, opts)|
       assert_equal(expected, Config::ARRAY_TYPE.call(val, opts))
     end
 
     data("hash" => [{"key" => "value"}, {}],
-         "hash w/ value_type" => [{"key" => "1"}, {value_type: :integer}])
-    test 'array w/ hash still raises' do |(val, opts)|
+         "hash w/ value_type" => [{"key" => "1"}, {value_type: :integer}],
+         "symbol" => [:foo, {}],
+         "time" => [Time.at(0), {}])
+    test 'array w/ unexpected type still raises' do |(val, opts)|
       assert_raise(Fluent::ConfigError.new("array required but got #{val.inspect}")) do
         Config::ARRAY_TYPE.call(val, opts)
       end
--- a/test/config/test_types.rb
+++ b/test/config/test_types.rb
@@ -383,6 +383,14 @@ class TestConfigTypes < ::Test::Unit::TestCase
       assert_equal(expected, Config::ARRAY_TYPE.call(val, opts))
     end
 
+    data("hash" => [{"key" => "value"}, {}],
+         "hash w/ value_type" => [{"key" => "1"}, {value_type: :integer}])
+    test 'array w/ hash still raises' do |(val, opts)|
+      assert_raise(Fluent::ConfigError.new("array required but got #{val.inspect}")) do
+        Config::ARRAY_TYPE.call(val, opts)
+      end
+    end
+
     data('["1","2"]' => [["1","2"], '["1","2"]'],
          '["3"]' => [["3"], '["3"]'])
     test 'array w/ default values' do |(expected, val)|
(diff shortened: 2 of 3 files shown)


test/config/test_types.rb
```diff
@@ -375,7 +375,10 @@ class TestConfigTypes < ::Test::Unit::TestCase
          "space in a value w/o qupte" => [["a a","b","c"], 'a a,b,c', {}],
          "integers" => [[1,2,1], '[1,2,1]', {}],
          "value_type: :integer w/ quote" => [[1,2,1], '["1","2","1"]', {value_type: :integer}],
-         "value_type: :integer w/o quote" => [[1,2,1], '1,2,1', {value_type: :integer}])
+         "value_type: :integer w/o quote" => [[1,2,1], '1,2,1', {value_type: :integer}],
+         "bare integer (YAML scalar)" => [[503], 503, {}],
+         "bare integer w/ value_type" => [[503], 503, {value_type: :integer}],
+         "already an array is untouched" => [[503], [503], {}])
```
kenhys, line 384: nitpick: conversion rule was changed, so it might be better to add another test case such asa `assert_raise(Fluent::ConfiguError.new("array required but got #{val.inspect}")` for known types.

Could you afford to write such a test case?

kenhys (changes requested):

arpitjain099:
Thanks for the review, good call. I've added a test case asserting that a Hash input still raises `Fluent::ConfigError` with the `array required but got #{val.inspect}` message, so the scalar-wrapping change doesn't accidentally accept a mapping under an array option. It covers a hash both with and without `value_type`. Verified green on Ruby 3.4.

test/config/test_types.rb
arpitjain099: [Responded](https://github.com/fluent/fluentd/pull/5413#issuecomment-4899808324)

lib/fluent/config/types.rb
```diff
@@ -229,8 +229,10 @@ def self.array_value(val, opts = {}, name = nil)
 
       param = if val.is_a?(String)
                 val.start_with?('[') ? JSON.parse(val, Fluent::DEFAULT_JSON_PARSE_OPTIONS) : val.strip.split(/\s*,\s*/)
-              else
+              elsif val.is_a?(Array) || val.is_a?(Hash)
                 val
+              else
+                [val]
```
Watson1978: One concern about the new `else` branch.
it wraps any object that isn't a String/Array/Hash, including Symbols, Time objects, or arbitrary objects that would previously have been rejected with `ConfigError`.
That defensive behavior is worth keeping for unexpected types.
Since the scalars Psych actually produces here are Integer/Float/true/false (nil is handled by the early return), how about limiting the wrapping to those types?

```ruby
elsif val.is_a?(Numeric) || val == true || val == false
  [val]
else
  val  # anything else still raises "array required" below
end
```

This fixes the reported issue while preserving the existing error for genuinely invalid values.

arpitjain099:
Good catch, thanks. You're right that the broad `else` swallowed types that should stay rejected. I've narrowed the wrapping to just the scalars Psych actually produces here (`Numeric`, `true`, `false`; `nil` still returns early), so anything unexpected like a Symbol or a Time now falls through and raises `array required` as before. Added test cases for a bare float/true/false wrapping, plus Symbol and Time still raising alongside the existing Hash case. Green on Ruby 3.4.

Watson1978:
The updated version looks good to me.

One non-blocking suggestion, it would be nice to also have an end-to-end test in `test/config/test_yaml_parser.rb` that exercises the actual YAML pipeline (e.g., a config with retryable_response_codes: 503 reaching a plugin's array parameter). The unit tests here verify ARRAY_TYPE directly, but the premise of this fix — that the YAML parser hands a bare Integer to the type-conversion layer — is only implicitly covered. If the YAML parser ever starts stringifying scalar values, an E2E test would catch the regression.

Watson1978 (dismissed):

arpitjain099:
Thanks @Watson1978, good idea. Added two tests to test/config/test_yaml_parser.rb that go through the real YAML pipeline: a config with retryable_response_codes: 503 on a match section, parsed with YamlParser and fed to a Configurable with an :array param (value_type: :integer). The bare scalar comes out as [503], and a [502, 503] sequence still works as before. Ran the file locally on Ruby 3.3, 33 tests green.

Watson1978 (approved):

kenhys (approved):

PR #5420, opened 2026-07-10
time: fix TimeFormatter returning nil for the Unix epoch

Watson1978:
Fixes #

`TimeFormatter` caches the last two formatted timestamps in two slots, and an empty slot was marked by its key being `0`. But `0` is a valid timestamp, so the epoch matched an empty slot and its `nil` string was returned as a cache hit:

```ruby
Fluent::TimeFormatter.new(nil, false, nil).format(0)  #=> nil
```

`format_with_subsec` compares slots with `EventTime.eq?`, which falls back to `EventTime#==` and compares seconds alone, so there the whole first second of the epoch is affected:

```ruby
fmt = Fluent::TimeFormatter.new("%Y%m%d %H%M%S.%N", false, nil)
fmt.format(Fluent::EventTime.new(0, 123456789))  #=> nil
```

The fix is to mark an empty slot by its cached string being `nil` rather than by its key.
`format_nocache` always returns a `String`, so a filled slot can never be mistaken for an empty one. Every other timestamp behaves exactly as before, since key `0` could never have matched it.

Impact is small, and I don't think this is urgent.
A slot keeps the key `0` only until it is written, so a formatter self-heals once it has formatted two distinct timestamps, within a second of the first record.
The exception is a stream whose timestamps are *all* the epoch: the early return never fills a slot, and `out_file` then writes every line with an empty time column.
That is reachable via `time_type unixtime`, which parses with `value.to_i` and turns an empty or non-numeric `time_key` into `EventTime.new(0)`, but such a pipeline is misconfigured anyway. The reason to fix it is simply that `nil` is never a correct result.

**Docs Changes**:

**Release Note**: 
* time: fix TimeFormatter returning nil for the Unix epoch

### DIFF
--- a/test/test_time_formatter.rb
+++ b/test/test_time_formatter.rb
@@ -196,6 +196,29 @@ class TimeFormatterTest < ::Test::Unit::TestCase
     assert_equal("20140927 0000.000000000", formatter.format(time))
   end
 
+  # The Unix epoch used to match an empty cache slot, whose string is nil.
+  def test_format_epoch
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
+
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(0))
+    assert_equal("1970-01-01T00:00:01Z", formatter.format(Fluent::EventTime.new(1)))
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
+  end
+
+  # Here the slot is compared with EventTime.eq?, which falls back to comparing
+  # seconds alone, so the whole first second of the epoch matched an empty slot.
+  def test_format_epoch_with_subsec
+    formatter = Fluent::TimeFormatter.new("%Y%m%d %H%M%S.%N", false, nil)
+    assert_equal("19700101 000000.000000000", formatter.format(Fluent::EventTime.new(0, 0)))
+    assert_equal("19700101 000000.123456789", formatter.format(Fluent::EventTime.new(0, 123456789)))
+    assert_equal("19700101 000000.000000000", formatter.format(Fluent::EventTime.new(0, 0)))
+
+    formatter = Fluent::TimeFormatter.new("%Y%m%d %H%M%S.%N", false, nil)
+    assert_equal("19700101 000000.999999999", formatter.format(Fluent::EventTime.new(0, 999999999)))
+  end
+
   sub_test_case 'TimeMixin::Formatter' do
     class DummyForTimeFormatter
       include Fluent::Configurable
--- a/test/test_time_formatter.rb
+++ b/test/test_time_formatter.rb
@@ -201,16 +201,38 @@ class TimeFormatterTest < ::Test::Unit::TestCase
     formatter = Fluent::TimeFormatter.new(nil, false, nil)
     assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
 
+    # Only the tc2 slot is used here, so the second 0 is a miss.
     formatter = Fluent::TimeFormatter.new(nil, false, nil)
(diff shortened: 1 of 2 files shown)


test/test_time_formatter.rb
```diff
@@ -196,6 +196,29 @@ def test_format_with_subsec
     assert_equal("20140927 0000.000000000", formatter.format(time))
   end
 
+  # The Unix epoch used to match an empty cache slot, whose string is nil.
+  def test_format_epoch
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
+
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
```
kenhys, line 205: It might be helpful to explain only one slot is used.

```
    # only tc2 cache is chosen the following test-case
    mock.proxy(formatter).format_nocache(satisfy { |t| t.to_i == 0}).twice
    mock.proxy(formatter).format_nocache(satisfy { |t| t.to_i == 1}).once
```

test/test_time_formatter.rb
```diff
@@ -196,6 +196,29 @@ def test_format_with_subsec
...
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(0))
+    assert_equal("1970-01-01T00:00:01Z", formatter.format(Fluent::EventTime.new(1)))
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
+  end
+
+  # Here the slot is compared with EventTime.eq?, which falls back to comparing
+  # seconds alone, so the whole first second of the epoch matched an empty slot.
+  def test_format_epoch_with_subsec
+    formatter = Fluent::TimeFormatter.new("%Y%m%d %H%M%S.%N", false, nil)
```
kenhys, line 234: ditto.

```
    # only tc2 cache is chosen the following test-case
    mock.proxy(formatter).format_nocache(satisfy { |t| t.to_i == 0 }).times(3)
```

test/test_time_formatter.rb
```diff
@@ -196,6 +196,29 @@ def test_format_with_subsec
     assert_equal("20140927 0000.000000000", formatter.format(time))
   end
 
+  # The Unix epoch used to match an empty cache slot, whose string is nil.
+  def test_format_epoch
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(Fluent::EventTime.new(0)))
+
+    formatter = Fluent::TimeFormatter.new(nil, false, nil)
+    assert_equal("1970-01-01T00:00:00Z", formatter.format(0))
```
kenhys, line 208: And no negative test case.

kenhys: Need one more test case tc1 and tc2 will be used.

kenhys (changes requested):

test/test_time_formatter.rb
Watson1978: Done — added a short comment (`# Only the tc2 slot is used here, so the second 0 is a miss.`) plus your suggested `mock.proxy` assertions: `format_nocache` runs twice for `0` and once for `1`.

test/test_time_formatter.rb
Watson1978: Done — same treatment: a short comment plus `mock.proxy` asserting `format_nocache` runs three times (`.times(3)`), since the three sub-second timestamps all share second `0` and reuse only the tc2 slot.

test/test_time_formatter.rb
Watson1978: Added `test_format_negative_timestamp` — a pre-epoch timestamp (`-1` → `1969-12-31T23:59:59Z`), the boundary neighbour of `0`. Let me know if you meant a different kind of negative case.

Watson1978: Added `test_format_uses_both_cache_slots` — `10` fills tc2, `20` fills tc1, and re-formatting each is a cache hit, so `format_nocache` runs once per distinct second; both slots are exercised.

kenhys (approved):

PR #5427, opened 2026-07-12
test: rework unused_port_tcp_udp to try random candidate ports

Watson1978:
Fixes #

`unused_port_tcp_udp()` collected 200 TCP ephemeral ports at once and returned the first one that could also be bound for UDP.
On Windows, Hyper-V/WinNAT/HNS reserve "excluded port ranges" inside the ephemeral port range (49152-65535), the reservations are managed separately for TCP and UDP, and binding a port inside such a range fails with `EACCES` instead of `EADDRINUSE`.
The TCP ephemeral allocator assigns ports sequentially and knows nothing about UDP exclusions, so the 200 candidates form a contiguous block that can land entirely inside a UDP-only excluded range. When that happens, every candidate fails the UDP bind check and test setup fails with:

```
  RuntimeError: can't find unused port
  test/helper.rb:94:in `unused_port_tcp_udp'
  test/helper.rb:81:in `unused_port'
  test/plugin/test_in_forward.rb:22:in `setup'
```

This was captured on a windows-2025 GitHub Actions runner whose UDP exclusion ranges (netsh interface ipv4 show excludedportrange protocol=udp) were 54315-54414 and 54415-54514: exactly 200 consecutive ports, while the TCP exclusions were elsewhere (49691-49890).

Instead, draw random candidate ports from a fixed high range and verify that each can be bound for both TCP and UDP, trying another candidate on failure. Because the candidates are independent draws, reserved or in-use ports only matter through their overall fraction of the range: contiguous reservation blocks cannot defeat the search, and the chance of exhausting the retry limit is negligible. The range is kept wide (10001 ports) so that reservation blocks, which come in units of about 100-200 ports, can only ever cover a small fraction of it.

This also removes the need to hold 200 sockets at once, which the old implementation capped because of file descriptor limits, and the checks now bind both protocols explicitly on the same IPv4 wildcard address.

**Docs Changes**:
N/A

**Release Note**: 
N/A

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -84,10 +84,6 @@ def unused_port(num = 1, protocol:, bind: "0.0.0.0")
   end
 end
 
-# Excluded port ranges reserved by Hyper-V/WinNAT/HNS on Windows come in
-# blocks of about 100-200 ports and can accumulate, so the range must be
-# wide enough that reservations can only ever cover a small fraction of it.
-PORT_RANGE_TCP_UDP = (55000..65000)
 # Windows components such as Hyper-V, WinNAT, and HNS may reserve dynamic
 # excluded port ranges. These exclusions are commonly observed in blocks of
 # roughly 100 ports and multiple ranges may coexist, so use a sufficiently
@@ -96,6 +92,8 @@ PORT_RANGE_TCP_UDP = (55000..65000)
 # About dynamic excluded port ranges, see:
 # > netsh interface ipv4 show excludedportrange protocol=tcp
 # > netsh interface ipv4 show excludedportrange protocol=ucp
+PORT_RANGE_TCP_UDP = (55000..65000)
+
 def unused_port_tcp_udp(num = 1, retries: 1000)
   raise "not support num > 1" if num > 1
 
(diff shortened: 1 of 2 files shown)


test/helper.rb
```diff
@@ -84,16 +84,38 @@ def unused_port(num = 1, protocol:, bind: "0.0.0.0")
   end
 end
 
-def unused_port_tcp_udp(num = 1)
+# Excluded port ranges reserved by Hyper-V/WinNAT/HNS on Windows come in
+# blocks of about 100-200 ports and can accumulate, so the range must be
+# wide enough that reservations can only ever cover a small fraction of it.
+PORT_RANGE_TCP_UDP = (55000..65000)
+
```
kenhys:
```suggestion
# Windows components such as Hyper-V, WinNAT, and HNS may reserve dynamic
# excluded port ranges. These exclusions are commonly observed in blocks of
# roughly 100 ports and multiple ranges may coexist, so use a sufficiently
# wide candidate range to reduce the probability that all candidates are
# excluded.
# About dynamic excluded port ranges, see:
# > netsh interface ipv4 show excludedportrange protocol=tcp
# > netsh interface ipv4 show excludedportrange protocol=ucp
```

kenhys: How about explaining like this?

kenhys (changes requested):

kenhys (approved):

PR #5436, opened 2026-07-16
out_forward: drop keepalive sockets with failed or mismatched acks

### LINKED ISSUE #4057: Fluentd pods getting flooded with warning messages in when used in "Forwarder-Aggregator" mode.
### Describe the bug

I am using Forwarder-Aggregator architecture of Fluentd to send logs to Splunk. I am using Fluentd image in both Forwarder and Aggregator. 
Forwarder --> Forwarder deployed as Daemonset which collects logs from node and send it to Aggregator. 
Aggregator --> Aggregator deployed as Deployment and collects logs from Forwarder to send it to Splunk. 

When Aggregator shuts down because of some reasons, I start getting below error message in Forwarder and then it never stops unless and until we don't kill forwarder pod.  

Error Message : 2023-02-15 13:57:29 +0000 [warn]: #0 ack in response and chunk id in sent data are different chunk_id="5f4bd71dab0718e2b31ad4c51fbaf556" ack="X0vXBdaKYzX7jY32oYdgHw==\n"

Fluentd Version :  fluent/fluentd:v1.15.1-debian-1.1

### To Reproduce

Run Fluentd in Forwarder-Aggregator mode and then shut down Aggregator

### Expected behavior

Forwarder should not continuously generate mentioned warning messages. 

### Your Environment

```markdown
- Fluentd version:  fluent/fluentd:v1.15.1-debian-1.1
- TD Agent version: 
- Operating system:
- Kernel version:
```

### Your Configuration

```apache
Worker-0 of Forwarder:

<match **>
        @type forward
        send_timeout 60s
        recover_wait 10s
        hard_timeout 60s
        require_ack_response true
        heartbeat_type tcp
        keepalive true
        keepalive_timeout 30s
        <service_discovery>
          @type srv
          service tcp
          proto tcp
          interval 20
          weight 60
          hostname fluentd-aggregator
        </service_discovery>
        <service_discovery>
          @type srv
          service tcp
          proto tcp
          interval 20
          hostname fluentd-aggregator
          weight 60
        </service_discovery>
        <buffer>
          @type file
          total_limit_size 5GB
          path /var/fluentd_out/buffer_1
          queue_limit_length 100000
          chunk_limit_size 30000000
          chunk_full_threshold 0.008
          flush_interval 10s
          flush_thread_count 8
          retry_forever false
          retry_timeout 1h
        </buffer>
        <secondary>
          @type file
          path /var/fluentd_out/out_plugin_secondary_1
        </secondary>
      </match>
```

### Your Error Log

```shell
2023-02-15 13:57:29 +0000 [warn]: #0 ack in response and chunk id in sent data are different chunk_id="5f4bd754fe2699106533a872fa077f94" ack="X0vXJzT0g8jJuvuHR243iw==\n"
```

### Additional context

_No response_

JSap0914:
Fixes #4057

In Forwarder-Aggregator mode with `keepalive true` and `flush_thread_count` > 1, when the aggregator restarts/reconnects the forwarder starts logging

```
[warn]: ack in response and chunk id in sent data are different chunk_id="..." ack="...\n"
```

endlessly, and never recovers on its own until the pod is killed.

Root cause: when a chunk's ack times out (`FAILED`) or comes back with a chunk id that does not match the one tracked for its socket (`CHUNKID_UNMATCHED`), `ack_check` returned the keepalive socket to the reuse pool via `ConnectionManager#close` (→ `SocketCache#checkin`). Such a connection may still carry an in-flight/stale ack for a previous chunk. When the socket is checked out again for a new chunk, that stale ack is read first and compared against the new chunk's id, so it never matches. From then on every ack on that connection is offset by one chunk, producing an unbounded stream of mismatch warnings.

This PR adds `ConnectionManager#discard`, which **revokes** a keepalive socket (or closes a non-keepalive one) instead of checking it back in, and uses it from `ack_check` on the `FAILED` and `CHUNKID_UNMATCHED` paths. The tainted connection is torn down, so the next chunk uses a fresh connection with a clean ack stream and the warning storm stops. Successful acks keep reusing the socket exactly as before.

**Docs Changes**:
None.

**Release Note**:
out_forward: stop the endless "ack in response and chunk id in sent data are different" warning storm by discarding (instead of reusing) a keepalive socket whose ack failed or came back with a mismatched chunk id.

---

Tests (`ruby 3.4.10`):

- `test/plugin/out_forward/test_connection_manager.rb` — new `#discard` cases: closes a non-keepalive socket, revokes a keepalive socket instead of checking it in, and confirms a discarded keepalive socket is not reused (a checked-in one is).
- `test/plugin/test_out_forward.rb` — new `ack_check` cases: the socket is discarded on `CHUNKID_UNMATCHED` and `FAILED`, and reused on `SUCCESS`.

```
$ bundle exec ruby -Itest test/plugin/out_forward/test_connection_manager.rb
10 tests, 16 assertions, 0 failures, 0 errors

$ bundle exec ruby -Itest test/plugin/test_out_forward.rb
74 tests, 147 assertions, 0 failures, 0 errors, 3 omissions (Windows-only)
```

The new tests fail on `master` (the mismatched/failed ack socket is reused) and pass with this change.

### DIFF
--- a/lib/fluent/plugin/out_forward/connection_manager.rb
+++ b/lib/fluent/plugin/out_forward/connection_manager.rb
@@ -78,6 +78,24 @@ module Fluent::Plugin
         end
       end
 
+      # Drop a socket instead of returning it to the keepalive reuse pool.
+      #
+      # Used when a chunk's ack timed out or came back with a mismatched chunk
+      # id: such a connection may still carry an in-flight/stale ack for a
+      # previous chunk. Reusing it would let that stale ack be read against a
+      # chunk sent on the reused socket, permanently offsetting every following
+      # ack and flooding the log with 'ack in response and chunk id in sent data
+      # are different' warnings (see #4057). Revoking removes it from the pool so
+      # the next chunk uses a fresh connection with a clean ack stream.
+      def discard(sock)
+        if @socket_cache
+          @socket_cache.revoke(sock)
+        else
+          sock.close_write rescue nil
+          sock.close rescue nil
+        end
+      end
+
       private
 
       def connect_keepalive(host:, port:, hostname:, ack: nil)
--- a/lib/fluent/plugin/out_forward.rb
+++ b/lib/fluent/plugin/out_forward.rb
@@ -533,17 +533,28 @@ module Fluent::Plugin
 
     def ack_check(select_interval)
       @ack_handler.collect_response(select_interval) do |chunk_id, node, sock, result|
-        @connection_manager.close(sock)
-
         case result
         when AckHandler::Result::SUCCESS
+          @connection_manager.close(sock)
           commit_write(chunk_id)
         when AckHandler::Result::FAILED
+          # The ack timed out or the peer closed the connection. A keepalive
+          # socket may still have an in-flight ack for this chunk, so drop it
+          # instead of returning it to the reuse pool; otherwise a late ack
+          # could later be read against a chunk sent on the reused socket (#4057).
+          @connection_manager.discard(sock)
           node&.disable!
           rollback_write(chunk_id, update_retry: false) if chunk_id
         when AckHandler::Result::CHUNKID_UNMATCHED
+          # The ack read from this socket belongs to a different chunk than the
+          # one tracked for it: the keepalive connection's ack stream is out of
+          # sync (e.g. a stale ack left over from before an aggregator restart).
+          # Discard the socket so the following acks are not each offset by one,
+          # which would otherwise flood the log with mismatch warnings (#4057).
+          @connection_manager.discard(sock)
           rollback_write(chunk_id, update_retry: false)
         else
+          @connection_manager.close(sock)
           log.warn("BUG: invalid status #{result} #{chunk_id}")
 
           if chunk_id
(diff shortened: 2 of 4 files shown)


lib/fluent/plugin/out_forward/connection_manager.rb
```diff
@@ -78,6 +78,24 @@ def close(sock)
...
+      # chunk sent on the reused socket, permanently offsetting every following
+      # ack and flooding the log with 'ack in response and chunk id in sent data
+      # are different' warnings (see #4057). Revoking removes it from the pool so
+      # the next chunk uses a fresh connection with a clean ack stream.
+      def discard(sock)
+        if @socket_cache
+          @socket_cache.revoke(sock)
+        else
+          sock.close_write rescue nil
+          sock.close rescue nil
```
Watson1978, line 95: [nits] These line now appears in

https://github.com/fluent/fluentd/blob/e83344221ba8eb15599ccf09e702bb898fb0d5de/lib/fluent/plugin/out_forward/connection_manager.rb#L76-L77

When you add a tiny private method (like a `close_socket`), it  would remove duplicates.

Watson1978 (approved):

kenhys (approved):
LGTM.

PR #5441, opened 2026-07-17
fix(in_tail): fix variable name bugs causing position file leak and NameError

magic-peach:
## Summary

Two variable name bugs in the tail input plugin (`lib/fluent/plugin/in_tail.rb`):

### 1. Wrong instance variable: `@follow_inode` vs `@follow_inodes` (line 650)

The condition references `@follow_inode` (singular), but the instance variable is defined as `@follow_inodes` (plural) on line 892. In Ruby, referencing an undefined instance variable silently returns `nil` rather than raising an error.

When `@follow_inodes` is enabled, the position file entry for unwatched tail watchers should always be cleaned up. But because `@follow_inode` evaluates to `nil`, the condition degenerates to `(!@tails[tw.path])`, so the entry is only unwatched if the path is not currently in `@tails`. During rotation, when the path IS still present, the position file entry leaks and is never cleaned up, causing unbounded growth with stale entries.

### 2. Undefined local variable `tail_watcher` (line 709)

In `flush_buffer`, the parameter is named `tw`, but line 709 references `tail_watcher` — an undefined local variable. Every other usage in this method (lines 700, 704, 711, 717) correctly uses `tw`.

This causes a `NameError: undefined local variable or method 'tail_watcher'` at runtime when `emit_unmatched_lines` is enabled with `path_key` in multiline mode, crashing the tail watcher's IOHandler and potentially losing log data.

### DIFF
--- a/lib/fluent/plugin/in_tail.rb
+++ b/lib/fluent/plugin/in_tail.rb
@@ -647,7 +647,7 @@ module Fluent::Plugin
 
       tw.close if close_io
 
-      if @pf && tw.unwatched && (@follow_inode || !@tails[tw.path])
+      if @pf && tw.unwatched && (@follow_inodes || !@tails[tw.path])
         target_info = TargetInfo.new(tw.path, ino)
         @pf.unwatch(target_info)
       end
@@ -751,7 +751,7 @@ module Fluent::Plugin
           else
             if @emit_unmatched_lines
               record = {'unmatched_line' => line}
-              record[@path_key] ||= tail_watcher.path unless @path_key.nil?
+            record[@path_key] ||= tw.path unless @path_key.nil?
               es.add(Fluent::EventTime.now, record)
             end
             log.warn { "pattern not matched: #{line.inspect}" }


lib/fluent/plugin/in_tail.rb
```diff
@@ -751,7 +751,7 @@ def convert_line_to_event(line, es, tail_watcher)
           else
             if @emit_unmatched_lines
               record = {'unmatched_line' => line}
-              record[@path_key] ||= tail_watcher.path unless @path_key.nil?
+            record[@path_key] ||= tw.path unless @path_key.nil?
```
Watson1978, line 754: This change looks inverted. In `convert_line_to_event(line, es, tail_watcher)` the parameter is named `tail_watcher`, and `tw` is not in scope here (the block parameters are `|time, record|`). So `tw.path` raises `NameError: undefined local variable or method 'tw'` whenever this branch runs (`emit_unmatched_lines` enabled with `path_key`).

Watson1978 (changes requested):
Thanks for the PR! Two quick things.

First, could you update the description to follow our pull request template? It helps with triage and release notes. Your current summary works well under "What this PR does / why we need it"; it's mainly about adding the remaining headings:

```
Fixes #

**Docs Changes**: 

**Release Note**: 
```

Second,
the line 754 change, however, is inverted — the enclosing method is `convert_line_to_event(line, es, tail_watcher)`, so `tw` is not in scope and `tw.path` raises `NameError`. `master` was already correct there, so that line should stay `tail_watcher.path`. Could you drop the line 754 hunk and keep only the 650 fix?

PR #5449, opened 2026-07-19
parser_syslog: accept a space after RFC3164 priority

### LINKED ISSUE #4158: in_syslog: consider space between PRI and HEADER valid for RFC3164

* #4157

As is the issue here, `in_syslog` considers the following record invalid as RFC3164 because there is a space between PRI and HEADER.

```
<14> Apr 25 16:43:29 PAA-SW1-1 General[procLOG]: main.c(257) 272264 %% Stopping System API  application
```

As I commented in https://github.com/fluent/fluentd/discussions/4157#discussioncomment-5748811, RFC3164 does not seem to say this format is invalid.

> https://www.rfc-editor.org/rfc/rfc3164#section-4.1.2 says:
>
>> The TIMESTAMP will immediately follow the trailing ">" from the PRI part
>
> "will" is used, not "must". (If there is no other possibility, "must" should be used.)

Consider space between PRI and HEADER valid for RFC3164.

We already have workarounds for this.

We can use `time_format` to solve the space between `PRI` and `HEADER`.

```xml
<source>
  @type syslog
  tag foo
  <parse>
    @type syslog
    time_format " %b %d %H:%M:%S"
  </parse>
</source>
```

### Additional context

At first, I thought we needed to use `parser_regexp` as a workaround, but it was just a matter of setting up `time_format`.
So there may be little need to support this.

AkashKumar7902:
Fixes #4158

Some syslog senders emit RFC3164 records with an ASCII space between the priority and header. Depending on whether `in_syslog` or the parser owns priority extraction, the RFC3164 parser sees that space either immediately after `>` or as the first character of its input, and currently attempts to parse it as the start of the timestamp.

This change accepts the RFC3164 boundary space in `rfc3164` and the RFC3164 path of `auto`, across both parser engines and both priority-ownership models. Canonical records without the space continue to parse unchanged, malformed priorities remain invalid, and RFC5424 parsing is unchanged.

For the default RFC3164 time format, regression coverage verifies that two boundary spaces and tabs remain invalid. An explicit RFC3164 `time_format` beginning with a space treats that leading space as timestamp syntax, so this change does not impose or claim a one-space maximum for that configuration.

This also corrects the string parser's dynamic timestamp slice cursor and the field-count accounting needed by RFC3164 time formats that begin with a space. Truncated subsecond records with the boundary space are rejected through the normal `[nil, nil]` path instead of raising while looking for the missing field delimiter.

**Docs Changes**:

None. This broadens input compatibility without adding or changing configuration.

**Release Note**: 

parser_syslog: accept one optional ASCII space between RFC3164 priority and header.

**Testing**:

- `bundle exec ruby -Itest test/plugin/test_parser_syslog.rb --name test_truncated_subsecond_rfc3164_is_rejected_without_raising` — 4 tests, 12 assertions, 0 failures, 0 errors across `rfc3164`/`auto` and parser-owned/input-owned priority
- `bundle exec ruby -Itest test/plugin/test_parser_syslog.rb` — 132 tests, 521 assertions, 0 failures, 0 errors
- `bundle exec ruby -Itest test/plugin/test_in_syslog.rb` — 41 tests, 280 assertions, 0 failures, 0 errors
- `bundle exec rake test` — 4,401 tests, 16,154 assertions, 0 failures, 0 errors, 3 pendings, 36 omissions
- `gem exec rubocop` — 460 files inspected, 0 offenses
- Balanced microbenchmark with 8 samples per revision and case — all 8 prespecified fixed regression gates passed. The narrowest throughput ratio was 0.953; this was used only as a regression gate, and no performance improvement is claimed.
- Installed-gem UDP validation — 20/20 scenarios passed. Both baseline and candidate gems were built and installed into isolated environments, real `fluentd --no-supervisor` processes were started, and packets were sent by external UDP clients. The original 16-scenario matrix covered both parser engines, both priority-ownership models, `rfc3164`/`auto`, canonical and spaced packets, whitespace rejection, RFC5424 compatibility, and explicit leading-space time formats. Four additional string-parser scenarios verified that the truncated subsecond packet takes the normal unmatched path without an exception and that a valid `.5` sibling is parsed by the same live process across `rfc3164`/`auto` and both priority-ownership modes.

### DIFF
--- a/lib/fluent/plugin/parser_syslog.rb
+++ b/lib/fluent/plugin/parser_syslog.rb
@@ -71,6 +71,7 @@ module Fluent
         @time_parser_rfc5424 = nil
         @space_count_rfc3164 = nil
         @space_count_rfc5424 = nil
+        @time_format_starts_with_space_rfc3164 = false
         @skip_space_count_rfc3164 = false
         @skip_space_count_rfc5424 = false
         @time_parser_rfc5424_without_subseconds = nil
@@ -123,10 +124,12 @@ module Fluent
 
       def setup_time_parser_3164(time_fmt)
         @time_parser_rfc3164 = time_parser_create(format: time_fmt)
+        @time_format_starts_with_space_rfc3164 = time_fmt.start_with?(SPLIT_CHAR)
         if ['%b %d %H:%M:%S', '%b %d %H:%M:%S.%N'].include?(time_fmt)
           @skip_space_count_rfc3164 = true
         end
         @space_count_rfc3164 = time_fmt.squeeze(' ').count(' ') + 1
+        @space_count_rfc3164 -= 1 if @time_format_starts_with_space_rfc3164
       end
 
       def setup_time_parser_5424(time_fmt)
@@ -162,6 +165,13 @@ module Fluent
       end
 
       SPLIT_CHAR = ' '.freeze
+      SPLIT_BYTE = 0x20
+      DOT_CHAR = '.'.freeze
+      DOT_BYTE = 0x2e
+      GREATER_THAN_BYTE = 0x3e
+      ZERO_BYTE = 0x30
+      NINE_BYTE = 0x39
+      RFC3164_PRI_MINIMUM = [nil, nil, nil, 1, 10, 100].freeze
 
       def parse_rfc3164_regex(text, &block)
         idx = 0
@@ -180,12 +190,21 @@ module Fluent
 
         i = idx - 1
         sq = false
+        first_time_field = true
         @space_count_rfc3164.times do
           while text[i + 1] == SPLIT_CHAR
+            if first_time_field
+              unless @time_format_starts_with_space_rfc3164
+                idx += 1
+                i += 1
+                break
+              end
+            end
             sq = true
             i += 1
           end
 
+          first_time_field = false
           i = text.index(SPLIT_CHAR, i + 1)
         end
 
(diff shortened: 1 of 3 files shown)


Watson1978 (changes requested):
The following test should pass. It works on the master branch, but on this PR the behavior changes and it fails.

```ruby
  data(
    'boundary space' => '<14> Apr 25 16:43:29.5',
  )
  def test_truncated_subsecond_rfc3164_is_rejected_without_raising(text)
    @parser.configure(
      'parser_engine'  => 'string',
      'message_format' => 'rfc3164',
      'with_priority'  => true,
      'time_format'    => '%b %d %H:%M:%S.%N',
    )

    result = :not_yielded
    assert_nothing_raised do
      @parser.instance.parse(text) do |time, record|
        result = [time, record]
      end
    end
    assert_equal([nil, nil], result)
  end
```

On master, the input `<14> Apr 25 16:43:29.5` (a truncated message with a boundary space, no host/message) is gracefully rejected as `[nil, nil]`. 
On this PR it raises `NoMethodError(<undefined method '-' for nil>)` instead.

Please add above test case in this PR.

AkashKumar7902:
Thanks for catching this. Addressed in `65e84da7`:

- The string parser now checks for the missing field delimiter before subtracting the cursor and rejects the truncated record through the normal `[nil, nil]` path.
- The requested `<14> Apr 25 16:43:29.5` case is covered. I extended the same table-driven test across `rfc3164`/`auto` and parser-owned/input-owned priority, with a valid `.5 host app: message` control for each route.

Validation on the published SHA:

- focused regression: 4 tests, 12 assertions
- syslog parser/input suites: 132/521 and 41/280
- full suite: 4,401 tests, 16,154 assertions, 0 failures, 0 errors
- RuboCop: 460 files, 0 offenses
- installed-gem UDP matrix: 20/20 scenarios, including no exception/error/backtrace for all four truncated-subsecond routes and successful recovery on the same process

There is no inline review thread to resolve, so I’m replying in the PR conversation. The branch is ready for re-review. Hosted `Test`, `Benchmark`, and `RuboCop` runs were created for the new head and are awaiting the repository’s maintainer-approval gate for fork workflows.

Watson1978 (approved):

PR #5467, opened 2026-08-09
fix(metrics): export-only clamp for buffer statistics

### LINKED ISSUE #5303: Buffer size metrics showing negative values
### Details
I have Fluentd set up to forward logs to Kafka. In some of the Prometheus metrics for the buffers of those Kafka output plugins, I'm seeing negative values. Specifically, I'm seeing negative values for the following metrics:
- `fluentd_output_status_buffer_total_bytes`
- `fluentd_output_status_buffer_stage_byte_size`

Eg.
```
fluentd_output_status_buffer_total_bytes{worker_id="3",plugin_id="kafka_output",type="kafka2"} -7897834.0
fluentd_output_status_buffer_stage_byte_size{worker_id="3",plugin_id="kafka_output",type="kafka2"} -7897834.0
```

This appears to be the same issue as fluent/fluent-plugin-prometheus#55, open since 2018.

### Prometheus Config
```
<source>
  @type prometheus
  port 24240
  metrics_path /metrics
  aggregated_metrics_path /aggregated_metrics
</source>

<source>
  @type prometheus_output_monitor
</source>
```

### Env Info
- Fluentd: `1.11.2`
- fluent-plugin-prometheus: `1.8.2`
- fluent-plugin-prometheus_pushgateway: `0.0.2`
- fluent-plugin-kafka: `0.14.1`
- Note that Fluentd has multi-process workers enabled

VedantMadane:
Fixes #5303

Buffer size gauges (`stage_byte_size`, `queue_byte_size`, and derived `total_queued_size`) can go transiently negative when Fluentd core under/over-subtracts during concurrent stage/queue transitions (the deferred `@stage_size_metrics.add` after chunk unlock vs `enqueue_chunk`'s `sub` — see #2712 / #2734). Those values are mirrored by the Prometheus plugin as `fluentd_output_status_buffer_total_bytes` / `fluentd_output_status_buffer_stage_byte_size`, which is what #5303 reports.

Clamping the **gauge store** on `sub`/`dec` is the wrong fix: it turns a self-correcting transient negative into a **permanent over-count**, so `Buffer#storable?` (which reads the raw gauge) eventually refuses every write. Thanks @Watson1978 for catching that.

This PR takes an **export-only** approach:

1. Leave `LocalMetrics` gauge `sub`/`dec`/`set` semantics unchanged (negatives still allowed so the deferred-add race can self-heal).
2. Clamp stage/queue sizes to `>= 0` only when building `statistics` (the path Prometheus and the monitor agent consume).
3. Clamp `available_buffer_space_ratios` to `[0, 100]` when counters overshoot `total_limit_size`, and treat `total_limit_size == 0` as 0% free without dividing (no NaN / no dead NaN guard).

**Docs Changes**:
None

**Release Note**:
* buffer: clamp exported buffer size metrics to non-negative values 

**General Checklist**:
- [x] I have read and followed [CONTRIBUTING.md](https://github.com/fluent/fluentd/blob/master/CONTRIBUTING.md)
- [x] tests pass (CI)
- [x] DCO signed-off

**Tests**:
- `test/plugin/test_buffer.rb` `#statistics`: negative underlying gauges export as 0; overshoot clamps ratio to 0; `total_limit_size == 0` does not raise and ratio stays finite.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -922,7 +922,9 @@ module Fluent
         # report negative buffer sizes.
         stage_size = [@stage_size_metrics.get, 0].max
         queue_size = [@queue_size_metrics.get, 0].max
+        # Keep available-space ratio in [0, 1] even if counters overshoot total_limit_size.
         buffer_space = 1.0 - ((stage_size + queue_size * 1.0) / @total_limit_size)
+        buffer_space = [[buffer_space, 0.0].max, 1.0].min
         @stage_length_metrics.set(@stage.size)
         @queue_length_metrics.set(@queue.size)
         @available_buffer_space_ratios_metrics.set(buffer_space * 100)
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -923,8 +923,15 @@ module Fluent
         stage_size = [@stage_size_metrics.get, 0].max
         queue_size = [@queue_size_metrics.get, 0].max
         # Keep available-space ratio in [0, 1] even if counters overshoot total_limit_size.
-        buffer_space = 1.0 - ((stage_size + queue_size * 1.0) / @total_limit_size)
-        buffer_space = [[buffer_space, 0.0].max, 1.0].min
+        # Avoid Array#max/min on NaN (e.g. 0/0 when total_limit_size is 0), which raises.
+        denom = @total_limit_size.to_f
+        if denom > 0.0
+          buffer_space = 1.0 - ((stage_size + queue_size).to_f / denom)
+          buffer_space = 0.0 if buffer_space.nan? || buffer_space < 0.0
+          buffer_space = 1.0 if buffer_space > 1.0
+        else
+          buffer_space = 0.0
+        end
         @stage_length_metrics.set(@stage.size)
         @queue_length_metrics.set(@queue.size)
         @available_buffer_space_ratios_metrics.set(buffer_space * 100)
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -917,7 +917,11 @@ module Fluent
       ]
 
       def statistics
-        stage_size, queue_size = @stage_size_metrics.get, @queue_size_metrics.get
+        # Clamp to non-negative: historical races can leave counters slightly
+        # negative (issues #5303, #2712). Exported Prometheus gauges must not
+        # report negative buffer sizes.
+        stage_size = [@stage_size_metrics.get, 0].max
+        queue_size = [@queue_size_metrics.get, 0].max
         buffer_space = 1.0 - ((stage_size + queue_size * 1.0) / @total_limit_size)
         @stage_length_metrics.set(@stage.size)
         @queue_length_metrics.set(@queue.size)
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -917,21 +917,15 @@ module Fluent
       ]
 
       def statistics
-        # Clamp to non-negative: historical races can leave counters slightly
-        # negative (issues #5303, #2712). Exported Prometheus gauges must not
-        # report negative buffer sizes.
+        # Export-only clamp: internal gauges may go transiently negative during
+        # the deferred stage_size add vs enqueue_chunk sub race (#5303, #2712).
+        # Clamping the gauge store itself would turn that into a permanent
+        # over-count and break Buffer#storable? -- keep raw gauge semantics.
         stage_size = [@stage_size_metrics.get, 0].max
         queue_size = [@queue_size_metrics.get, 0].max
-        # Keep available-space ratio in [0, 1] even if counters overshoot total_limit_size.
-        # Avoid Array#max/min on NaN (e.g. 0/0 when total_limit_size is 0), which raises.
         denom = @total_limit_size.to_f
-        if denom > 0.0
-          buffer_space = 1.0 - ((stage_size + queue_size).to_f / denom)
-          buffer_space = 0.0 if buffer_space.nan? || buffer_space < 0.0
-          buffer_space = 1.0 if buffer_space > 1.0
-        else
-          buffer_space = 0.0
-        end
+        # denom > 0 already excludes 0/0 NaN; stage/queue are floored above.
+        buffer_space = denom > 0.0 ? (1.0 - (stage_size + queue_size).to_f / denom).clamp(0.0, 1.0) : 0.0
         @stage_length_metrics.set(@stage.size)
         @queue_length_metrics.set(@queue.size)
         @available_buffer_space_ratios_metrics.set(buffer_space * 100)
--- a/lib/fluent/plugin/metrics_local.rb
+++ b/lib/fluent/plugin/metrics_local.rb
@@ -62,10 +62,7 @@ module Fluent
 
       def dec_gauge
         @monitor.synchronize do
-          # Buffer size / length gauges must never go negative even if a race
-          # causes sub/dec to run more times than add/inc (see #5303).
           @store -= 1
-          @store = 0 if @store < 0
         end
       end
 
@@ -77,10 +74,7 @@ module Fluent
 
       def sub_gauge(value)
         @monitor.synchronize do
-          # Prevent negative values that leak into Prometheus buffer metrics
-          # (fluentd_output_status_buffer_total_bytes, etc.). See #5303 / #2712.
           @store -= value
-          @store = 0 if @store < 0
         end
       end
 
--- a/lib/fluent/plugin/metrics_local.rb
+++ b/lib/fluent/plugin/metrics_local.rb
@@ -62,7 +62,10 @@ module Fluent
 
       def dec_gauge
         @monitor.synchronize do
+          # Buffer size / length gauges must never go negative even if a race
+          # causes sub/dec to run more times than add/inc (see #5303).
           @store -= 1
+          @store = 0 if @store < 0
         end
       end
 
@@ -74,7 +77,10 @@ module Fluent
 
       def sub_gauge(value)
         @monitor.synchronize do
+          # Prevent negative values that leak into Prometheus buffer metrics
+          # (fluentd_output_status_buffer_total_bytes, etc.). See #5303 / #2712.
           @store -= value
+          @store = 0 if @store < 0
         end
       end
 
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -1565,5 +1565,18 @@ class BufferTest < Test::Unit::TestCase
       assert stats['available_buffer_space_ratios'] <= 100.0
       assert_equal 0.0, stats['available_buffer_space_ratios']
     end
+
+    test 'available_buffer_space_ratios is safe when total_limit_size is zero' do
+      # 0/0 would be NaN; Array#max/min on NaN raises — must not crash export.
+      @p.instance_variable_set(:@total_limit_size, 0)
+      @p.stage_size_metrics.set(0)
+      @p.queue_size_metrics.set(0)
+      stats = @p.statistics['buffer']
+      assert_equal 0, stats['stage_byte_size']
+      assert_equal 0, stats['queue_byte_size']
+      assert stats['available_buffer_space_ratios'] >= 0.0
+      assert stats['available_buffer_space_ratios'] <= 100.0
+      refute stats['available_buffer_space_ratios'].to_f.nan?
+    end
   end
 end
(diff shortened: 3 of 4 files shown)


Watson1978:
Thanks for digging into #5303. However, I think clamping at zero is the wrong direction for the buffer counters: it converts a transient, self-healing negative into a **permanent over-count**.

`Buffer#write` intentionally defers the `add`. As part of the #2712 fix, `staged_bytesizes_by_chunk` is accumulated while each chunk is locked, but `@stage_size_metrics.add(bytesize)` only runs *after* the chunk locks are released (`chunk.mon_exit` at `lib/fluent/plugin/buffer.rb:386` / `:400`, and the deferred `add` at `:416`).

That leaves a window in which the enqueue thread can run `enqueue_chunk` → `@stage_size_metrics.sub(chunk.bytesize)` (`buffer.rb:504`) while `stage_size` is still 0.

Take a 100-byte chunk in that window:

* **Before this PR**: the store goes to `-100`, then the pending `add(100)` brings it back to the correct `0`. The negative reading is transient and self-correcting — ugly in the export, but the accounting converges.
* **After this PR**: `sub` clamps the store to `0`, then the pending `add(100)` leaves `stage_size == 100` while the real staged size is `0`. The `+100` never decays.

Under load this repeats on every flush cycle and accumulates. Note that `storable?` (`buffer.rb:300`) reads the raw `.get`, not the clamped `statistics` value:

```ruby
@total_limit_size > @stage_size_metrics.get + @queue_size_metrics.get
```

So an effectively empty buffer eventually fails this check and the output starts raising `BufferOverflowError` permanently. That would turn a cosmetic metrics problem into a back-pressure / data-loss problem.

Two related points:

1. The negative value is currently the only signal operators have that add/sub pairing broke. Clamping silently removes the alert without removing the drift — an over-subtraction that does not cross zero was never visible to begin with, and after this change nothing is left to alert on.
2. `set_gauge` is not clamped, so `buffer.stage_size = -1` still stores a negative. `statistics` would report `0` while `storable?` computes with `-1`, i.e. the exported number and the number the buffer actually acts on disagree.

I would suggest fixing the add/sub pairing in `buffer.rb` (the deferred `add` at `:416` versus `enqueue_chunk`'s `sub` at `:504`) rather than clamping the generic gauge type. If a defensive clamp for alternate metrics backends is still wanted, doing it once in the `stage_size` / `queue_size` accessors (`buffer.rb:204-218`) would cover `statistics`, `storable?` and `in_monitor_agent` in one place, and it should emit a `log.warn` so the underlying race stays diagnosable.

One more thing unrelated to the design question: `@store = 0 if @store < 0` is now the last expression of the `synchronize` block, so `dec_gauge` / `sub_gauge` return `nil` on the non-clamped path. `m.set(10); m.dec` returned `9` before and returns `nil` now, while `inc` / `add` / `set` still return the value (and the existing tests assert that contract, e.g. `assert_equal 1, @m.inc`). Worth an explicit `@store` at the end of the block, plus a return-value assertion in the new tests.

lib/fluent/plugin/buffer.rb
```diff
@@ -917,7 +917,11 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
       ]
 
       def statistics
-        stage_size, queue_size = @stage_size_metrics.get, @queue_size_metrics.get
+        # Clamp to non-negative: historical races can leave counters slightly
+        # negative (issues #5303, #2712). Exported Prometheus gauges must not
+        # report negative buffer sizes.
+        stage_size = [@stage_size_metrics.get, 0].max
+        queue_size = [@queue_size_metrics.get, 0].max
```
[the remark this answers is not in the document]
VedantMadane: Addressed in 3b17bb57b96585ae6b379c7a739a9440218b0de7.

Added `test/plugin/test_buffer.rb` coverage under `#statistics`: forces `stage_size_metrics`/`queue_size_metrics` negative via `set`, then asserts exported `stage_byte_size`/`queue_byte_size`/`total_queued_size` are 0 (and `available_buffer_space_ratios` stays non-negative). This exercises the export-path `[metrics.get, 0].max` clamp independently of LocalMetrics `sub`/`dec` flooring.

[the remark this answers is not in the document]
VedantMadane: Addressed:

1. **Export-path tests** (already on the branch since `ce7211a`): `test/plugin/test_buffer.rb` asserts `stage_byte_size` / `queue_byte_size` / `total_queued_size` export as `0` when underlying gauges are negative.
2. **Ratio clamp** (new commit): `available_buffer_space_ratios` is now clamped to `[0, 100]` when counters overshoot `total_limit_size`, with a dedicated unit test.

Thanks for the review.

VedantMadane:
Addressed remaining Copilot note on NaN when `total_limit_size` is 0: compute the ratio only when denom > 0, clamp with comparisons (not Array#max/min on NaN), and treat zero/invalid capacity as 0.0 free-space ratio. Added a unit test that export does not raise and the ratio is finite.

lib/fluent/plugin/metrics_local.rb
```diff
@@ -62,7 +62,10 @@ def inc
 
       def dec_gauge
         @monitor.synchronize do
+          # Buffer size / length gauges must never go negative even if a race
+          # causes sub/dec to run more times than add/inc (see #5303).
           @store -= 1
+          @store = 0 if @store < 0
```
Watson1978: Clamping `sub_gauge`/`dec_gauge` at zero discards a transient negative excursion that used to self-correct, so `stage_size` stays permanently inflated; `Buffer#storable?` reads that raw gauge, so the buffer refuses every further write even when it is empty.

lib/fluent/plugin/metrics_local.rb
```diff
@@ -74,7 +77,10 @@ def add(value)
 
       def sub_gauge(value)
         @monitor.synchronize do
+          # Prevent negative values that leak into Prometheus buffer metrics
+          # (fluentd_output_status_buffer_total_bytes, etc.). See #5303 / #2712.
           @store -= value
+          @store = 0 if @store < 0
```
Watson1978: Clamping `sub_gauge`/`dec_gauge` at zero discards a transient negative excursion that used to self-correct, so `stage_size` stays permanently inflated; `Buffer#storable?` reads that raw gauge, so the buffer refuses every further write even when it is empty.

lib/fluent/plugin/metrics_local.rb
Watson1978: `@store = 0 if @store < 0` is the last expression in the synchronize block, and a false guard evaluates to `nil`, so `dec`/`sub` return `nil` on the normal (non-clamped) path instead of the new store value.

lib/fluent/plugin/metrics_local.rb
Watson1978: `@store = 0 if @store < 0` is the last expression in the synchronize block, and a false guard evaluates to `nil`, so `dec`/`sub` return `nil` on the normal (non-clamped) path instead of the new store value.

lib/fluent/plugin/buffer.rb
```diff
@@ -917,8 +917,21 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
+        # Keep available-space ratio in [0, 1] even if counters overshoot total_limit_size.
+        # Avoid Array#max/min on NaN (e.g. 0/0 when total_limit_size is 0), which raises.
+        denom = @total_limit_size.to_f
+        if denom > 0.0
+          buffer_space = 1.0 - ((stage_size + queue_size).to_f / denom)
+          buffer_space = 0.0 if buffer_space.nan? || buffer_space < 0.0
+          buffer_space = 1.0 if buffer_space > 1.0
+        else
+          buffer_space = 0.0
+        end
```
Watson1978: The `denom > 0.0` branch already covers the case this comment describes, so the NaN
guard inside it is dead code.

- `total_limit_size == 0` no longer reaches the division at all — it takes the `else` branch.
- Inside the `if`, `stage_size` and `queue_size` are already floored at 0 by the
  `[..., 0].max` calls above, and `denom` is positive and finite. So
  `(stage_size + queue_size).to_f / denom` is always `>= 0` and never `NaN`, which means
  neither `buffer_space.nan?` nor `buffer_space > 1.0` can ever be true.

I'd drop both and keep just the lower bound:

```ruby
denom = @total_limit_size.to_f
buffer_space = denom > 0.0 ? (1.0 - (stage_size + queue_size) / denom).clamp(0.0, 1.0) : 0.0

**Review by Watson1978 — CHANGES_REQUESTED** [review]

**Comment by VedantMadane** [comment]
Thanks @Watson1978 — you were right about the gauge-store clamp.

**Addressed in 3fe8d18:**

1. **Reverted `LocalMetrics` `sub`/`dec` zero-clamping** entirely. Flooring the store discarded the self-correcting transient negative from the deferred `stage_size` `add` vs `enqueue_chunk` `sub` race, leaving a permanent over-count that `Buffer#storable?` (raw `.get`) would eventually treat as full. Internal gauge semantics are unchanged again; the `nil` return-value bug on the non-clamped path is gone with the clamp.

2. **Export-only clamp** remains in `Buffer#statistics` via `[metrics.get, 0].max` for `stage_byte_size` / `queue_byte_size` / `total_queued_size`, so Prometheus still never sees negatives.

3. **`available_buffer_space_ratios`** simplified per your suggestion:
   ```ruby
   denom = @total_limit_size.to_f
   buffer_space = denom > 0.0 ? (1.0 - (stage_size + queue_size).to_f / denom).clamp(0.0, 1.0) : 0.0
   ```
   Dead NaN guard inside the `denom > 0` branch removed.

4. Dropped the LocalMetrics floor unit tests; kept / clarified the statistics export tests (negative gauges, overshoot ratio, `total_limit_size == 0`).

I did **not** move the clamp into the `stage_size` / `queue_size` accessors or fix the deferred-add pairing in this PR — export-only keeps the cosmetic metrics fix without changing `storable?` behavior. Happy to follow up on the race itself in a separate change if that is preferred.

File: lib/fluent/plugin/metrics_local.rb (hunk 2 above)
Comment (VedantMadane) [reply lib/fluent/plugin/metrics_local.rb]: Addressed in 3fe8d18: reverted LocalMetrics sub/dec zero-clamp entirely so the deferred-add vs enqueue_chunk-sub race can still self-correct. Export-only clamp remains in Buffer#statistics.

File: lib/fluent/plugin/metrics_local.rb (hunk 3 above)
Comment (VedantMadane) [reply lib/fluent/plugin/metrics_local.rb]: Addressed in 3fe8d18: same as sibling — LocalMetrics clamp removed; raw gauge semantics restored.

File: lib/fluent/plugin/metrics_local.rb (hunk 2 above)
Comment (VedantMadane) [reply lib/fluent/plugin/metrics_local.rb]: Addressed in 3fe8d18: clamp (and the bad last-expression return) removed with the revert of LocalMetrics flooring.

File: lib/fluent/plugin/metrics_local.rb (hunk 3 above)
Comment (VedantMadane) [reply lib/fluent/plugin/metrics_local.rb]: Addressed in 3fe8d18: clamp (and the bad last-expression return) removed with the revert of LocalMetrics flooring.

File: lib/fluent/plugin/buffer.rb (hunk 4 above)
Comment (VedantMadane) [reply lib/fluent/plugin/buffer.rb]: Addressed in 3fe8d18: simplified to your suggested form — `denom > 0.0 ? (...).clamp(0.0, 1.0) : 0.0`; dead NaN guard dropped.

File: lib/fluent/plugin/buffer.rb [hunk 5]
```diff
@@ -917,8 +917,14 @@ def write_step_by_step(metadata, data, format, splits_count, &block)
...
       def statistics
-        stage_size, queue_size = @stage_size_metrics.get, @queue_size_metrics.get
+        # Clamp to non-negative: historical races can leave counters slightly
+        # negative (issues #5303, #2712). Exported Prometheus gauges must not
+        # report negative buffer sizes.
+        stage_size = [@stage_size_metrics.get, 0].max
+        queue_size = [@queue_size_metrics.get, 0].max
+        # Keep available-space ratio in [0, 1] even if counters overshoot total_limit_size.
         buffer_space = 1.0 - ((stage_size + queue_size * 1.0) / @total_limit_size)
+        buffer_space = [[buffer_space, 0.0].max, 1.0].min
```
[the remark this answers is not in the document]
VedantMadane: Addressed in 3fe8d18: no longer uses Array#max/min on the ratio. Zero capacity takes the else branch (0.0); positive denom uses Numeric#clamp on a non-NaN value.

Watson1978:
The export-only approach looks right to me for #5303. Just noting for the record that the underlying race is still open: `#write` defers `@stage_size_metrics.add` until after the chunk locks are released (buffer.rb L414-L419), so a concurrent `enqueue_chunk` can `sub` the full `chunk.bytesize` first (L503-L505) and leave the gauge transiently negative. It should be rare and it isn't blocking this PR — let's track it in a separate issue.

Filed in https://github.com/fluent/fluentd/issues/5479

Watson1978 (approved):
Thanks

PR #5469, opened 2026-08-12
supervisor: reduce memory usage of cleanup_lock_dir with huge number of lock files

Watson1978:
Fixes #

`cleanup_lock_dir` collects all lock file paths with `Dir.glob` at shutdown, which materializes the whole path list in memory.

With that many files, the `Dir.glob` array causes a multi-GB memory spike during shutdown,
which increases the risk of hitting systemd's `TimeoutStopSec` and getting SIGKILLed in the middle of cleanup.

This PR switches to `Dir.each_child` to stream directory entries one by one.

Measurement (removing 1,000,000 lock files, RSS):

|                | RSS growth | elapsed |
|----------------|-----------:|--------:|
| `Dir.glob`     |  884.5 MiB |  26.0 s |
| `Dir.each_child` |  1.8 MiB |  25.4 s |

**Docs Changes**:
N/A

**Release Note**: 
* supervisor: reduce memory usage of cleanup_lock_dir with huge number of lock files

### DIFF
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -102,7 +102,7 @@ module Fluent
     def cleanup_lock_dir
       begin
         Dir.each_child(@fluentd_lock_dir) do |name|
-          FileUtils.rm(File.join(@fluentd_lock_dir, name)) if File.fnmatch?("fluentd-*.lock", name)
+          FileUtils.rm_f(File.join(@fluentd_lock_dir, name)) if File.fnmatch?("fluentd-*.lock", name)
         end
       rescue Errno::ENOENT
         # Directory is already missing. Fall through and let rmdir below
--- a/lib/fluent/supervisor.rb
+++ b/lib/fluent/supervisor.rb
@@ -100,7 +100,14 @@ module Fluent
     end
 
     def cleanup_lock_dir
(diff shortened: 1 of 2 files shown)


lib/fluent/supervisor.rb
```diff
@@ -100,7 +100,14 @@ def after_run
     end
 
     def cleanup_lock_dir
-      FileUtils.rm(Dir.glob(File.join(@fluentd_lock_dir, "fluentd-*.lock")))
+      begin
+        Dir.each_child(@fluentd_lock_dir) do |name|
+          FileUtils.rm(File.join(@fluentd_lock_dir, name)) if File.fnmatch?("fluentd-*.lock", name)
```
kenhys:
```suggestion
          FileUtils.rm_f(File.join(@fluentd_lock_dir, name)) if File.fnmatch?("fluentd-*.lock", name)
```

If another (third party) process removed .lock file unexpectedly, the rest of .lock files are not removed.

kenhys (changes requested):

lib/fluent/supervisor.rb
Watson1978: Thanks

kenhys (approved):

PR #5473, opened 2026-08-18
Add evacuate_limit_size to bound chunk evacuation

### LINKED ISSUE #5352: Evacuate chunk on networking issues need limits
### Describe the bug

We output to multiple locations, one of the locations was down for a long period of time and fluent began evacuating chunks to the root_dir until it filled up the disk.  It would be good to add an option to limit the size of the evacuated chunks or disable the functionality

### To Reproduce

Make an output unreachable/timeout and let the logs flow until root_dir is full.

### Expected behavior

Evacuated chunks should have a limit to prevent filling up the disk or the feature should be able to be disabled. Or both

### Your Environment

```markdown
- Fluentd version: 1.19.0
- Package version:
- Operating system:
- Kernel version:
```

### Your Configuration

```apache
Any config with an output that can be unavailable/unreachable
```

### Your Error Log

```shell
2026-05-04 15:15:29 +0000 [warn]: #0 chunk files are evacuated to /tmp/fluent/buffer/object_570. chunk_id="650ff4c25bf36ff8208adc20ecb6e68d"
```

### Additional context

_No response_

ChrisJr404:
Fixes #5352

Adds a `evacuate_limit_size` buffer param so chunk evacuation can't fill up the disk when an output stays down.

Right now, when the retry limit is hit and `clear_queue!` runs, every queued chunk gets evacuated to the backup dir. If an output is unreachable for a long time that keeps piling files onto disk with no bound, which is exactly what #5352 ran into (evacuated chunks filled `root_dir`).

The new param caps the total size that gets evacuated in one `clear_queue!` pass. Chunks are evacuated in queue order until the next one wouldn't fit in the remaining budget; after that the rest are purged without evacuating, so disk usage stays under the limit. Setting it to `0` turns evacuation off completely, and the default is `nil` which keeps the current unbounded behavior, so nobody's setup changes unless they opt in.

The size accounting and the skip/disable decision live in the base `Buffer` class (`clear_queue!` plus a small `evacuate_chunk?` helper), so it applies to every buffer type that overrides `evacuate_chunk` (file, file_single). While I was in that loop I also fixed the trace log line there, which referenced an undefined `chunk` instead of `q`.

Tests: added three cases to `test/plugin/test_buffer.rb` covering the default (all chunks evacuated), `evacuate_limit_size 0` (nothing evacuated, everything still purged), and a positive limit (evacuation stops once the budget is spent). Ran `bundle exec ruby -Itest test/plugin/test_buffer.rb` -> 72 tests, 420 assertions, 0 failures, 0 errors.

**Docs Changes**:
The new `evacuate_limit_size` buffer parameter should be documented in the buffer section of the docs. Happy to open a PR on fluentd-docs-gitbook.

**Release Note**:
Add `evacuate_limit_size` buffer parameter to bound (or disable, with `0`) the total size of chunks evacuated when the retry limit is reached. Default `nil` keeps the current unbounded behavior.

### DIFF
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -72,6 +72,9 @@ module Fluent
       desc 'If true, chunks are thrown away when unrecoverable error happens'
       config_param :disable_chunk_backup, :bool, default: false
 
+      desc 'The total size limit for chunks evacuated on unrecoverable errors. Once evacuation reaches this size, remaining chunks are purged without being evacuated. Set 0 to disable evacuation. nil (default) means no limit.'
+      config_param :evacuate_limit_size, :size, default: nil
+
       Metadata = Struct.new(:timekey, :tag, :variables, :seq) do
         def initialize(timekey, tag, variables)
           super(timekey, tag, variables, 0)
@@ -625,11 +628,15 @@ module Fluent
         log.on_trace { log.trace "clearing queue", instance: self.object_id }
 
         synchronize do
+          evacuated_size = 0
           until @queue.empty?
             begin
               q = @queue.shift
-              evacuate_chunk(q)
-              log.trace("purging a chunk in queue"){ {id: dump_unique_id_hex(chunk.unique_id), bytesize: chunk.bytesize, size: chunk.size} }
+              if evacuate_chunk?(q.bytesize, evacuated_size)
+                evacuate_chunk(q)
+                evacuated_size += q.bytesize
+              end
+              log.trace("purging a chunk in queue"){ {id: dump_unique_id_hex(q.unique_id), bytesize: q.bytesize, size: q.size} }
               q.purge
             rescue => e
               log.error "unexpected error while clearing buffer queue", error_class: e.class, error: e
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/buffer.rb
```diff
@@ -625,11 +628,15 @@ def clear_queue!
...
         synchronize do
+          evacuated_size = 0
           until @queue.empty?
             begin
               q = @queue.shift
-              evacuate_chunk(q)
-              log.trace("purging a chunk in queue"){ {id: dump_unique_id_hex(chunk.unique_id), bytesize: chunk.bytesize, size: chunk.size} }
+              if evacuate_chunk?(q.bytesize, evacuated_size)
+                evacuate_chunk(q)
+                evacuated_size += q.bytesize
```
Watson1978, line 637: In `clear_queue!`, `evacuated_size += q.bytesize` runs unconditionally after `evacuate_chunk(q)`, without checking whether the backup actually succeeded. Since `evacuate_chunk` in both `buf_file.rb` and `buf_file_single.rb` rescues its own exceptions instead of re-raising, a failed backup copy (e.g. `ENOSPC`, permission error) still counts against the budget.

Worth confirming success (or having `evacuate_chunk` report bytes actually written) before counting a chunk toward `evacuated_size`.

lib/fluent/plugin/buffer.rb
```diff
@@ -625,11 +628,15 @@ def clear_queue!
         log.on_trace { log.trace "clearing queue", instance: self.object_id }
 
         synchronize do
+          evacuated_size = 0
```
Watson1978, line 631: [Nit] since the budget actually resets on every `clear_queue!` call rather than being a true cumulative cap, the name `evacuate_limit_size` and the desc's "total size limit" phrasing read as more absolute than what's implemented. Worth clarifying in the desc that this is per-`clear_queue!` call, not a lifetime/cumulative limit — otherwise the naming suggestion (e.g. something like `evacuate_limit_size_per_clear`) is up to you.

Watson1978 (changes requested):

simrangera99:
Is there any update, when can we expect this to go live as this is a major issue when used at larger scale and there is no current ways supported to disable this Chunk's Backup Completely.

PR #5475, opened 2026-08-21
fix(buffer): decrement queue_size in ensure so failed purge does not leak

### LINKED ISSUE #5468: buffer: queue_size leaks permanently when chunk.purge fails after a successful write, eventually causing spurious BufferOverflowError
### Describe the bug

In Fluent::Plugin::Buffer#purge_chunk, the queued byte counter (@queue_size_metrics) is only decremented inside the begin/rescue block, after chunk.purge succeeds. If chunk.purge raises, the rescue swallows the error and @queue_size_metrics.sub(bytesize) is skipped — but the chunk has already been removed from @dequeued at the top of the method, so nothing ever retries the purge. The result is a permanent leak of @queue_size (and, if the failure was in the file unlink, an orphaned buffer file left on disk).

Because storable? is

def storable?
  @total_limit_size > @stage_size_metrics.get + @queue_size_metrics.get
end
each failed purge ratchets @queue_size upward and never comes back down. Over a long-lived process this drifts toward total_limit_size, at which point storable? returns false and every subsequent emit raises Fluent::Plugin::Buffer::BufferOverflowError even though the real staged + queued data is small or empty. Input plugins then start rejecting incoming data (e.g. in_tail stops advancing), so records are dropped. Only a restart resets the counter.

This is the queue-side sibling of the staged-side accounting leak fixed in https://github.com/fluent/fluentd/issues/5439. Same user-visible symptom — spurious BufferOverflowError on a near-empty buffer — but a different trigger and code path.

    def purge_chunk(chunk_id)
    metadata = nil
    synchronize do
      chunk = @dequeued.delete(chunk_id)   # (1) removed from @dequeued up front
      return nil unless chunk

    metadata = chunk.metadata
    begin
      bytesize = chunk.bytesize
      chunk.purge                        # (2) can raise (unlink/close on the buffer path)
      @queue_size_metrics.sub(bytesize)  # (3) SKIPPED when (2) raises
    rescue => e
      log.error "failed to purge buffer chunk", chunk_id: dump_unique_id_hex(chunk_id), error_class: e.class, error: e
      log.error_backtrace
    end

    @dequeued_num[chunk.metadata] -= 1
    ...
    end
    nil
    end
If (2) raises:

(3) never runs → @queue_size is never decremented for those bytes.
The chunk is already gone from both @queue (dequeued before the write) and @dequeued (step 1) → no retry path, the leak is permanent.
If the raise came from the file unlink itself, the buffer file also stays on disk permanently.
To be precise about terminology: the leak is in the queued byte size (@queue_size / buffer_queue_byte_size), which is what storable? checks — not the queue element count.
If (2) raises:

(3) never runs → @queue_size is never decremented for those bytes.
The chunk is already gone from both @queue (dequeued before the write) and @dequeued (step 1) → no retry path, the leak is permanent.
If the raise came from the file unlink itself, the buffer file also stays on disk permanently.
To be precise about terminology: the leak is in the queued byte size (@queue_size / buffer_queue_byte_size), which is what storable? checks — not the queue element count.

### To Reproduce

Self-contained script driving the real Buffer API (`write → enqueue_chunk → dequeue_chunk → purge_chunk`), with `chunk.purge` forced to raise to simulate an unlink/`EIO` failure on the buffer path. Deterministic — no flush-thread timing.

```ruby
require 'fluent/version'
require 'fluent/test'
require 'fluent/test/driver/output'
require 'fluent/plugin/buffer'
require 'fluent/plugin/buffer/memory_chunk'

# simulate the physical purge (unlink/close) failing
module FailingPurge
  def purge; raise IOError, "simulated purge failure (unlink EIO)"; end
end
Fluent::Plugin::Buffer::MemoryChunk.prepend(FailingPurge)

class PurgeLeakOut < Fluent::Plugin::Output
  def format(tag, time, record); record['msg'] + "\n"; end
  def write(chunk); end   # write to the OUTPUT succeeds; only purge fails
end
Fluent::Plugin.register_output('purge_leak_out', PurgeLeakOut)

d = Fluent::Test::Driver::Output.new(PurgeLeakOut)
d.configure(%[
  <buffer>
    @type memory
    chunk_limit_size 1k
    total_limit_size 16k
    flush_at_shutdown false
  </buffer>
])
out = d.instance; out.start; buf = out.buffer

1.upto(40) do |i|
  begin
    meta = buf.metadata()
    buf.write({ meta => ["x" * 900 + "\n"] })  # stage ~900B (succeeds)
    buf.enqueue_chunk(meta)                     # staged -> queue (queue_size += ~900)
    c = buf.dequeue_chunk                       # queue -> dequeued (flush picks it up)
    buf.purge_chunk(c.unique_id) if c           # commit -> purge; chunk.purge raises
    note = ""
  rescue Fluent::Plugin::Buffer::BufferOverflowError
    note = "BufferOverflowError (buffer refuses write)"
  end
  real = buf.queue.map(&:bytesize).sum + buf.instance_variable_get(:@dequeued).values.map(&:bytesize).sum
  printf("iter=%-3d queue_size=%-6d real_queued=%-3d storable?=%-6s %s\n", i, buf.queue_size, real, buf.storable?, note)
end
```

Run it (no local ruby needed):

```
docker run --rm -v "$PWD:/t" ruby:3.2 sh -c \
  'gem install fluentd -v 1.19.3 --no-document -q; ruby /t/repro.rb'
```

### Observed output (fluentd 1.19.3)

```
total_limit_size: 16384 bytes
iter=1   queue_size=901    real_queued=0   storable?=true
iter=2   queue_size=1802   real_queued=0   storable?=true
...
iter=18  queue_size=16218  real_queued=0   storable?=true
iter=19  queue_size=17119  real_queued=0   storable?=false
iter=20  queue_size=17119  real_queued=0   storable?=false  BufferOverflowError (buffer refuses write)
iter=21  queue_size=17119  real_queued=0   storable?=false  BufferOverflowError (buffer refuses write)
... (every subsequent emit raises)
```

`queue_size` climbs by each failed chunk's bytesize and never comes back down, while `real_queued` (the bytes actually held in `@queue` + `@dequeued`) stays **0**. Once `queue_size` crosses `total_limit_size`, `storable?` is permanently false and every emit raises `BufferOverflowError` on an empty buffer. Each iteration also logs `failed to purge buffer chunk`. A restart clears it.

Moving `@queue_size_metrics.sub(bytesize)` so it runs even when `chunk.purge` fails (i.e. in an `ensure`) makes `queue_size` stay at 0 and the overflow never occurs — verified in the same harness.

[fluentd-purge-leak-repro.zip](https://github.com/user-attachments/files/30892530/fluentd-purge-leak-repro.zip)

### Expected behavior

A failed chunk.purge should not permanently inflate @queue_size. The queued bytes should be reconciled (decremented, or the purge retried/deferred) so that total_queued_size stays consistent with the chunks the buffer is actually holding, and storable? does not become permanently false due to accounting drift.

### Your Environment

```markdown
- Fluentd version:reproduced by source inspection on master; the purge_chunk code path is identical in v1.18.0, v1.19.0, v1.19.3
- Package version:
- Operating system:
- Kernel version:
```

### Your Configuration

```apache
<system>
  log_level info
</system>

<source>
  @type sample
  tag test
  sample {"msg":"hello world hello world hello world"}
  rate 200
</source>

<match test>
  @type noop_format
  <buffer>
    @type file
    path /tmp/buf/test
    chunk_limit_size 1k
    total_limit_size 16k
    flush_mode immediate
    flush_thread_count 1
    retry_type periodic
    retry_wait 1s
    overflow_action throw_exception
    flush_at_shutdown false
  </buffer>
</match>
```

### Your Error Log

```shell
# Trigger: chunk.purge raises; purge_chunk rescues it, so queue_size is never decremented
2026-08-10 08:43:50 +0000 [error]: #0 failed to purge buffer chunk chunk_id="658ad59b0aea892ad9316cb427274a28" error_class=IOError error=#<IOError: simulated purge failure (unlink EIO on buffer path)>
  2026-08-10 08:43:50 +0000 [error]: #0 /path/fail_purge.rb:9:in `purge'
  2026-08-10 08:43:50 +0000 [error]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/buffer.rb:606:in `block in purge_chunk'
  2026-08-10 08:43:50 +0000 [error]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/buffer.rb:597:in `purge_chunk'
  2026-08-10 08:43:50 +0000 [error]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/output.rb:1139:in `commit_write'
  2026-08-10 08:43:50 +0000 [error]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/output.rb:1260:in `try_flush'
  2026-08-10 08:43:50 +0000 [error]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/output.rb:1571:in `flush_thread_run'

# Consequence: leaked queue_size crosses total_limit_size -> storable? false -> every emit rejected
2026-08-10 08:43:50 +0000 [warn]: #0 failed to write data into buffer by buffer overflow action=:throw_exception
2026-08-10 08:43:50 +0000 [warn]: #0 emit transaction failed: error_class=Fluent::Plugin::Buffer::BufferOverflowError error="buffer space has too many data" location="/usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/buffer.rb:335:in `write'" tag="test"
  2026-08-10 08:43:50 +0000 [warn]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/output.rb:921:in `emit_buffered'
  2026-08-10 08:43:50 +0000 [warn]: #0 /usr/local/bundle/gems/fluentd-1.19.3/lib/fluent/plugin/in_sample.rb:115:in `emit'
```

### Additional context

_No response_

mehrdadbn9:
Fixes #5468

In `Fluent::Plugin::Buffer#purge_chunk`, `@queue_size_metrics` is only decremented inside `begin/rescue`. When `chunk.purge` raises (e.g. unlink/close failure on the buffer path), the error is logged and swallowed but the `sub` is skipped. The chunk is already removed from `@dequeued`, so it is never retried — the queued byte counter ratchets toward `total_limit_size`, making `storable?` permanently `false` and causing a spurious `BufferOverflowError` on a near-empty buffer.

This moves `bytesize = chunk.bytesize` above the `begin` and moves `@queue_size_metrics.sub(bytesize)` into an `ensure` block so the counter is always released exactly once per `purge_chunk` call, regardless of whether `purge` succeeds.

Why safe:
- `ensure` runs exactly once per call → no double-decrement.
- Every chunk reaching `purge_chunk` was previously enqueued (dequeue/emit path), so the `sub` is always balanced.
- If `chunk.purge` raises, the chunk is lost on filesystem error (already a logged, separate failure); this change only preserves counter integrity so `storable?` no longer spuriously goes false.

**Docs Changes**:
None (behavior fix; inline comment added).

**Release Note**:
* buffer: fix spurious `BufferOverflowError` caused by `queue_size` leaking when a chunk purge fails

### DIFF
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -435,6 +435,32 @@ class BufferTest < Test::Unit::TestCase
       assert_equal({}, @p.dequeued)
     end
 
+    test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      failing_purge = Module.new do
+        def purge
+          raise IOError, 'simulated purge failure (unlink EIO)'
+        end
+      end
+      @p.buffer_class::Chunk.include(failing_purge)
+
+      m1 = @p.dequeue_chunk
+      assert_equal [@dm0, @dm1, @dm1], @p.queue.map(&:metadata)
+      assert_equal({m1.unique_id => m1}, @p.dequeued)
+
+      queued_before = @p.queue_size
+      assert queued_before > 0
+
+      # purge_chunk swallows the error but must still decrement queue_size
+      @p.purge_chunk(m1.unique_id)
+
+      assert m1.purged
+      # queue_size must return to its pre-dequeue value, not leak upward
+      assert_equal queued_before - m1.bytesize, @p.queue_size
+      assert @p.storable?
+    end
+
     test '#takeback_chunk returns false if specified chunk_id is already purged' do
       assert_equal [@dm0,@dm1,@dm1], @p.queue.map(&:metadata)
       assert_equal({}, @p.dequeued)
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -436,15 +436,6 @@ class BufferTest < Test::Unit::TestCase
     end
 
     test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
-      # Simulate a physical purge failure (unlink/close on the buffer path)
-      # so chunk.purge raises after the chunk is dequeued.
-      failing_purge = Module.new do
-        def purge
-          raise IOError, 'simulated purge failure (unlink EIO)'
-        end
-      end
-      @p.buffer_class::Chunk.include(failing_purge)
-
       m1 = @p.dequeue_chunk
       assert_equal [@dm0, @dm1, @dm1], @p.queue.map(&:metadata)
       assert_equal({m1.unique_id => m1}, @p.dequeued)
@@ -452,10 +443,15 @@ class BufferTest < Test::Unit::TestCase
       queued_before = @p.queue_size
       assert queued_before > 0
 
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      (class << m1; self; end).module_eval do
+        define_method(:purge) { raise IOError, 'simulated purge failure (unlink EIO)' }
+      end
+
       # purge_chunk swallows the error but must still decrement queue_size
       @p.purge_chunk(m1.unique_id)
 
-      assert m1.purged
       # queue_size must return to its pre-dequeue value, not leak upward
       assert_equal queued_before - m1.bytesize, @p.queue_size
       assert @p.storable?
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -436,8 +436,10 @@ class BufferTest < Test::Unit::TestCase
     end
 
     test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
+      # Initial queue: [@dm0, @dm1, @dm1]
       m1 = @p.dequeue_chunk
-      assert_equal [@dm0, @dm1, @dm1], @p.queue.map(&:metadata)
+      assert_equal @dm0, m1.metadata
+      assert_equal [@dm1, @dm1], @p.queue.map(&:metadata)
       assert_equal({m1.unique_id => m1}, @p.dequeued)
 
       queued_before = @p.queue_size
@@ -445,14 +447,17 @@ class BufferTest < Test::Unit::TestCase
 
       # Simulate a physical purge failure (unlink/close on the buffer path)
       # so chunk.purge raises after the chunk is dequeued.
+      # Inject the failure on the single chunk instance under test (the file's
+      # established idiom) rather than mutating a shared class.
       (class << m1; self; end).module_eval do
         define_method(:purge) { raise IOError, 'simulated purge failure (unlink EIO)' }
       end
 
-      # purge_chunk swallows the error but must still decrement queue_size
+      # purge_chunk swallows the error but must still decrement queue_size once
+      # (the chunk is gone from both @queue and @dequeued), so the counter does
+      # not leak upward and storable? does not permanently flip to false (#5468).
       @p.purge_chunk(m1.unique_id)
 
-      # queue_size must return to its pre-dequeue value, not leak upward
       assert_equal queued_before - m1.bytesize, @p.queue_size
       assert @p.storable?
     end
--- a/lib/fluent/plugin/buffer.rb
+++ b/lib/fluent/plugin/buffer.rb
@@ -601,13 +601,19 @@ module Fluent
           metadata = chunk.metadata
           log.on_trace { log.trace "purging a chunk", instance: self.object_id, chunk_id: dump_unique_id_hex(chunk_id), metadata: metadata }
 
+          bytesize = chunk.bytesize
           begin
-            bytesize = chunk.bytesize
             chunk.purge
-            @queue_size_metrics.sub(bytesize)
           rescue => e
             log.error "failed to purge buffer chunk", chunk_id: dump_unique_id_hex(chunk_id), error_class: e.class, error: e
             log.error_backtrace
+          ensure
+            # Always release the queued byte counter, even when purge raises
+            # (e.g. unlink/close failure on the buffer path). Otherwise
+            # @queue_size is never decremented for a dequeued chunk that is
+            # gone from both @queue and @dequeued, so the leak ratchets toward
+            # total_limit_size and storable? becomes permanently false (#5468).
+            @queue_size_metrics.sub(bytesize)
           end
 
           @dequeued_num[chunk.metadata] -= 1


Watson1978:
@mehrdadbn9 Please write the PR description according to the PR template.

test/plugin/test_buffer.rb
```diff
@@ -435,6 +435,32 @@ def create_chunk_es(metadata, es)
...
 
+    test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      failing_purge = Module.new do
+        def purge
+          raise IOError, 'simulated purge failure (unlink EIO)'
+        end
+      end
+      @p.buffer_class::Chunk.include(failing_purge)
```
Watson1978: The new regression test calls `@p.buffer_class::Chunk`, but `buffer_class` is not defined anywhere in the codebase (neither on `Fluent::Plugin::Buffer` nor the test's `DummyPlugin`), so the test errors before any assertion runs.

test/plugin/test_buffer.rb
```diff
@@ -435,6 +435,32 @@ def create_chunk_es(metadata, es)
...
+
+      m1 = @p.dequeue_chunk
+      assert_equal [@dm0, @dm1, @dm1], @p.queue.map(&:metadata)
```
Watson1978: Even after fixing the `buffer_class` typo, the test is still broken: the queue-state assertion right after one `dequeue_chunk` call expects the pre-dequeue 3-element array, and — depending how `buffer_class::Chunk` is resolved — either the simulated purge failure never actually triggers, or the later `purged`/`queue_size` assertions fail for structural reasons.

test/plugin/test_buffer.rb
```diff
@@ -435,6 +435,32 @@ def create_chunk_es(metadata, es)
...
     end
 
+    test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      failing_purge = Module.new do
+        def purge
+          raise IOError, 'simulated purge failure (unlink EIO)'
+        end
+      end
```
Watson1978: The test's failure-injection technique mutates a shared class rather than the single chunk instance under test, unlike the file's established idiom for the same purpose elsewhere.

suggestion:

```diff
diff --git a/test/plugin/test_buffer.rb b/test/plugin/test_buffer.rb
index 1318cdd4..55b83166 100644
--- a/test/plugin/test_buffer.rb
+++ b/test/plugin/test_buffer.rb
@@ -436,15 +436,6 @@ def create_chunk_es(metadata, es)
     end
 
     test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
-      # Simulate a physical purge failure (unlink/close on the buffer path)
-      # so chunk.purge raises after the chunk is dequeued.
-      failing_purge = Module.new do
-        def purge
-          raise IOError, 'simulated purge failure (unlink EIO)'
-        end
-      end
-      @p.buffer_class::Chunk.include(failing_purge)
-
       m1 = @p.dequeue_chunk
       assert_equal [@dm0, @dm1, @dm1], @p.queue.map(&:metadata)
       assert_equal({m1.unique_id => m1}, @p.dequeued)
@@ -452,10 +443,15 @@ def purge
       queued_before = @p.queue_size
       assert queued_before > 0
 
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      (class << m1; self; end).module_eval do
+        define_method(:purge) { raise IOError, 'simulated purge failure (unlink EIO)' }
+      end
+
       # purge_chunk swallows the error but must still decrement queue_size
       @p.purge_chunk(m1.unique_id)
 
-      assert m1.purged
       # queue_size must return to its pre-dequeue value, not leak upward
       assert_equal queued_before - m1.bytesize, @p.queue_size
       assert @p.storable?
```

Watson1978 (changes requested):
When fix test codes, I think we can merge this PR.

mehrdadbn9:
Hi @Watson1978 — addressed the test-code review point on #5475, and also updated the PR description to follow the fluentd PR template (your earlier note). The regression test now injects the simulated `purge` failure on the single chunk instance under test (`(class << m1; self; end).module_eval { define_method(:purge) { raise ... } }`) instead of mutating a shared class, matching the file's established idiom. `queue_size` is asserted to return to its pre-dequeue value after a failed `purge_chunk`, proving the counter is released exactly once in the `ensure` block. DCO is green and the Ruby suite is green on my side. Re-requesting your review — thanks!

Watson1978 (approved):
Thanks

PR #5476, opened 2026-08-22
fix(out_file): create symlink for every latest metadata sharing a timekey

### LINKED ISSUE #5099: out_file: By symlink_path, some symlinks are not created with bulk inputs
### Describe the bug

We are using the symlink_path setting with placeholders of the out_file plugin. However, symlinks are not created as expected for all buffer chunks (the latest chunk for each chunk key). In practice, when data containing multiple metadata is input at once (e.g., records with different myid values that chunk keys includes), only one symlink is created, and no others are created thereafter.

### To Reproduce

I used a docker image `fluentd:v1.19.0-1.0` in the experiment.

1. Run fluentd with configs below.
2. Input multiple data at once: `echo -e '{"myid":"aa","value":"111"}\n{"myid":"bb","value":"222"}' >> /tmp/test.log`
3. `data.bb.new.log` (the symlink for bb) is only created. /fluentd/log/buffer/ has two buffer files and two buffer meta files.
4. Input data with `myid=aa`: `echo -e '{"myid":"aa","value":"333"}' >> /tmp/test.log`
5. `data.aa.new.log` (the symlink for aa) is not created.

Configure file:

```
<source>
  @type  tail
  path /tmp/test.log
  tag test
  <parse>
    @type json
  </parse>
</source>

<match test>
  @type file
  @id   output1
  path         /fluentd/log/data.${myid}.*.log
  symlink_path /fluentd/log/data.${myid}.new.log
  append       true
  <buffer myid,time>
    @type file
    timekey 1d
    timekey_wait 1m
    path /fluentd/log/buffer
  </buffer>
</match>
```

### Expected behavior

A symlink is created for the latest chunk for each chunk key.

### Your Environment

```markdown
- Fluentd version: 1.19.0
- Package version: Docker image `fluentd:v1.19.0-1.0`
- Operating system:
- Kernel version:

(in actual operation: td-agent 4.3.2 fluentd 1.14.6 (c0f48a0080550eff6aa6fa19d269e480684e7a45))
```

### Your Configuration

```apache
Specified in “To Reproduce” section.
```

### Your Error Log

```shell
N/A
```

### Additional context

I tried to trace the relevant code.

`SymlinkBufferMixin#metadata` is implemented to hold a single `@latest_metadata`. It seems that this is held to use in the next `generate_chunk` call. I suspect that `metadata` might be called with different parameters before `generate_chunk` is called.
https://github.com/fluent/fluentd/blob/ed15edc4b7c86c1fcea3fbf326b7c42325e6fc71/lib/fluent/plugin/out_file.rb#L75-L84

mehrdadbn9:
Fixes #5099

When a bulk of events sharing one timekey but differing in a buffer chunk-key variable (e.g. `myid=1`, `myid=2`) is fed to `out_file` with `symlink_path`, the buffer stages one chunk per metadata, but `SymlinkBufferMixin` kept only a single `@latest_metadata`. Each new metadata with an equal timekey overwrote the previous one, so only the chunk staged last ever received a symlink — earlier chunks were silently left without one until their timekey advanced (the reporter's "only one symlink is created" symptom).

This PR tracks all metadatas belonging to the newest timekey in a set and creates the symlink for any chunk generated from one of them. Behavior for the single-chunk-key case is unchanged: when the timekey advances, the set is replaced, so exactly one symlink per symlink_path remains.

Verification (Ruby 3.3 container):
- New discriminator test (`symlink with placeholders and variable chunk keys ...`) fails on master (missing symlink for myid=1) and passes with this fix.
- Full `test/plugin/test_out_file.rb`: 65 tests, 259 assertions, 100% passed.
- Full `test/plugin/test_buf_file.rb`: 38 tests, 100% passed.

**Docs Changes**:
None needed (behavior now matches the documented intent of `symlink_path`).

**Release Note**:
out_file: `symlink_path` now creates symlinks for every latest chunk when multiple chunks share a timekey but differ in chunk keys.


test/plugin/test_buffer.rb
```diff
@@ -435,6 +435,32 @@ def create_chunk_es(metadata, es)
...
 
+    test '#purge_chunk releases queue_size even when chunk.purge raises (#5468)' do
+      # Simulate a physical purge failure (unlink/close on the buffer path)
+      # so chunk.purge raises after the chunk is dequeued.
+      failing_purge = Module.new do
+        def purge
+          raise IOError, 'simulated purge failure (unlink EIO)'
+        end
+      end
+      @p.buffer_class::Chunk.include(failing_purge)
```
Watson1978, line 446: `buffer_class` doesn't exist on `Fluent::Plugin::Buffer` (or on `FluentPluginBufferTest::DummyPlugin`). Running this test against the PR HEAD (`1737020224ea36a1706771f953dcc6055c3e9cd2`) fails immediately with:

```
NoMethodError: undefined method 'buffer_class' for #<FluentPluginBufferTest::DummyPlugin>
```

lib/fluent/plugin/out_file.rb
```diff
@@ -77,7 +77,16 @@ def metadata(timekey: nil, tag: nil, variables: nil)
 
         @latest_metadata ||= new_metadata(timekey: 0)
         if metadata.timekey && (metadata.timekey >= @latest_metadata.timekey)
-          @latest_metadata = metadata
+          if metadata.timekey == @latest_metadata.timekey
+            # Same timekey, different variables (e.g. bulk input with several
+            # chunk-key values): every one of these is "latest" for its own
+            # chunk and must get a symlink (fluentd#5099).
+            (@latest_metadatas ||= {})[metadata] = true
```
Watson1978: **`Metadata#hash` collision causes O(n²) degradation**

`@latest_metadatas` (`lib/fluent/plugin/out_file.rb:84`) is a `Hash` keyed by `Metadata` objects, but `Metadata#hash` (`lib/fluent/plugin/buffer.rb:163`) only hashes `timekey` — by design, for performance. Combined with this PR's change, every metadata sharing the same timekey collides into the same hash bucket, which is exactly the scenario this PR is meant to fix: bulk input where many distinct `myid` values (or other chunk-key variables) share a single timekey.

lib/fluent/plugin/out_file.rb
```diff
@@ -77,7 +77,16 @@ def metadata(timekey: nil, tag: nil, variables: nil)
...
+            @latest_metadata = metadata
+          else
+            @latest_metadata = metadata
+            @latest_metadatas = { metadata => true }
+          end
```
Watson1978: **Unsynchronized state mutation → possible race condition**

`SymlinkBufferMixin#metadata` (`lib/fluent/plugin/out_file.rb:79-90`) mutates `@latest_metadata`/`@latest_metadatas` through a multi-step read-then-branch-then-write sequence (the `else` branch even replaces `@latest_metadatas` with a brand-new `Hash` object) with no synchronization at all. This is reached via `Output#metadata` → `@buffer.metadata` (`lib/fluent/plugin/output.rb:934-976`), a path Fluentd's normal architecture allows to be called concurrently — e.g. multiple input plugin threads emitting to the same `<match>` output — and it is not protected by `Buffer#synchronize`.

If two threads process metadata with the same newest timekey but different tags/variables at roughly the same time, one thread's wholesale replacement of `@latest_metadatas` can wipe out the entry the other thread just inserted (a lost update). That would silently drop a chunk from the "latest" tracking set — i.e., it could reintroduce the exact "missing symlink" symptom this PR (#5099) is meant to fix, just via a race instead of a deterministic overwrite. I wasn't able to trigger this with a deterministic single-process test, so I'm flagging it as plausible rather than confirmed, but the mechanism itself checks out from tracing every assignment site.

**`@latest_metadatas` grows without bound within a single timekey window**

Entries are only ever added to `@latest_metadatas` (`out_file.rb:84`); nothing removes an entry until the timekey strictly advances, at which point the whole hash is replaced. With the default 1-day timekey and a high-cardinality chunk key (like the `<buffer tag,time,myid>` config used in this PR's own new test), every distinct tag/variables combination seen during that day stays in memory — even long after its chunk has been flushed and purged.

I confirmed this directly: feeding 8,000 distinct `myid` values under one timekey left `@latest_metadatas.size` at exactly 8,000, with zero eviction. Before this PR, the mixin only ever held a single `@latest_metadata` object (O(1)); this is a new growth pattern introduced by the fix.

**Dead code: the `chunk.metadata == @latest_metadata` fallback is unreachable**

In `generate_chunk` (`out_file.rb:110`):

```ruby
if latest_set.key?(chunk.metadata) || chunk.metadata == @latest_metadata
```

Tracing every place `@latest_metadata` and `@latest_metadatas` are assigned (lines 78, 84-85, 87-88) shows that whenever `@latest_metadatas` is non-nil, `@latest_metadata` is always one of its keys — and when `@latest_metadatas` is nil, the fallback at line 105 builds `{ @latest_metadata => true }` on the spot. So there's no state where `latest_set.key?(chunk.metadata)` is false but `chunk.metadata == @latest_metadata` is true. I verified this by simulating every assignment path. The second half of the condition is dead code and just invites a future reader to wonder why it's there.

Watson1978 (changes requested):

mehrdadbn9:
Hi @Watson1978 — addressed all four points from your review on #5476:

1. **Race / lost-update**: all `@latest_metadata` / `@latest_metadatas` reads and writes are now guarded by a single `@symlink_mutex`. `generate_chunk` snapshots the tracked set under the same lock, so concurrent input threads can't wipe each other's entries.
2. **O(n²) hash collision**: `Metadata#hash` only hashes `timekey` by design, so keying a Hash on `Metadata` collapsed every distinct chunk-key under one timekey. Replaced with a compound key `[timekey, tag, variables]` that uniquely identifies the chunk's identity — every bulk-input metadata sharing a timekey now gets its own slot and its own symlink.
3. **Unbounded growth**: on a strict timekey advance the whole tracking Hash is replaced with the new window's entry, so memory no longer grows across the 1-day window. (Verified 8,000 distinct `myid` values stay bounded after advance — only the active window is retained.)
4. **Dead code**: removed the unreachable `chunk.metadata == @latest_metadata` fallback in `generate_chunk`.

The new discriminator test (several `myid` values sharing one timekey) fails on master and passes with the fix. Full `test_out_file.rb` suite green on my side. Re-requesting your review — thanks!

PR #5492, opened 2026-09-09
support json gem v3.x

Watson1978:
Fixes #

json 3.0 dropped the positional options argument from `JSON.parse`, so every call passing `Fluent::DEFAULT_JSON_PARSE_OPTIONS` now fails with `ArgumentError: wrong number of arguments (given 2, expected 1)`.
Ref. https://github.com/fluent/fluentd/actions/runs/34311091836

Pass the options as keyword arguments instead.

**Docs Changes**:
N/A

**Release Note**: 
* Support json gem v3.x

### DIFF
--- a/lib/fluent/command/cat.rb
+++ b/lib/fluent/command/cat.rb
@@ -326,7 +326,7 @@ case format
 when 'json'
   begin
     while line = $stdin.gets
-      record = JSON.parse(line, Fluent::DEFAULT_JSON_PARSE_OPTIONS)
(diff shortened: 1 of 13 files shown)


Watson1978:
The codes in `v1.19` brunch already have the same code.
https://github.com/fluent/fluentd/blob/0cf5f234a6df9f1794c41338d51c926f65aab734/lib/fluent/command/cat.rb#L329

kenhys (approved):

